Issue #19, opened 2019-02-12
expose cluster services


rigazilla:
Currently nothing is exposed outside openshift. Operator must expose services (hotrod and rest). Configuration should be provided via CR file.

galderz:
Indeed. To be more precise, we need two services: one for hotrod and rest, and the other to enable DNS ping to work. For details, see how product defines services [here](https://github.com/jboss-container-images/jboss-datagrid-7-openshift-image/blob/datagrid73-dev/services/cache-service-template.yaml#L29-L63).

PR #22, opened 2019-02-13
Expose http and hotrod as ClusterIp services #19

galderz:
My first PR 🙈

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -273,6 +273,7 @@
   packages = [
     "authorization/v1",
     "pkg/serialization",
+    "route/v1",
     "user/v1",
   ]
   pruneopts = "NT"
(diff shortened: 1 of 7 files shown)


galderz:
Abandoning the idea of Hot Rod testing for the service. Client is limited in functionality and I've not figured out the steps to make it accessible externally when unencrypted. So instead I'll just rely on the HTTP port service, which we can easily test and we know how to expose externally. I'll remove the Hot Rod testing commit.

galderz:
@rigazilla This is ready to be integrated in CI agrees. Added HTTP testing of the service exposed via a route.

galderz:
And CI agrees :)

rigazilla:
merged, thanks!

PR #24, opened 2019-02-25
ISPN-9996 removed clustername field

rigazilla:
This PR removes clustername field form CR Spec. Cluster are identified by metadata.name now.

### DIFF
--- a/deploy/cr/cr_minimal.yaml
+++ b/deploy/cr/cr_minimal.yaml
@@ -4,4 +4,3 @@ metadata:
   name: example-infinispan
 spec:
   size: 3
-  clustername: helloworldcluster
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -3,15 +3,16 @@ package e2e
 import (
 	"bytes"
 	"fmt"
-	ispnv1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
-	"github.com/infinispan/infinispan-operator/test/e2e/util"
 	"io/ioutil"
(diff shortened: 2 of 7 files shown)


deploy/cr/cr_minimal.yaml
```diff
@@ -1,7 +1,6 @@
 apiVersion: infinispan.org/v1
 kind: Infinispan
 metadata:
-  name: example-infinispan
```
galderz: Any particular reason to change this name? I'd leave it as is (and also in `cr_with_configmap`). You'd normally run one or the other as you try it out things, so it's fine to have the same example name.

galderz: I consider these files as example files, mostly for documentation or our tutorials, so I think it's easier to stick just one name here.

test/e2e/main_test.go
```diff
@@ -172,54 +171,51 @@ func TestCreateWithInternalConfig(t *testing.T) {
 			Kind:       "Infinispan",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: "cache-infinispan-1",
+			Name: "cache-infinispan-minimal",
```
galderz: I think using more descriptive names here is fine.

deploy/cr/cr_minimal.yaml
rigazilla: I can't remember what was the idea... so I guess it's ok to use just one name

rigazilla:
> On top of comments, I see a few formatting changes here (or change of import orders). Did you add those or did golang's fmt introduce them?

vscode does it automatically with fmt

PR #25, opened 2019-02-25
ISPN-10002 Image configurable via CR spec

rigazilla:
This PR makes configurable the Image name for pods containers

### DIFF
--- a/pkg/apis/infinispan/v1/infinispan_types.go
+++ b/pkg/apis/infinispan/v1/infinispan_types.go
@@ -6,7 +6,8 @@ import (
 
 // InfinispanSpec defines the desired state of Infinispan
 type InfinispanSpec struct {
-	Size int32 `json:"size"`
+	Size  int32  `json:"size"`
+	Image string `json:"image"`
 }
 
 // InfinispanStatus defines the observed state of Infinispan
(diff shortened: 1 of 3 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -165,6 +165,16 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	// Check if pods container runs the right image
```
galderz: What's this check about?

galderz:
@rigazilla I've seen both PRs fail with timeout waiting for first pods to appear. I've seen that happening locally when there are leftover things. Dunno why might be happening here. I'll try to test locally and see if I can add more logging when timeout happens.

galderz:
Needs rebasing

pkg/controller/infinispan/infinispan_controller.go
rigazilla: Suppose that we have a 9.4.0 running cluster and the user changes the image to 9.4.5 and deploys cr. The Reconcile policy needs to handle this in some way, but we need some use cases definition first: i.e. is this a cluster upgrade use case?

Basically I added this check as a todo flag atm.

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -6,7 +6,9 @@ import (
 
 // InfinispanSpec defines the desired state of Infinispan
 type InfinispanSpec struct {
-	Size int32 `json:"size"`
+	Size        int32  `json:"size"`
+	ClusterName string `json:"clustername"`
```
galderz: This shouldn't be here, it was removed by ba950d6860bae1bcaa795cfee407f6ff8175e874

rigazilla: weird, fixed

PR #27, opened 2019-03-04
Increase single pod timeout

galderz:
This should fix CI and PR errors.

* Since this is the first test to execute,
  downloading image might take some time,
  so increase timeout to allow for image downloads.
* When pod number is not the one required,
  add some info to know more about the pods.


rigazilla:
@galderz looks good, but needs rebase

galderz:
Rebased

rigazilla:
merged, thank you!

PR #28, opened 2019-03-06
ISPN-10005 Dns ping capability

rigazilla:
Enabling dns ping protocol for cluster formation. 
Now the jgroups discovery protocol can be configured in the resource as spec.jgroups-stack. Default is kubernetes to preserve compatibility with the current Infinispan image.


rigazilla:
@galderz the cloud.xml with Infinispan doesn't have a dns_ping stack definition atm. So we need this cloud-dns.xml until we add the stack to the image.

galderz:
Ok, please create a JIRA and send a PR to community ASAP adding a JGroups stack that uses DNS ping (9.4.x and 10.x).

In the mean time, any use of the `cloud-dns.xml` should be hidden from the user. This means the user should not need to provide a custom XML configuration to get DNS based clustering working. The operator should internally hook this transparently. So `cr_cloud-dns-2.yaml` and `cr_cloud-dns.yaml` should not be needed. Existing CR examples should work and use DNS based clustering out of the box.

rigazilla:
Added PRs: [6749](https://github.com/infinispan/infinispan/pull/6749), [6750](https://github.com/infinispan/infinispan/pull/6750)
Let's wait if we can have them merged quickly and fix this pr just once

galderz:
@rigazilla Needs rebasing. The Infinispan PRs are integrated, so let's wait for next 10.x release

galderz:
Hmmm, Beta3 is currently scheduled for 29th March which is a bit late for this. Checking with Tristan for options.

galderz:
@rigazilla We're expecting a release this Friday (tomorow)

galderz:
@rigazilla Can you try with latest? It now has 10.0.0.Beta3

rigazilla:
@galderz tried. Looks good to me

rigazilla:
@galderz WAIT. missing a commit

rigazilla:
set dns-ping as default, now is ok

galderz:
@rigazilla You've not addressed my initial concerns with the PR. We don't need more XML definitions in the operator:

> I don't want to keep adding XML into this repo because it's a maintenance burden (with each update on the server you'd have to update the XML). We should be able to use alternative ping protocols passing in either system properties or environment variables.

rigazilla:
@galderz, cleaned up

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -8,6 +8,7 @@ import (
 type InfinispanSpec struct {
 	Size  int32  `json:"size"`
 	Image string `json:"image"`
+	JGroupsStack string `json:"jgroups-stack"`
```
galderz: Hmm, does adding this mean the user can configure the jgroups-stack to use? If so I'd leave it out, the user should not touch this.

rigazilla: @galderz you mean there's no gain for the user to choose an alternative discovery protocol?... sounds

galderz: Yeah, I think that's fine.

rigazilla: ok, removed

galderz:
Looks good now, integrating :)

PR #31, opened 2019-03-15
ISPN-10053 Mount secret volume for authentication

rigazilla:
User can mount secret into pod declaring its name in the
config.cluster-secrets customer resource yaml file.
Volume will be mounted as folder under the 'standalone/configuration'.
Secret must be created in advance and then user can setup a custom
configuration .xml that uses the secret's content:
user, password, keystore...

https://issues.jboss.org/browse/ISPN-10053

### DIFF
--- a/deploy/cr/cr_authentication.yaml
+++ b/deploy/cr/cr_authentication.yaml
@@ -1,0 +1,11 @@
+apiVersion: infinispan.org/v1
+kind: Infinispan
+metadata:
+  name: example-infinispan
+config:
+  sourceType: ConfigMap
+  sourceRef:  infinispan-app-configuration
+  name: cloud-ephemeral.xml
+  secret: cluster-secrets
+spec:
+  size: 3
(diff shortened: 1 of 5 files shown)


galderz:
@rigazilla Do you have a test for this? It's unclear if you need something to be implemented server side.

I do like the idea in principle, but it'd be easier for us to align things with server if we use the app user/pwd approach that's in community template and product. Eventually we'd like to go away from that in favour of a solution like this.

rigazilla:
@galderz I would like to keep this general design and if possible implement an helper layer that provides an user exp similar to the template and product approach.
Working on that and on a test that shows something

rigazilla:
added test

galderz:
@rigazilla I think we should close this PR. I don't think mounting a volume goes well with the approach of keystore or credentials secrets that we're going for #44 and more formally specified in #46.

rigazilla:
old proposal, closing

PR #33, opened 2019-03-27
ISPN-10053 COnfigurable user/pass for cloud.xml

rigazilla:
EDITED: this discussion is obsolete up to this message
https://github.com/infinispan/infinispan-operator/pull/33#issuecomment-479824070

This allows the user to configure user/pass both for application and
management.
This works for the default cloud.xml provided in the ispn image, fields
are grouped under the cloud: struct. see cr_cloud_auth.yaml file


rigazilla:
I don't like so much to force the user to write the password in a yaml file, but I don't have any better idea atm

Crumby:
> I don't like so much to force the user to write the password in a yaml file, but I don't have any better idea atm

@rigazilla Can't we use secret for holding user/pass and reference the secret in the yaml instead typing user/pass directly?

rigazilla:
@Crumby yeah, much better!

rigazilla:
This PR allows the user to configure application and management user/password for a cluster running with cloud.xml via secrets.

If the cloud.secret field is present in the Infinispan resource specification, this configuration method is used for authentication.

cloud.secret must contain the name of an accessible secret in the same namespace. The secret must contain all the required keys: APP_USER, APP_PASS, MGMT_USER, MGMT_PASS.
If something is missing pods will not start (CreateContainerConfigError).

See deploy/cr/cr_cloud_secret.yaml and cr_cloud_auth.yaml.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -213,6 +216,79 @@ func (r *ReconcileInfinispan) deploymentForInfinispan(m *infinispanv1.Infinispan
...
 	}
+
+	falseVal := false
+	var appUser, appPass, mgmtUser, mgmtPass corev1.EnvVar
+	if infinispanCloud.Secret != "" {
+		appUser = corev1.EnvVar{
+			Name: "APP_USER",
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{Name: "cloudsecret"},
```
Crumby: `Name: "cloudsecret"` -> `Name: infinispanCloud.secret`?

rigazilla: sure thanks!

Crumby:
@rigazilla I'm curious, what is the motivation for having three different specification fields `cloud`, `config` and `spec`? Seems to go against Kubernetes api conventions. Usually objects have `metada`, `spec` and `status` where all the user desired configuration should be under the `spec` field.

Second though I'm having is whether `cloud.secret` is enough. Are we going to use this one secret for all possible secrets we want to keep? As an example if we would like to give an user opportunity to provide a keystore for encrypting its endpoint we may want to keep that keystore in different secret. So having `cloud.authSecret` and `cloud.encryptSecret` would be more sufficient.

rigazilla:
> @rigazilla I'm curious, what is the motivation for having three different specification fields `cloud`, `config` and `spec`? Seems to go against Kubernetes api conventions. Usually objects have `metada`, `spec` and `status` where all the user desired configuration should be under the `spec` field.

I've added the `cloud` field just to keep together all the stuff that can work with the cloud.xml configuration.

But yeah, my idea is to spend some time looking to guidelines to reorganize the yaml.
Please add links in comment if you have good docs on this, thanks!

rigazilla:
> Second though I'm having is whether `cloud.secret` is enough. Are we going to use this one secret for all possible secrets we want to keep? As an example if we would like to give an user opportunity to provide a keystore for encrypting its endpoint we may want to keep that keystore in different secret. So having `cloud.authSecret` and `cloud.encryptSecret` would be more sufficient.

Is this something that can work with cloud.xml? 

If it needs a custom configuration server.xml I would like to provide the feature to mount a secret as a folder
here:
 https://github.com/infinispan/infinispan-operator/pull/31
The user can store there all the needed sensitive info.

Crumby:
@rigazilla Community's API conventions [doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md) is a good place where to start. Part about spec and status then confirms what I though about resource specifications.

rigazilla:
> Second though I'm having is whether `cloud.secret` is enough. Are we going to use this one secret for all possible secrets we want to keep? As an example if we would like to give an user opportunity to provide a keystore for encrypting its endpoint we may want to keep that keystore in different secret. So having `cloud.authSecret` and `cloud.encryptSecret` would be more sufficient.

Is this something that can work with cloud.xml? 

If it needs a custom configuration server.xml I would like to provide the feature to mount a secret as a folder
here:
 https://github.com/infinispan/infinispan-operator/pull/31
The user can store there all the needed sensitive info.

rigazilla:
@Crumby you are right, we should move everything under spec but I would like to use a different PR for that to avoid confusion.

rigazilla:
As specified in [k8s doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds) desired status of an object must be under `spec` field.

rigazilla:
I had a look to the docker-entrypoint.sh and to the cloud.xml and seems to me that things can be configured only via env variables and I can see any easy way to pass other kind of secrets (i.e. files or folders) that can be easily consumed by cloud.xml.

So cloud.secret seems enough, atm.

After https://github.com/infinispan/infinispan-operator/pull/31
If a user needs more configurability he can provide it's own .xml configuration via `config.name` and then provide all the required secrets files via `config.secret`.

deploy/cr/cr_cloud_auth.yaml
```diff
@@ -0,0 +1,19 @@
...
+#       MGMT_PASS
+# the cr_cloudsecret.yaml file is an example and can be used like this:
+# oc apply -f cloud_secret.yaml
+#
+
+apiVersion: infinispan.org/v1
+kind: Infinispan
+metadata:
+  name: example-infinispan
+cloud:
```
galderz: Why is `cloud` outside of `spec`? When does it make sense to do that? Couldn't easily find info on this.

rigazilla: `cloud` should be in `spec`, I've tracked that task in a separate issue.  See discussion above with Crumby.

Actually the _status_ of an Infinispan object is something undefined to me, but I don't it's a problem atm.

The only design choice here is to put under `cloud` all the setup data that works with the cloud.xml configuration.

galderz: > The only design choice here is to put under cloud all the setup data that works with the cloud.xml configuration.

Not sure I agree with that. Although we allow cloud XML file to be provided by the user, we want to move towards a situation where we don't expose that at all, hence I don't even think calling it `cloud` makes sense. 

I'm happy to leave it as `cloud`, but in the future it might not indicative. Something like `server` or `config` might be more suitable.

PR #36, opened 2019-04-08
ISPN-10117 Ensure a complete pods cleanup between tests

rigazilla:
This PR ensure that all the pods spawned by a test terminates before the next test start.

https://issues.jboss.org/browse/ISPN-10117

Plus: get the pod name from okd instead of suppose is "cache-infinispan-0"

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -70,7 +70,7 @@ func TestClusterFormation(t *testing.T) {
 	}
 	// Register it
 	okd.CreateInfinispan(&spec, Namespace)
-	defer okd.DeleteInfinispan("cache-infinispan", Namespace)
+	defer okd.DeleteInfinispan("cache-infinispan", Namespace, "app=infinispan-pod", SinglePodTimeout)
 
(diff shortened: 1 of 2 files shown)


test/e2e/main_test.go
```diff
@@ -70,17 +70,23 @@ func TestClusterFormation(t *testing.T) {
 	}
 	// Register it
 	okd.CreateInfinispan(&spec, Namespace)
-	defer okd.DeleteInfinispan("cache-infinispan", Namespace)
+	defer okd.DeleteInfinispanAndWaitPodsCleanup("cache-infinispan", Namespace, "app=infinispan-pod", SinglePodTimeout)
```
galderz: What about we leave the name of the method as is, `DeleteInfinispan`? Does it make sense to have a `DeleteInfinispan` method that does not wait for pod cleanup?

rigazilla: yeah ok, I can do that

rigazilla: done @galderz

PR #37, opened 2019-04-08
ISPN-10118 Fix operator status not being shown

galderz:
https://issues.jboss.org/browse/ISPN-10118

Fix so that operator status is shown.

Also, add ability to run operator and Infinispan on Minikube. We're already having to test operatorhub submission on Minikube (see separate PR), so I can simply reuse the same virtual machine for this too (can't run oc cluster straight on OSX).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 IMAGE ?= jboss/infinispan-operator
 TAG ?= latest
+GOOS ?= linux
 PROG  := infinispan-operator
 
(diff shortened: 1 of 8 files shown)


galderz:
Do not integrate yet, I still need double check against the operator scorecard.

galderz:
@rigazilla I've verified this with the scorecard and the fix looks good.

galderz:
Btw, let me close this PR cos I'm merging it with ISPN-10126.

PR #39, opened 2019-04-30
ISPN-10158 In docker multistage build

rigazilla:
This PR allows the image for distros that provide docker without multistage build.

https://issues.jboss.org/browse/ISPN-10158


galderz:
Looks good, thanks @rigazilla

PR #40, opened 2019-05-07
Expose detailed status info in the spec.status field

rigazilla:
This is an experimental PR that shows how detailed info can be provided to the user in the `spec.status` field.

spec.status contains:

- a `summary` field (an overall status for the cluster. `Ready` iff all the nodes are ready)
- a `nodes` filed, which is the list of the pods, containing:
. a `name` field with the pod name
. an `image` field with the running image
. a `clusterMembers` with the cluster members list
. a `clusterSize`
. a 'conditions` field, which is a list of condition flags coming from the pod

clusterMembers and clusterSize are populated with the values read via the Infinispan cli (exec on pod)


deploy/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: infinispan-operator
       containers:
         - name: infinispan-operator
-          image: jboss/infinispan-operator:latest
+          image: docker.io/rigazilla/rigazilla-operator:i0.02
```
galderz: Why change this to your repo?

galderz:
I'm not sure you need to see the image that each pod uses. If all pods are created by the operator, then they all should be using the same image, so I'd leave the image in the status but only present it once.

Also, showing the cluster view that each pod sees might get too noisy? I can see this potentially be of some use if say a node has a different view to the others, but it'd be hard to spot if say you have a cluster of 100 pods. If we have a big cluster and one of the nodes has a slightly different view to the others, I'd want that to be shown in the status, but in a way that it does not become too nosiy. What was your intention with this exactly?

deploy/operator.yaml
rigazilla: uh, yeah oversight

rigazilla:
> I'm not sure you need to see the image that each pod uses. If all pods are created by the operator, then they all should be using the same image, so I'd leave the image in the status but only present it once.
> 
Maybe this could be useful to detect some fault. Check the version on a pod is probably the first step a devop wants to do.

> Also, showing the cluster view that each pod sees might get too noisy? I can see this potentially be of some use if say a node has a different view to the others, but it'd be hard to spot if say you have a cluster of 100 pods. If we have a big cluster and one of the nodes has a slightly different view to the others, I'd want that to be shown in the status, but in a way that it does not become too nosiy. What was your intention with this exactly?

The general idea in this PR was to provide raw data to the consumer.

The followings were the reference use cases:
- consumer is a devOp bot. In this case is probably easier for the bot to have the complete info in one shot.
- consumer is a presentation layer that can organize/filter info for the real user.

In the output there's also a `summary` field as an example aggregated info produced by the operator, but my preference goes to leave to the consumer the possibility to build rules from raw data.

rigazilla:
> Do you have an example of what the output looks like?

```
12:10 $ oc get pod
NAME                   READY     STATUS    RESTARTS   AGE
example-infinispan-0   1/1       Running   0          3m
example-infinispan-1   1/1       Running   0          3m
example-infinispan-2   1/1       Running   0          2m
✔ ~/go/src/github.com/infinispan/infinispan-operator [status_exp L|…30⚑ 8] 
12:11 $ oc get infinispan -o yaml
apiVersion: v1
items:
- apiVersion: infinispan.org/v1
  kind: Infinispan
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"infinispan.org/v1","kind":"Infinispan","metadata":{"annotations":{},"name":"example-infinispan","namespace":"myproject"},"spec":{"image":"jboss/infinispan-server:latest","size":3}}
    creationTimestamp: 2019-05-16T10:07:42Z
    generation: 1
    name: example-infinispan
    namespace: myproject
    resourceVersion: "5210"
    selfLink: /apis/infinispan.org/v1/namespaces/myproject/infinispans/example-infinispan
    uid: 717ef9b8-77c2-11e9-aeb1-482ae332698b
  spec:
    image: jboss/infinispan-server:latest
    size: 3
  status:
    nodes:
    - clusterMembers: '[example-infinispan-0, example-infinispan-1, example-infinispan-2]'
      clusterSize: 3
      conditions:
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:07:42Z
        status: "True"
        type: Initialized
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:08:00Z
        status: "True"
        type: Ready
      - lastProbeTime: null
        lastTransitionTime: null
        status: "True"
        type: ContainersReady
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:07:42Z
        status: "True"
        type: PodScheduled
      image: docker.io/jboss/infinispan-server:latest
      name: example-infinispan-0
    - clusterMembers: '[example-infinispan-0, example-infinispan-1, example-infinispan-2]'
      clusterSize: 3
      conditions:
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:08:00Z
        status: "True"
        type: Initialized
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:08:24Z
        status: "True"
        type: Ready
      - lastProbeTime: null
        lastTransitionTime: null
        status: "True"
        type: ContainersReady
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:08:00Z
        status: "True"
        type: PodScheduled
      image: docker.io/jboss/infinispan-server:latest
      name: example-infinispan-1
    - clusterMembers: '[example-infinispan-0, example-infinispan-1, example-infinispan-2]'
      clusterSize: 3
      conditions:
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:08:24Z
        status: "True"
        type: Initialized
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:08:51Z
        status: "True"
        type: Ready
      - lastProbeTime: null
        lastTransitionTime: null
        status: "True"
        type: ContainersReady
      - lastProbeTime: null
        lastTransitionTime: 2019-05-16T10:08:24Z
        status: "True"
        type: PodScheduled
      image: docker.io/jboss/infinispan-server:latest
      name: example-infinispan-2
    summary: Ready
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

galderz:
@rigazilla What are conditions useful for exactly? 

I do still think those cluster memberships are going to get noisy. The only time you care about individual pod cluster memberships is when one of the pods has a different view. To repeat what I said before, if you have 100 pods, how is a user visually going to see if there are any discrepancies in the views? I do agree about providing raw data but there will also be eyes on users looking at this.

Again I still think there's no value in each individual pod indicating the image run. This is all controlled centrally by the operator. I can't see a situation where a pod should/would run a different image.

Much more interesting is to see what each pod is consuming memory wise. This would for example allow us to see if any pods is consuming more memory than others, and see if we want to do something on it? e.g. get a heap dump or something. This is valuable short information.

galderz:
@rigazilla I think you should spec this out and share it with the rest of the team. We need some consensus on what to show.

rigazilla:
> @rigazilla I think you should spec this out and share it with the rest of the team. We need some consensus on what to show.

Yeah, according to k8s conventions, I think that all the pod specific info can be removed, so only cluster members and a reference to pod object will be shown. 
Adding formal status spec to https://github.com/infinispan/infinispan-operator/pull/46

rigazilla:
This is the new output:
```15:18 $ oc get infinispan -o yaml
apiVersion: v1
items:
- apiVersion: infinispan.org/v1
  kind: Infinispan
  metadata:
    annotations:
      kubectl.kubernetes.io/last-applied-configuration: |
        {"apiVersion":"infinispan.org/v1","kind":"Infinispan","metadata":{"annotations":{},"name":"example-infinispan","namespace":"myproject"},"spec":{"image":"jboss/infinispan-server:latest","size":3}}
    creationTimestamp: 2019-05-29T13:11:07Z
    generation: 1
    name: example-infinispan
    namespace: myproject
    resourceVersion: "488044"
    selfLink: /apis/infinispan.org/v1/namespaces/myproject/infinispans/example-infinispan
    uid: 38a3be28-8213-11e9-888b-482ae332698b
  spec:
    image: jboss/infinispan-server:latest
    size: 3
  status:
    nodes:
    - clusterMembers: '[example-infinispan-0, example-infinispan-1, example-infinispan-2]'
      clusterSize: 3
      podName: example-infinispan-0
    - clusterMembers: '[example-infinispan-0, example-infinispan-1, example-infinispan-2]'
      clusterSize: 3
      podName: example-infinispan-1
    - clusterMembers: '[example-infinispan-0, example-infinispan-1, example-infinispan-2]'
      clusterSize: 3
      podName: example-infinispan-2
    summary: Ready
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

Crumby:
@rigazilla @galderz Having `clusterMembers` and `clusterSize` in operator status showed per node seems like lot of redundant information. It will be same for most of the times. I'm not sure about having memory usage per pod in operator status either. Theoretically that information can be shown directly in OKD console or retrieved with `oc` CLI if you have monitoring enabled (not 100% sure however).

What I think would make sense in here is to use `conditions` in similar way as is done for `statefulSets` and `deploymentConfigs`. As an example there would be a condition that infinispan-operator is running and ready meaning that all the nodes are up and ready to retrieve requests. Another condition could check cluster consistence -> all the nodes have same `clusterMemebers` and `clusterSize`. Later one wouldn't increase output exponentially with `clusterSize` compared to actual design.

I'm putting here example status for `deploymentConfig` with 4 replicas for reference:
```
status:
  availableReplicas: 4
  conditions:
  - lastTransitionTime: 2019-06-05T11:25:13Z
    lastUpdateTime: 2019-06-05T11:25:13Z
    message: Deployment config has minimum availability.
    status: "True"
    type: Available
  - lastTransitionTime: 2019-06-05T11:23:39Z
    lastUpdateTime: 2019-06-05T11:25:14Z
    message: replication controller "jdg-cache-1" successfully rolled out
    reason: NewReplicationControllerAvailable
    status: "True"
    type: Progressing
  details:
    causes:
    - type: ConfigChange
    message: config change
  latestVersion: 1
  observedGeneration: 1
  readyReplicas: 4
  replicas: 4
  unavailableReplicas: 0
  updatedReplicas: 4
```

rigazilla:
@galderz @Crumby my perplexity with the "aggregate" approach is that in case of trouble we provide very little help to the human devop. I.e.: suppose that the cluster has 100 nodes but is malformed, the operator queries all the nodes and detects the faults and desipite it has all the info just returns "malformed: true". How? Which node?
How we can help more?
More specific aggregate conditions? (which ones?)
Helper api to investigate faults?

Crumby:
@rigazilla Every condition has a `message` field to give user an indication of why the condition occurred. This can be used to propagate more info about how cluster is malformed. (eg: "There are 2 different views in 4 nodes cluster. [ispn-0, ispn-2] and [ispn-1, ispn-3]").

rigazilla:
> @rigazilla Every condition has a `message` field to give user an indication of why the condition occurred. This can be used to propagate more info about how cluster is malformed. (eg: "There are 2 different views in 4 nodes cluster. [ispn-0, ispn-2] and [ispn-1, ispn-3]").

This sounds good.
I can add a condition and message with all the different views

rigazilla:
A example of output with conditions:
```
10:24 $ KUBECONFIG=/home/rigazilla/openshift-on-aws/auth/kubeconfig  oc get infinispan/example-infinispan -o yaml
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"infinispan.org/v1","kind":"Infinispan","metadata":{"annotations":{},"name":"example-infinispan","namespace":"myproject"},"spec":{"image":"jboss/infinispan-server:latest","size":2}}
  creationTimestamp: "2019-06-19T09:56:04Z"
  generation: 1
  name: example-infinispan
  namespace: myproject
  resourceVersion: "34431"
  selfLink: /apis/infinispan.org/v1/namespaces/myproject/infinispans/example-infinispan
  uid: 739ffc37-9278-11e9-b80e-0244faf5516a
spec:
  image: jboss/infinispan-server:latest
  size: 2
status:
  conditions:
  - message: 'View: [example-infinispan-0, example-infinispan-1]'
    status: "True"
    type: well-formed
  statefulSetName: example-infinispan
```

galderz:
@rigazilla Had a quick look and that looks much better. As @Crumby also said, the initial proposals had too much information on them. I will have a deeper look into this in the next couple of days.

galderz:
>  I'm not sure about having memory usage per pod in operator status either. Theoretically that information can be shown directly in OKD console or retrieved with oc CLI if you have monitoring enabled (not 100% sure however).

^ Prometheus exposes this when monitoring is enabled. I've seen this in action but I've not set up any OpenShift clusters with this myself.

galderz:
@rigazilla Some comments:

* `type` should be camel cased, e.g. `WellFormed` as opposed to `well-formed`. That seems to be the consensus looking at [pod conditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions).

* What condition type would be shown if there were divergences between the cluster views? Another thing to consider where there are divergences in cluster views is what to put in the `reason` field. The [pod lifecycle](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/) has suggestions on that too.

* Also, I think we should have some unit tests that guarantee that we adhere to the spec, even for situations when the cluster views differ. Doing this might be difficult to do with an integration test, but could use a unit test for that?

rigazilla:
>     * `type` should be camel cased, e.g. `WellFormed` as opposed to `well-formed`. That seems to be the consensus looking at [pod conditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions).

ok

rigazilla:
>     * What condition type would be shown if there were divergences between the cluster views?

There's only wellFormed condition atm, with this semantic:

- True: no error getting cluster info from every node via CLI && just one view
- False: no error getting cluster info from every node via CLI && different views
- Unknown: error getting cluster info from nodes via CLI

rigazilla:
>     * Also, I think we should have some unit tests that guarantee that we adhere to the spec, even for situations when the cluster views differ. Doing this might be difficult to do with an integration test, but could use a unit test for that?

yeah, good idea introducing unit test

galderz:
Looks good to me. @Crumby Any more feedback?

Crumby:
@galderz No, PR LGTM now.

galderz:
Fantastic! Thx @rigazilla for the great effort and @Crumby for the excellent feedback :)

PR #41, opened 2019-05-07
Enabling statefulSet rolling up

rigazilla:
This experimental PR enables statefulSet rolling upgrade.

When an Infinispan cluster is deployed, user can upgrade/change the running image changing the `spec.image` field of the Infinispan resource.

The simplest upgrade policy is enable atm: single pod swich off/change/swich on


rigazilla:
Idea here has been ported to #156. I think we can close this PR.

Issue #42, opened 2019-05-17
Set UBI as image base for operator


galderz:

tristantarrant:
We should use ubi-minimal of possible

galderz:
Oh yeah

galderz:
@tristantarrant Just about to try with ubi-minimal actually...

PR #43, opened 2019-05-17
Set UBI minimal as base image for operator #42

galderz:


rigazilla:
merged, thank you @galderz

PR #47, opened 2019-05-27
Enable control of non standard Infinispan image

rigazilla:
This PR provide some flexibility in the setup of the Infinispan image for users that want to run non standard Infinispan image. 

Following a list of the available customization:
- **names of the environment variables** The operator defines in the pod all the env variables needed by Infinispan. The variables names are tuned for standard Infinispan, if some name mapping is needed can be done in the operator environment. I.e. if a specific image uses _APPLICATION_USER_ instead of _APP_USER_ the operator environment must have a `APP_USER=APPLICATION_USER` setting;

- **arguments for entry point command**: args can be passed into the `ENTRY_POINT_ARGS` var as a json array of strings. Default is not the empty array so if needed it must be explicitly set;

- **probes entries**: custom entry points can be passed into the `PROBES` var as a json map. Standard image only uses `liveness` and `readiness` probe.

- **appendVolumes**: operator will try to mount all the entries in the `VOLUME_MOUNTS` env vars. `VOLUME_REFS` var must contains all the referred volume name. `VOLUME_CLAIMS` must contains all the templates for the claimed volume.
Please see `misc/launch.json` for an example.


rigazilla:
@oraNod I've added an how-to start custom Infinispan image here:
https://github.com/rigazilla/infinispan-operator/blob/dg_startup/documentation/asciidoc/stories/spinning-up-derived-clusters.adoc

deploy/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: infinispan-operator
       containers:
         - name: infinispan-operator
-          image: jboss/infinispan-operator:latest
+          image: docker.io/rigazilla/rigazilla-operator:i0.03
```
Crumby: Is this change intentional?

misc/launch.json
```diff
@@ -0,0 +1,29 @@
...
+        {
+            "name": "Launch",
+            "type": "go",
+            "request": "launch",
+            "mode": "auto",
+            "program": "${fileDirname}",
+            "env": {
+                "WATCH_NAMESPACE": "myproject",
+                "APP_USER": "APPLICATION_USER",
+                "APP_PASS": "APPLICATION_PASSWORD",
```
Crumby: Wouldn't it be better to update non-standard image to understand operator set-up? We want to bring them closer to upstream anyway.

deploy/operator.yaml
rigazilla: Uh no, my fault... actually is the second time I made this mistake, I begin to think that I want to release a rigazilla-operator :)

Crumby:
@rigazilla We should ensure that tests are easily executable against non-standard image as well. (See [ClusterFormationTest](https://github.com/infinispan/infinispan-operator/blob/9969e348299dea0d9ce18c7dcdb7986096529f42/test/e2e/main_test.go#L112) as an example that will fail)

rigazilla:
> @rigazilla We should ensure that tests are easily executable against non-standard image as well. (See [ClusterFormationTest](https://github.com/infinispan/infinispan-operator/blob/9969e348299dea0d9ce18c7dcdb7986096529f42/test/e2e/main_test.go#L112) as an example that will fail)

good point, we need to deal with that

misc/launch.json
rigazilla: Yeah it would be better, but atm we need that defines. When images will be closer we can just remove the env vars and the operator will run with defaults

deploy/operator.yaml
galderz: Why are you making the change in the first place?

galderz:
@rigazilla Needs rebasing

deploy/operator.yaml
rigazilla: no just a mistake, the fix is in the commit below

deploy/dg_operator.yaml
```diff
@@ -0,0 +1,48 @@
...
+      name: infinispan-operator
+  template:
+    metadata:
+      labels:
+        name: infinispan-operator
+    spec:
+      serviceAccountName: infinispan-operator
+      containers:
+        - name: infinispan-operator
+          image: docker.io/rigazilla/rigazilla-operator:i0.03
```
rigazilla: this points to my repo to ease quick testing. should be changed when an official operator image with this pr will be available

rigazilla:
> > @rigazilla We should ensure that tests are easily executable against non-standard image as well. (See [ClusterFormationTest](https://github.com/infinispan/infinispan-operator/blob/9969e348299dea0d9ce18c7dcdb7986096529f42/test/e2e/main_test.go#L112) as an example that will fail)
> 
> good point, we need to deal with that

Test suite now is ok, non standard cli path can be configured in CLI_CMD env var

galderz:
I've tried this locally and looked good. @rigazilla Can you please sort out the commits and squash them?

galderz:
@rigazilla I've created issue #54 for this. When you squash commits, add the `#54` to the end of the Git commit summary.

rigazilla:
Just a resume of the latest version.

A set of variables can be set in the operator environment to make it run with custom Infinispan images.

- APP_USER: must contains the name of the environment variable (in the Infinispan image) where Infinispan is expecting to find the application user name
- APP_PASS: same as above for app user password
- MGMT_USER: same for admin username
- MGMT_PASS: and admin password
- ENTRY_POINT_ARGS: json array with additional arguments for the entry point command
- PROBES: json maps with probes. "readiness" and "liveness" probes avaible
- VOLUME_MOUNTS: volumes needed by the image. Json struct, sample "[{\"MountPath\": \"/opt/datagrid/standalone/data\", \"Name\": \"srv-data\"},{\"MountPath\": \"/var/run/secrets/java.io/keystores\", \"Name\": \"keystore-volume\"},{\"MountPath\": \"/var/run/secrets/openshift.io/serviceaccount\", \"Name\": \"services-certs\"}]"
- VOLUME_REFS: references to volumes provided by the environment. Sample "[\"srv-data\",\"keystore-volume\",\"services-certs\"]"
- VOLUME_CLAIMS: reference to volumes that must be claimed "[{\"metadata\": {\"Name\": \"srv-data\"}, \"Spec\":{\"AccessModes\": [\"ReadWriteOnce\"], \"Resources\": {\"Requests\": {\"storage\": \"1Gi\"}}}}]"
- CLI_CMD: path to the cli command
- DEFAULT_IMAGE: default image if not provided in the cr instance description

deploy/cr/cr_minimal_with_auth.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: example-infinispan
 spec:
   size: 2
-  image: jboss/infinispan-server:latest
+  image: registry.access.redhat.com/jboss-datagrid-7/datagrid73-openshift:1.0
```
galderz: This should not be here.

test/e2e/main_test.go
```diff
@@ -137,7 +147,8 @@ func TestExternalService(t *testing.T) {
 			Name: "cache-infinispan-0",
 		},
 		Spec: ispnv1.InfinispanSpec{
-			Size: 1,
+			Size:  1,
```
galderz: We should avoid cosmetic changes like this to align things. Did we create this or does `fmt` enforce it?

deploy/cr/cr_minimal_with_auth.yaml
rigazilla: ok, removed

test/e2e/main_test.go
rigazilla: yeah fmt wants it

PR #49, opened 2019-05-29
Switch testsuite to use pure Kubernetes APIs

galderz:
And with that, be able to run testsuite on Minikube. To test:

```bash
make minikube-config
make minikube-start
make minikube-test
```

Has been tested with latest Minikube (1.1.0).

I think we should embrace vanilla Kubernetes APIs, switch to Minikube completely and remove OpenShift references. Vanilla Kubernetes support is required to make it to operatorhub.io anyway.

Note: CI script would need updated, so CI runs for this PR will fail for sure.


galderz:
I've created issue #50 for this.

galderz:
@rigazilla This is no longer preview, it's ready to go in now that CI has been switched to minikube too. /cc @gustavonalle

gustavonalle:
LGTM! One suggestion is to trigger a lot of builds on travis (like 10) to see if there aren't random failures, as k8s and openshift are complicated beasts :)

Makefile
```diff
@@ -90,6 +90,9 @@ minikube-start:
 minikube-run-local: build
 	build/minikube/run-local.sh ${NAMESPACE}
 
+minikube-test: build
```
gustavonalle: TIP: If add a comment on top of the minikube related tasks, when someone types ```make``` it shows a summary of available targets.

galderz:
Moving this back to preview so that we can add testing on okd as well to CI. This will give us guarantees that we're not doing anything that is not supported in okd.

@gustavonalle If you have time, can you see if our okd set up can be updated?

gustavonalle:
You mean, gustavonalle/oc-cluster-up?

galderz:
@gustavonalle Yup

gustavonalle:
I will take a look

galderz:
@rigazilla This is ready to go in again. CI now runs with both Minikube and OpenShift and the jobs run concurrently to save up time. I've also completely removed the `okd` client code since we can now guarantee that vanilla Kubernetes based client code works well regardless of the target platform.

galderz:
@rigazilla Can you integrate it then?

galderz:
Ah yes, missing adding comments on top of make targets

rigazilla:
merged, thanks @galderz!

PR #57, opened 2019-06-07
removed Openshift Deps

rigazilla:
No more need to depend from openshift go api.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -279,33 +279,6 @@
   version = "1.0.1"
 
 [[projects]]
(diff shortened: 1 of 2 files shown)


rigazilla:
@galderz I've some build warning with master and I've notice that we still have openshift dep in Gopkg.toml.
Please check if this PR sounds ok

galderz:
The PR looks good. There was a problem with etcd server on the minikube run. Seemed to be caused by some kind of resource issue since liveness probe failed there. I'll remove minikube memory settings in another PR..

PR #58, opened 2019-06-07
Remove memory settings with minikube driver=none

galderz:


galderz:
Some jobs are failling on Minikube, the problems seem to be resource related, in particular around the cluster formation test. I've seen similar resource issues when Minikube 2 cpus and running 2 node cluster with OLM. It could be that sometimes with 2 cpus there are not enough resources to run things with Minikube (e.g. current java+cli based probes can take a lot of CPU). Ideally we'd have 4 cpus and 4gb memory, but that's not the case with default travis settings.

So if things continue to fail after this PR, we should remove Minikube from CI and leave OpenShift alone which works pretty reliably. Once we have probes that use less resources (/cc @gustavonalle), we can try Minikube again.

rigazilla:
merged, thanks @galderz

PR #60, opened 2019-06-13
59/expose service

rigazilla:
This replaces NodePort with LoadBalancer as type of the service that exposes the infinispan cluster during tests.  NodePort doesn't work on some platforms (os on aws).
On the other hand some platforms create the LoadBalancer but do not expose it to the outside (minikube, oc-cluster-up).
So this PR tries both NodePort and LoadBalancer access.

### DIFF
--- a/test/e2e/util/k8s/k8s_client.go
+++ b/test/e2e/util/k8s/k8s_client.go
@@ -2,17 +2,19 @@ package k8s
 
 import (
 	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"path"
+
 	api "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
 	ispnv1 "github.com/infinispan/infinispan-operator/pkg/generated/clientset/versioned/typed/infinispan/v1"
-	"io/ioutil"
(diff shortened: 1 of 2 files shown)


test/e2e/util/k8s/k8s_client.go
```diff
@@ -439,64 +441,107 @@ func debugPods(required int, pods []v1.Pod) {
...
+	// If the cluster exposes external IP then return it
+	if len(router.Status.LoadBalancer.Ingress) == 1 {
+		if router.Status.LoadBalancer.Ingress[0].IP != "" {
+			return router.Status.LoadBalancer.Ingress[0].IP + ":8080"
+		}
+		if router.Status.LoadBalancer.Ingress[0].Hostname != "" {
+			return router.Status.LoadBalancer.Ingress[0].Hostname + ":8080"
+		}
+	}
+	// Return empty address if nothing available
```
galderz: Even though you return empty here, I don't see there's any checks to see if the external address is empty in the code that calls it. So, why not just throw an error if no external address can be computed?

rigazilla: ok, added error there and simplified checkExternalAddress fail on error

Issue #63, opened 2019-06-19
ServiceAccount for Infinispan pods should be configurable


rigazilla:
Currently Infinispan pods run with the `infinispan-operator` account.
This account can do almost everything inside the namespace: i.e. read and write all the secrets, create/delete objects.
The result is that anyone that can login to the Infinispan pod can do dangerous actions.

The user should be able to specify the service account that will be used to run the Infinispan cluster.

galderz:
Decided not to make service account configurable, and just use default service account.

PR #65, opened 2019-06-20
Use default service account #63

rigazilla:
This PR makes configurable the ServiceAccountName of the Infinispan pods setting `Infinispan.Spec.ServiceAccountName` field. If empty pods will run with the `default` service account.

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -295,7 +295,6 @@ func (r *ReconcileInfinispan) deploymentForInfinispan(m *infinispanv1.Infinispan
 						Resources: corev1.ResourceRequirements{Requests: corev1.ResourceList{"cpu": resource.MustParse("0.5"),
 							"memory": resource.MustParse("512Mi")}},
 					}},
-					ServiceAccountName: "infinispan-operator",
 				},
 			},
 		},


rigazilla:
Update pr to always use default service account

galderz:
@rigazilla If we don't need the service account, shouldn't you also remove it from `deploy/rbac.yaml` definition?

galderz:
Isn't it the case that we do need to define a service account which is linked with a role that has just enough rights to do what it needs to do?

rigazilla:
> @rigazilla If we don't need the service account, shouldn't you also remove it from deploy/rbac.yaml definition?

@galderz the operator itself runs as service-account:infinispan-operator and it needs the rights in the rbac settings.

> Isn't it the case that we do need to define a service account which is linked with a role that has just enough rights to do what it needs to do?

The rbac.yaml has been generated by the sdk, do you mean we have to do a fine tuning and check if all the rights are really needed?

galderz:
> The rbac.yaml has been generated by the sdk, do you mean we have to do a fine tuning and check if all the rights are really needed?

Probably. @Crumby mentioned something related to this, plus there's a document around this.

rigazilla:
> > The rbac.yaml has been generated by the sdk, do you mean we have to do a fine tuning and check if all the rights are really needed?
> 
> Probably. @Crumby mentioned something related to this, plus there's a document around this.

The operator is namespace scoped and runs with the default `restricted` SCC, so (reading through the doc) it's safe enough to be defined a `non-privileged` operator. My opinion is that we could stay with the default rbac atm.

Issue #61, opened 2019-06-18
Run go vet


galderz:
Running `go vet` throws error:

```
2019-06-18 07:07:19,972 - atomic_reactor.util - DEBUG - /root/go/src/github.com/infinispan/infinispan-operator
2019-06-18 07:07:20,110 - atomic_reactor.util - DEBUG - [91m+ go vet ./...
2019-06-18 07:07:20,110 - atomic_reactor.util - DEBUG - [0m
2019-06-18 07:07:29,663 - atomic_reactor.util - DEBUG - [91m# github.com/infinispan/infinispan-operator/pkg/generated/listers/infinispan/v1
2019-06-18 07:07:29,663 - atomic_reactor.util - DEBUG - pkg/generated/listers/infinispan/v1/infinispan.go:91:34: undefined: "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1".Resource
```

PR #66, opened 2019-06-20
Fix go vet errors and add it as a build stage #61

galderz:


rigazilla:
merged. thanks @galderz

Issue #67, opened 2019-06-21
Implement InfinispanContainerSpec


rigazilla:
As described [here](https://github.com/infinispan/infinispan-operator/blob/f7250d0fb4e44cf81b0e03d51acc13cf5d8b38fe/doc/design.adoc#infinispancontainerspec)

PR #68, opened 2019-06-26
InfinispanContainerSpec implementation #67

rigazilla:
This PR allows the user to configure the InfinispanContainerSpec (infinispan.spec.container) properties.
See https://github.com/infinispan/infinispan-operator/pull/46, for a formal definition.

- `infinispan.spec.container.memory`: memory reserved to the pod (512Mi)
- `infinispan.spec.container.storage`: storage space for the `srv-data` volume (2Gi)
- `infinispan.spec.container.jvmOptionsAppend`: additional java option for the Infinispan server (""). The value of this field is by default passed to the pod into the `JAVA_OPTS` variable. If needed (i.e. in case of custom Infinispan image),  the variable name can be changed setting the `JAVA_OPTS_VARNAME` in the operator env.


galderz:
@rigazilla `infinispan.spec.container.memory` and `infinispan.spec.container.jvmOptionsAppend` are good, but in v2 of the configuration design I'm moving `storage` option to be under a `datagrid` element. Also, I'm adding `infinispan.spec.container.cpu`. Could you rework the PR to only add memory/jvmOptionsAppend/cpu to container section?

rigazilla:
done, @galderz

PR #74, opened 2019-07-01
Random generated credentials if not provided #72

rigazilla:
Generate a random user/pass credential for application and management authentication, if not provided.
Generated values can be read in the pod environment variables (MGMT_USER, MGMT_PASS, APP_USER, APP_PASS)


rigazilla:
Credentials are stored in the container environment variables (`infinispan.spec.containers[0].env`) and user can obtain them via kubectl/oc command line. 
Example:
```
$ oc get pod
NAME                   READY     STATUS    RESTARTS   AGE
example-infinispan-0   1/1       Running   0          1m
example-infinispan-1   1/1       Running   0          49s
$ oc get pod/example-infinispan-0 -o yaml
apiVersion: v1
kind: Pod
...
spec:
  containers:
  - args:
    - cloud.xml
    - -Djboss.default.jgroups.stack=dns-ping
    - -Djgroups.dns_ping.dns_query=example-infinispan-ping.myproject.svc.cluster.local
    env:
    - name: MGMT_USER
      value: Ns=UsiPUIgooJ.kr
    - name: MGMT_PASS
      value: xRqqMCpwgXNNwwTY
    - name: APP_USER
      value: FmQxggYuh/WB/PiY
    - name: APP_PASS
      value: hmvU/TcsguzNPGEa
...
```
some branded Infinispan images could use different variable names, i.e.:
```
oc get pod/example-infinispan-0 -o yaml
apiVersion: v1
kind: Pod
...
spec:
  containers:
  - env:
    - name: ADMIN_USER
      value: zrINkHGtR,EkKDaU
    - name: ADMIN_PASSWORD
      value: jdqdfPhsDRnuETL\
    - name: APPLICATION_USER
      value: sOqNzleAh=G=mrEM
    - name: APPLICATION_PASSWORD
      value: YDvE\MgjzqKUhJ-h
...
```

galderz:
Looks good to me, @Crumby any comments?

Crumby:
@galderz Do we need randomized default usernames? While randomized passwords are common practice, randomized usernames aren't. (Commonly defaults to 'admin' ('kubeadmin' for openshift/kubernetes installation)). It feels also more convinient to have reasonable default admin's/user's username for developer trying out the application as it's clear what are the authorization rights of the user. (That will be hard to tell with random username)

galderz:
@Crumby Good point. @rigazilla is away now, so I'll send a new PR with your suggested changes. Any thoughts on names? What about `developer`?

galderz:
I've fixed conflict and added `developer` / `admin` as default users for app and mgmt respectively. 

There's a further improvement we can do. Having passwords as plain in environment variables is going to be problematic. Instead, we should just generate a secret on the fly with the randomly generated password, and reference that from the env variable (e.g. see [here](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-container-environment-variables-using-secret-data)).

galderz:
This is superseded by #78.

PR #91, opened 2019-07-23
Fix status retrieving with enhanced image #89

rigazilla:
Operator by default runs the cli command `/opt/jboss/infinispan-server/bin/ispn-cli.sh`, for customized Infinispan image this value can be overridden setting the CLI_CMD operator environment variable.


galderz:
@rigazilla What's the value for enhanced image?

galderz:
Looks good, thx @rigazilla

Issue #92, opened 2019-07-24
Application and admin refer to same secret


galderz:
```
spec:
  containers:
  - env:
    - name: ADMIN_USER
      valueFrom:
        secretKeyRef:
          key: username
          name: connect-secret
    - name: ADMIN_PASSWORD
      valueFrom:
        secretKeyRef:
          key: password
          name: connect-secret
    - name: APPLICATION_USER
      valueFrom:
        secretKeyRef:
          key: username
          name: connect-secret
    - name: APPLICATION_PASSWORD
      valueFrom:
        secretKeyRef:
          key: password
          name: connect-secret
```

Admin username/password should be referencing `mgmt-secret`

PR #94, opened 2019-07-25
Fixed mgmt authentication plus tests #92

rigazilla:
Fixed credential retrieving for management authentication. Plus connection test to management via digest-md5 authentication

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -122,14 +122,14 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: infinispan.Name, Namespace: infinispan.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
 		// Define or locate application user secret
-		appSecret, err := r.secretForUser("developer", "app", infinispan, reqLogger)
+		appSecret, err := r.secretForUser("developer", "app", infinispan, infinispan.Spec.Connector.Authentication, reqLogger)
 		if err != nil {
 			reqLogger.Error(err, "could not find or create application Secret")
 			return reconcile.Result{}, err
 		}
 
 		// Define or locate management user secret
-		mgmtSecret, err := r.secretForUser("admin", "mgmt", infinispan, reqLogger)
+		mgmtSecret, err := r.secretForUser("admin", "mgmt", infinispan, infinispan.Spec.Management.Authentication, reqLogger)
 		if err != nil {
 			reqLogger.Error(err, "could not find or create management Secret")
 			return reconcile.Result{}, err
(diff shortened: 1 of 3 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -459,7 +459,24 @@ func (r *ReconcileInfinispan) secretForUser(user string, realm string, m *infini
 			return nil, err
 		}
 	}
+	return r.createSecretForAuthentication(user, realm, m, reqLogger)
+}
 
+func (r *ReconcileInfinispan) secretForMgmt(user string, realm string, m *infinispanv1.Infinispan, reqLogger logr.Logger) (*corev1.Secret, error) {
```
galderz: Hmmmm, seems to me that adding another method is a bit too much here. 

Couldn't we just make `secretForUser` take an extra parameter? For management use, the parameter value would be `m.Spec.Management.Authentication.SecretName`. For application user `m.Spec.Connector.Authentication.SecretName`?

test/e2e/main_test.go
```diff
@@ -204,20 +224,26 @@ func TestExternalServiceWithAuth(t *testing.T) {
...
+	okd.CreateRoute(Namespace, "cache-infinispan-0", 8080, "http")
 	defer okd.DeleteRoute(Namespace, "cache-infinispan-0-http")
 
+	okd.CreateRoute(Namespace, "cache-infinispan-0", 9990, "mgmt")
+	defer okd.DeleteRoute(Namespace, "cache-infinispan-0-mgmt")
+
 	client := &http.Client{}
 	hostAddr := okd.WaitForRoute(client, Namespace, "cache-infinispan-0-http", RouteTimeout, "connectorusr", "connectorpass")
 
+	hostAddrMgmt := okd.WaitForRoute(client, Namespace, "cache-infinispan-0-mgmt", RouteTimeout, "connectorusr", "connectorpass")
```
galderz: Is this right? Waiting for a management service route to be available shouldn't require connector user/password. If anything it should need management user/pass, but I feel this method is not really using these credentials? Or under which circumstances does it use it?

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I choosed that way because I don't like a lot the idea of passing `m` ad also a substruct of the same object `m.Spec...` in the same function call.

test/e2e/main_test.go
rigazilla: `checkExternalAddress` does a check wait basic authentication, I can't rememer why... need to check. I would go for simplifing this in a separated issue.

galderz: Make sure you track that.

rigazilla: https://github.com/infinispan/infinispan-operator/issues/95

pkg/controller/infinispan/infinispan_controller.go
rigazilla: ok, removed dup code

PR #96, opened 2019-07-26
Delete the created tag on dry run

rigazilla:
trivial fix to remove unwanted tag when dry running

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -40,6 +40,11 @@ tag() {
 }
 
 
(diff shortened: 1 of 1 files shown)


galderz:
I've included a fix for this in #110 and did it in a different way. You might want to inspect the tag after the dry run finishes, so instead I remove it before tagging. That means that branches/tags can be left around in the end but you can clean those up easily.

Issue #97, opened 2019-07-29
A way to propagate additional variables from operator to image


rigazilla:
Custom/enhanced images may need to propagate several environment variable to the server image. A generic way to handle this without having to modify the operator code is welcome.

Approach: ADDITIONAL_VARS env var should contain the variable names of all the vars that need to be propagated

PR #98, opened 2019-07-29
A way to pass additional env vars to the image #97

rigazilla:
User must define in the operator env the ADDITIONAL_VARS envvar containing the list of the env var names that the operator must pass to the image (as a json array of string). A variable in the list will be passed only if defined in the operator environment.


rigazilla:
trivial PR, merging

Issue #99, opened 2019-07-29
Simplify syntax for VOLUME_KEYSTORE_NAME, VOLUME_SECRET_NAME


rigazilla:
No need for json syntax since they just name

galderz:
Reverted in 0.x branch due to causing issues.

Issue #85, opened 2019-07-15
Add tests for enhanced images


galderz:
A bug was introduced in our enhanced image support (see #84). A test needs to be added to verify the minimal CR works with enhanced images.

galderz:
Reverted in 0.x branch due to causing issues.

PR #100, opened 2019-07-29
No need for json in VOLUME_*_NAME vars #99

rigazilla:


galderz:
@rigazilla The PR looks fine, but we really need #85 to test this. Can you address #85 first so that we can add a CI step that tests the operator testsuite with an enhanced image?

galderz:
Superseded by #106

PR #108, opened 2019-08-09
Formal Operator configuration design v3

galderz:
Version 3 of the operator configuration design here. Noteworthy changes:

General
--------

* Renamed `.spec.container.jvmOptionsAppend` to `.spec.container.extraJvmOpts`.
* Prometheus always enabled, for all profiles. So option for enable/disable it in `.spec.management` gone.
* Datasources gone, these are configured per cache, including their credentials.
* Renamed `Secured` profile to `Default` profile.
* Data grid/cache service selection cannot be combined, so reworked configuration so that you choose one or the other. Both cannot be configured at the same time.

Security
--------

* Reworked security with a secret containing all private information related to identities. This means that management specific authentication has been removed.
* `.spec.management` removed since no separate options are configured for it anymore.
* Secret defines identities that can be `credentials`, `certificate`, `token` or `oauth` type of authentication.
* Each identity can have a list of roles associated to it, except for oauth (roles are defined in keycloak).
* Security configuration added in operator which predefined list of roles.
* Data grid storage has been moved to be under `.spec.service.container.storage`. This has been done because storage is a per-container setting only for data grid service. Adding `.container` makes it clear that is per-container, and not a global storage option for all pods to share.
* Added role definitions in operator along with permissions for each role.
* Identities (except oauth and token) can optionally define the roles that are associated with this identity.

X-Site
------

* xsite static configuration has become url: `xsite://...` format, where the host:port combination define the static external host:port combination.
* xsite dynamic configuration has become url: `openshift://...` format, where host:port combinations define the dynamic nature of remote openshift instances. Operator will connect remotely and figure out what the external host:port is, and it will configure infinispan with that.
* Remote sites that use openshift scheme must have a secret defined. The secret can contain `credentials` and/or `token` to connect to openshift instances

Notes
------
* It's desirable for the operator to be configurable to expose external routes, for HTTP REST and Hot Rod endpoints. However, doing this is non-trivial. So, this is not considered in the formal specification at this stage. The user can always create (based on documentation) external routes,
so not having easy to configure external routes in operator is not a show stopper. More details to come.

### DIFF
--- a/doc/design.adoc
+++ b/doc/design.adoc
@@ -57,25 +57,19 @@ This document describes the types introduced by the Infinispan Operator to be co
 
 | `profile`
 | Profile in use. See <<infinispanprofiles,profiles>> for details.
-| `Secured` / `Performance` / `Development`
+| `Default` / `Performance` / `Development`
 | false
-| `Secured`
+| `Default`
 
-| `cache`
-| Cache configuration.
-| <<infinispancachespec>>
-| false
-|
-
-| `datagrid`
-| Datagrid configuration.
-| <<infinispandatagridspec>>
+| `service`
+| Service specific configuration.
+| <<infinispanservicespec>>
 | false
 |
 
-| `connector`
-| Connector configuration.
-| <<infinispanconnectorspec>>
+| `security`
+| Security configuration.
+| <<infinispansecurityspec>>
 | false
 |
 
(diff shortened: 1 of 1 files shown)


galderz:
For reference, link to formal specification v2 is #77.

galderz:
Further changes:

* Added role definitions in operator along with permissions for each role.
* Identities (except oauth and token) can optionally define the roles that are associated with this identity.

galderz:
@ryanemerson @tristantarrant @Crumby Infinispan 10 is already in master and so we'll be moving fast implementing this configuration. So, last chance saloon to review this ;)

rigazilla:
@galderz I would merge this if it's ok to you

galderz:
@rigazilla Yeah let's merge it

PR #109, opened 2019-08-09
React on changes to Connector, Management, Container #81

rigazilla:
Changes to the Spec.<Connector|Management|Container> are propagated to the StatefulSet and applied with a RollingUpgrade strategy.

This PR watches only the infinispan CR fields: i.e. secretName change is detected, but changes in the secret content aren't.

### DIFF
--- a/pkg/apis/infinispan/v1/infinispan_types.go
+++ b/pkg/apis/infinispan/v1/infinispan_types.go
@@ -49,8 +49,11 @@ type InfinispanCondition struct {
 // InfinispanStatus defines the observed state of Infinispan
 type InfinispanStatus struct {
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	Conditions      []InfinispanCondition `json:"conditions"`
-	StatefulSetName string                `json:"statefulSetName"`
+	Conditions      []InfinispanCondition    `json:"conditions"`
+	StatefulSetName string                   `json:"statefulSetName"`
+	Connector       InfinispanConnectorInfo  `json:"connector"`
+	Management      InfinispanManagementInfo `json:"management"`
+	Container       InfinispanContainerSpec  `json:"container"`
 }
 
 type Config struct {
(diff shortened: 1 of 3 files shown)


galderz:
@rigazilla Did you check the operator framework documentation on how to achieve this? The code in the PR looks more complicated than what is suggested in their [docu](https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#resources-watched-by-the-controller).

galderz:
@rigazilla Seems to me we just need to take each and every resource we create and pass it to `Reconciler.Watch` method and that should do it.

rigazilla:
@galderz the `Controller.Watch` just enables the controller to receive events for a specific kind of objects. The actual reconciliation logic is application specific and needs to implemented.
The approach in this PR is:
- add to Status all the Spec structs that the user can change
- on every event
    - if Spec.field1 != Status.field1 do rec logic and update the Status

 Which is a simple implementation of what explained for standard k8s objects in [spec-ad-status](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/#object-spec-and-status).

The reconciliation logic is obsolete after the latest refactoring, but we can discuss if we want to keep the .Spec .Status match approach.

galderz:
@rigazilla Makes sense, can you please rebase the PR?

galderz:
Btw, I think I know how we can test this in an efficient way. Instead of having each test start/stop a cluster, we could do this: start the testsuite with the minimal cluster deployed, then in the cluster formation tests, you take the spec and change num nodes to 2, then run the tests that we do in that test, then drop it to 1, then add custom secret...etc.

We'd essentially do the same testing we do today, but you'd be testing also that the reconciler detects changes.

rigazilla:
> @rigazilla Makes sense, can you please rebase the PR?

It can't be rebased, needs to be rewritten on top of the 1.0 design. I'll look into

rigazilla:
@galderz  new PR is #156. Closing this

PR #115, opened 2019-08-22
Integrating with Infinispan 10

galderz:
Initial PR. 

All should be fine except for `infinispan_cli_helper`, `k8s_client` and `operator_helper` go classes. These need to be better modularized so that they don't become dumping grounds for utility methods for anything.


pkg/controller/infinispan/infinispan_controller_test.go
```diff
@@ -42,20 +42,20 @@ type mockCluster struct{}
 // GetClusterMembers returns a fake cluster view, produced returning the substring
 // after the - char of the `name` arg. If the substring doesn't start with View and error
 // will be also returned
-func (m mockCluster) GetClusterMembers(ns, name string) (string, error) {
+func (m mockCluster) GetClusterMembers(ns, name, secretName string) ([]string, error) {
```
Crumby: What's the point of `secretName` if it's not used in the function?

galderz: This file contains a test that mocks Kubernetes, it's a unit test. Hence, secretName is unused, but it's used in the production version, where the secret is used to query the operator password and do the curl to obtain the cluster health.

galderz: Btw, I've changed this now to use simply use a mock function when testing.

galderz:
Of course, once CR1 is out and we have a proper image location I'll update the code.

galderz:
Pushed a commit to re-enable Minikube runs on CI, since Infinispan 10 can run with much less resources.

rigazilla:
> 
>     * I would prefer to see `Identities,Credentials...` types around instead of unstructured string with yaml inside... but I see that only yaml is need atm
> 

i.e. [CreateIdentities](https://github.com/infinispan/infinispan-operator/blob/34acdb1c56b0f839f7822ac7d22909f053026c0c/pkg/controller/infinispan/util/security.go#L25) returns a string containing a yaml structure, wouldn't be better to return an [Identities](https://github.com/infinispan/infinispan-operator/blob/34acdb1c56b0f839f7822ac7d22909f053026c0c/pkg/controller/infinispan/util/security.go#L16)? Same for the other funcs that returns string with yaml

rigazilla:
@galderz as discussed via chat I'm going to merge this PR so we can go on with devs. We can fix things in separated issues if needed.

rigazilla:
merged, thank you!

PR #116, opened 2019-08-22
85/enhanced tests

rigazilla:
Misc PR with:

- fix for VOLUME_* env vars parsing
- tests for enhanced image
- get service-certs secrets from OS if service is available


galderz:
@rigazilla I think there might be a problem with the travis yaml descriptor because no CI runs appear, can you double check?

rigazilla:
It seems it just needed a kick. should be ok @galderz

galderz:
Thx @rigazilla :)

Issue #122, opened 2019-08-27
Refactor .spec.connector.authentication to .spec.security


galderz:
Following proposals in configuration [v3](https://github.com/infinispan/infinispan-operator/pull/108).

Mostly it's just a matter of taking the secret name and setting it to `.spec.security.endpointSecret`. Then, remove any leftover.

Issue #121, opened 2019-08-27
Make security and configuration functions return typed instances


galderz:
From the PR comments for #111:

> I would prefer to see `Identities`, `Credentials`... types around instead of unstructured string with yaml inside... but I see that only yaml is need atm.

> i.e. CreateIdentities returns a string containing a yaml structure, wouldn't be better to return an Identities? Same for the other funcs that returns string with yaml

PR #140, opened 2019-08-30
Post Infinispan 10 integration tasks

### LINKED ISSUE #120: Address production code Go linter warnings
Related to the comments in the PR for #111.

galderz:
Resolve #120, #121 and #122 

Also refactored all Kubernetes code to use just one client API: the one provided by the controller. This way, we can load up the Kubernetes helper method from either the controller manager, or external configuration.

The controller API is less typesafe, which might be harder to use compared to the typed one, but it also offers more possibilities for further simplification of the API, since a lot of calls are very similar and only differ in parameters.

Documentation for the security changes will be added in a later commit (either in this PR or a separate one)


galderz:
Fixing CI...

galderz:
Added documentation for security changes (cc @oraNod), and refactored CR examples to highlight each independent set of examples.

galderz:
Forced pushed branch to fix issue that was making CI fail

galderz:
p.s. Even if the server image is both on quay.io and docker, I've decided to use docker since that's more reliable. Quay.io was unavailable at times during the development of this PR :\

galderz:
The issue I'm having with creating CRDs using the controller runtime API might be related to #321. I'm exploring alternatives proposed there. Otherwise can always revert to the previous method for creating/instantiating CRDs.

galderz:
That worked, I'll tidy up the commits.

galderz:
Squashed CI fix to the refactoring commit. Let's see if CI agrees :)

galderz:
CI looks good @rigazilla

rigazilla:
@galderz ok. Testing locally

rigazilla:
tests are ok locally in all the k8s flavours: minikube, os 3.11, os 4.1.
Quick reviewing the code now

pkg/controller/infinispan/util/cluster.go
```diff
@@ -8,10 +8,23 @@ import (
...
+
+func NewCluster(kubernetes *Kubernetes) *Cluster {
+	return &Cluster{Kubernetes: kubernetes}
+}
+
+type ClusterInterface interface {
+	GetClusterMembers(secretName, podName, namespace string) ([]string, error)
+}
+
+// ClusterMembers is a lambda function for getting cluster members from server
```
rigazilla: to be deleted?

pkg/controller/infinispan/util/yaml.go
```diff
@@ -0,0 +1,15 @@
+package util
+
+import (
+	"gopkg.in/yaml.v2"
+)
+
+// ToYaml converts a type instance into serialized yaml
+func ToYaml(in interface{}) ([]byte, error) {
```
rigazilla: Can't we just call yaml funcs directly?

galderz: Hmm, maybe... 

I was trying to isolate the use of them in case we end up using a different third party library at some point (e.g. go-yaml 3). I think it provides some good encapsulation, but I can remove it if it's unclear.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -360,36 +369,35 @@ func (r *ReconcileInfinispan) configMapForInfinispan(m *infinispanv1.Infinispan)
...
+			return nil, err
 		}
 
-		return string(secret.Data["identities.yaml"]), nil
+		return secret.Data["identities.yaml"], nil
 	}
 
-	identities, err := ispnutil.CreateIdentities()
+	identities := ispnutil.CreateIdentities()
+	data, err := ispnutil.ToYaml(&identities)
```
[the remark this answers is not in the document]
galderz: Hmmm, true, should probably not be necessary since converting to yaml should not need to modify the object of origin.

galderz:
@rigazilla Fixed you comments and pushed. Let's see CI...

rigazilla:
thank you!

PR #141, opened 2019-09-03
ISPN-10554 aligning with Infinispan doc set

oraNod:
https://issues.jboss.org/browse/ISPN-10554


documentation/asciidoc/topics/attributes/product-attributes.adoc
```diff
@@ -1,77 +1,111 @@
...
+:underscores: ___
+
+//
+// Community attributes
+// In the main Infinispan repo, these attributes are defined in
+// documentation/pom.xml
+//
+:brandname: Data Grid
+:fullbrandname: Red Hat Data Grid
+:wildflybrandname: EAP
```
galderz: Can these wildfly references be removed? They're not relevant nor to the server nor the operator any more.

oraNod: @galderz I need to investigate before removing. This wildfly attribute is in the community docs as well. 

https://github.com/infinispan/infinispan/blob/68170ad67914ac29ee3194c3e9e7c2e006c6460d/documentation/pom.xml#L170 

And here is an example of the attribute in the docs: https://github.com/infinispan/infinispan/blob/master/documentation/src/main/asciidoc/topics/integrations.adoc#server-mode 

I'd much rather remove this attribute as part of a larger effort to scrub wildfly/eap references across the entire doc set. I've created ISPN-10564 to handle that.

oraNod: @galderz ignore the previous comment. I was thinking about the main Infinispan repo. I've removed those wildfly attributes. cheers.

PR #143, opened 2019-09-04
Fix image version to 9.x for 0.x operator

rigazilla:
0.x operator can't run with `latest` server image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ jobs:
         - make unit-test
     - stage: "Test"
       env:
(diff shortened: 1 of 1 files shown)


galderz:
@rigazilla Hmmm, rather than just fix CI, we should fix all the references to `latest` to the latest 9.x release, see my alternative PR. We don't have a moving `9.x` docker tag, but the latest 9.4.x release is just fine here since we won't be making many more updates on this branch.

rigazilla:
problem has been fixed by #144

PR #147, opened 2019-09-05
fixing cr urls and adding tech preview changes for product

oraNod:


galderz:
@oraNod Looks good. One more thing that should be added. Since the complexity of the secret structure has increased, I added the `connect_secret.yaml` example. Could you reference it in the docu and explain that it needs to be applied before applying `cr_minimal_with_auth.yaml`? Thanks!

galderz:
Thx for changing that @oraNod :)

Issue #131, opened 2019-08-27
Add connector encryption


galderz:

rigazilla:
To enable connector encryption the Infinispan server needs TLS certificates.
Three ways to do this:

1. self generation (unsure and only for devs)
2. user provided
3. external or infrastructure service

### Possible approaches
#### 1. Self generation
User can select self generated certs when deploying an infinispan CRD. This needs:

- appropriate field in the CRD
- appropriate field the infinispan.yaml image configuration
- the conf-generator tool must add the xml fragment that configure self-generation in the infinispan.xml

#### 2. User provided
The Infinispan image can be configured to use either a keystore or a key,cert pair. The operator must provide these files to the image. This needs:

- user must create secret with keystore or key,cert inside
- appropriate field in the CRD that refers to secret
- the operator must mount the secret as folder (/etc/encrypt)
- the operator must pass the path to the keystore to the Infinispan image

#### 3. External
A general solution for the integration with a cert service provider is TODO.
Specific solutions are strongly platform dependent.
##### Openshift
If the certificates service is available Openshift can provide a secret with key,cert pair in it. This way needs:
- the operator must configure the service in the Infinispan CRD and mount the served secret in the appropriate folder
- the operator must pass the path to the cert,key folder to the Infinispan image

ryanemerson:
@**Vittorio Rigamonti** Overall LGTM. I would say the self-signed certificates is more of nice to have for now, but 2 & 3 are essential

rigazilla:
Ok, working plan is: 2, 3 and 1

Issue #132, opened 2019-08-27
Complete default profile


galderz:
Default profile should enable connector authentication and encryption, and cluster authentication and encryption.

Connector authentication is already in place. Connector encryption will come from #131. Cluster authentication support will come with #129, and cluster encryption with #130.

galderz:
Although connector encryption has been added, this is not enabled by default yet. This should come as part of #190.

galderz:
With #190 included, this is now complete.

PR #154, opened 2019-09-11
Enabling connector encryption #131

rigazilla:
This enables connector encryption for methods 2 and 3 of issue #131 
There's a new deploy/cr/tls forlder with secret and cr for a tryout:
```
oc apply -f deploy/cr/tls/tls_secret.p12
oc apply -f deploy/cr/tls/cr_minimal_with_tls.yaml
```
yaml are commented for all the possible configurations


deploy/cr/tls/cr_minimal_with_tls.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: infinispan.org/v1
+kind: Infinispan
+metadata:
+  name: example-infinispan
+spec:
+  replicas: 2
+  security:
+    endpointEncryption:
```
galderz: This configuration makes it hard to choose one or the other option. What happens if someone puts both `certSecretName` and `certService`?

I think this would benefit from a `endpointEncryption.type` attribute that defines which type of encryption is chosen, and based on that we can validate that the right properties are set. It's the same we're doing with `spec.service` where we'll have either cache or datagrid...etc.

rigazilla: Yaml comments explain the policy (with a typo actually). The choice is implicit: OS service if certService not null else certSecretName is used.
But I think it's OK to add the type field and uniform encryption to spec.service

rigazilla: @galderz added a test for cluster formation with TLS

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -354,19 +360,101 @@ func (r *ReconcileInfinispan) deploymentForInfinispan(m *infinispanv1.Infinispan
...
+	}
+	return nil
+}
+
 func (r *ReconcileInfinispan) configMapForInfinispan(m *infinispanv1.Infinispan) (*corev1.ConfigMap, error) {
 	name := m.ObjectMeta.Name
 	namespace := m.ObjectMeta.Namespace
 
 	config := ispnutil.CreateInfinispanConfiguration(name, namespace)
+	setupConfigForEncryption(m, &config, r.client)
```
galderz: Unhandled error

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -354,19 +360,101 @@ func (r *ReconcileInfinispan) deploymentForInfinispan(m *infinispanv1.Infinispan
...
+			secretName := getEncryptionSecretName(m)
+			if ser.ObjectMeta.Annotations == nil {
+				ser.ObjectMeta.Annotations = map[string]string{}
+			}
+			ser.ObjectMeta.Annotations[ee.CertService] = secretName
+		}
+	}
+}
+
+func setupVolumesForEncryption(m *infinispanv1.Infinispan, dep *appsv1beta1.StatefulSet) error {
```
galderz: Why does this function return error? It always returns nil. Unless a non-nil error can be thrown, I'd make it a void function.

galderz:
Thx for adding the test btw!

rigazilla:
> Small comments about error handling.
> 
> One more thing, remember what we agreed [here](https://github.com/galderz/infinispan-operator/blob/t_development_profile_128/doc/design.adoc#profiles). In particular, in the default profile (or no profile defined), we'd want connector encryption enabled by default. This might require self-signed certificates in vanilla Kubernetes, but in an OpenShift environment we have the certificate authority? FYI, this is being tracked by #132

With the default profile on Openshift we can rely on the [serving certificates service](https://docs.openshift.com/container-platform/4.1/authentication/certificates/service-serving-certificate.html#understanding-service-serving_service-serving-certificate). This can be done just setting `EncryptionEndpoint.certService=service.beta.openshift.io/serving-cert-secret-name`

pkg/controller/infinispan/infinispan_controller.go
rigazilla: ok

deploy/cr/tls/cr_minimal_with_tls.yaml
oraNod: @rigazilla having `type: service` commented out means community users have to manually edit the yaml instead of doing `oc apply -f github/raw/path/to/cr_minimal_with_tls.yaml`

what if we broke this into two files? 

`cr_minimal_service_encryption.yaml` - uses `type: service` 
`cr_minimal_custom_encryption.yaml` - uses `type: secret`

galderz:
> With the default profile on Openshift we can rely on the serving certificates service. This can be done just setting EncryptionEndpoint.certService=service.beta.openshift.io/serving-cert-secret-name.

It would be nice if the operator could detect if it's running on OpenShift or vanilla Kubernetes, to decide what to do when user sets no encryption and we have to enable it by default.

Great work @rigazilla!

PR #156, opened 2019-09-13
Update on Credential secretname change #155

### LINKED ISSUE #155: React on Identities (endpoint auth) changes
Relates to #81 (subset)

rigazilla:
This PR enables rolling update (restart) if identities (security.endpointSecret) changes
Partially fixes #155 : doesn't detect secret content change atm

Replaces #109

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -2,6 +2,7 @@ package e2e
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"github.com/infinispan/infinispan-operator/pkg/controller/infinispan/util"
 	testutil "github.com/infinispan/infinispan-operator/test/e2e/util"
@@ -16,6 +17,7 @@ import (
 	ispnv1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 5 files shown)


rigazilla:
@galderz added auth test with credentials change

galderz:
@rigazilla I don't think there's a need to add yet another test. You could just expand `TestExternalServiceWithAuth` so that at the end it tries to modify the auth config...etc. Each test creates a new Infinispan operator instance (for the time being), so the quicker it runs the better.

At some point it'd be nice to have a single operator instance start up before all tests, and then run each test where you scale up to 2 nodes, add security...etc. I think this would both save execution time and would test both the logic and the capabilities of the operator to handle changes.

test/e2e/main_test.go
```diff
@@ -224,6 +226,101 @@ func TestExternalServiceWithAuth(t *testing.T) {
...
+	er := kubernetes.Kubernetes.Client.Update(context.TODO(), &spec)
+	if er != nil {
+		panic(fmt.Errorf(err.Error()))
+	}
+	kubernetes.WaitForPods("app=infinispan-pod", 1, SinglePodTimeout, Namespace)
+
+	// Test authentication
+	routeName := fmt.Sprintf("%s-external", name)
+	client := &http.Client{}
+	hostAddr := kubernetes.WaitForExternalService(routeName, RouteTimeout, client, Namespace)
```
galderz: CI failed here with Minikube. Can you have a look?

galderz: Run: https://travis-ci.org/infinispan/infinispan-operator/jobs/588876584

galderz: To be precise, it got stuck within there.

rigazilla:
> @rigazilla I don't think there's a need to add yet another test. You could just expand `TestExternalServiceWithAuth` so that at the end it tries to modify the auth config...etc. Each test creates a new Infinispan operator instance (for the time being), so the quicker it runs the better.
> 
> At some point it'd be nice to have a single operator instance start up before all tests, and then run each test where you scale up to 2 nodes, add security...etc. I think this would both save execution time and would test both the logic and the capabilities of the operator to handle changes.

ok done @galderz 
Tried on local minikube and test are fine.
Let's see the new run on ci... tomorrow :)

rigazilla:
Minikube test is still failing on CI. Test seems to run until the end, so it must be something in the clean up.
Looking into

rigazilla:
@galderz reset timeout to default and squashed history

PR #159, opened 2019-09-23
Cekit build environment #158

rigazilla:
This enable the operator build with Cekit.
Just go under build/cekit and run:
`cekit build docker`

### DIFF
--- a/build/cekit/image.yaml
+++ b/build/cekit/image.yaml
@@ -1,0 +1,30 @@
+- name: "builder-infinispan-operator"
+  description: "Builder for Infinispan Operator"
+  version: latest
+  from: "fedora:29"
+  modules:
+    repositories:
+      - path: modules
+        name: modules
+    install:
+      - name: builder-infinispan-operator
+
+- name: "jboss/infinispan-operator"
+  description: "Infinispan Operator"
+  version: latest
+  from: "registry.access.redhat.com/ubi8/ubi-minimal"
+  labels:
+    - name: "name"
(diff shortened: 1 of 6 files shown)


build/cekit/image.yaml
```diff
@@ -0,0 +1,50 @@
+- name: "builder-infinispan-operator"
+  description: "Builder for Infinispan Operator"
+  version: latest
+  from: "fedora:29"
+  labels:
+    - name: "maintainer"
+      value: "jboss-support-dg6@redhat.com"
```
galderz: Most of this values are product specific. I'd leave those out for the community image and add them in the product overrides, is that possible?

build/cekit/image.yaml
```diff
@@ -0,0 +1,50 @@
...
+      value: "datagrid,java,jboss,xpaas,operator"
+    - name: "com.redhat.delivery.appregistry"
+      value: "true"
+  modules:
+    repositories:
+      - path: modules
+        name: modules
+    install:
+      - name: builder-infinispan-operator
+        #osbs:
```
galderz: You can remove commented sections from here. They should be added in product overrides.

build/cekit/image.yaml
rigazilla: yes sure, done

build/cekit/image.yaml
rigazilla: done. thanks @galderz !

Issue #128, opened 2019-08-27
Add development profile


galderz:
Add profile configuration to operator. 

Using `Development` profile should deactivate connector authentication.

galderz:
Initial work done [here](https://github.com/galderz/infinispan-operator/tree/t_development_profile_128), but some work is required in the image so that we can instruct the image to not generated a user/pass if we don't pass identities path. See [Zulip discussion](https://infinispan.zulipchat.com/#narrow/stream/185835-infinispan-cloud/topic/Operator.20.26.20Infinispan.2010/near/175302130).

ryanemerson:
Required image work:
https://github.com/infinispan/infinispan-images/issues/10
https://github.com/infinispan/infinispan-images/issues/16

PR #172, opened 2019-10-08
Self signed cert handling #131

rigazilla:
This enables the generation of self signed certificates by Infinispan.
To have self signed certificates user must:
- set spec.profile = Development
- not provide any certificates, keystore nor platform service

This must go after:
 https://github.com/infinispan/infinispan-image-artifacts/pull/1

Relates to:
https://github.com/infinispan/infinispan-images/issues/15
https://github.com/infinispan/infinispan-operator/issues/131

### DIFF
--- a/pkg/apis/infinispan/v1/infinispan_types.go
+++ b/pkg/apis/infinispan/v1/infinispan_types.go
@@ -35,6 +35,7 @@ type InfinispanSpec struct {
 	Image     string                  `json:"image"`
 	Security  InfinispanSecurity      `json:"security"`
 	Container InfinispanContainerSpec `json:"container"`
+	Profile   string                  `json:"profile"`
 }
 
 // InfinispanCondition define a condition of the cluster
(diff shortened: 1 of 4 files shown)


galderz:
Why are self signed certificates required for development profile? With development profile I assumed connectivity would not be encrypted, but maybe that's not possible with the new single port set up which runs on top of HTTP/2?

Btw, I had been working in parallel on development profile myself (#128 is assigned to me), so I had expanded development support to not generate developer user/pass and was still missing a way for the image to instruct it not to need authentication... Are you taking care of this now? Or shall I?

galderz:
See my ongoing development profile work in the last 2 commits of [this branch](https://github.com/galderz/infinispan-operator/commits/t_development_profile_128).

rigazilla:
> Why are self signed certificates required for development profile? With development profile I assumed connectivity would not be encrypted, but maybe that's not possible with the new single port set up which runs on top of HTTP/2?
My fault, self signed should be and additional option only for Development. Fixed in the last commit. Thanks!

> 
> Btw, I had been working in parallel on development profile myself (#128 is assigned to me), so I had expanded development support to not generate developer user/pass and was still missing a way for the image to instruct it not to need authentication... Are you taking care of this now? Or shall I?

Not in this PR, I can open another PR for authentication

rigazilla:
replaced by #233

Issue #126, opened 2019-08-27
Add local persistent volume storage


galderz:
Implement local persistent volume storage support. As well as adding the corresponding configuration (see formal specification [v3](https://github.com/infinispan/infinispan-operator/pull/108)), it also requires porting the tests in the openshift template for verifying that: permanent cache definitions survive restarts, and both data and query results also survive (integration testsuite #124).

galderz:
Requires datagrid service configuration to be in place #135.

PR #176, opened 2019-10-10
Persistent volume #126

rigazilla:
Persistent volume implemented here + tests.


galderz:
@rigazilla Persistent volume information should go under `.spec.service` and should only be used when Data Grid service is used. The design doc has an example of this already:

```
spec:
  image: jboss/infinispan-server:latest
  replicas: 6
  profile: Performance
  service:
    type: Data Grid
    container:
      storage: 2Gi
```

And the [design doc](https://github.com/infinispan/infinispan-operator/blob/master/doc/design.adoc#infinispanservicecontainerspec) already specifies how this should be configured. Any reason to do it in a different way? Or just an oversight?

galderz:
The rest looks good, nice tests :)

rigazilla:
> ```
> 
> ```
> 
> 
> And the [design doc](https://github.com/infinispan/infinispan-operator/blob/master/doc/design.adoc#infinispanservicecontainerspec) already specifies how this should be configured. Any reason to do it in a different way? Or just an oversight?

The need to have persistence (at least for permanent caches) seems quite general, to we really want it to be enable only for Data Grid?

galderz:
> The need to have persistence (at least for permanent caches) seems quite general, do we really want it to be enable only for Data Grid?

We discussed this during the 7.3 development. Cache service does of course need persistence for permanent cache definitions...etc, but this is only for metadata and the persistence size of this should not be configurable by the user. If you look at the 7.3 cache service template, you won't see an option to configure the persistence size, even if we test permanent caches with cache service.

In data grid service, persistence is used for actual cached data and there we do want to have it configurable. In 7.3 data grid service template you'll see an option to configure persistence size.

rigazilla:
> We discussed this during the 7.3 development. Cache service does of course need persistence for permanent cache definitions...etc, but this is only for metadata and the persistence size of this should not be configurable by the user. If you look at the 7.3 cache service template, you won't see an option to configure the persistence size, even if we test permanent caches with cache service.
> 

uhm ok, so your idea is: always create a persistent volume with predefined size for metadata. Right?

galderz:
> uhm ok, so your idea is: always create a persistent volume with predefined size for metadata. Right?

Yeah, for cache service we need that. For now you can just do it by default and that's it. Then, if using datagrid service you just configure the size of it.

rigazilla:
@galderz changed according design.

PR #183, opened 2019-10-23
Use Kubernetes-in-docker (kind) in CI

galderz:
* Use custom kind configuration so that port forward can be enabled,
to enable external services to be acessible.
* Custom kind configuration exposes nodeport 30222 on port 11222.
* If running kind, check external services via 127.0.0.1:11222.
* Disable OpenShift (dind) 4 testing.
* Remove CPU/memory limitations for Infinispan with kind.
* No longer running a VM, so such strict Infinispan server pod limitations
should not be needed.
* Fix check address so that Infinispan 10 endpoints are checked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: go
 
+go:
+  - '1.11.x'
+
 sudo: required
 
 services:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -62,27 +49,20 @@ jobs:
...
-#        - kubectl cluster-info
-#        - df -h
-#        - sudo minikube status
-#        - sudo minikube logs
-#        - docker ps -a
-#        - for log in $(docker ps -qa | xargs); do docker logs --tail 500 $log; done
-#        - sudo minikube stop
+
+    - stage: "Test"
+      name: "Minikube"
```
gustavonalle: Why "minikube" ?

galderz: Leftover from previous configuration, let me fix that

rigazilla:
merged, thanks @galderz !

PR #195, opened 2019-10-29
Integration test suite base

Crumby:
Java test-suite holding integration tests with OpenShift. So far two tests are present:

- Default Authentication test
- Custom Authentication test

### DIFF
--- a/test-integration/operator-tests/src/test/java/org/infinispan/operator/security/AuthTest.java
+++ b/test-integration/operator-tests/src/test/java/org/infinispan/operator/security/AuthTest.java
@@ -1,0 +1,89 @@
+package org.infinispan.operator.security;
+
+import java.io.IOException;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
+import org.apache.http.entity.ContentType;
+import org.assertj.core.api.Assertions;
+import org.infinispan.TestServer;
+import org.junit.jupiter.api.Test;
+
+import cz.xtf.client.Http;
+import cz.xtf.core.openshift.OpenShift;
+import cz.xtf.core.openshift.OpenShifts;
+
+abstract class AuthTest {
+   static final OpenShift openShift = OpenShifts.master();
+
+   static final TestServer testServer = TestServer.get();
+   static final String testServerHost = testServer.host();
+
+   static String appName;
+   static String hostName;
+
+   static String user;
+   static String pass;
+
+   @Test
+   void restCorrectCredentials() throws IOException {
+      String putUrl = "http://" + hostName + "/rest/testcache/correct";
+      String getUrl = "http://" + hostName + "/rest/testcache/correct";
+
+      int putCode = Http.put(putUrl).basicAuth(user, pass).data("credentials", ContentType.TEXT_PLAIN).execute().code();
+      int getCode = Http.get(getUrl).basicAuth(user, pass).execute().code();
+
+      Assertions.assertThat(putCode).isEqualTo(204);
+      Assertions.assertThat(getCode).isEqualTo(200);
+   }
+
+   @Test
+   void restInvalidPassword() throws IOException {
+      String putUrl = "http://" + hostName + "/rest/testcache/invalid";
+
+      int putCode = Http.put(putUrl).basicAuth(user, "invalid").data("credentials", ContentType.TEXT_PLAIN).execute().code();
+
+      Assertions.assertThat(putCode).isEqualTo(401);
+   }
+
+   @Test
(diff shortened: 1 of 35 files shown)


test-integration/operator-tests/src/test/java/org/infinispan/operator/security/CustomAuthTest.java
```diff
@@ -0,0 +1,63 @@
...
+      host = openShift.generateHostname(appName);
+
+      Secret secret = new SecretBuilder("connect-secret").addData("identities.yaml", Paths.get("src/test/resources/secrets/identities.yaml")).build();
+      Route route = new RouteBuilder(appName).forService(appName).targetPort(11222).exposedAsHost(host).build();
+
+      openShift.secrets().create(secret);
+      openShift.customResource(infinispanContext).create(openShift.getNamespace(), infinispan.getInputStream());
+      openShift.createRoute(route);
+
+      openShift.waiters().areExactlyNPodsReady(2, "clusterName", appName).timeout(TimeUnit.MINUTES, 5).waitFor();
```
galderz: This is waiting for 2 pods to be ready, right? Which 2 pods? 1 for the operator and 1 for infinispan server? Or 2 for infinispan server? I think it's the former but I'm unsure...

test-integration/operator-tests/src/test/java/org/infinispan/operator/security/AuthTest.java
```diff
@@ -0,0 +1,72 @@
...
+      int putCode = Http.put("http://" + host + "/rest/testcache/correct").basicAuth(user, pass).data("credentials", ContentType.TEXT_PLAIN).execute().code();
+      int getCode = Http.get("http://" + host + "/rest/testcache/correct").basicAuth(user, pass).execute().code();
+
+      Assertions.assertThat(putCode).isEqualTo(204);
+      Assertions.assertThat(getCode).isEqualTo(200);
+   }
+
+   @Test
+   void restInvalidPassword() throws IOException {
+      Assertions.assertThat(Http.put("http://" + host + "/rest/testcache/invalid").basicAuth(user, "invalid").data("credentials", ContentType.TEXT_PLAIN).execute().code()).isEqualTo(401);
```
galderz: Could you make these lines a bit narrower? Easier to review the PR without having to scroll ;)

test-integration/operator-tests/src/test/java/org/infinispan/operator/security/CustomAuthTest.java
Crumby: It's waiting for two Infinispan pods. Test suite assumes that operator is already up and running in different namespace before test execution. Would be nicer to make this waiting function part of the `Infinispan` so it deduces the amount of pod that it should wait on from the CR but writing more tests is higher priority right now.

test-integration/operator-tests/src/test/java/org/infinispan/operator/security/AuthTest.java
Crumby: I'll see what I can do with that :)

test-integration/operator-tests/src/test/java/org/infinispan/operator/security/CustomAuthTest.java
galderz: Rather than waiting for 2 pods, could you wait for a cluster of 2 pods to be formed? That's what `wellFormed` status is about. It verifies that a cluster of N nodes (as defined in `.spec.replicas`) has been created. That's a better condition to wait for.

rigazilla:
@Crumby can I run the tests locally? I need to setup the env first?

test-integration/operator-tests/src/test/java/org/infinispan/operator/security/CustomAuthTest.java
Crumby: That's a good idea. This way at least we verify that cluster forms in all tested cases of deployment.

Crumby:
@rigazilla Sure you can. Just execute eg.:`mvn clean test` and pass:
`xtf.openshift.url`
`xtf.openshift.master.username`
`xtf.openshift.master.password`

You can pass `token` instead `username` and `password`. Also you can create a `test.properties` file with this properties in root directory and tests will catch that up.

Crumby:
@rigazilla I forgot on `xtf.openshift.namespace` to set up a namespace used for running the tests

galderz:
@rigazilla Are you Ok with this now?

galderz:
Great work @Crumby! :)))

PR #196, opened 2019-10-29
design doc updates for encryption

rigazilla:
update with encryption fields

### DIFF
--- a/doc/design.adoc
+++ b/doc/design.adoc
@@ -231,21 +231,23 @@ The format of the metadata in the secret can be found <<identities,here>>.
 | Field | Description | Scheme | Required | Default
 
 | `type`
-| Certificates provider type
+| Certificates provider type: `service` if the user wants to use a platform serving
+certificates service. `secret` if the user provides a secret with certs inside
 | `service`/`secret`
-| false
+| true
 |
 
 | `certServiceName`
-| Name of the certs service provider
+| Name of the serving certificates service. Only `service.beta.infinispan.io` is
+currently supported
 | string
 | false
 |
 
 | `certSecretName`
-| Name of the secret containing the certificates
+| Name of the secret containing the certificates (both for service and secret type)
 | string
-| false
+| true
 |
 
 |=======================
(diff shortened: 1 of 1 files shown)


Crumby:
Can we extend `Full Cache/DataGrid example` with the encryption, one with the service type and second with the secret type? Aside from that several things are not quite clear to me. Is some format expected from the secret? What should I paste as value for `certServiceName` to use OpenShift certs serving service?

galderz:
I agree with @Crumby, it'd be good to have a full example of encryption. I also agree with that if you know of any example values for these attributes, you should document them.

rigazilla:
@galderz @Crumby I've added some more info

galderz:
Thx @rigazilla for the extra details.

I think it looks ok in general, but wanted to make a couple of notes:

First, for consistency I'd make `Service` and `Secret` in `.spec.endpointEncryption` start with capital letters. We do that with `.spec.service.type` and you also see it with Kubernetes Service types.

Second, now that I see the design, I see something that could be improved in the future:

Currently as a user, seems like you have to have 2 decisions: Decide on value for `.spec.endpointEncryption.type`, and if type `secret` then you need to decide what to store in the secret (keystore or key+crt)...

Aside from the split decision, you have validation issues with this approach: what happens if the secret contains a `keystore.p12` field and only `tls.crt` fields? Do you take the keystore and ignore the crt? Or do you fail saying that `tls.key` field is missing? Either choice would be valid

Instead I would have merged the two decisions into one 3-way decision, e.g.

1. If `.spec.endpointEncryption.type=SecretKeystore`, then create a secret and add a `.p12` formatted keystore into the secret in field x.
2. If `.spec.endpointEncryption.type=SecretKey`, then create a secret and add a `.key` and `.crt` fields x and y.
3. If `.spec.endpointEncryption.type=Service`...

I think this would be more neat from a user perspective, and makes it easier to validate expectations. If the type is `SecretKeystore` you know exactly what to expect in the secret and you can safely ignore any fields other than `keystore.p12`

I've followed a similar approach x-site, where the contents of the secrets depend on the type of remote x-site master to which it talks to.

galderz:
There is another alternative, to only support `keystore.p12`. This would be a good starting point, and if people demand for it, add crt+key support. I think this would simplify the design further. How easy would be for the user to create a `kesytore.p12` out of crt+key?

galderz:
Create a separate issue to fine tune this #215.

Issue #181, opened 2019-10-17
Configure external service


galderz:
Add configuration to allow external service to be enabled. By default, for security reasons, it should be disabled but it should be easy to just enable it. The external service has been available since #127.

Also, it should be possible to select whether to create `NodePort`, `LoadBalancer`...etc. A similar thing is done with the [external service created for x-site](https://github.com/infinispan/infinispan-operator/blob/master/doc/design.adoc#infinispanlocalsitespec).

PR #198, opened 2019-11-01
X-Site OpenShift support and Logging categories

galderz:
Added support for `LoadBalancer` type x-site service and ability to communicate with OpenShift clusters.

There's also a manual example for OpenShift set ups that work with LoadBalancer.

I've also added logging configuration to operator.

Documentation for both x-site and logging still needs to be added. I'll coordinate with @oraNod for that.

**NOTE:** X-Site OpenShift example does not yet support creating caches and storing data, because external service is not yet configurable. That'll be added in #181.


galderz:
@rigazilla One more thing to do in this PR. In similar vain to what we did with encryption, we should rename `secret` to `secretName` in remote site backup, so that it's consistent. I'll add now.

galderz:
Added commit to rename `.secret` to `.secretName`, and updated design document with minor changes done during development.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -924,38 +1000,52 @@ func (r *ReconcileInfinispan) createSiteService(siteServiceName string, infinisp
...
-		}
-		return nil
+	err = appendKubernetesBackupSite(xsite, infinispan, backup, remoteKubernetes, logger)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func GetRemoteSiteRESTConfig(infinispan *infinispanv1.Infinispan, backup *infinispanv1.InfinispanSitesBackupSpec, logger logr.Logger) (*restclient.Config, error) {
```
rigazilla: Does this need to be exportable? Also GetMinikubeRESTConfig and GetOpenShiftRESTConfig

galderz: Doesn't need to, I'll fix it.

galderz:
@rigazilla Fixed and squashed some smaller commits

Issue #137, opened 2019-08-28
Enable graceful shutdown of operator managed Infinispan instances


galderz:

rigazilla:
Splitting this issue in the following tasks:
- [ ] implement cluster shutdown avoiding rebalancing during single node shutdown
- [ ] ensure all the persistent volumes are preserved and correctly remounted at restart
- [ ] ensure clean shutdown ad startup for caches with persistent data

PR #200, opened 2019-11-05
Graceful shutdown #137

rigazilla:
Graceful shutdown implementation.
To graceful shutdown a cluster user has to set `spec.replicas=0` and change it to the value before he shutdown when he wants to restart. Data will be preserved for all the caches with file-store persistence, i.e.:
```
<infinispan>
<cache-container>
<distributed-cache name ="pcache">
<persistence>
<file-store/>
</persistence>
</distributed-cache>
</cache-container>
</infinispan>
```

Shutdown procedure is:
- disable pods restart
- set `stopping` condition to `True`
- send graceful shutdown to one pod
- wait for all the pods to be not-ready
- save the statefulset.spec.replicas into the infinispan.status.ReplicasWantedAtRestart field
- scale to 0
- wait for statefulset.status.currentReplicas=0
- set `gracefulShutdown` to `True` and `stopping` to `False`

Startup procedure:
- set `infinispan.spec.replicas`=same replicas num before shutdown
- set  `gracefulShutdown` to `False`

### DIFF
--- a/pkg/apis/infinispan/v1/types_util.go
+++ b/pkg/apis/infinispan/v1/types_util.go
@@ -1,0 +1,42 @@
+package v1
+
+// GetCondition return true if condition == status
+func (ispn *Infinispan) GetCondition(condition string) *string {
+	for _, c := range ispn.Status.Conditions {
+		if c.Type == condition {
+			return &c.Status
+		}
+	}
+	return nil
+}
+
+// SetCondition set condition to status
+func (ispn *Infinispan) SetCondition(condition, status, message string) bool {
+	changed := false
+	for idx := range ispn.Status.Conditions {
+		c := &ispn.Status.Conditions[idx]
+		if c.Type == condition {
+			if c.Status != status {
+				c.Status = status
+				changed = true
+			}
+			if c.Message != message {
+				c.Message = message
+				changed = true
+			}
+
+			return changed
+		}
+	}
+	ispn.Status.Conditions = append(ispn.Status.Conditions, InfinispanCondition{Type: condition, Status: status, Message: message})
+	return true
+}
+
+// SetConditions set provided conditions to status
+func (ispn *Infinispan) SetConditions(conds []InfinispanCondition) bool {
+	changed := false
+	for _, c := range conds {
+		changed = changed || ispn.SetCondition(c.Type, c.Status, c.Message)
+	}
+	return changed
+}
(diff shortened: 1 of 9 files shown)


rigazilla:
updated with the new shutdown procedure. now user can gracefully shutdown setting `Spec.replicas=0`

test/e2e/util/test_kubernetes.go
```diff
@@ -147,6 +147,50 @@ func (k TestKubernetes) DeleteInfinispan(infinispan *ispnv1.Infinispan, timeout
...
+			return false, nil
+		}
+		return true, nil
+	})
+
+	ExpectNoError(err)
+}
+
+// GracefulRestartInfinispan deletes the infinispan resource
+// and waits that all the pods are gone
```
galderz: Comment needs updating

test/e2e/main_test.go
```diff
@@ -307,6 +306,71 @@ func genericTestForPersistenceVolume(clusterName string, modifier func(*ispnv1.I
 	verifier(spec)
 }
 
+func TestCheckDataSurviveToShutdown(t *testing.T) {
+	name := "test-data-shutdown-cache"
+	usr := "developer"
+	cacheName := "test"
+	template := "<infinispan>" +
```
galderz: golang has string literals that make this more palatable, see [here](https://github.com/infinispan/infinispan-operator/pull/210/files#diff-9b3d5baa2e1ef758d21388446f725c84R456).

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -0,0 +1,42 @@
+package v1
+
+// GetCondition return true if condition == status
```
galderz: The comment does not match the method API.

galderz:
Needs rebase

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -0,0 +1,42 @@
+package v1
```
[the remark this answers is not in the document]
rigazilla: I would like to keep them as methods and methods must be in the same receiver package.

pkg/apis/infinispan/v1/types_util.go
rigazilla: yep thanks

test/e2e/util/test_kubernetes.go
rigazilla: thanks

PR #201, opened 2019-11-05
ISPN-10738 xsite repl docs

oraNod:
https://issues.jboss.org/browse/ISPN-10738


documentation/asciidoc/topics/proc_creating_sa_tokens.adoc
```diff
@@ -0,0 +1,63 @@
...
+. Add the service account tokens to secrets on each site so it can authenticate with the backup site.
++
+For example, in the preceding example you created a service account token for SiteA. You should add that token to a secret on SiteB. Doing so allows SiteB to authenticate with SiteA.
++
+.. Log in to {ocp} for SiteB.
+.. Add the service account token for SiteA to a secret.
++
+[source,options="nowrap",subs=attributes+]
+----
+oc create secret generic site-a-token --from-literal=token=eyJhbGciOiJSUzI1NiIsImtpZCI6IiJ9...
```
galderz: In the minikube example you create secrets for site-b. I think both examples should refer to same site for consistency?

documentation/asciidoc/topics/proc_configuring_backup_sites.adoc
```diff
@@ -0,0 +1,65 @@
...
+endif::productized[]
+//Product
+ifdef::productized[]
+<3> specifies `LoadBalancer` as the service that exposes the {openshiftshort} cluster to externally network traffic.
+endif::productized[]
+<4> names the site to which the {brandname} cluster replicates data, the backup site.
+//Community
+ifndef::productized[]
+<5> specifies the URL to access the backup site.
+* Use `xsite://` if the backup site has a static hostname and port.
```
galderz: Leave `xsite://` out since it's not yet implemented.

documentation/asciidoc/topics/proc_configuring_backup_sites.adoc
```diff
@@ -0,0 +1,65 @@
...
+* Use `minikube://` if the backup site is a Minikube instance.
```
galderz: Not sure if you want to do it here or elsewhere, but for `minikube://` and `openshift://` I'd stress that the type of remote cluster, as defined in the URL scheme, has an impact on what the secret must contain. If the remote is minikube, then it expects the secret to contain ..., and if openshift then .... I think you might already mention it elsewhere, but the key is the scheme the URL uses.

oraNod: thx @galderz updated the docs.

galderz: Where did you add that info? I cannot find it in the pushed update

PR #207, opened 2019-11-07
ISPN-10923 external service docs

oraNod:
https://issues.jboss.org/browse/ISPN-10923


documentation/asciidoc/topics/ref_external_service.adoc
```diff
@@ -0,0 +1,21 @@
...
+[source,options="nowrap",subs=attributes+]
+----
+include::yaml_annotated/cr_external_service.yaml[]
+----
+
+<1> exposes an external service at port `11222`.
+<2> specifies one of the following service resource types:
+
+`NodePort`::
+Suitable for locally running {k8s} clusters.
```
galderz: If the user wants to find out more about NodePort services, I wondered if we should link to reference documentation in Kubernetes and/or OpenShift. I know for Kubernetes NodePort services are further explained in [here](https://kubernetes.io/docs/concepts/services-networking/service/#nodeport). Below that `LoadBalancer` service type is explained.

oraNod: good suggestion @galderz there's deeper knowledge in those resources so I've added the links

Issue #205, opened 2019-11-07
Pod cpu/memory limits have to be set


galderz:
We're currently only adding resource requests, but more importantly limits have to be set to!

```
      resources:
        requests:
          cpu: 500m
          memory: 512Mi
```

Issue #134, opened 2019-08-27
Add cache service


galderz:
Add cache service configuration.

PR #210, opened 2019-11-11
Pod limits, cache service and Go 1.13 compatibility

galderz:
Several pieces of work:

* Set pod cpu/memory limits (#205), this is required to build the default cache used by cache service (#134).
* Add cache service support (#134).
* Update code to StatefulSet `app/v1` and make it compile/test on Go 1.13.

### DIFF
--- a/pkg/controller/infinispan/util/cluster.go
+++ b/pkg/controller/infinispan/util/cluster.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"fmt"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"strconv"
+	"strings"
 )
 
 var log = logf.Log.WithName("cluster_util")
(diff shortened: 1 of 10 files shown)


galderz:
@rigazilla Squashed commits. If CI passes this is ready to go in.

pkg/controller/infinispan/util/cluster.go
```diff
@@ -90,5 +92,96 @@ func (c Cluster) GetClusterMembers(secretName, podName, namespace, protocol stri
...
+	execOut, execErr, err := c.Kubernetes.ExecWithOptions(execOptions)
+	if err != nil {
+		return fmt.Errorf("unexpected error creating cache, stderr: %v, err: %v", execErr, err)
+	}
+
+	// Reverse sort the standard output so that HTTP status code is first
+	execOutLines := strings.Split(execOut.String(), "\n")
+	reverse(execOutLines)
+
+	httpCode, err := strconv.ParseUint(execOutLines[0], 10, 64)
```
rigazilla: Not a bug but
seems you only use the last execOutLines string. Do you really need to reverse the array?

galderz: Yeah, that's on purpouse. The curl as defined shows the output and then a line break and the HTTP status code. So, to get the status code I split the entire output by line and reverse it. Then, the first element would be the HTTP status code alone.

galderz: The output of the command can itself have line breaks.

rigazilla:
merged, thanks!

PR #223, opened 2019-11-26
ISPN-10755 graceful shutdown

oraNod:
https://issues.jboss.org/browse/ISPN-10755

### DIFF
--- a/documentation/asciidoc/topics/proc_shut_down_gracefully.adoc
+++ b/documentation/asciidoc/topics/proc_shut_down_gracefully.adoc
@@ -1,0 +1,29 @@
+[id='shutting_down-{context}']
+= Shutting Down {brandname} Clusters
+Gracefully stop all {brandname} pods in the cluster. The {ispn_operator}
+ensures that all permanent cache configurations are stored in the persistent
+volume.
+
+Data Distribution::
+{brandname} saves the distribution of data across the cluster when you stop
+pods. For this reason, when you restore {brandname} clusters, `replicas` must
+match the number of pods that were in the cluster before you shut it down.
++
+For example, you shut down a cluster of 6 pods. When you restart the cluster,
+you must set the value of `replicas` to `6`.
+
+Data Passivation::
+The {ispn_operator} also writes all data to persistent volumes before shutting down {datagridservice} pods.
++
+You must ensure that the persistent volume has enough available space for your
+data. If the size of the persistent volume is not large enough to store all 
+data in memory, {brandname} writes an exception to logs and you lose data.
+
+.Procedure
+
+* Set the value of `replicas` to `0` and apply your custom resources.
++
+----
+spec:
+  replicas: 0
+----
--- a/documentation/asciidoc/stories/assembly_creating_clusters.adoc
+++ b/documentation/asciidoc/stories/assembly_creating_clusters.adoc
@@ -1,6 +1,6 @@
 [id='spinning_up_clusters']
+:context: create
 = Creating {brandname} Clusters
-:context: create_cluster
 You create, configure, and manage {brandname} clusters with the {ispn_operator}
 by defining custom resource objects in `yaml` format and applying them to
 {brandname} pods. Get started here and learn the basics for using the
@@ -26,6 +26,7 @@ include::{topics}/ref_crd.adoc[leveloffset=+1]
 include::{topics}/ref_container_resources.adoc[leveloffset=+2]
 include::{topics}/proc_create_cluster_minimal.adoc[leveloffset=+1]
 include::{topics}/proc_verify_cluster.adoc[leveloffset=+1]
+include::{topics}/proc_shut_down_gracefully.adoc[leveloffset=+1]
 
 // Restore the parent context.
 ifdef::parent-context[:context: {parent-context}]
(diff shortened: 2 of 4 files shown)


galderz:
@rigazilla Could you help @oraNod add more info here? For example:

* The need for the persistent volume configuration to be big enough. Also, graceful shutdown mostly makes sense for data grid service where persistent volume size is configurable.
* How restore from a graceful shutdown works. On restore, the number of pods has to be the same as when graceful shutdown was executed.

documentation/asciidoc/topics/proc_graceful_shutdown.adoc
```diff
@@ -0,0 +1,12 @@
+[id='shutting_down-{context}']
+= Shutting Down {brandname} Clusters
+Gracefully stop all {brandname} pods in the cluster so that permanent cache definitions and persisted data is available the next time pods start.
+
```
rigazilla: MORE INFO:
The operator makes sure that the capacity of the PV volume mounted on the pod is enough for storing all the data during the shutdown procedure. The user must take care of sizing the PV, if he explicitly specify a PV size (DataGrid only).

documentation/asciidoc/topics/proc_graceful_shutdown.adoc
```diff
@@ -0,0 +1,12 @@
...
+.Procedure
+
+* Set the value of `replicas` to `0` and apply the custom resources.
++
+----
+spec:
+  replicas: 0
+----
```
rigazilla: MORE INFO
The cluster must wake up from a shutdown with the same number of pods, so to restart the cluster replicas must be set to the number of pods that were running just before the shutdown

documentation/asciidoc/topics/ref_crd.adoc
```diff
@@ -16,7 +16,7 @@ include::yaml_annotated/cr_minimal.yaml[]
 <1> declares the version.
 <2> sets the resource type that the {ispn_operator} uses to handle the lifecycle of {brandname} clusters.
 <3> names the {brandname} cluster.
-<4> sets the number of pods in the {brandname} cluster.
+<4> sets the number of pods in the {brandname} cluster. If you set the value to `0` the {ispn_operator} stops the running cluster.
```
rigazilla: see comment above for the restart procedure

documentation/asciidoc/topics/proc_graceful_shutdown.adoc
oraNod: gotcha. so if there were 6 nodes in the cluster before shutdown, you must set `replicas: 6` when you bring it back up.

what are the consequences when you don't do that? say 6 nodes. shutdown. restart the cluster with 8 nodes. there isn't an automatic rebalance?

documentation/asciidoc/topics/proc_graceful_shutdown.adoc
oraNod: if the capacity of the PV is less than the size of the cache what happens? data loss or does the operator not complete the shutdown unless all data is persisted? so the shutdown hangs if the PV is too small.

rigazilla: I think you get a no space left on device. iirc @ryanemerson implemented that part and maybe can help us.

documentation/asciidoc/topics/proc_graceful_shutdown.adoc
rigazilla: I don't know the details, maybe @tristantarrant can help on this

documentation/asciidoc/topics/proc_graceful_shutdown.adoc
ryanemerson: I think it was Tristan who implemented graceful shutdown, but I assume the same as you @rigazilla that an exception will just be thrown. Kubernetes now supports dynamic expansion of persistent volumes, so that's a potential workaround https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/.

documentation/asciidoc/topics/ref_crd.adoc
oraNod: I think it's better to just keep this simple. better to say it sets the number of pods. we can talk about graceful shutdown in its own topic.

documentation/asciidoc/topics/proc_shut_down_gracefully.adoc
```diff
@@ -0,0 +1,29 @@
...
+For example, you shut down a cluster of 6 pods. When you restart the cluster,
+you must set the value of `replicas` to `6`.
+
+Data Passivation::
+The {ispn_operator} also writes all data to persistent volumes before shutting down {datagridservice} pods.
++
+You must ensure that the persistent volume has enough available space for your
+data. If the size of the persistent volume is not large enough to store all 
+data in memory, {brandname} writes an exception to logs and you lose data.
+
```
rigazilla: I would add that, for the Cache Service or if the user doesn't specify a spec.service.container.storage, the operator set the PV size to be at least equal to the pod memory

oraNod: @rigazilla thanks for the comment. I think we should update [this PR](https://github.com/infinispan/infinispan-operator/pull/231) with these details. 

my preference is to merge this PR and rebase the other on top. it'd be good to link the two sections about graceful shutdown and container storage.

rigazilla: ok with me. let's go with this and use the other PR to join things together

rigazilla:
merged, thanks oraNod!

PR #225, opened 2019-11-26
Use httpGet probe. github/infinispan-images#29

rigazilla:
Use httpGet probes instead of the ones provided by the image
https://github.com/infinispan/infinispan-images/issues/29


ryanemerson:
Thanks @rigazilla

PR #231, opened 2019-12-04
ISPN-10994 service type configuration

oraNod:
https://issues.jboss.org/browse/ISPN-10994


documentation/asciidoc/topics/config_examples/cache_service_default_cache.xml
```diff
@@ -0,0 +1,6 @@
+<distributed-cache name="default" mode="SYNC" owners="1">
+  <memory>
+    <off-heap size="91100283" eviction="MEMORY" strategy="REMOVE" address-count="1048576"/>
```
galderz: `address-count` no longer in use.

I'd also remove `size` since that's calculated dynamically by the operator given the memory given to each pod.

galderz:
Thx @oraNod :)

Issue #190, opened 2019-10-25
Enable connector encryption by default


galderz:
In some Kubernetes environments, such as OpenShift, it's possible to enable connector encryption by default since a CA can be accessed.

The operator should check if running on OpenShift, and if running with the default profile (or any other profile that requires it), enable connector encryption by default.

In vanilla Kubernetes environments, such as Minikube, connector encryption could be disabled by default, or let's encrypt could be used in those situations.

rigazilla:
We could check if `/apis/operator.openshift.io/v1/servicecas` has entries, in that case a serving certificates service should be available and we can use that service by default.
In the other cases the user must provide the secret with certificates to enable TLS.

@galderz , @Crumby wdyt?

galderz:
@rigazilla Sounds good

PR #233, opened 2019-12-06
TLS on by default with service-ca available #190

rigazilla:
Enable TLS by default if running on openshift and service-ca is available.
Certificates are stored in the secret specified by the field `infinispan.spec.security.endpointEncryption.certSecretName`, if not provided it will be set to `<clustername>-cert-secret`

If empty, fields under `infinispan.spec.security.endpointEncryption` will be populated accordingly. I.e.:
````
    security:
      endpointEncryption:
        certSecretName: example-infinispan-cert-secret
        certServiceName: service.beta.openshift.io
        type: service
````

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -151,6 +151,26 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	found := &appsv1.StatefulSet{}
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: infinispan.Name, Namespace: infinispan.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
+
+		// Populate EndpointEncryption if serving cert service is available
+		if getServingCertsMode(kubernetes) == "openshift.io" {
+			requeue := false
+			ee := &infinispan.Spec.Security.EndpointEncryption
+			if ee.Type == "" {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pkg/controller/infinispan/infinispan_controller.go [hunk 1]
```diff
@@ -1416,3 +1442,15 @@ func getEnvWithDefault(name, defVal string) string {
 	}
 	return defVal
 }
+
+// getServingCertsMode returns a label that identify the kind of serving
+// certs service is available. Returns 'openshift.io' for service-ca on openshift
+func getServingCertsMode(remoteKubernetes *ispnutil.Kubernetes) string {
+	if remoteKubernetes.HasServiceCAsCRDResource() {
+		return "openshift.io"
```
galderz: Why return this string? Seems a little arbitrary and not sure what other situation you might have in mind where this method would return something other than `""` or `"openshift.io".

Couldn't you just skip this method and ask `remoteKubernetes` if openshift service CA exists, and use that in the caller in this function?

rigazilla: This is to easily add other certs service in the future. Service discovery can be isolated in this method instead of mix the code with the reconciliaiton stuff.
String labels are needed instead of bool because different serviced may require different reconciliation policies in the caller.

rigazilla: @galderz if the comments are ok, can we go on and merge?

Issue #237, opened 2019-12-11
Probes not working with 10.1.0.CR1


galderz:
```
2m         Warning   Unhealthy               pod/example-infinispan-0                                        Readiness probe failed: Get http://172.17.0.5:11222/rest/v2/cache-managers/DefaultCacheManager/health/status: dial tcp 172.17.0.5:11222: connect: connection refused
14m         Warning   Unhealthy               pod/example-infinispan-0                                        Readiness probe failed: HTTP probe failed with statuscode: 404
20m         Warning   Unhealthy               pod/example-infinispan-0                                        Liveness probe failed: HTTP probe failed with statuscode: 404
```

`DefaultCacheManager` is named `default` since 10.0, but the image had not been updated.

dmvolod:
@galderz , @rigazilla  
I can see that the issue has fixed, but on my OKD still occurs 

Readiness probe failed: Get http://172.17.0.7:11222/rest/v2/cache-managers/DefaultCacheManager/health/status: dial tcp 172.17.0.7:11222: connect: connection refused

Could you please to comment this issue. Thanks in advance.

rigazilla:
@dmvolod it seems that your operator hasn't the fix. Are you running from master branch?

dmvolod:
@rigazilla , yes, I just build it, build docker image and run with make run-local on master branch as I'm working to fix some issues and that was applied, but this fix not.

rigazilla:
That's weird.
@dmvolod probe URL is defined [here](https://github.com/infinispan/infinispan-operator/blob/468a18451664b67efcaa2ac86448c6a2c61252b2/pkg/controller/infinispan/util/cluster.go#L17)
check with your local code.

dmvolod:
Thanks. @rigazilla Missed just one commit and rebase on master.

PR #238, opened 2019-12-11
Adjust probe URLs for Infinispan 10.1 #237

galderz:


galderz:
Not yet there, need to also change the cluster members. I'll centralise this.

rigazilla:
merged, thanks

PR #242, opened 2019-12-16
Move build from the 'dep' to 'mod' golang dependencies manager

### LINKED ISSUE #234: Move build from the 'dep' to 'mod' golang dependencies manager
It would be nice to move build from 'dep' to 'mod' golang dependencies manager.
@rigazilla what do you think about it. I can take care on migration, if community decide to do this.
Thanks in advance.

dmvolod:
Fixes #234


dmvolod:
@rigazilla could you please to look at the PR. All tests are passed.

rigazilla:
@dmvolod, yep I'll do asap. 
Maybe @slaskawi wants to have a look also.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - '1.11.x'
+  - '1.12.x'
```
slaskawi: Perhaps 1.13?

.travis.yml
```diff
@@ -19,8 +19,10 @@ jobs:
   include:
     - stage: "Unit Test"
       name: "Unit Test"
+      before_script:
+        - export GO111MODULE=on
       script:
-        - go get -u github.com/golang/dep/cmd/dep
+        - go mod tidy
```
slaskawi: Perhaps this should be `go mod vendor`?

BTW, In Keycloak Operator we pushed the whole vendor directory. This made Travis builds blazing fast.

dmvolod: Will change this for travis build, however the build time less than test several times

.travis.yml
dmvolod: Thanks, @slaskawi 
I will change this for travis build

.travis.yml
galderz: Our tests are not yet that fast (we need to implement mocking), so speed of adding vendor will probably be not so noticeable cos tests are the primary time consumer at the moment.

pkg/generated/clientset/versioned/clientset.go
```diff
@@ -19,6 +19,8 @@ limitations under the License.
 package versioned
```
rigazilla: Are all the changes in the `pkg/generated` folder related to the new deps manager?

dmvolod: It's looks like related to changes in k8s.io/code-generator invocation and used version as in dep it was bind to kubernetes version not fixed (v0.17.0 // indirect) right now.

rigazilla: uh ok, thanks!

build/boilerplate.go.txt
```diff
@@ -0,0 +1,16 @@
+/*
```
galderz: What's this file for?

dmvolod: This is header for auto-generated files, as method moves from vendor/k8s.io/code-generator/generate-groups.sh invoke to the direct go run k8s.io/code-generator/cmd/client-gen call

galderz:
Thx @dmvolod! Having this integrated should path the way for you to continue with the unit test work?

dmvolod:
Yes, @galderz , sure. Thanks for merging this.

PR #243, opened 2019-12-20
Auto-generated secret present even if user provides own secret

### LINKED ISSUE #239: Auto-generated secret present even if user provides own secret
If creating a secret and passing it via the `.spec.security.endpointSecretName`, operator still generates `*-generated-secret`, which seems to contain same information found in the user provided secret name. There should be no need for this extra secret.

dmvolod:
Fixes #239

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,4 @@ vendor/
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 ### Backup files ###
 *.backup
+/cmd/manager/debug
--- a/build/run-local.sh
+++ b/build/run-local.sh
@@ -8,4 +8,5 @@ oc login -u system:admin
 oc project default
 oc apply -f deploy/rbac.yaml
 oc apply -f deploy/crd.yaml
+oc wait --for condition=established crd infinispans.infinispan.org --timeout=60s
 WATCH_NAMESPACE="default" ./build/_output/bin/infinispan-operator -kubeconfig $KUBECONFIG
--- a/build/run-local.sh
+++ b/build/run-local.sh
@@ -1,12 +1,14 @@
 #!/usr/bin/env bash
 
 KUBECONFIG=${1-openshift.local.clusterup/openshift-apiserver/admin.kubeconfig}
+PROJECT_NAME=${PROJECT_NAME-default}
 
 echo "Using KUBECONFIG '$KUBECONFIG'"
+echo "Using PROJECT_NAME '$PROJECT_NAME'"
 
 oc login -u system:admin
-oc project default
+oc project "$PROJECT_NAME"
 oc apply -f deploy/rbac.yaml
 oc apply -f deploy/crd.yaml
 oc wait --for condition=established crd infinispans.infinispan.org --timeout=60s
-WATCH_NAMESPACE="default" ./build/_output/bin/infinispan-operator -kubeconfig $KUBECONFIG
+WATCH_NAMESPACE="$PROJECT_NAME" ./build/_output/bin/infinispan-operator -kubeconfig "$KUBECONFIG"
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -184,10 +184,13 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 
 		secret := r.secretForInfinispan(identities, infinispan)
-		err = r.client.Create(context.TODO(), secret)
-		if err != nil {
-			reqLogger.Error(err, "could not find or create identities Secret")
-			return reconcile.Result{}, err
+		if infinispan.Spec.Security.EndpointSecretName == "" {
+			reqLogger.Info("Creating a new Secret", "Secret.Namespace", secret.Namespace, "Secret.Name", secret.Name)
+			err = r.client.Create(context.TODO(), secret)
+			if err != nil {
+				reqLogger.Error(err, "could not find or create identities Secret")
+				return reconcile.Result{}, err
+			}
 		}
 
 		reqLogger.Info("Creating a new ConfigMap", "ConfigMap.Namespace", configMap.Namespace, "ConfigMap.Name", configMap.Name)
@@ -284,7 +287,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 				// if there's no ready pod we're in trouble
 				for _, pod := range podList.Items {
 					if pod.Status.ContainerStatuses[0].Ready {
-						err = cluster.GracefulShutdown(ispnutil.GetSecretName(infinispan.Name), pod.GetName(), pod.GetNamespace(), protocol)
+						err = cluster.GracefulShutdown(ispnutil.GetSecretName(infinispan), pod.GetName(), pod.GetNamespace(), protocol)
 						if err != nil {
 							reqLogger.Error(err, "failed to exec shutdown command on pod", "Infinispan.Namespace", infinispan.Namespace, "Infinispan.Name", infinispan.Name)
 							return reconcile.Result{}, err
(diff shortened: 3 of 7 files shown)


dmvolod:
@galderz , @rigazilla could you please to look at this PR
Thanks a lot.

pkg/controller/infinispan/util/security.go
```diff
@@ -75,7 +78,11 @@ func FindPassword(usr string, descriptor []byte) (string, error) {
 }
 
 // GetSecretName returns the secret name associated with a server
-func GetSecretName(name string) string {
-	return fmt.Sprintf("%v-generated-secret", name)
+func GetSecretName(m *infinispanv1.Infinispan) string {
+	if m.Spec.Security.EndpointSecretName == "" {
+		return fmt.Sprintf("%v-generated-secret", m.GetName())
+	} else {
```
rigazilla: go-lint suggests that the else keyword is not needed (due to the return in the if condition)

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -184,10 +184,13 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 
 		secret := r.secretForInfinispan(identities, infinispan)
```
rigazilla: `secretForInfinispan` calls `SetControllerReference`, I think it's ok to call it also on a user provided secret, but I'm not 100% sure

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -184,10 +184,13 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
-		if err != nil {
-			reqLogger.Error(err, "could not find or create identities Secret")
-			return reconcile.Result{}, err
+		if infinispan.Spec.Security.EndpointSecretName == "" {
+			reqLogger.Info("Creating a new Secret", "Secret.Namespace", secret.Namespace, "Secret.Name", secret.Name)
+			err = r.client.Create(context.TODO(), secret)
+			if err != nil {
+				reqLogger.Error(err, "could not find or create identities Secret")
+				return reconcile.Result{}, err
+			}
```
rigazilla: How about updating `infinispan.Spec.Security.EndpointSecretName` with the new name? So we inform the user on which secret is now in use

.gitignore
```diff
@@ -94,3 +94,4 @@ vendor/
 # End of https://www.gitignore.io/api/go,vim,emacs,visualstudiocode
 ### Backup files ###
 *.backup
+/cmd/manager/debug
```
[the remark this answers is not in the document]
dmvolod: this auto-generates while remote debug with dlv util. It will delete on exit, but can occurs on code clearance while hang or before exit.

build/run-local.sh
```diff
@@ -5,7 +5,8 @@ KUBECONFIG=${1-openshift.local.clusterup/openshift-apiserver/admin.kubeconfig}
 echo "Using KUBECONFIG '$KUBECONFIG'"
 
 oc login -u system:admin
-oc project default
+oc project myproject
 oc apply -f deploy/rbac.yaml
 oc apply -f deploy/crd.yaml
-WATCH_NAMESPACE="default" ./build/_output/bin/infinispan-operator -kubeconfig $KUBECONFIG
+oc wait --for condition=established crd infinispans.infinispan.org --timeout=60s
```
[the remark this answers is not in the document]
dmvolod: I was checking, but looks like no direct assign on namespace/project in RBAC config

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Good point, @rigazilla . But if we will do this, user provided secret will remove if user delete Infinispan resource

pkg/controller/infinispan/infinispan_controller.go
dmvolod: I will check it, but looks like this will be updating in this code https://github.com/infinispan/infinispan-operator/blob/8c14d3e9631241648b6b785a4d011d5b164a5119/pkg/controller/infinispan/infinispan_controller.go#L352
during next reconcile

pkg/controller/infinispan/util/security.go
dmvolod: Thanks, @rigazilla , will update.

.gitignore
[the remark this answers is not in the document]
rigazilla: ok, thanks!

pkg/controller/infinispan/infinispan_controller.go
rigazilla: that code applies only when the infinispan resources is updated. When the cluster is created with copy the Spec.Security into status.Security here:
https://github.com/infinispan/infinispan-operator/blob/8c14d3e9631241648b6b785a4d011d5b164a5119/pkg/controller/infinispan/infinispan_controller.go#L249

this is to prevent an immediate restart of the cluster.

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I've just checked: currently the user secret is deleted with the infinispan resource, due to this call:

https://github.com/dmvolod/infinispan-operator/blob/8c14d3e9631241648b6b785a4d011d5b164a5119/pkg/controller/infinispan/infinispan_controller.go#L1070

we probably need to fix the logic here: we may want to watch for the secret changes but not the final "cleanup"

build/run-local.sh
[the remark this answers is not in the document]
rigazilla: yep, it also seems that Role, RoleBind and ServiceAcc haven't any conditions attached

pkg/controller/infinispan/infinispan_controller.go
rigazilla: actually, there's probably something to check because I think I saw a restart

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Thanks, we will need to change logic to set reference just for auto-generated secrets

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Thanks, I will check

dmvolod:
@rigazilla , could you please to look at this again after code refactoring.
Thanks in advance.

build/run-local.sh
```diff
@@ -5,7 +5,8 @@ KUBECONFIG=${1-openshift.local.clusterup/openshift-apiserver/admin.kubeconfig}
 echo "Using KUBECONFIG '$KUBECONFIG'"
 
 oc login -u system:admin
-oc project default
```
galderz: Why change the project name? If anything, make it a parameter of env variable that can be tweaked 😉

dmvolod: Thanks, @galderz . Will change to the env var with 'default' project name by default :)

galderz: Thanks! Looks good to me :)

galderz:
Just need @rigazilla to check the updates to his comment and this should be good to go in.

rigazilla:
@dmvolod I still see restarts with this PR:
````
✔ ~/go/src/github.com/infinispan/infinispan-operator [pr243dmv L|…283⚑ 58] 
18:13 $ oc get pods
No resources found.
✔ ~/go/src/github.com/infinispan/infinispan-operator [pr243dmv L|…283⚑ 58] 
18:13 $ oc apply -f deploy/cr/minimal/cr_minimal.yaml 
infinispan.infinispan.org/example-infinispan created
✔ ~/go/src/github.com/infinispan/infinispan-operator [pr243dmv L|…283⚑ 58] 
18:13 $ oc get pods -w
NAME                   READY     STATUS              RESTARTS   AGE
example-infinispan-0   0/1       ContainerCreating   0          1s
example-infinispan-0   0/1       Running   0         3s
        example-infinispan-0   1/1       Running   0         1m
example-infinispan-1   0/1       Pending   0         0s
example-infinispan-1   0/1       Pending   0         0s
example-infinispan-1   0/1       ContainerCreating   0         0s
example-infinispan-1   0/1       Running   0         2s
example-infinispan-1   1/1       Running   0         58s
example-infinispan-0   1/1       Terminating   0         2m
example-infinispan-0   0/1       Terminating   0         2m
example-infinispan-0   0/1       Terminating   0         2m
example-infinispan-0   0/1       Terminating   0         2m
example-infinispan-0   0/1       Pending   0         0s
example-infinispan-0   0/1       Pending   0         0s
example-infinispan-0   0/1       ContainerCreating   0         0s
example-infinispan-0   0/1       Running   0         3s

````

rigazilla:
@dmvolod I've added a commit that seems to fix the restart. Pls have a look

galderz:
@rigazilla CI failed with your last commit

dmvolod:
Thanks, @rigazilla, LGTM
@galderz looks like this is slow CI problem, I got it previously and solve it with rerun tests.

rigazilla:
@galderz now CI is green. Probably @dmvolod is right on slow CI, I'm not able to reproduce the failure locally.

PR #245, opened 2020-01-03
Tests fix: Root web page no longers returns 401

Crumby:


rigazilla:
merged, thanks @Crumby !

PR #248, opened 2020-01-09
Handling operator upgrades

galderz:
The PR is complete.

As mentioned in the code, better resource management is required to avoid having to delete things manually. Currently all objects are owned by Infinispan, so deleting Infinispan deletes everything else. In an upgrade however, you don't want to delete Infinispan but you want to delete certain resources so that they're recreated based on new operator logic. An intermediate ownership would work, e.g. all upgradable objects are owned by stateful set, and if deleting stateful set, all dependant objects are removed. Stateful set would be owned by Infinispan to keep current logic. I'll create a separate issue for this.


rigazilla:
merged, thanks @galderz

Issue #249, opened 2020-01-13
Extra Java options not applied correctly on update


galderz:
Extra JVM options updates are not being correctly applied. Although the value is updated in the Infinispan object and `EXTRA_JAVA_OPTIONS` environment variable, this is not applied to `JAVA_OPTIONS` which is what really matters. If Infinispan object is created new this is applied as expected.

PR #252, opened 2020-01-13
Update JAVA_OPTIONS when updating extra options #249

galderz:
Version 2 of the fix for #249


rigazilla:
merged, thanks @galderz

Issue #173, opened 2019-10-09
Check/update info on the operatorhub home page


rigazilla:

galderz:
We control that information now under `deploy/olm-catalog`. I'll update that information to align it with the latest changes.

PR #255, opened 2020-01-14
Update OLM metadata information #173

galderz:
I've added missing spec information. WRT status, I've only listed the conditions since the rest of status information is really only for internal use.


rigazilla:
merged, thanks @galderz

Issue #261, opened 2020-01-16
Decrease cpu requests


galderz:
Resource requests are the amount of resources requested for a pod to be scheduled. In some situations, not enough CPU might be available if 500m are requested. Instead, we should drop requests to 250m and keep the limit as is. Looking at startup resource consumption, the server does not seem to need more than 80-100m to do its job, so we could further drop requests to 125m potentially, but I think 250m is a safe choice.

PR #263, opened 2020-01-16
Half CPU requests to better fit into systems #261

galderz:


rigazilla:
merged, thanks @galderz

PR #268, opened 2020-01-21
Change extract from Data section to StringData for tls secret data extract

dmvolod:
Looks like the section must be stringData instead of data according to design doc https://github.com/infinispan/infinispan-operator/blame/master/doc/design.adoc#L703 
Found this during unit tests developing
Related to issue #215

PR #272, opened 2020-01-23
Change extract from Data section to StringData for tls secret

rigazilla:
1.1.x backport for pr #268


dmvolod:
I'm sorry, @rigazilla . I've tested more deeply this issue and this PR and commit to the master should be reverted.

rigazilla:
> I'm sorry, @rigazilla . I've tested more deeply this issue and this PR and commit to the master should be reverted. I will change the unit-test according this issue. Thanks a lot.

no worries, what's the problem?

dmvolod:
The problem was in FakeClient implementation which is using for unit-test as it place and store both Data and StringData fields "as is", but in real k8s and OpenShift StringData is write only fields and can't be use for reading. I will think how to write test for this issue to prevent problem like this.

dmvolod:
@rigazilla , let's close this PR as I send revert for the commit

PR #273, opened 2020-01-23
219/min pods num

rigazilla:
This PR forbids that a cluster size is set below the minimum cluster size required to preserve data safety


deploy/cr/minimal/cr_minimal.yaml
```diff
@@ -3,4 +3,5 @@ kind: Infinispan
 metadata:
   name: example-infinispan
 spec:
+  image: infinispan/server:latest
```
galderz: `image` is optional, so no longer a minimal example 😉. What's the reason for this change?

rigazilla:
closing for now

PR #280, opened 2020-01-30
Wrong restart and test stabilization

rigazilla:
Extended timeout from 30m to 45m
Checked error code on status and resource update
fixed check on cpu resources changed

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -133,6 +133,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
+		reqLogger.Info("Error reading the object")
 		return reconcile.Result{}, err
 	}
 
@@ -144,16 +145,18 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	// because remote site host:port combinations need to be injected into Infinispan.
 	xsite, err := r.computeXSite(infinispan, reqLogger)
 	if err != nil {
+		reqLogger.Error(err, "Error in computeXSite", "Infinispan.Namespace")
 		return reconcile.Result{}, err
 	}
 
-	// Check if the deployment already exists, if not create a new one
+	// Check if the StatefulSet already exists, if not create a new one
 	found := &appsv1.StatefulSet{}
 	err = r.client.Get(context.TODO(), types.NamespacedName{Name: infinispan.Name, Namespace: infinispan.Namespace}, found)
 	if err != nil && errors.IsNotFound(err) {
-
+		reqLogger.Info("Configuring the StatefulSet")
 		// Populate EndpointEncryption if serving cert service is available
 		if getServingCertsMode(kubernetes) == "openshift.io" {
+			reqLogger.Info("Serving certificate service present. Configuring into CRD")
 			requeue := false
 			ee := &infinispan.Spec.Security.EndpointEncryption
 			if ee.Type == "" {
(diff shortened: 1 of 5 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -197,7 +209,20 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 				reqLogger.Error(err, "could not create a new Secret")
 				return reconcile.Result{}, err
 			}
+			// Set the endpoint secret name and update .Spec.Security
 			infinispan.Spec.Security.EndpointSecretName = secret.GetName()
+			err = r.client.Update(context.TODO(), infinispan)
```
dmvolod: @rigazilla what do you think about to move this update whole infinispan and Security into the separate function as looks like this is common code for the update?

rigazilla: That could be fine, but first I would like to find a solution to this kind of errors:
https://travis-ci.org/infinispan/infinispan-operator/jobs/644097070#L925
we have a lot of them. I though it was enough to reload the object before apply the update, but it doesn't seem the definitive solution.
The error above comes from here:
https://github.com/rigazilla/infinispan-operator/blob/71cefaba4a2dff40a050045005e482352bd532a6/pkg/controller/infinispan/infinispan_controller.go#L325
and in this case the infinispan object is reloaded just two lines above.

rigazilla: Uh no, in this case the actual error was here:
https://github.com/rigazilla/infinispan-operator/blob/71cefaba4a2dff40a050045005e482352bd532a6/test/e2e/util/test_kubernetes.go#L179

rigazilla:
@dmvolod now this PR should be ok if you want to have a look

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -413,15 +467,25 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 				found.Spec.Template.Spec.Volumes[i].Secret.SecretName = secret.GetName()
 			}
 		}
-		err = r.client.Update(context.TODO(), found)
 
+		err = r.client.Update(context.TODO(), found)
```
dmvolod: @rigazilla can we use common updateSecurity function for this?
Maybe we can use common flow for the all Security or Status updates: Update, Get latest CRD, Update whole Status or just Security?

rigazilla: @dmvolod in this case we cannot use updateSecurity since we're updating the StatefulSet and only the status for the Infinispan CRD.

dmvolod: Sorry, @rigazilla pointing on wrong string. I think about this.
https://github.com/infinispan/infinispan-operator/blob/3eb83d98400b9aa286259ed4816927eb0d597e80/pkg/controller/infinispan/infinispan_controller.go#L413
I not sure and would like to check that the Update without context.Background() cause extra invoke Reconcile

rigazilla: I can't see an advantage in wrapping the Update call in another function since we still need the error checking outside to return and requeue from the Reconcile method... but maybe I didn't get your point.

> I not sure and would like to check that the Update without context.Background() cause extra invoke Reconcile

Yeah that's something interesting.
Also a check on how we deal with Update() failure would be useful: the general police is to raise an error and restart the Reconcile loop. This imply that after an error the system is leaved in a clean state and I'm not sure this is true for all our Update() cases.

dmvolod:
Thanks @rigazilla , you've fixed all issues and also created a very good logging.

dmvolod:
@rigazilla we could also add log for arePodIPsReady function and probably add RequeueAfter delay in this method as well as notClusterFormed to reduce reconciliation tries

dmvolod:
@rigazilla maybe I'm not right, but looks like this code which dives into container invokes before notClusterFormed which is checking container state  
// Inspect the system and get the current Infinispan conditions
currConds := getInfinispanConditions(podList.Items, infinispan, protocol, cluster) 

However, container status Ready can be determined with only following condition 
cond.Type == corev1.ContainersReady && cond.Status == corev1.ConditionTrue

rigazilla:
> @rigazilla we could also add log for arePodIPsReady function and probably add RequeueAfter delay in this method as well as notClusterFormed to reduce reconciliation tries

@dmvolod good shot! I've added some delay around

rigazilla:
> @rigazilla maybe I'm not right, but looks like this code which dives into container invokes before notClusterFormed which is checking container state
> // Inspect the system and get the current Infinispan conditions
> currConds := getInfinispanConditions(podList.Items, infinispan, protocol, cluster)
> 
> However, container status Ready can be determined with only following condition
> cond.Type == corev1.ContainersReady && cond.Status == corev1.ConditionTrue

Checking the container status is not enough: an Infinispan node can be Ready but the Infinispan cluster not formed yet. That's why we dig into the system to check that all the nodes have the same cluster view.

dmvolod:
> Checking the container status is not enough: an Infinispan node can be Ready but the Infinispan cluster not formed yet. That's why we dig into the system to check that all the nodes have the same cluster view.
Yes, I know and understood, but I can see that we are trying to dig into the system to check that all the nodes have the same cluster view too early before at least one container not in ContainersReady status.

rigazilla:
> > Checking the container status is not enough: an Infinispan node can be Ready but the Infinispan cluster not formed yet. That's why we dig into the system to check that all the nodes have the same cluster view.
> 
> Yes, I know and understood, but I can see that we are trying to dig into the system to check that all the nodes have the same cluster view too early before at least one container not in ContainersReady status.

Uh yeah, now I see. Let me see if we can improve this. Thanks!

dmvolod:
Thanks, looks very good

rigazilla:
merged

Issue #224, opened 2019-11-26
Cache CRD


rigazilla:
Implement a new Custom Resource Definition **Cache** that looks like this:

    apiVersion: infinispan.org/v2alpha1
        kind: Cache
    metadata:
        name: cache-name
    spec:
        clustername: mycluster
        cacheTemplate: my-template
        cacheXML: |
      <distributed-cache name="mycache">
         <encoding>
            <key media-type="application/x-protostream"/>
            <value media-type="application/x-protostream"/>
         </encoding>
      </distributed-cache>
    status:
        connectionInfo: info needed by the user for connecting. Structure is TBD

With the following behaviour:
if an Infinispan cluster named `mycluster` is not present the operator must create it with a configuration that fits the cache needs.

The operator must create the cache  `mycache` and return in status.connectionInfo the correct info on how to use it.

Crumby:
@rigazilla What is the relation between `cacheTemplate` and `cacheXML`? Are they mutually exclusive or will the `cacheXML` creates also `cacheTemplate` for later reference?

> If an Infinispan cluster named mycluster is not present the operator must create it with a configuration that fits the cache needs.

Do you have some examples that would require to instantiate differently configured Infinispan clusters? It would be good to have some reasonable default specified as part of the issue too.

Btw. I like the way you solved getting connection info to the user.

rigazilla:
> @rigazilla What is the relation between `cacheTemplate` and `cacheXML`? Are they mutually exclusive or will the `cacheXML` creates also `cacheTemplate` for later reference?
> 

I was thinking to force them to be exclusive or one to override the other, but your option is good too if the template doesn't exist on the cluster (or the cluster is new)

> > If an Infinispan cluster named mycluster is not present the operator must create it with a configuration that fits the cache needs.
> 
> Do you have some examples that would require to instantiate differently configured Infinispan clusters? It would be good to have some reasonable default specified as part of the issue too.
> 

I'm referring to proper cache options like the 'number of owners' or 'maximum size'... 
I would like to see security aspect applied later by composition: like TLS terminated on the service for encryption... no idea for authentication atm.

> Btw. I like the way you solved getting connection info to the user.

thank you :)

rigazilla:
Hi @slaskawi, this is the issue for the cache CRD, please share here anything

slaskawi:
I think there might be a better way than submitting XML. 

[...]

Issue #218, opened 2019-11-25
Provide v2alpha1 API for further devs


rigazilla:
A new API version is needed for future devs that may break compatibility

galderz:
I was thinking about this the other day. Once you move to v2alpha1, you should probably target 2.0.0 as version for the operator (as opposed to 1.2.0). This would make a natural match between the configuration the operator version.

PR #284, opened 2020-02-10
Preparation for fix issue #229

### LINKED ISSUE #229: Convert golang e2e test to unit tests
**Issue:** Current GoLand e2e tests used within CI pipeline spin up Minikube and go through process of redeploying Infinispan cluster using Operator per test case which is causing long tests execution per PR.

**Proposal:** Reimplement the tests as unit tests that'd be sending actual and desired state to the Operator/`Reconciler`. The resulting operations from the Operator/`Reconciler` would be then evaluated against expected output.

dmvolod:
As an base operator code frequently changes, it's a preparation for unit testing to fix #229
@rigazilla could you please to look a this small PR.

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -56,7 +56,7 @@ var DefaultPVSize = resource.MustParse("1Gi")
 var kubernetes *ispnutil.Kubernetes
 
 // Cluster object
-var cluster *ispnutil.Cluster
+var cluster ispnutil.ClusterInterface
 
 const CacheServiceFixedMemoryXmxMb = 200
 const CacheServiceJvmNativeMb = 220
@@ -808,7 +808,7 @@ func areAllPodsReady(podList *corev1.PodList) bool {
 	return true
 }
 
-func existsCacheServiceDefaultCache(podName string, infinispan *infinispanv1.Infinispan, cluster *ispnutil.Cluster) bool {
+func existsCacheServiceDefaultCache(podName string, infinispan *infinispanv1.Infinispan, cluster ispnutil.ClusterInterface) bool {
 	namespace := infinispan.ObjectMeta.Namespace
 	secretName := ispnutil.GetSecretName(infinispan)
 	protocol := infinispanProtocol(infinispan)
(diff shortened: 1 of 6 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -75,6 +75,13 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 	return &ReconcileInfinispan{client: mgr.GetClient(), scheme: mgr.GetScheme()}
 }
 
+// NewFakeReconciler creates a new fake Reconciler for unit testing
```
rigazilla: Is this just for testing? It would be great to keep it separated from implementation code. A separated file would be good, if possible

dmvolod: to move it to the separate file, it would be need to move kubernetes and cluster vars inside ReconcileInfinispan struct and have all fields to be exported

rigazilla: Is this need also if moved in the same package?

dmvolod: Thanks, @rigazilla , just requires RestClient rest.Interface to be changed and exported

dmvolod:
Also, @rigazilla what do you think about moving GetMemoryLimitBytes and GetMaxMemoryUnboundedBytes to the cluster package (from kubernetes) as this is more related to Infinispan cluster pods rather than kubernetes entities?
I would be better for mocking with cluster as not possible to fake it.

rigazilla:
> Also, @rigazilla what do you think about moving GetMemoryLimitBytes and GetMaxMemoryUnboundedBytes to the cluster package (from kubernetes) as this is more related to Infinispan cluster pods rather than kubernetes entities and move GetPassword from cluster to kubernetes and also use ClusterInterface widely?
> I would be better for mocking with cluster as not possible to fake it.

At first sight looks good to me. To avoid to work on code that we'll have to change in the future, please check the your design with the roadmap, mainly: #218, #224, #288. There's nothing defined for them, just try to guess the possible impacts.

dmvolod:
@rigazilla looks like Cache CRD and Operator SDK can potential impact on this, but it could be and merge code quickly, as this is just tests

dmvolod:
@rigazilla . #228 is completed.
Right now, I think this PR will relate to #284 implementation, but without new cache CRD model it's a little bit difficult to guess which methods we will need to implement.

rigazilla:
merged thanks @dmvolod !

Issue #285, opened 2020-02-10
Add label to service


rigazilla:
Service hasn't any label.
It would be good to have at least:
- a label to select all services related to all the infinispan CRs
- a label to select the service related to a specific infinispan CR

dmvolod:
@rigazilla do we need to label another auto generated items, like configmap and secret?

rigazilla:
> @rigazilla do we need to label another auto generated items, like configmap and secret?

Yeah right, we should have labels to select every object related to an Infinispan cluster.

dmvolod:
Do we need to update this issue or create a new one?

rigazilla:
I think this one is ok

PR #286, opened 2020-02-10
Added label to services #285

rigazilla:
This PR adds label to services, to keep backward compatibility I've reused the `app` keyword. All the objects have a set of common labels:
`"clusterName":"<name-of-the-cluster>"`
`"infinispan_cr":"<name-of-the-cr>"`
atm the cluster name is equal to the cr name.

In addition all the object have an `"app"` label with different values:
- infinispan pod `"app":"infinispan-pod"`
- service has `"app":"infinispan-service"`
- dns-ping service has `"app":"infinispan-service-ping"`
- external service has  `"app":"infinispan-service-external"`
- xsite service has  `"app":"infinispan-service-xsite"`

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -307,7 +307,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 
 	// List the pods for this infinispan's deployment
 	podList := &corev1.PodList{}
-	labelSelector := labels.SelectorFromSet(labelsForInfinispanSelector(infinispan.Name))
+	labelSelector := labels.SelectorFromSet(labelsForInfinispan(infinispan.Name, ""))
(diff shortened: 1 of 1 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1492,7 +1486,10 @@ func isPodReady(pod corev1.Pod) bool {
 }
 
 func (r *ReconcileInfinispan) serviceForInfinispan(m *infinispanv1.Infinispan) *corev1.Service {
-	ls := labelsForInfinispan(m.ObjectMeta.Name)
+	lsPodSelector := labelsForInfinispan(m.ObjectMeta.Name)
```
dmvolod: @rigazilla what do you think about change labelsForInfinispan func to add advanced parameters for adding flexible object label inside it like labelsForInfinispan(name, type) or add some logic inside to reduce number of code lines.

rigazilla: yeah ok, done

rigazilla:
merged

PR #287, opened 2020-02-11
notes for Prometheus integration

rigazilla:
steps for integrating Infinispan and Prometheus.
Needs #286

### DIFF
--- a/documentation/asciidoc/stories/prometheus_integration.adoc
+++ b/documentation/asciidoc/stories/prometheus_integration.adoc
@@ -1,0 +1,77 @@
+## Prometheus integration
+This chapter is about sending Infinispan statistics to Prometheus in an operator managed environment running on Openshift.
+
+#### Prerequisites
+1. Access to an Openshift console
+2. A running Prometheus server, installed via Operator (see )
+3. A running Infinispan cluster
+
+#### Deploy the service monitor
+A service monitor is a customer resource (ServiceMonitor) that represent the application to be monitored.
+The Prometheus Operator, when a ServiceMonitor cr is deployed to the system, genetares all the configuration needed to scrap the application.
+The information needed to define a ServiceMonitor are:
+
+1. how to select the workload to scrap. This will be represented by a selector item;
+2. how to connect to selected workload. This will be represented by an endpoint item.
+
+The ServiceMonitor template to populate is the following:
+....
+name: infinispan-monitor-docs
+ namespace: <Prometheus namespace>
+spec:
+ endpoints:
+   - basicAuth:
+       password:
+         key: password
+         name: <authentication secret>
+       username:
+         key: username
+         name: <authentication secret>
+     interval: 30s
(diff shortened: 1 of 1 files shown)


rigazilla:
@oraNod, some notes about Prometheus

documentation/asciidoc/stories/prometheus_integration.adoc
```diff
@@ -0,0 +1,77 @@
...
+
+#### Deploy the service monitor
+A service monitor is a customer resource (ServiceMonitor) that represent the application to be monitored.
+The Prometheus Operator, when a ServiceMonitor cr is deployed to the system, genetares all the configuration needed to scrap the application.
+The information needed to define a ServiceMonitor are:
+
+1. how to select the workload to scrap. This will be represented by a selector item;
+2. how to connect to selected workload. This will be represented by an endpoint item.
+
+The ServiceMonitor template to populate is the following:
```
Crumby: Is this example intentionally "compressed"? It's missing `kind` and `metadata`.

documentation/asciidoc/stories/prometheus_integration.adoc
```diff
@@ -0,0 +1,77 @@
...
+     scheme: https
+     tlsConfig:
+       insecureSkipVerify: true
+       serverName: <certificate CN>
+ namespaceSelector:
+   matchNames:
+     - <Infinispan namespace>
+ selector:
+   matchLabels:
+     app: infinispan-service
```
Crumby: If I'm reading #286 correctly shouldn't this be `<InfinispanCRName>: infinsipan-service`? Also won't this match all the cluster that will possibly run in the same namespace? How about adding `cluster: <InfinispanCRName>`? If I'm not mistaken then all the resources created by Operator get this label...

rigazilla: yeah oversight, we need to identify the cluster. Thanks!

documentation/asciidoc/stories/prometheus_integration.adoc
rigazilla: bad copy/paste. thanks!

documentation/asciidoc/stories/prometheus_integration.adoc
Crumby: Btw, reading it one more time after myself you can ignore the second part of the comment as it's invalid. `<InfinispanCRName>: infinsipan-service` should be enough for the verification. Second applies only if the we would be matching `app: infinispan-service`.

documentation/asciidoc/stories/prometheus_integration.adoc
rigazilla: done

documentation/asciidoc/stories/prometheus_integration.adoc
rigazilla: done

PR #292, opened 2020-02-14
Convert golang e2e test to unit tests

dmvolod:


galderz:
Looks good to me. Just double check what @Crumby said about keeping some tests as unit tests and others as integration tests.

galderz:
@Crumby @rigazilla Thoughts?

build/run-tests.sh
```diff
@@ -5,4 +5,4 @@ KUBECONFIG=${1-openshift.local.clusterup/openshift-apiserver/admin.kubeconfig}
 echo "Using KUBECONFIG '$KUBECONFIG'"
 
 go clean -testcache ./test/e2e
-go test -v ./test/e2e -timeout 45m
+go test -v ./test/e2e -run TestExternalServiceWithAuth -timeout 45m
```
Crumby: This should be omitted from the PR so the CI doesn't run only one test.

dmvolod: Yes, sure. But we need to decide which set of the e2e test will keep for CI. What do you think about these?

Crumby: With unit tests in place we can get rid off some e2e execution heavy duplicates and keep executing all in e2e. See my suggestion:

**Remove:** TestSimple, TestNodeStartup, TestClusterFormationWithTls, TestContainerCPUUpdate, TestContainerMemoryUpdate, TestContainerJavaOptsUpdate

**Keep:** TestClusterFormation, TestCacheService, TestPernamentCache, TestCheckDataSurviveToShutdown, TestExternalService, TestExternalServiceWithAuth

Basically we can remove all the `*Update` tests. As far as we are concerned they have equivalent unit test at this moment. `TestSimple` isn't a test. `TestNodeStartup` is subtest of `TestClusterFormation` and `TestClusterFormationWithTLS` doesn't provide any more assurance as clustering uses different protocol meaning that TLS enabled doesn't affect it.

@rigazilla What do you think about this?

test/unit/main_test.go
```diff
@@ -0,0 +1,567 @@
...
+	}
+	return nil, nil
+}
+
+func TestBasicConfiguration(t *testing.T) {
+	infinispan := basicInfinispan
+	r := testDefaultConfiguration(t, infinispan, basicRestMock)
+	sset := tstutil.GetStatefulSet(r.GetClient(), types.NamespacedName{operatorNamespace, operatorName})
+
+	assert.Equal(t, corev1.URIScheme("HTTP"), sset.Spec.Template.Spec.Containers[0].LivenessProbe.HTTPGet.Scheme, "Liveness probe Scheme not defined as expected")
```
Crumby: Should this checks be part of `testDefaultConfiguration`?

build/run-tests.sh
dmvolod: Thanks for response, @Crumby 
In case of the issue #311 we should need to keep one of the *Update tests with replicas=2 or create a new one. In other cases, the list looks very good.

test/unit/main_test.go
dmvolod: I will look into it. Thanks.

Crumby:
@dmvolod I haven't noticed noticed a deletion presence in the unit tests. We should also have such when user deletes infinispan resource then then operator deletes all the related resource it created.

dmvolod:
Thanks, @Crumby , will do this test.

test/unit/main_test.go
dmvolod: @Crumby looks like it's a correct test, as testDefaultConfiguration uses for both URI schema's: HTTP and HTTPS

dmvolod:
@Crumby I've done some research and we can't create test for deletion as some of the resources are creating and deleting by foreign controllers but not fake client and we can just test them if will create/delete/update by yourself.

Crumby:
@dmvolod Aren't we supposed to send the deletion request for some resources upon infinispan CR deletion? Eg. service, statefulsets, custom secrets...

dmvolod:
> @dmvolod Aren't we supposed to send the deletion request for some resources upon infinispan CR deletion? Eg. service, statefulsets, custom secrets...

@Crumby no, all auto-generated resources are referenced by Infinispan CRD
`controllerutil.SetControllerReference(m, dep, r.scheme)`

Issue #290, opened 2020-02-13
Move entities model outside the infinispan_controller code


dmvolod:
@rigazilla , @galderz what do you think about some refactor to move entities model (create and delete services, configmaps and statefulset ) outside the infinispan_controller.go to the separate model package?

The stages are:

- [x] Stage I - move only  some simple methods and introduce constants
- [x] Stage II - move common functions from infinispan_controller to the separate packages. Arrange Status structs for Cluster and Cache (change from string to metav1.ConditionStatus in Cluster)
- [ ] Stage III - implement model (secrets, services, configmap) outside infinispan_controller. Use common list function kubernetes.GetK8sResources (possible give a new name), use common functions as much as possible (change IsConditionTrue in autoscale.go to IsWellFormed), move all entities (Service, ConfigMaps, etc.) name templates to the constants and templating instead of the values concatenation
- [ ] Stage IV - try to split kubernetes and cluster entities
- [ ] Stage V - simplify using kubernetes client and possible to use https://github.com/redhat-cop/operator-utils

rigazilla:
@dmvolod look great to me! infinispan_controller.go really need a downsizing

Issue #288, opened 2020-02-12
Update code to the latest operator-sdk


rigazilla:
Current operator code is based on an skeleton generated by an old operator-sdk. Assess whether is possible to update it.

PR #300, opened 2020-02-20
299/creds secret

### LINKED ISSUE #299: Credentials secret is deleted during upgrade
The secret could be provided by the user and we don't want to delete it. Also in case it has been generated by the operator it's ok to keep the same credentials.

rigazilla:
Do not delete secret during upgrade. fixes #299


rigazilla:
> Thanks, @rigazilla looks good
> I'm not sure, we would like to introduce a new marker filed (flag) for auto-generated entities which is possible to customize for further logic.

@dmvolod thanks for the feedback, I'm going to build a 1.1.1 rel with this fix.
A way to select autogenerated stuff look a goods idea, but let see first how things will looks like after #224, #288, #290

rigazilla:
merged

PR #302, opened 2020-02-22
Fix log info version print

### LINKED ISSUE #301: Print the operator version on startup
It would be nice to store info about operator version and also print it as well as OS version on the operator startup
Also need to print version info without WATCH_NAMESPACE installed

dmvolod:
Fixes #301
@rigazilla please pay attention, as this a little bit related to release process. Now we cant define which operator version is running inside the cluster and docker image as this info is missing in logs.

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -8,7 +8,7 @@ GOOS=${1-linux}
 BIN_DIR=${2-$(pwd)/build/_output/bin}
 PROJECT_NAME="infinispan-operator"
 BUILD_PATH="./cmd/manager"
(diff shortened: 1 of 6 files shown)


rigazilla:
@dmvolod looks good. I've added a commit to print the version at the beginning of the reconcile loop. Have a look how it sounds to you.

dmvolod:
@rigazilla , looks good. Thanks for adding this to reconcile stage.

rigazilla:
merged, thanks @dmvolod !

PR #311, opened 2020-03-06
Cluster with two or more replicas never starts after Spec.Container properties changes

dmvolod:
This is base changes for switching form static source to function provided and for testing with our CI

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -57,9 +57,23 @@ var DefaultSpec = ispnv1.Infinispan{
 	},
 }
 
+var MinimalSpec = ispnv1.Infinispan{
+	TypeMeta: metav1.TypeMeta{
+		APIVersion: "infinispan.org/v1",
+		Kind:       "Infinispan",
+	},
+	ObjectMeta: metav1.ObjectMeta{
+		Name: DefaultClusterName,
+	},
+	Spec: ispnv1.InfinispanSpec{
+		Image:    getEnvWithDefault("IMAGE", "registry.hub.docker.com/infinispan/server"),
+		Replicas: 2,
+	},
+}
+
 func TestMain(m *testing.M) {
 	namespace := strings.ToLower(Namespace)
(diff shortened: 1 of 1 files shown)


build/run-tests.sh
```diff
@@ -5,4 +5,4 @@ KUBECONFIG=${1-openshift.local.clusterup/openshift-apiserver/admin.kubeconfig}
 echo "Using KUBECONFIG '$KUBECONFIG'"
 
 go clean -testcache ./test/e2e
-go test -v ./test/e2e -timeout 45m
+go test -v ./test/e2e -run TestContainerCPUUpdateWithTwoReplicas -timeout 45m
```
Crumby: Isn't this used to run all the tests via Travis CI? This change would drop all but one this one change. (Also the file is named `run-tests`)

dmvolod: @Crumby , yes, you are right, but it just for one issue reproducer. I will drop -run when this issue will be fixed, but I'm thinking about it nature.

dmvolod:
@rigazilla , it's a very strange, but reproducer looks like fails after fix

rigazilla:
@dmvolod the test seems to fail during the first cluster startup:

https://github.com/dmvolod/infinispan-operator/blob/ac727d307c1c73e08820d4a989253853d035c523/test/e2e/main_test.go#L199

I've added a commit, but I'm not sure it's the fix... let's see CI

dmvolod:
Thanks, @rigazilla , I'm testing now on my local env.

dmvolod:
Oh, @rigazilla excellent. Now all are working fine!!!
What do you think about to keep this test in e2e set?

rigazilla:
@dmvolod looks good to me!
Can you squash the commits? I can merge then

dmvolod:
Done, @rigazilla 
Please merge, thanks

rigazilla:
I've squashed my commit also

rigazilla:
merged. Thanks @dmvolod , great job!

Issue #312, opened 2020-03-06
Unable to run some of the e2e tests against OpenShift


Crumby:
Some of the e2e tests do no work with OpenShift (tested with 4.2 and 4.3). The issue started appearing after migration to Operator SDK 0.15.2 PR was merged (#293).

So far this applies for all the `*Update` tests. Request to change the parameter won't hit the OpenShift.

This issue is probably caused by Kubernetes client update.

dmvolod:
@Crumby could you please provide full error stack from the test env.

dmvolod:
I'm not sure that all tests could be running on any openshift remotely as for some envs is not clear and possible to expose service port. It's necessary to create a routes for this.
@rigazilla , @galderz what do you think about this?

Crumby:
@dmvolod Tests were running fine before #293 for OpenShift on AWS you just need to set `EXPOSE_SERVICE_TYPE=LoadBalancer` to be able access the ispn cluster in there.

Here is the stacktrace of failing CPUUpdate test but it's just plain timeout:
```
$ go test -v -run CPUUpdate ./test/e2e/ -timeout 30m
=== RUN   TestContainerCPUUpdate
--- FAIL: TestContainerCPUUpdate (357.71s)
panic: timed out waiting for the condition [recovered]
	panic: timed out waiting for the condition

goroutine 216 [running]:
testing.tRunner.func1(0xc0001ce300)
	/usr/lib/golang/src/testing/testing.go:874 +0x3a3
panic(0x14238e0, 0xc0001d4830)
	/usr/lib/golang/src/runtime/panic.go:679 +0x1b2
github.com/infinispan/infinispan-operator/test/e2e.genericTestForContainerUpdated(0x16e9898, 0x16e98a0)
	/home/pdrobek/work/GIT/DG_QE/infinispan-operator/test/e2e/main_test.go:213 +0x4bc
github.com/infinispan/infinispan-operator/test/e2e.TestContainerCPUUpdate(0xc0001ce300)
	/home/pdrobek/work/GIT/DG_QE/infinispan-operator/test/e2e/main_test.go:144 +0x39
testing.tRunner(0xc0001ce300, 0x16e98a8)
	/usr/lib/golang/src/testing/testing.go:909 +0xc9
created by testing.(*T).Run
	/usr/lib/golang/src/testing/testing.go:960 +0x350
FAIL	github.com/infinispan/infinispan-operator/test/e2e	358.053s
FAIL
```

[...]

PR #318, opened 2020-03-08
Check err and validate that issue #312 runs on k8s & ocp

dmvolod:


Crumby:
@dmvolod @rigazilla Can we make this one priority and backport it to 1.1.x?

dmvolod:
@Crumby , @rigazilla i'm not sure that we need to backport this to the 1.1.x as we don't have an updated Operator SDK version 0.15.2 there

Crumby:
@dmvolod Makes sense. It doesn't. I haven't realized that.

rigazilla:
@dmvolod can you check the failure in CI?

I think I can merge this. If there are no conflict I can port on 1.1.x also to keep branches closer. Ok?

dmvolod:
Thanks, @rigazilla , add some fixes and waiting for CI completion.

dmvolod:
@rigazilla this is ready for merging
We don't need to port it to the 1.1.x as this branch using o old SDK

rigazilla:
> @rigazilla this is ready for merging
> We don't need to port it to the 1.1.x as this branch using o old SDK

I know it's not needed but If it works on 1.1.x and doesn't generate conflicts I would do the port to reduce the differences between branches and ease the subsequent backports

dmvolod:
Oh, thanks,  @rigazilla  didn't think about it.
Let's merge this PR and i will create a PR for 1.1.x

rigazilla:
merged, thanks @dmvolod !

PR #320, opened 2020-03-09
Added Infinispan finalizer for cleaning PVC #315

### LINKED ISSUE #315: Operator doesn't clean-up PVC upon cluster removal
PVC should be delete on cluster removal (Infinispan CR deletion)

rigazilla:
A finalizer has been added to delete all PVC that match the pod selector for an Infinispan resource.

Fixes #315

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -63,6 +63,7 @@ const CacheServiceJvmNativeMb = 220
 const CacheServiceMaxRamMb = CacheServiceFixedMemoryXmxMb + CacheServiceJvmNativeMb
 const CacheServiceAdditionalJavaOptions = "-Dsun.zip.disableMemoryMapping=true -XX:+UseSerialGC -XX:MinHeapFreeRatio=5 -XX:MaxHeapFreeRatio=10"
 const CacheServiceJvmNativePercentageOverhead = 1
+const infinispanFinalizer = "finalizer.infinispan.org"
 
 // Add creates a new Infinispan Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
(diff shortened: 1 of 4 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1832,3 +1855,44 @@ func getServingCertsMode(remoteKubernetes *ispnutil.Kubernetes) string {
 	}
 	return ""
 }
+
+func contains(list []string, s string) bool {
```
dmvolod: @rigazilla it's just a syntax sugar, but it' wold be nice to use already implemented library such as https://github.com/thoas/go-funk

rigazilla: @dmvolod I've just copy-paste the code from the [sdk-doc](https://github.com/operator-framework/operator-sdk/blob/master/doc/user-guide.md#handle-cleanup-on-deletion).

Does go-funk help us to simplify other part of code? Otherwise I would prefer to limit library dependences.

dmvolod: @rigazilla not sure about current code, maybe in the future.

dmvolod: But we should also avoid of copy/paste code if possible to use a simple libraries if possible.
We should be use also https://github.com/elliotchance/pie auto-generation for reducing custom search functions also.  
@rigazilla what do you think about it?

rigazilla: @dmvolod I think that it would be good to see a POC in a separate PR, so we can evaluate pro and cons

dmvolod: good point, thanks @rigazilla . I will try to do this during refactoring PR.

test/e2e/main_test.go
```diff
@@ -260,69 +260,6 @@ func TestCacheService(t *testing.T) {
 	}
 }
 
-// TestPermanentCache creates a permanent caches the stop/start
-// the cluster and checks that the cache is still there
-func TestPermanentCache(t *testing.T) {
```
Crumby: Are we deleting the test case as it's duplication of `TestCheckDataSurviveToShutdown`?

rigazilla: > Are we deleting the test case as it's duplication of `TestCheckDataSurviveToShutdown`?

actually no, it tests permanent cache... let me see if I can fix it, instead of removing.
thanks @Crumby

rigazilla: @Crumby should be ok now

rigazilla:
@Crumby, @dmvolod please approve the PR if you think it's ok

rigazilla:
thanks for the review @dmvolod , @Crumby ! merged

Issue #325, opened 2020-03-11
NodePort is hardcoded and exposed also when not needed


rigazilla:
Operator must set explicitly a node port only if needed and use port number from `spec.Expose` if available or default to 30222

Crumby:
@rigazilla Can't we let OpenShift to pick up the port by default? Defaulting to 30222 will force users to set the number if they are about to deploy second cluster or the port is already taken by different app running in that environment.

rigazilla:
@Crumby that is because of or kind setup for testing (not sure but I guess it from a comment). But you're right, let's openshift choose the port and hardcode 30222 in our tests

galderz:
@Crumby I fixed NodePort for easier debugging.

dmvolod:
Yes, @galderz sure, but for tests against OCP on AWS we need a flexible set of NodePort, as look like AWS LoadBalancer release port not immediately with the service remove.
@rigazilla do you have a plans to fix this for tests?

Crumby:
@dmvolod Tests won't be an issue won't be causing an issue with #326 on AWS and using LoadBalancer. As with LoadBalancer Operator will not set node port at all.

dmvolod:
Yes, sure @Crumby , thank you.

PR #326, opened 2020-03-11
NodePort fix. #325

rigazilla:
Now NodePort is explicitly specified only for NodePort service type. If not specified by the user port is no more defaulted to 30222, but is left to the cluster platform selection policy.

### DIFF
--- a/doc/design.adoc
+++ b/doc/design.adoc
@@ -94,13 +94,16 @@ is gracefully shut down.
 | v1
 
 | `expose`
-| Expose Infinispan for external access over port 11222.
-At the very least,
-the user must indicate via `type` attribute how the
+| If not empty the operator will expose Infinispan port 11222 externally.
+The `spec.type` attribute must indicate how the
 https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#service-v1-core[Service]
-is exposed.
-`NodePort` is often used while testing locally,
-while `LoadBalancer` is used in production-like environments where an external load-balancer is available.
+is exposed. Two service types are supported: `LoadBalancer, NodePort`.
+
+For the `NodePort` service type the `Spec.Ports[0].nodePort` attribute can be specified, if not port will be selected by the platform.
+
+This is just an helper for a quick service exposition, user can setup it's own service exposition
+if more configurability is needed.
+
 | https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#servicespec-v1-core[core/v1/ServiceSpec]
 | false
 | Infinispan not exposed externally.
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -282,12 +282,23 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 
 		if isExposed(infinispan) {
 			externalService := r.serviceExternal(infinispan)
-			err = r.client.Create(context.TODO(), externalService)
-			if err != nil && !errors.IsAlreadyExists(err) {
-				reqLogger.Error(err, "failed to create external Service", "Service", ser)
-				return reconcile.Result{}, err
+			if externalService != nil {
+				err = r.client.Create(context.TODO(), externalService)
+				if err != nil && !errors.IsAlreadyExists(err) {
+					reqLogger.Error(err, "failed to create external Service", "Service", ser)
+					return reconcile.Result{}, err
+				}
+				infinispan.Spec.Expose = externalService.Spec
+				err = r.client.Update(context.TODO(), infinispan)
+				if err != nil {
+					reqLogger.Error(err, "failed to Infinispan with Service spec", "Service", ser)
+					return reconcile.Result{}, err
+				}
+				reqLogger.Info("Created External Service", "Service", externalService)
+			} else {
+				reqLogger.Info("External Service NOT created. Unsupported type?", "Service", nil)
+
 			}
-			reqLogger.Info("Created External Service", "Service", externalService)
 		}
 
 		// Infinispan resource must have a finalizer
@@ -1571,7 +1582,6 @@ func (r *ReconcileInfinispan) serviceExternal(m *infinispanv1.Infinispan) *corev
 	lsPodSelector := labelsForInfinispan(m.ObjectMeta.Name, "infinispan-pod")
 	lsService := labelsForInfinispan(m.ObjectMeta.Name, "infinispan-service-external")
 	externalServiceType := m.Spec.Expose.Type
-
 	// An external service can be simply achieved with a LoadBalancer
 	// that has same selectors as original service.
 	externalServiceName := getServiceExternalName(m)
@@ -1592,13 +1602,23 @@ func (r *ReconcileInfinispan) serviceExternal(m *infinispanv1.Infinispan) *corev
 				{
 					Port:       int32(11222),
 					TargetPort: intstr.FromInt(11222),
-					// Fix NodePort to match exported port in kind configuration
-					NodePort: int32(30222),
 				},
 			},
 		},
 	}
 
+	if externalServiceType == corev1.ServiceTypeLoadBalancer {
+		// Nothing to do here, keeping the if else struct
+	} else if externalServiceType == corev1.ServiceTypeNodePort {
+		var portNum int32
+		if m.Spec.Expose.Ports != nil || len(m.Spec.Expose.Ports) > 0 {
+			portNum = m.Spec.Expose.Ports[0].NodePort
+			externalService.Spec.Ports[0].NodePort = portNum
+		}
+	} else {
+		// Service type currently unsupported
+		return nil
+	}
 	// Set Infinispan instance as the owner and controller
 	controllerutil.SetControllerReference(m, externalService, r.scheme)
 	return externalService
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -449,7 +449,8 @@ func TestExternalService(t *testing.T) {
 
 func exposeServiceSpec() corev1.ServiceSpec {
 	return corev1.ServiceSpec{
-		Type: exposeServiceType(),
+		Type:  exposeServiceType(),
+		Ports: []corev1.ServicePort{{NodePort: 30222}},
 	}
 }
 


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -282,12 +282,17 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 
 		if isExposed(infinispan) {
 			externalService := r.serviceExternal(infinispan)
-			err = r.client.Create(context.TODO(), externalService)
-			if err != nil && !errors.IsAlreadyExists(err) {
-				reqLogger.Error(err, "failed to create external Service", "Service", ser)
-				return reconcile.Result{}, err
+			if externalService != nil {
+				err = r.client.Create(context.TODO(), externalService)
```
dmvolod: @rigazilla what do you think to return back NodePort to the Infinispan structure after create? I must be more visible for users.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1585,18 +1589,35 @@ func (r *ReconcileInfinispan) serviceExternal(m *infinispanv1.Infinispan) *corev
 			Namespace: m.ObjectMeta.Namespace,
 			Labels:    lsService,
 		},
-		Spec: corev1.ServiceSpec{
+	}
+
+	if externalServiceType == "LoadBalancer" {
```
dmvolod: It would be nice to use constants from the typesv1 for better supporting and validation

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1585,18 +1589,35 @@ func (r *ReconcileInfinispan) serviceExternal(m *infinispanv1.Infinispan) *corev
...
+			Selector: lsPodSelector,
+		}
+	} else if externalServiceType == "NodePort" {
+		var portNum int32
+		if m.Spec.Expose.Ports != nil || len(m.Spec.Expose.Ports) > 0 {
+			portNum = m.Spec.Expose.Ports[0].NodePort
+		} else {
+			portNum = 0
+		}
+		externalService.Spec = corev1.ServiceSpec{
```
dmvolod: Do you think to change Expose  from v1.ServiceSpec to the custom one, as we are using just two parameters, but customer need to create a complex one for real system. Or another parameters will be using in the fuster?

pkg/controller/infinispan/infinispan_controller.go
rigazilla: Maybe we could add to the infinispan RC a reference to the created service. wdyt @dmvolod ?

pkg/controller/infinispan/infinispan_controller.go
rigazilla: yeah sure, thanks!

pkg/controller/infinispan/infinispan_controller.go
rigazilla: but yeah, also update the struct with the actual values is a good idea

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I would prefer to use standard types, there are other kind of services like "External Name" (?) that maybe we want to support.

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I would go for your solution @dmvolod, we already provide selector to join service to Infinispan

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, sure. If we will cut to custom types, we will restrict it in the future. Let's keep the current implementation.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -282,12 +282,23 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+				err = r.client.Update(context.TODO(), infinispan)
+				if err != nil {
+					reqLogger.Error(err, "failed to Infinispan with Service spec", "Service", ser)
+					return reconcile.Result{}, err
+				}
+				reqLogger.Info("Created External Service", "Service", externalService)
 			}
-			reqLogger.Info("Created External Service", "Service", externalService)
+		} else {
+			reqLogger.Info("External Service NOT created. Unsupported type?", "Service", nil)
```
Crumby: This part triggers when `isExposed == false` and `isExposed` checks `infinispan.Spec.Expose.Type != ""`.

This means that if user doesn't want service to be exposed (he doesn't define the expose type) then the Operator log will inform him that it didn't really created a service and it's maybe because it's type is unsupported. I think this may cause a confusion.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1585,18 +1595,34 @@ func (r *ReconcileInfinispan) serviceExternal(m *infinispanv1.Infinispan) *corev
 			Namespace: m.ObjectMeta.Namespace,
 			Labels:    lsService,
 		},
-		Spec: corev1.ServiceSpec{
+	}
+
+	if externalServiceType == corev1.ServiceTypeLoadBalancer {
```
Crumby: Isn't the `LoadBalancer` type missing ports definition? Especially `targetPort`?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1585,18 +1595,34 @@ func (r *ReconcileInfinispan) serviceExternal(m *infinispanv1.Infinispan) *corev
...
 					TargetPort: intstr.FromInt(11222),
-					// Fix NodePort to match exported port in kind configuration
-					NodePort: int32(30222),
+					NodePort:   portNum,
 				},
 			},
-		},
+		}
+	} else {
+		// Service type currently unsupported
```
Crumby: This looks like the place for unsupported type log entry.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1585,18 +1595,34 @@ func (r *ReconcileInfinispan) serviceExternal(m *infinispanv1.Infinispan) *corev
...
+		externalService.Spec = corev1.ServiceSpec{
 			Type:     externalServiceType,
 			Selector: lsPodSelector,
 			Ports: []corev1.ServicePort{
 				{
 					Port:       int32(11222),
 					TargetPort: intstr.FromInt(11222),
-					// Fix NodePort to match exported port in kind configuration
-					NodePort: int32(30222),
+					NodePort:   portNum,
```
Crumby: Have you considered not defining the NodePort at all? This part will be added automatically by OpenShift if the service type will be NodePort. This would allow you to have the definition same way as before and would have save a lot of code lines.

Crumby: If the node port specification would be required feature present in Infinispan CR then you could just check that the service is of type NodePort with node port set explicitly and add it extra. Also this would allow you too keep the support for all the service types as it was.

test/e2e/main_test.go
```diff
@@ -449,7 +449,8 @@ func TestExternalService(t *testing.T) {
 
 func exposeServiceSpec() corev1.ServiceSpec {
 	return corev1.ServiceSpec{
-		Type: exposeServiceType(),
+		Type:  exposeServiceType(),
+		Ports: []corev1.ServicePort{{NodePort: 30222}},
```
Crumby: Having NodePort hard-coded in the tests is what could cause issues on particular platforms. Wouldn't it be better to retrieve actual port used after service creation?

pkg/controller/infinispan/infinispan_controller.go
Crumby: And if the NodePort should be part of the Infinispan CR then design doc should be updated.

pkg/controller/infinispan/infinispan_controller.go
rigazilla: yeah, the `else` should be in the inner block. Thanks!

pkg/controller/infinispan/infinispan_controller.go
rigazilla: yep, tnx!

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I've added logging in the caller, we have no logger here

test/e2e/main_test.go
rigazilla: I have to check how kind works in CI, it seems it need 30222

rigazilla: yep it's configured at cluster startup. I would keep this and provide more configurability in a separated PR if needed

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I've reorganized the switch code.
The `.Expose` field is a v1.ServiceSpec so it implicitly contains NodePort info.

tbh I would leave to the user the task to expose things in the k8s proper way. atm we just save the user from the effort to write 11222 two times

Crumby:
@rigazilla The code itself looks good too me. But we should still update [design.doc](https://github.com/infinispan/infinispan-operator/blob/master/doc/design.adoc) so the possibility to set up NodePort in `infinispan` cr explicitly is clear?

rigazilla:
> @rigazilla The code itself looks good too me. But we should still update [design.doc](https://github.com/infinispan/infinispan-operator/blob/master/doc/design.adoc) so the possibility to set up NodePort in `infinispan` cr explicitly is clear?

@pdrobek yeah, I've updated the doc

doc/design.adoc
```diff
@@ -94,13 +94,16 @@ is gracefully shut down.
...
-the user must indicate via `type` attribute how the
+| If not empty the operator will expose Infinispan port 11222 externally.
+The `spec.type` attribute must indicate how the
 https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#service-v1-core[Service]
-is exposed.
-`NodePort` is often used while testing locally,
-while `LoadBalancer` is used in production-like environments where an external load-balancer is available.
+is exposed. Two service types are supported: `LoadBalancer, NodePort`.
+For the `NodePort` service type the `Spec.Ports[0].nodePort` attribute can be specified, otherwise services
+are exposed with default value.
```
Crumby: `with default value` seems a bit misleading as a node port is selected from the free ports by OpenShift. (What is the default value?)

PR #330, opened 2020-03-12
[1.1.x] NodePort fix. #325

### LINKED ISSUE #325: NodePort is hardcoded and exposed also when not needed

rigazilla:
backport #326.
fixes #325.


rigazilla:
merged, thanks

Issue #329, opened 2020-03-11
Test suite should use https when TLS is autoenable (ie openshift with CA service)


rigazilla:

dmvolod:
@rigazilla I've already done isOpenShift for some tests in keberetes package and can cherry-pick this commit to a new branch, so we can start to work on it not from the scratch.
What do you think about it. Or we need just env variable for this option?

rigazilla:
@dmvolod the idea is to check if infinispan.Spec.Security.EndpointEncryption.Type is not nil,
the operator should set it to "service" if a CA service is available.
I think it's the most safe way to check if TLS is enabled also can survive if we add support for CA service on other platform.

dmvolod:
Oh, @rigazilla . It's much easiest that I thought

rigazilla:
@dmvolod yeah all the discovery work is already done by the operator

PR #331, opened 2020-03-12
Using correct schema protocol for test with auto TLS #329

rigazilla:
This PR select the correct schema protocol (http or https) for testing.
If infinispan.spec.security.endpointencryption.type is not null https is selected.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -90,7 +90,7 @@ func TestNodeStartup(t *testing.T) {
 	// Register it
 	kubernetes.CreateInfinispan(spec, Namespace)
 	defer kubernetes.DeleteInfinispan(spec, SinglePodTimeout)
-	waitForPodsOrFail(spec, "http", 1)
+	waitForPodsOrFail(spec, 1)
 }
 
 // Test if the cluster is working correctly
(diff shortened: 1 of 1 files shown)


dmvolod:
@rigazilla I tried this test on OCP 4.3, but TestClusterFormation still keep PersistentVolumeClaim and didn't pass.
test-cluster-formation-test-cluster-formation-0   Bound     pvc-6122c5de-efca-4fb8-bf46-5cdeaa05fd89
test-cluster-formation-test-cluster-formation-1   Bound     pvc-047434c8-8367-4af4-833f-d1f0dd7756d3
[dvolodin@dvolodin manager]$ oc get infinispan
No resources found.
That is strange, as this test passes on travis.

rigazilla:
@dmvolod I don't think it's related to this PR. How does the test fail?
Can you check if the infinispan resource has the "finalizer.infinispan.org" in the metadata.finalizers?

dmvolod:
Yes, sure, not to this PR. I will check, but block this PR test, however may be problem with my env also.

dmvolod:
Oh, I've found regression in #315 , it creates a finalizer to often, if someone happens with external service creating, it has never reached finalizer, but already require it, as StatefulSet already present and PVC need to be finalized. @rigazilla please pay attention to this issue.

rigazilla:
merged thanks!

PR #335, opened 2020-03-13
Change r.addFinalizer invocation place

### LINKED ISSUE #334: r.addFinalizer invokes to late in Reconcile flow
This is regression of #315 , it creates a finalizer to often, if someone happens with external service creating, it has never reached finalizer, but always require it, as StatefulSet already present and PVC need to be finalized.
It need to be called right away StatefulSet create

dmvolod:
Fixes #334


dmvolod:
merged, thanks @rigazilla

PR #345, opened 2020-03-23
Cache crd

rigazilla:
First step for  #224

Added Cache api 
Added Cache controller
Adjusted files to the new sdk generated file struct
Fixed tags on golang types to work with openAPI
Cleaned up demo login in Reconcile loop.

This PR doesn't provide functional features, these will be added later once this PR will be merged


build/minikube/run-local.sh
```diff
@@ -8,6 +8,9 @@ KUBECONFIG=${2-${HOME}/.kube/config}
 echo "Using KUBECONFIG '$KUBECONFIG'"
 
 kubectl create ns ${NAMESPACE} || true
-kubectl apply -f deploy/rbac.yaml -n ${NAMESPACE}
-kubectl apply -f deploy/crd.yaml -n ${NAMESPACE}
+kubectl apply -f deploy/role.yaml -n ${NAMESPACE}
```
dmvolod: @rigazilla what do you think about to simplify with one command against a folder 
```suggestion
kubectl apply -f deploy/ -n ${NAMESPACE}
```
If you agree, it can be use in multiple places

rigazilla: this will deploy `operator.yaml` also, which is not required by the script.
Overall it seems to me a little bit unsafe to deploy everything is in a folder.

dmvolod: Yes, it's a little bit unsafe for if we will have a large number of files, manual install will be harder
May be just change in the doc only or for **crds** folder only?

rigazilla: I would prefer to provide precise steps in the doc and let the user to go for shortcuts.
`crds` folder contains cache CR example (generated by the sdk?)

I'm say this because we have only 2 CRDs atm, in the future we will probably want to move to your approach.

dmvolod: Ok, good point.

dmvolod:
@rigazilla are you ready to merge this changes?
I'm preparing some code refactoring to fit this PR with new folder structure.

rigazilla:
merged, thanks!

PR #348, opened 2020-03-25
Change partition strategy for safe restart

rigazilla:
This PR changes partition handling policy to ALLOW_READ_WRITE which is more RollingUpdate friendly


rigazilla:
trivia, merged

Issue #305, opened 2020-03-04
Increase RequeueAfter interval as constant or multiplicator each reconcile 


dmvolod:
On the huge cluster the startup time of the one Infinispan node with default CPU limits still about
**Infinispan Server 10.1.2.Final started in 15916ms**
We should increase RequeueAfter interval as constant or multiply it by 2 every reconcile state as we can see a large number of iterations on bootstrap in the operator log

{"level":"info","ts":1583304923.3194888,"logger":"controller_infinispan","msg":"Reconciling Infinispan. Operator Version: 8.0.0.ER4-6-g3b15baf","Request.Namespace":"my-project","Request.Name":"example-infinispan"}
{"level":"info","ts":1583304923.319542,"logger":"controller_infinispan","msg":"Configuring the StatefulSet","Request.Namespace":"my-project","Request.Name":"example-infinispan"}
{"level":"info","ts":1583304923.4219995,"logger":"controller_infinispan","msg":"Creating identity secret","Request.Namespace":"my-project","Request.Name":"example-infinispan"}
{"level":"info","ts":1583304923.4220781,"logger":"controller_infinispan","msg":"Creating a new Secret","Request.Namespace":"my-project","Request.Name":"example-infinispan","Secret.Name":"example-infinispan-generated-secret"}
{"level":"info","ts":1583304923.5307214,"logger":"controller_infinispan","msg":"Security set","Request.Namespace":"my-project","Request.Name":"example-infinispan","Spec.EndpointSecretName":"example-infinispan-generated-secret","Status.EndpointSecretName":"example-infinispan-generated-secret"}
{"level":"info","ts":1583304923.5307539,"logger":"controller_infinispan","msg":"Creating a new ConfigMap","Request.Namespace":"my-project","Request.Name":"example-infinispan","ConfigMap.Name":"example-infinispan-configuration"}
{"level":"info","ts":1583304923.536818,"logger":"controller_infinispan","msg":"Creating a new StatefulSet","Request.Namespace":"my-project","Request.Name":"example-infinispan","StatefulSet.Name":"example-infinispan"}
{"level":"info","ts":1583304923.5770628,"logger":"controller_infinispan","msg":"Created Service","Request.Namespace":"my-project","Request.Name":"example-infinispan","Service":{"namespace":"my-project","name":"example-infinispan-ping"}}
[...]

PR #351, opened 2020-03-27
Extended delay, removed some requeue #305

rigazilla:
Removed some requeue since the Reconcile loop is already triggered from events coming from the StatefulSet.
Extended delay on cluster formation check.


rigazilla:
merged

PR #352, opened 2020-03-27
Push docker image to the OCP|OKD4 and deploy operator

dmvolod:


build/push-okd4.sh
```diff
@@ -0,0 +1,56 @@
...
+  validateROOT
+  sudo -E mkdir -p "${DOCKER_CERTS_FOLDER}${REGISTRY_ROUTE}"
+fi
+
+if [ -f "${DOCKER_CERTS_FOLDER}${REGISTRY_ROUTE}/tls.crt" ]; then
+  echo "Instance file detected"
+else
+  echo "Instance file not found, extracting..."
+  validateROOT
+  sudo -E oc extract secret/router-ca --keys=tls.crt -n openshift-ingress-operator --to="${DOCKER_CERTS_FOLDER}${REGISTRY_ROUTE}"
```
rigazilla: I have oc installed under $HOME/bin and it's not found by sudo:
```
$ sudo -E oc
sudo: oc: command not found
```
I think is quite common to have oc not in the default path, how about change to:
```
OC=$(which oc)
sudo -E $OC
```
?

build/push-okd4.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+validateOC() {
+  OC_USER_NAME=$(oc whoami 2>/dev/null)
```
rigazilla: Strange problem here for me:
```
$ oc whoami 
kube:admin
```
but 
```
docker login -u kube:admin
```
fails and
```
docker login -u kubeadmin
```
works

build/push-okd4.sh
```diff
@@ -0,0 +1,56 @@
...
+
+if [ -f "${DOCKER_CERTS_FOLDER}${REGISTRY_ROUTE}/tls.crt" ]; then
+  echo "Instance file detected"
+else
+  echo "Instance file not found, extracting..."
+  validateROOT
+  sudo -E oc extract secret/router-ca --keys=tls.crt -n openshift-ingress-operator --to="${DOCKER_CERTS_FOLDER}${REGISTRY_ROUTE}"
+fi
+
+docker login -u "$OC_USER_NAME" -p "${OC_USER_TOKEN}" "${REGISTRY_ROUTE}"
```
Crumby: `oc registry login --skip-check` does the same job once you are logged in and you can drop the env vars.

dmvolod: Thanks, @Crumby will update
It'a strange, that out KB note related to this problem, doesn't know about this option

build/push-okd4.sh
dmvolod: @rigazilla, updated without sudo with solution provided by @Crumby

dmvolod: Yes, thanks for suggestion, will update

build/push-okd4.sh
dmvolod: @rigazilla will update with solution provided by Pavel

build/push-okd4.sh
```diff
@@ -0,0 +1,54 @@
...
+  fi
+}
+
+validateOC
+
+DOCKER_CERTS_FOLDER="/etc/docker/certs.d/"
+REGISTRY_ROUTE=$(oc get route default-route -n openshift-image-registry -o json | jq -r '.spec.host')
+OC_PROJECT_NAME=$(oc project -q)
+PROJECT_NAME=${PROJECT_NAME-$OC_PROJECT_NAME}
+VERSION=$(git describe --tags --always --dirty)
```
rigazilla: Can we make it configurable as in build.sh?
`VERSION=${RELEASE_NAME-$(git describe --tags --always --dirty)}`

dmvolod: Thanks, done

PR #358, opened 2020-03-30
Fix push docker image to the OCP|OKD4 and deploy operator

dmvolod:


dmvolod:
@rigazilla could you please confirm that we need to delete deploy/crd.yaml from master?

rigazilla:
> @rigazilla could you please confirm that we need to delete deploy/crd.yaml from master?

yeah @dmvolod, all the crd are now in the `crds` folder

dmvolod:
Thanks @rigazilla , as you confirmed I'm merging with deploy/crd.yaml file removing

PR #360, opened 2020-03-31
Create cache 356

rigazilla:
This PR allows to create a cache on an existing cluster.

Procedure:
- check that `deploy/cr/cache/auth_secret.yaml` has the correct credential and deploy it
- check that `deploy/cr/cache/cache_1.yaml` has the correct cluster name and deploy it

once created cache will be described as:
```
apiVersion: infinispan.org/v2alpha1
kind: Cache
spec:
  adminAuth:
    secretName: basic-auth
  cacheName: mycache
  clusterName: example-infinispan
status:
  conditions:
  - status: "True"
    type: Ready
  serviceName: example-infinispan
```
Only if the target cluster is working in DataGrid mode, either a `templateName` or an xml `template` can be specified in the .Spec. `templateName` is used first if present.
The `basic-auth` secret must be deployed and must contain two entries: `username` and `password`, optionally `AdminAuth` field could be expressed as a pair of SecretKeySelector:
```
adminAuth
    password:
      key: password
      name: basic-auth
    username:
      key: username
      name: basic-auth
```


rigazilla:
@tristantarrant please check the use case above. More sofisticated cases will follow.

pkg/controller/cache/cache_controller.go
```diff
@@ -92,10 +111,163 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
...
+		if errors.IsNotFound(err) {
+			reqLogger.Error(err, "Infinispan cluster "+ispnInstance.Name+" not found")
+			return reconcile.Result{RequeueAfter: time.Second * 10}, err
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// Cluster must be well formed
+	for i, c := range ispnInstance.Status.Conditions {
```
dmvolod, line 130: @rigazilla what do you think if we will create IsWellFormed method during refactoring on v1.Infinispan type?

pkg/controller/cache/cache_controller.go
```diff
@@ -92,10 +111,163 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
...
+		return reconcile.Result{}, err
+	}
+	cacheName := instance.Name
+	if instance.Spec.Name != "" {
+		cacheName = instance.Spec.Name
+	}
+	caches, err := cluster.CacheNames(user, pass, podList.Items[0].Name, instance.Namespace, ispnctrl.Protocol(ispnInstance))
+	if err == nil {
+		for i, cn := range caches {
+			if cn == cacheName {
```
dmvolod, line 160: Can we check that cache is exists via API?
https://infinispan.org/docs/stable/titles/rest/rest.html#rest_v2_cache_exists

pkg/controller/cache/cache_controller.go
```diff
@@ -92,10 +111,163 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
...
+				xmlTemplate := instance.Spec.Template
+				if xmlTemplate == "" {
+					xmlTemplate, err = ispnctrl.GetDefaultCacheTemplateXML(podName, ispnInstance, cluster, reqLogger)
+				}
+				if err != nil {
+					reqLogger.Error(err, "Error getting default XML")
+					return reconcile.Result{}, err
+				}
+				reqLogger.Info(xmlTemplate)
+				err = cluster.CreateCacheWithAuth(user, pass, instance.Spec.Name, xmlTemplate, podName, instance.Namespace, ispnctrl.Protocol(ispnInstance))
```
dmvolod, line 177: Can we  create a Cache with different parameters more generic as we need more methods after Auth, Template, Transactions, etc.

pkg/controller/cache/cache_controller.go
```diff
@@ -92,10 +111,163 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	// Search the service associated to the cluster
+	serviceList := &corev1.ServiceList{}
+	labelSelector = labels.SelectorFromSet(ispnutil.GetLabelsForInfinispan(ispnInstance.Name, "infinispan-service"))
+	listOps = &client.ListOptions{Namespace: ispnInstance.Namespace, LabelSelector: labelSelector}
+	err = r.client.List(context.TODO(), serviceList, listOps)
+	if err != nil {
+		reqLogger.Error(err, "failed to select cluster service")
+		return reconcile.Result{}, err
+	}
+	if len(serviceList.Items) != 1 {
```
dmvolod, line 200: @rigazilla why we need this check? If needed, we should did this check on the start?

pkg/controller/cache/cache_controller.go
```diff
@@ -92,10 +111,163 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
...
+		if err != nil {
+			reqLogger.Error(err, "Unable to update Cache "+instance.Name+" status")
+			return reconcile.Result{}, err
+		}
+	}
 	return reconcile.Result{}, nil
 }
+
+// SetCondition set condition to status
+func setCondition(cache *v2alpha1.Cache, condition string, status metav1.ConditionStatus, message string) bool {
```
dmvolod, line 223: Looks like methods like this may be on top of the v2alpha1.Cache entity?

pkg/controller/cache/cache_controller.go
```diff
@@ -92,10 +111,163 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
...
+	// List the pods for this infinispan's deployment
+	podList := &corev1.PodList{}
+	labelSelector := labels.SelectorFromSet(util.GetLabelsForInfinispan(ispnInstance.Name, ""))
+	listOps := &client.ListOptions{Namespace: ispnInstance.GetClusterName(), LabelSelector: labelSelector}
+	err = r.client.List(context.TODO(), podList, listOps)
+	if err != nil || (len(podList.Items) == 0) {
+		reqLogger.Error(err, "failed to list pods", "Infinispan.Namespace")
+		return reconcile.Result{}, err
+	}
+	cacheName := instance.Name
```
dmvolod, line 153: We should create GetName on to of the v2alpha1.Cache or will done during refactor

pkg/controller/cache/cache_controller.go
rigazilla: sounds good

pkg/controller/cache/cache_controller.go
rigazilla: Do we really need a getter in this specific case?

pkg/controller/cache/cache_controller.go
rigazilla: yeah, good idea. I'll do that

pkg/controller/cache/cache_controller.go
rigazilla: shouldn't the cache configuration completely described in the template?

rigazilla: There are other cache options outside the template?

pkg/controller/cache/cache_controller.go
rigazilla: yeah it's probably superfluous

pkg/controller/cache/cache_controller.go
rigazilla: ok ,will do

rigazilla:
@dmvolod I would like to eval the effort of rebasing this on top of your #363. I'll do that in the afternoon

pkg/controller/cache/cache_controller.go
dmvolod: Yes, @rigazilla . If we will not create a new CreateCacheWith* it's ok. But maybe create generic CreateCache with auth as no required options?

pkg/controller/cache/cache_controller.go
dmvolod: I'm not sure, but if we will use more than once, it's better to have a separate method. Or we could potential change a logic of any complex getter like this in the future.

dmvolod:
Thanks, @rigazilla . You are welcome to merge it "as is" after review and it's not to hard to rebase with refactoring as it on the Stage I only and has not short story.

dmvolod:
@rigazilla I will wait until you are not rebasing and merging this PR into the master for the new refactoring stage.

pkg/controller/cache/cache_controller.go
rigazilla: done

pkg/controller/cache/cache_controller.go
rigazilla: I'm adding *WIthAuth methods when needed because current methods imply the access to the secret containing all the credentials configured on the cluster. That secret should be used only to configure the users.properties file on the pod and caller should provide it's own credential.
I would like to remove old methods in a next PR and provide optional authentication if needed

pkg/controller/cache/cache_controller.go
rigazilla: removed

pkg/controller/cache/cache_controller.go
rigazilla: done

pkg/controller/cache/cache_controller.go
rigazilla: I would postpone this evaluation: until we're only r/w field I think we can avoid field accessors.

rigazilla:
@dmvolod I've added some changes and comments. Let me know, if ok I'll squash the commits before merging

pkg/controller/infinispan/util/cluster.go
```diff
@@ -153,6 +156,62 @@ func (c Cluster) ExistsCache(cacheName, secretName, podName, namespace, protocol
 	}
 }
 
+// ExistsCacheWithAuth returns true if cacheName cache exists on the podName pod
+func (c Cluster) ExistsCacheWithAuth(user, pass, cacheName, podName, namespace, protocol string) (bool, error) {
```
dmvolod, line 160: @rigazilla looks like **user** var is not using here. Is it correct or not?

pkg/controller/infinispan/util/cluster.go
```diff
@@ -250,3 +351,12 @@ func ClusterStatusHandler(scheme corev1.URIScheme) corev1.Handler {
 			Port:   intstr.FromInt(11222)},
 	}
 }
+
+// GetLabelsForInfinispan returns the labels that must me applied to the resource
+func GetLabelsForInfinispan(name, resourceType string) map[string]string {
```
dmvolod, line 356: @rigazilla are you agree that we will move GetLabelsForInfinispan outside this package as it's not only related to cluster utils?

pkg/controller/cache/cache_controller.go
dmvolod: Ok ,let's keep it as is, but it would be nice in the future or during refactoring to have a common model for all methods.

pkg/controller/cache/cache_controller.go
dmvolod: @rigazilla have you decide to create a IsWellFormed function for this?

rigazilla: yeah,  Let me add it. (I though it was in your refactoring plans.)

rigazilla: @dmvolod done

pkg/controller/infinispan/util/cluster.go
dmvolod: @rigazilla may be you've missed this comment due to large number of interactions?

rigazilla: @dmvolod yeah, thanks. my fault. should bo ok now

pkg/controller/infinispan/util/cluster.go
rigazilla: sounds good, let me find a place in common

dmvolod: yes, sure

pkg/controller/infinispan/util/cluster.go
dmvolod: yes, looks good

pkg/controller/infinispan/util/cluster.go
rigazilla: uh, we already have the right method in common. removed this

dmvolod:
Looks good, Vittorio!
But I've a couple of questions:
- do we need also create Cache entity with name **default** in k8s on Infinispan reconciliation to be consistence with others caches
- not sure, but caches k8s entities could be orphans when infinispan will be deleted

rigazilla:
> Looks good, Vittorio!
> But I've a couple of questions:
> 
>     * do we need also create Cache entity with name **default** in k8s on Infinispan reconciliation to be consistence with others caches
Good point, I think it's ok to reconcile **default** since it has been created by the operator. 

> 
>     * not sure, but caches k8s entities could be orphans when infinispan will be deleted
Yeah, this need a better thought, but we probably need to setup a policy to report cluster status into the cache CR

dmvolod:
thanks, @rigazilla as it were just point of info, rather that this PR implementation cases. We can implement them in the next PR's

rigazilla:
@oraNod from the user pov, do you like the use case (yaml format, steps...) at the beginning?

deploy/cr/cache/cache_1.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: example-cache
+spec:
+  adminAuth:
+    username:
+      key: username
+      name: basic-auth
+    password:
+      key: password
+      name: basic-auth
+  clusterName: example-infinispan
```
oraNod, line 13: do we need `clusterName`?

maybe this is somewhat future looking as the operator only controls one cluster at a time. in the case of controlling multiple clusters across different namespaces, then this could be useful. but, in that case, will it be mandatory that cluster names are unique? clusters in different namespaces cannot have the same name or the operator throws an error?

I just think this is an extra field that is not really needed. less fields, better. if the goal is to make sure that the cache gets created in the right cluster then maybe the k8s namespace is what should be used. make sense?

deploy/crds/infinispan.org_caches_crd.yaml
```diff
@@ -30,9 +30,82 @@ spec:
...
+              type: object
+            clusterName:
+              description: Name of the cluster where to create the cache
+              type: string
+            name:
+              description: Name of the cache to be created. If empty ObjectMeta.Name
+                will be used
+              type: string
+            template:
+              description: Cache template in XML format
```
oraNod, line 80: embedding xml in yaml? 

how about pointing to the location of an xml file instead?

```
template:
  path: /path/to/mycache.xml
```

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -261,9 +261,6 @@ spec:
                   type: object
                 endpointSecretName:
                   type: string
-              required:
-              - endpointEncryption
-              - endpointSecretName
```
oraNod, line 266: just to check but is this change going to affect endpoint encryption CRs?

pkg/apis/infinispan/v2alpha1/cache_types.go
```diff
@@ -1,24 +1,48 @@
...
-	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
-	// Add custom validation using kubebuilder tags: https://book-v1.book.kubebuilder.io/beyond_basics/generating_crd.html
+	// Authentication info
+	AdminAuth AdminAuth `json:"adminAuth,omitempty"`
+	// Name of the cluster where to create the cache
+	ClusterName string `json:"clusterName,omitempty"`
+	// Name of the cache to be created. If empty ObjectMeta.Name will be used
+	Name string `json:"name,optional,omitempty"`
+	// Cache template in XML format
+	Template string `json:"template,optional,omitempty"`
```
oraNod, line 27: can you pass a cache name only without any xml template and get a local, basic cache? it makes no sense and I'm not sure how it would work. maybe it would just create the cache on the node that gets the request. now that I think about it that's a really bad idea.

so no cache template = no cache gets created?

will there be any sort of check to ensure the template is appropriate? as in my example, if you pass a local cache definition, will the operator create it on the cluster or throw an error?

what about creating caches from the infinispan.org templates?

with the CLI you can do "create cache --template=org.infinispan.DIST_SYNC mycache" 

for the Operator, I'd like to be able to do something like:

```
apiVersion: infinispan.org/v2alpha1
kind: Cache
metadata:
  name: mycache
spec:
  template: org.infinispan.DIST_SYNC
```

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -855,11 +848,17 @@ func createCacheServiceDefaultCache(podName string, infinispan *infinispanv1.Inf
...
+func createCacheServiceDefaultCache(podName string, infinispan *infinispanv1.Infinispan, cluster ispnutil.ClusterInterface, logger logr.Logger) error {
+	defaultCacheXML, err := GetDefaultCacheTemplateXML(podName, infinispan, cluster, logger)
+	if err != nil {
+		return err
+	}
 	secretName := infinispan.GetSecretName()
-	protocol := infinispanProtocol(infinispan)
-	return cluster.CreateCache("default", defaultCacheXML, secretName, podName, namespace, protocol)
+	protocol := infinispan.GetEndpointScheme()
+	return cluster.CreateCache("default", defaultCacheXML, secretName, podName, infinispan.Namespace, string(protocol))
```
oraNod, line 861: not sure what's going on here but it leads me to another question. for Cache Service, we will create copies of the default cache only, right?

so in effect, for Cache Service all that is allowed is to pass cache_1.yaml which creates a copy of the default cache with a new name. if you attempt to add an XML definition or cache template, the operator throws an error?

deploy/cr/cache/cache_1.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: example-cache
+spec:
+  adminAuth:
+    username:
+      key: username
+      name: basic-auth
+    password:
+      key: password
+      name: basic-auth
```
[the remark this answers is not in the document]
rigazilla: You're right I think I can go for your second proposal.
I've implemented a common method to access credential (Prometheus does something similar), it's probably a way to give the user more flexibility. I'm going to change the code adding some default values so for the user will be enough to provide only the secret name.

deploy/cr/cache/cache_1.yaml
rigazilla: Operator is (should be) able to handle several clusters in one namespace, that's why the cluster name is mandatory.
In the future we would probably add some more intelligence to the operator so it will be able to pick up the right cluster for the user cache.

deploy/crds/infinispan.org_caches_crd.yaml
rigazilla: this is something that needs implementation on the oc (or k8s client)

pkg/apis/infinispan/v2alpha1/cache_types.go
rigazilla: If no template is provided a cache is created with the same `default` cache template.
I'm not sure if a template name like `org.infinispan.DIST_SYNC` can be passed as REST argument, I'll work on this

oraNod: maybe:

----
POST /rest/v2/caches/mycache?template=org.infinispan.DIST_SYNC
----

deploy/crds/infinispan.org_infinispans_crd.yaml
rigazilla: No, those fields are not actually required. We have cr verification in 2.0.0 se we had to correctly specify they are optional. Usage is the same.

deploy/cr/cache/cache_1.yaml
oraNod: thanks! several clusters in one namespace makes sense.

deploy/crds/infinispan.org_caches_crd.yaml
oraNod: maybe the work @tristantarrant has done on the CLI could help us here? if my understanding is correct this CRD lets users wrap infinispan.xml in yaml, which just feels sort of hideous for complex configuration. I think that is for the power users who would be more likely to adopt a CLI approach for the heavy configuration.

to create caches via the Operator through the Infinispan CR maybe it should just be with one of the Infinispan cache templates for the DG service or a copy of the default cache for the Cache service?

deploy/cr/cache/cache_1.yaml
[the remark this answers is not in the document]
rigazilla: @oraNod I've updated the use case

pkg/controller/infinispan/infinispan_controller.go
rigazilla: The change itself is just cosmetic.
The `default` cache is created by default when the cluster is deployed, user can create caches of any kind later.

pkg/apis/infinispan/v2alpha1/cache_types.go
rigazilla: yeah good, going to implement

rigazilla: @oraNod done

pkg/controller/infinispan/util/cluster.go
```diff
@@ -153,6 +157,62 @@ func (c Cluster) ExistsCache(cacheName, secretName, podName, namespace, protocol
...
+}
+
+// CacheNames return the names of the cluster caches available on the pod `podName`
+func (c Cluster) CacheNames(user, pass, podName, namespace, protocol string) ([]string, error) {
+	podIP, err := c.Kubernetes.GetPodIP(podName, namespace)
+	if err != nil {
+		return nil, err
+	}
+	httpURL := fmt.Sprintf("%s://%s:11222/rest/v2/caches", protocol, podIP)
+	commands := []string{"curl", "--insecure", "-u", fmt.Sprintf("operator:%v", pass), httpURL}
```
dmvolod, line 202: @rigazilla please change with user instead of operator constant

pkg/controller/infinispan/util/cluster.go
```diff
@@ -198,6 +258,88 @@ func (c Cluster) CreateCache(cacheName, cacheXml, secretName, podName, namespace
...
+	}
+
+	httpURL := fmt.Sprintf("%s://%s:11222/rest/v2/caches/%s", protocol, podIP, cacheName)
+	commands := []string{"curl",
+		"--insecure",
+		"--http1.1",
+		"-w", "\n%{http_code}", // add http status at the end
+		"-d", fmt.Sprintf("%s", cacheXML),
+		"-H", "Content-Type: application/xml",
+		"-u", fmt.Sprintf("operator:%v", pass),
```
dmvolod, line 275: @rigazilla please change with user instead of operator constant

pkg/controller/infinispan/util/cluster.go
```diff
@@ -198,6 +258,88 @@ func (c Cluster) CreateCache(cacheName, cacheXml, secretName, podName, namespace
...
+	if err != nil {
+		return err
+	}
+
+	httpURL := fmt.Sprintf("%s://%s:11222/rest/v2/caches/%s?template=%s", protocol, podIP, cacheName, templateName)
+	commands := []string{"curl",
+		"--insecure",
+		"--http1.1",
+		"-w", "\n%{http_code}", // add http status at the end
+		"-u", fmt.Sprintf("operator:%v", pass),
```
dmvolod, line 315: @rigazilla please change with user instead of operator constant

pkg/controller/infinispan/util/cluster.go
```diff
@@ -198,6 +258,88 @@ func (c Cluster) CreateCache(cacheName, cacheXml, secretName, podName, namespace
...
+}
+
+// CreateCacheWithTemplateName create cluster cache on the pod `podName`
+func (c Cluster) CreateCacheWithTemplateName(user, pass, cacheName, templateName, podName, namespace, protocol string) error {
+	podIP, err := c.Kubernetes.GetPodIP(podName, namespace)
+	if err != nil {
+		return err
+	}
+
+	httpURL := fmt.Sprintf("%s://%s:11222/rest/v2/caches/%s?template=%s", protocol, podIP, cacheName, templateName)
```
dmvolod, line 310: @rigazilla, looks like we need to check which error will display if template is not exists or incorrect otherwise check with 
GET /rest/v2/cache-managers/{cacheManagerName}/cache-configs/templates before
or we can keep and show available templates list before

pkg/controller/infinispan/util/cluster.go
```diff
@@ -198,6 +258,88 @@ func (c Cluster) CreateCache(cacheName, cacheXml, secretName, podName, namespace
...
+
+	// Split lines in standard output, HTTP status code will be last
+	execOutLines := strings.Split(execOut.String(), "\n")
+
+	httpCode, err := strconv.ParseUint(execOutLines[len(execOutLines)-1], 10, 64)
+	if err != nil {
+		return err
+	}
+
+	if httpCode > 299 || httpCode < 200 {
```
dmvolod, line 334: Oh, looks like we will display very common error if template is not exists or incorrect

pkg/controller/infinispan/util/cluster.go
```diff
@@ -30,9 +30,13 @@ type ClusterInterface interface {
 	GracefulShutdown(secretName, podName, namespace, protocol string) error
 	GetClusterMembers(secretName, podName, namespace, protocol string) ([]string, error)
 	ExistsCache(cacheName, secretName, podName, namespace, protocol string) bool
+	ExistsCacheWithAuth(user, pass, cacheName, podName, namespace, protocol string) (bool, error)
 	CreateCache(cacheName, cacheXml, secretName, podName, namespace, protocol string) error
+	CreateCacheWithAuth(user, pass, cacheName, cacheXML, podName, namespace, protocol string) error
+	CreateCacheWithTemplateName(user, pass, cacheName, templateName, podName, namespace, protocol string) error
```
dmvolod, line 36: @rigazilla maybe change name with common pattern (CreateCacheWith*), like CreateCacheWithTemplate
As we have already a large number of combinations create a common one function with all attrs, like user, pass and templateName and define logic inside depends of the provided values (nil or not nil)
wdyt?

rigazilla: ok for the name change, but I would prefer to keep small separated methods with clear interface instead of having 1 big func with all the logic mixed inside.

dmvolod: Ok, good point, let's keep your strategy as it can be easy to maintain

pkg/controller/infinispan/infinispan_controller.go
oraNod: oh really? I thought users could only create copies of the default when using Cache service. so you could create caches with any definition? imo that seems to defeat the purpose of the Cache service.

pkg/controller/infinispan/util/cluster.go
rigazilla: the ISPN error is reported in the err object and logged by the operator.
I would leave improved error handling in a next PR, maybe we want to show error also in the cr status.

pkg/controller/infinispan/util/cluster.go
rigazilla: see comment above

pkg/controller/infinispan/util/cluster.go
rigazilla: since this impact method used in the infinispan reconciler I would do this in a next PR together with:
- removing methods that access secret
- removing hardcoded `operator` user

pkg/controller/infinispan/infinispan_controller.go
rigazilla: mmm, I see your concern. We can add a check that blocks CR cache creation with arbitrary template, but user will still be able to create them programmatically

rigazilla: @oraNod I've added the check to block cache with templates on CacheService.

pkg/controller/infinispan/util/cluster.go
dmvolod: @rigazilla sorry for disturb, I don't remember if you decide to update this  in future PR as we have very long discussion

rigazilla: @dmvolod yeah I'll clean up all the methods in a next refactoring PR. 
This PR is trying to get out of control :)

pkg/controller/infinispan/infinispan_controller.go
rigazilla: @oraNod I've update the use case at the top with the limitation for the CacheService.

oraNod: @rigazilla thanks. that looks good!

rigazilla:
merge thanks all!

rigazilla:
Arrgh... I forgot to squash the commits :/

dmvolod:
@rigazilla I think you can squash this locally on maater and re-push with -force option

rigazilla:
yeah thanks @dmvolod but I won't take the habit to force push to origin/master. Not a big problem, let's leave my mistake there

dmvolod:
Oh, thanks @rigazilla , I don't think that is mistake. But yes, you are right, force push on origin/master is a little bit dangerous if it's not personal repo.

PR #363, opened 2020-04-01
Move entities model outside the infinispan_controller code (Stage I)

dmvolod:
This is the just first stage (Stage I) of the code refactoring (only move some simple methods and introduce constants)
This is part of the issue #290
Next stages will be:
Stage II - move common functions from infinispan_controller to the separate pakcages
Stage III - implement model (secrets, services, configmap) outside infinispan_controller
Stage IV - try to split kubernetes and cluster entities
Stage V - simplify using kubernetes client and possible to use https://github.com/redhat-cop/operator-utils

PR #365, opened 2020-04-07
Cache creation (use case 1)

rigazilla:
rebasing #360 on top of #363


rigazilla:
changes moved back to #360 to maintain discussion.
closing this

PR #378, opened 2020-04-24
160 scale memory

rigazilla:
This allows the user to enable autoscale for the default cache of the CacheService. 
Infinispan CR has an additional field like this:
```
spec:
  autoscale:
    maxMemUsagePercent: 70
    maxReplicas: 5
    minMemUsagePercent: 30
    minReplicas: 2
```
Autoscaling is based on this metric:
` statistics_data_memory_used/configuration_eviction_size`

if metric is over `maxMemUsagePercent` for one pod the a new pod is started
if metric is below `minMemUsagePercent` for all the pods a pod is shutdown

User can enable/disable autoscaling adding or deleting the `autoscale` field from the CR.

@oraNod usage info for the `CacheService default cache autoscaling`


pkg/controller/infinispan/autoscale.go
```diff
@@ -0,0 +1,174 @@
...
+		time.Sleep(defaultMinimumPollPeriod)
+		ispnList := infinispanv1.InfinispanList{}
+		// Check all the cluster in the namespace for autoscaling
+		err := mgr.GetClient().List(context.TODO(), &ispnList)
+		if err != nil {
+			log.Error(err, "Unable to select cluster pods list")
+			continue
+		}
+		for _, iItem := range ispnList.Items {
+			if !iItem.IsConditionTrue("wellFormed") {
```
dmvolod, line 36: @rigazilla looks like we should check that auto-scaling is enabled ASAP early, than cluster is formed to prevent unnecessary work.
Wdyt to move Autoscale and/or any other optional parameters options to the separate CRD and apply it if only needed?

rigazilla: @dmvolod yeah thanks! I've also added a check that the cluster is running a CacheService. Since only that is supported atm.

What do you mean with separated CRD?

dmvolod: We can create a new resource, for example called InfinispanOptions or InfinispanFeatures or InfinispanControl or anything else and define optional parameters there to keep Infinispan cluster CR simple.
Wdyt?

rigazilla: I would try to reach this using the Cache/Infinispan CRDs pair:
- promoting the Cache CR as simple interface for user/application developer
- leaving all the configuration stuff in the Infinispan CR

Actually, I would like to simplify Infinispan CRD itself, but before split data in an external struct I wuold like to try:
- a proper usage of json `omitempty,optional` tags
- a proper definition of optional fields as pointers instead of structs

dmvolod: Ok, @rigazilla sure. Let's try to optimize this option.

pkg/controller/infinispan/autoscale.go
```diff
@@ -0,0 +1,178 @@
...
+	for true {
+		time.Sleep(defaultMinimumPollPeriod)
+		ispnList := infinispanv1.InfinispanList{}
+		// Check all the cluster in the namespace for autoscaling
+		err := mgr.GetClient().List(context.TODO(), &ispnList)
+		if err != nil {
+			log.Error(err, "Unable to select cluster pods list")
+			continue
+		}
+		for _, iItem := range ispnList.Items {
```
dmvolod, line 35: @rigazilla looks like sequential processing is a little bit dangerous if one of the working step in multiple cluster env is hung. Wdyt to split every cluster processing as they are independent of each other in separate goroutines and interact with channels?

rigazilla: I see your point. Yeah, maybe we can make things more robust.

rigazilla: @dmvolod wdyt about a goroutine per cluster? We can use the recocile loop to start and stop the scale monitor goroutine during cluster creation/destruction

dmvolod: Yes, @rigazilla it'a s good options. I collect some ideas, we have check on implementation
- verify that only one thread per cluster is working simultaneously
- avoid conflicts between auto- and manual scaling
- validate up-scaling/descaling rate to avoid it doing very frequently

pkg/controller/infinispan/autoscale.go
```diff
@@ -0,0 +1,204 @@
+package infinispan
```
dmvolod, line 1: @rigazilla wdyt to place it in separate package outside the controller, for example **autoscale**?

pkg/controller/infinispan/autoscale.go
```diff
@@ -0,0 +1,204 @@
...
+	autoscaleThreadPool.Unlock()
+	if !ok {
+		autoscaleThreadPool.Lock()
+		autoscaleThreadPool.m[clusterNsn] = 1
+		autoscaleThreadPool.Unlock()
+		go autoscalerLoop(clusterNsn, r)
+	}
+}
+
+var defaultMinimumPollPeriod = time.Second
```
dmvolod, line 36: maybe place to constants package?

pkg/controller/infinispan/autoscale.go
```diff
@@ -0,0 +1,204 @@
...
+
+func autoscalerLoop(clusterNsn types.NamespacedName, r *ReconcileInfinispan) {
+	log.Info(fmt.Sprintf("Starting loop for autoscaling on cluster %v", clusterNsn))
+	for true {
```
dmvolod, line 40: Still worries about to heavy load on this loop side
The initial loop must be as lighter as possible, and pass huge work to the separate thread for processing

pkg/controller/infinispan/autoscale.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+var autoscaleThreadPool = struct {
+	sync.RWMutex
+	m map[types.NamespacedName]int
+}{m: make(map[types.NamespacedName]int)}
+
+// Starting a go routine that does polling autoscaling on an Infinispan cluster
+func addAutoscalingEquipment(clusterNsn types.NamespacedName, r *ReconcileInfinispan) {
+	autoscaleThreadPool.Lock()
```
dmvolod, line 25: @rigazilla looks like this is good implementation and avoid conflicts between separate clusters

pkg/controller/infinispan/autoscale.go
dmvolod: But we could check all cases with manual tests later and detect all possibilities.

pkg/controller/infinispan/autoscale.go
rigazilla: right, done

pkg/controller/infinispan/autoscale.go
rigazilla: this creates some `import *` issues and I don't know if the autoscale has some utility outside from infinispan atm

pkg/controller/infinispan/autoscale.go
rigazilla: this is a tradeoff between simplicity and concurrency. I'm not sure the loop is using a lot of resources, most of the time is waiting for response from the net.

pkg/controller/infinispan/autoscale.go
dmvolod: Oh, yes, cyclic dependency issue.

pkg/controller/infinispan/autoscale.go
dmvolod: Yes, as core logic is implemented, always can change it if problems will occurs.

oraNod:
@rigazilla hey, not sure from the code but this seems like it could potentially lead to data loss if users set bad combinations. I know the cache service is volatile anyway but say you set the following:

```
spec:
  autoscale:
    maxMemUsagePercent: 70
    maxReplicas: 2
    minMemUsagePercent: 30
    minReplicas: 2
```

this is the same as setting `replicas: 2` and configuring the service to evict entries when the data container reaches 70% capacity, right?

in which case existing entries get removed to make space for new entries, but the data container only uses 70%. does the Operator check for weird conditions like this and throw any sort of warning?

rigazilla:
> @rigazilla hey, not sure from the code but this seems like it could potentially lead to data loss if users set bad combinations. I know the cache service is volatile anyway but say you set the following:
> 
> ```
> spec:
>   autoscale:
>     maxMemUsagePercent: 70
>     maxReplicas: 2
>     minMemUsagePercent: 30
>     minReplicas: 2
> ```
> 
> this is the same as setting `replicas: 2` and configuring the service to evict entries when the data container reaches 70% capacity, right?
> 
> in which case existing entries get removed to make space for new entries, but the data container only uses 70%. does the Operator check for weird conditions like this and throw any sort of warning?

No the thresholds impact only autoscaling: when data reaches 70% a new pod is spawned if possible; if not possible (that's your case) eviction will happen when data reaches 100% as usual.

rigazilla:
merged, thanks all!

Issue #380, opened 2020-04-29
Change external exposition struct from v1.ServiceSpec to an adhoc one


rigazilla:
Defining an adhoc struct should:
- simplify configuration, showing only configurable fields
- allow exposition with Ingress and Route

PR #381, opened 2020-05-05
380 expose

rigazilla:
This PR allows the user to configure an external service in the Infinispan CR.
Syntax is:
```
spec:
  expose:
    type: LoadBalancer | NodePort
    port: 32033
```
`type` must be `LoadBalancer` or `NodePort`.
`port` is optional, if omitted will be `11222` for LoadBalancer or assigned by the platform for NodePort
This solves #380 point 1. Point 2 is postponed to 2.0.x


Crumby:
@rigazilla I wonder if the `port` should be rather `nodePort` for the transparency? Or do you plan add other expose type in the feature and reuse `port` for non `nodePort` ports?

rigazilla:
> @rigazilla I wonder if the `port` should be rather `nodePort` for the transparency? Or do you plan add other expose type in the feature and reuse `port` for non `nodePort` ports?

@Crumby, I think that `nodePort` is fine, it's strictly related to services, but I don't think we're going to provide in the Infinispan CR a way to customize port for future exposition options (i.e. route or ingress).

I'm going to apply you hint, thanks!

rigazilla:
@Crumby changed to `nodePort`

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -71,14 +71,37 @@ type InfinispanSiteLocationSpec struct {
...
+	// every node, in addition to 'ClusterIP' type.
+	ExposeTypeNodePort ExposeType = "NodePort"
+
+	// ExposeTypeLoadBalancer means a service will be exposed via an
+	// external load balancer (if the cloud provider supports it), in addition
+	// to 'NodePort' type.
+	ExposeTypeLoadBalancer ExposeType = "LoadBalancer"
+
+	// ExposeTypeRoute means Infinispan cluster will be exposed via Route resource
+	ExposeTypeRoute ExposeType = "Route"
```
Crumby, line 96: This new option doesn't seem to be handled later in the code.

test/e2e/main_test.go
```diff
@@ -468,24 +468,22 @@ func TestExternalService(t *testing.T) {
...
-		return corev1.ServiceTypeClusterIP
 	case "NodePort":
-		return corev1.ServiceTypeNodePort
+		return ispnv1.ExposeTypeNodePort
 	case "LoadBalancer":
-		return corev1.ServiceTypeLoadBalancer
-	case "ExternalName":
-		return corev1.ServiceTypeExternalName
+		return ispnv1.ExposeTypeLoadBalancer
+	case "ExternalName", "ClusterIP":
```
Crumby, line 485: Could thsi option be just dropped as it's behaviour is same as `default`?

test/e2e/main_test.go
```diff
@@ -468,24 +468,22 @@ func TestExternalService(t *testing.T) {
...
+	return ispnv1.ExposeSpec{
+		Type:     exposeServiceType(),
+		NodePort: 30222,
 	}
 }
 
-func exposeServiceType() corev1.ServiceType {
+func exposeServiceType() ispnv1.ExposeType {
 	exposeServiceType := getEnvWithDefault("EXPOSE_SERVICE_TYPE", "NodePort")
 	switch exposeServiceType {
```
Crumby, line 480: Missing `Route` option?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -285,23 +285,26 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
-				if err != nil && !errors.IsAlreadyExists(err) {
-					reqLogger.Error(err, "failed to create external Service", "Service", ser)
-					return reconcile.Result{}, err
-				}
-				infinispan.Spec.Expose = externalService.Spec
-				err = r.client.Update(context.TODO(), infinispan)
-				if err != nil {
-					reqLogger.Error(err, "failed to Infinispan with Service spec", "Service", ser)
-					return reconcile.Result{}, err
+			switch infinispan.Spec.Expose.Type {
```
Crumby, line 288: Missing `ExposeTypeRoute` option?

pkg/apis/infinispan/v1/infinispan_types.go
rigazilla: I've tried to add support for Route exposition, but openshift packages conflict with the current dependencies set, so I've posponed this feature to 2.x.
It's probably better to remove all the references to Route for now. Going to do.

test/e2e/main_test.go
rigazilla: se above

pkg/controller/infinispan/infinispan_controller.go
rigazilla: see above

test/e2e/main_test.go
rigazilla: yeah, fine

rigazilla:
@oraNod I've updated the docs with this changes

rigazilla:
merged, thanks!

PR #386, opened 2020-05-06
Omit empty struct in yaml output #382

### LINKED ISSUE #382: Do not shown empty values in Infinispan CR
Even with minimal Infinispan deployment spec shows whole possible cr definition with lot of empty string and null values. Common approach is not to write these fields if they are not explicitly specified and consider the defaults.

This can be resolved adding proper json annotation "optional,omitempty" where needed in the types definition

rigazilla:
This partially fixes #382.
`omitempty` doesn't work on struct, so I've changed some `.Spec` fields from struct to pointer to struct.
I've tried to minimize the changes.


pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -47,16 +47,16 @@ const (
...
+	Sites     *InfinispanSitesSpec            `json:"sites,optional,omitempty"`
 }
 
 // InfinispanContainerSpec specify resource requirements per container
 type InfinispanContainerSpec struct {
-	ExtraJvmOpts string `json:"extraJvmOpts"`
-	Memory       string `json:"memory"`
-	CPU          string `json:"cpu"`
+	ExtraJvmOpts string `json:"extraJvmOpts,omitempty"`
+	Memory       string `json:"memory,omitempty"`
```
Crumby, line 58: I don't agree with CPU and Memory being `omitempty` as they are always set to some value. Either user provided or default but never empty. It'd be good to propagate this information back to user.

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -81,13 +81,13 @@ type InfinispanLoggingSpec struct {
...
-	Security  InfinispanSecurity      `json:"security"`
-	Container InfinispanContainerSpec `json:"container"`
-	Service   InfinispanServiceSpec   `json:"service"`
-	Logging   InfinispanLoggingSpec   `json:"logging"`
-	Expose    v1.ServiceSpec          `json:"expose"`
+	Replicas  int32                    `json:"replicas"`
+	Image     string                   `json:"image,optional,omitempty"`
+	Security  InfinispanSecurity       `json:"security,optional,omitempty"`
+	Container *InfinispanContainerSpec `json:"container,optional,omitempty"`
+	Service   InfinispanServiceSpec    `json:"service,optional,omitempty"`
```
Crumby, line 88: I'm not sure making this `omitempty` makes sense. Isn't this field always set either  to `cache` (by default) or `datagrid`?

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -81,13 +81,13 @@ type InfinispanLoggingSpec struct {
...
-	Image     string                  `json:"image"`
-	Security  InfinispanSecurity      `json:"security"`
-	Container InfinispanContainerSpec `json:"container"`
-	Service   InfinispanServiceSpec   `json:"service"`
-	Logging   InfinispanLoggingSpec   `json:"logging"`
-	Expose    v1.ServiceSpec          `json:"expose"`
+	Replicas  int32                    `json:"replicas"`
+	Image     string                   `json:"image,optional,omitempty"`
+	Security  InfinispanSecurity       `json:"security,optional,omitempty"`
+	Container *InfinispanContainerSpec `json:"container,optional,omitempty"`
```
Crumby, line 87: Container is never empty. See above.

rigazilla:
@Crumby I've removed some omitempty around

rigazilla:
merged, thanks!

Issue #383, opened 2020-05-05
CPU and memory field should reflect actual settings


rigazilla:
If user creates an Infinispan resource without specifying cpu and memory fields then after creation the yaml related fields look as follows (they were missing in previous versions):
```
spec: 
  container: 
    cpu: ''
    memory: ''
```
In OpenShift environment this generally indicates that no restrictions are applied on the deployment but the restrictions are used as the fields were setup to 500m for CPU and 512Mi for memory.

Expected: cpu and memory field reflect current setup.

rigazilla:
with #382 `spec.container` field with default value is no more shown.
User can check resources details in the podtemplate struct.

PR #390, opened 2020-05-11
383 resources

rigazilla:
fix for #383 , defaults Infinispan values are now showed in the CR


rigazilla:
added on top of #382

rigazilla:
merged, thanks

PR #391, opened 2020-05-12
feedback port number to Infinispan cr

rigazilla:
Update infinispan.spec.expose.nodeport with the actual service nodeport if available

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -294,7 +294,13 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 						reqLogger.Error(err, "failed to create external Service", "Service", ser)
 						return reconcile.Result{}, err
 					}
-					err = r.client.Update(context.TODO(), infinispan)
+					r.client.Get(context.TODO(), types.NamespacedName{Namespace: externalService.Namespace, Name: externalService.Name}, externalService)
+					if err == nil {
+						if len(externalService.Spec.Ports) > 0 {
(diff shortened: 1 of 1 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -294,7 +294,13 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
-					err = r.client.Update(context.TODO(), infinispan)
+					r.client.Get(context.TODO(), types.NamespacedName{Namespace: externalService.Namespace, Name: externalService.Name}, externalService)
+					if err == nil {
+						if len(externalService.Spec.Ports) > 0 {
+							infinispan.Spec.Expose.NodePort = externalService.Spec.Ports[0].NodePort
+							r.client.Update(context.TODO(), infinispan)
+						}
+					}
 					if err != nil {
 						reqLogger.Error(err, "failed to Infinispan with Service spec", "Service", ser)
```
dmvolod, line 305: @rigazilla looks like the verb **update** is missing in error message

rigazilla: @dmvolod yeah right thanks. I've also fixed some other details and removed the requeue, a requeue will not fix this since the statefulset has been already created

rigazilla:
merged, thanks!

rigazilla:
> we should combine err == nil and len(externalService.Spec.Ports) > 0 in one if with &&

Sorry I've missed this comment, yeah it should be one statement with joined conditions

dmvolod:
Oh, it can be implemented in another 1.1.x branch PR if possible

PR #395, opened 2020-05-13
Rollback to sdk15

rigazilla:
We need to stay on sdk 0.15.2 because 0.17 requires golang 1.13 which is not available on some platform (i.e. ubi-8)


rigazilla:
merged, thanks!

PR #396, opened 2020-05-16
Fix a xsite script issue when using oc or kubectl

ryanzhang:
I think the bash part has some problem, it's not working in my terminal( Fedora 31, bshell).
Please check if this is ok.

### DIFF
--- a/manual/xsite-minikube/Makefile
+++ b/manual/xsite-minikube/Makefile
@@ -126,7 +126,7 @@ create-cache:
 developer-password = $$(kubectl get secret \
     example-infinispan-generated-secret \
     -o jsonpath="{.data.identities\.yaml}" \
-    | base64 -D \
+    | base64 --decode \
     | yq -r .credentials[0].password)
 
 url = $$(minikube service \
--- a/manual/xsite-openshift/Makefile
+++ b/manual/xsite-openshift/Makefile
@@ -99,7 +99,7 @@ create-cache:
 developer-password = $$(oc get secret \
    example-infinispan-generated-secret \
    -o jsonpath="{.data.identities\.yaml}" \
-   | base64 -D \
+   | base64 --decode \
    | yq -r .credentials[0].password)
 
 url = $$(oc get service \
--- a/documentation/asciidoc/topics/cmd_examples/oc_get_secret_tls.adoc
+++ b/documentation/asciidoc/topics/cmd_examples/oc_get_secret_tls.adoc
@@ -1,2 +1,2 @@
 $ {oc_get_secret} {example_crd_name}-cert-secret \
--o jsonpath='{.data.tls\.crt}' | base64 -d > tls.crt
+-o jsonpath='{.data.tls\.crt}' | base64 --decode > tls.crt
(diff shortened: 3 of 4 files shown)


dmvolod:
Looks good, but we should check that this works fine on macOS and RHEL also.
@rigazilla what do you think?

manual/xsite-minikube/Makefile
```diff
@@ -126,8 +126,8 @@ create-cache:
 developer-password = $$(kubectl get secret \
     example-infinispan-generated-secret \
     -o jsonpath="{.data.identities\.yaml}" \
-    | base64 -D \
-    | yq -r .credentials[0].password)
+    | base64 -d \
```
rigazilla, line 129: MacOS uses `-d` for debug. `base64 --decode` it's probably the portable syntax.

manual/xsite-minikube/Makefile
```diff
@@ -126,8 +126,8 @@ create-cache:
...
+    | yq r - credentials[0].password)
```
rigazilla, line 130: I'm not sure about this, on my FC31 doesn't work. With the fix to the base64 opts the old version is ok.

manual/xsite-openshift/Makefile
```diff
@@ -99,8 +99,8 @@ create-cache:
 developer-password = $$(oc get secret \
    example-infinispan-generated-secret \
    -o jsonpath="{.data.identities\.yaml}" \
-   | base64 -D \
-   | yq -r .credentials[0].password)
+   | base64 -d \
+   | yq r - credentials[0].password)
```
rigazilla, line 103: see above comments

rigazilla:
@ryanzhang thanks for the PR, please see my review.

rigazilla:
@ryanzhang I've replace `-d` with the longopt version `--decode` and fixed the filters chain. Please have a look if it sounds good to you, thanks!

manual/xsite-minikube/Makefile
ryanzhang: Yes, --decode should be the best option here.

manual/xsite-minikube/Makefile
ryanzhang: I need to double check if yq -r would work for me. It seemed to me that yq command is like this:
yq read things.yaml a.b.c
yq r - a.b.c (reads from stdin)

my client info:
uname -a
Linux localhost 5.5.7-200.fc31.x86_64 #1 SMP Fri Feb 28 17:18:37 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
 yq --version
yq version 2.4.1

ryanzhang:
@rigazilla It looks good to me. Thanks!
(I am not  sure about the "yq -r " part, it is ok to me if it works in your end.)

manual/xsite-minikube/Makefile
ryanzhang: I see what's going on here. 
It turn out that I used a different "yq" tool which behave slightly differently. 
1) https://github.com/mikefarah/yq accept parameter  "yq r -"
2) https://github.com/kislyuk/yq acept parameter "yq -r "

rigazilla:
> @rigazilla It looks good to me. Thanks!
> (I am not sure about the "yq -r " part, it is ok to me if it works in your end.)

yeah seems ok on my side

rigazilla:
merged, thanks @ryanzhang !

PR #401, opened 2020-05-20
Service expose redesign #380

rigazilla:
This is for #380 

- [x] backport of 1.1.x
- [x] add exposition for route and ingress with default or minimal setup


rigazilla:
User can expose Infinispan cluster using NodePort, LoadBalancer (same as in 1.1.x).
User can now use also:
**Route**
```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: example-infinispan
spec:
  replicas: 2
  expose:
    type: Route
```
**Ingress**
```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: example-infinispan
spec:
  replicas: 2
  expose:
    type: Ingress
    host: myapplication.local
```

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -267,23 +269,84 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+						return reconcile.Result{}, err
+					}
+					reqLogger.Info("Created External Service", "Service", externalService)
+				} else {
+					reqLogger.Info("External Service NOT created. Unsupported type?", "Service", nil)
+
+				}
+			case infinispanv1.ExposeTypeRoute:
+				lsService := ispncom.LabelsResource(infinispan.ObjectMeta.Name, "infinispan-service-external")
+				route := openshiftv1.Route{
```
dmvolod, line 293: @rigazilla what do you think about to check that we are running on OCP here? I'm not sure that on k8s this will print correct error during Route struct construction. We can implement isOpenshift flag in the ReconcileInfinispan and detect it at the early init phase.

test/e2e/main_test.go
```diff
@@ -453,24 +454,20 @@ func TestExternalService(t *testing.T) {
...
+func exposeServiceSpec() ispnv1.ExposeSpec {
+	return ispnv1.ExposeSpec{
+		Type:     exposeServiceType(),
+		NodePort: 30222,
 	}
 }
 
-func exposeServiceType() corev1.ServiceType {
-	exposeServiceType := tconst.ExposeServiceType
+func exposeServiceType() ispnv1.ExposeType {
```
dmvolod, line 464: @rigazilla have you plan to add specific OCP route test, which will be skipped on k8s running?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -267,23 +269,84 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
 		if infinispan.IsExposed() {
-			externalService := r.serviceExternal(infinispan)
-			if externalService != nil {
-				err = r.client.Create(context.TODO(), externalService)
-				if err != nil && !errors.IsAlreadyExists(err) {
-					reqLogger.Error(err, "failed to create external Service", "Service", ser)
+			switch infinispan.Spec.Expose.Type {
+			case infinispanv1.ExposeTypeLoadBalancer, infinispanv1.ExposeTypeNodePort:
+				externalService := r.serviceExternal(infinispan)
+				if externalService != nil {
```
dmvolod, line 275: @rigazilla Looks like serviceExternal can return nil not case of Unsupported type as it doesn't have any error handlers and mainly in case of problem inside code.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -267,23 +269,84 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+					TypeMeta: metav1.TypeMeta{
+						APIVersion: "v1",
+						Kind:       "Route",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      infinispan.Name,
+						Namespace: infinispan.Namespace,
+						Labels:    lsService,
+					},
+					Spec: openshiftv1.RouteSpec{To: openshiftv1.RouteTargetReference{Kind: "Service",
```
dmvolod, line 303: Looks like struct formatting issue

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -17,9 +17,11 @@ import (
 	ispnutil "github.com/infinispan/infinispan-operator/pkg/controller/infinispan/util"
 	ispncom "github.com/infinispan/infinispan-operator/pkg/controller/utils/common"
 	"github.com/infinispan/infinispan-operator/version"
+	openshiftv1 "github.com/openshift/api/route/v1"
```
dmvolod, line 20: @rigazilla may be use routev1 for package alias?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -267,23 +269,84 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+					reqLogger.Info("Created External Service", "Service", externalService)
+				} else {
+					reqLogger.Info("External Service NOT created. Unsupported type?", "Service", nil)
+
+				}
+			case infinispanv1.ExposeTypeRoute:
+				lsService := ispncom.LabelsResource(infinispan.ObjectMeta.Name, "infinispan-service-external")
+				route := openshiftv1.Route{
+					TypeMeta: metav1.TypeMeta{
+						APIVersion: "v1",
```
dmvolod, line 295: @rigazilla it would be nice to use routev1.SchemeGroupVersion.String() for better maintaining

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -267,23 +269,84 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+						Kind:       "Route",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      infinispan.Name,
+						Namespace: infinispan.Namespace,
+						Labels:    lsService,
+					},
+					Spec: openshiftv1.RouteSpec{To: openshiftv1.RouteTargetReference{Kind: "Service",
+						Name: infinispan.Name},
```
dmvolod, line 304: @rigazilla do we need to define port for Route?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -267,23 +269,84 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+						TLS: &openshiftv1.TLSConfig{Termination: openshiftv1.TLSTerminationPassthrough},
```
dmvolod, line 305: @rigazilla , it would be nice to have an option to configure Host for route. We may implement it in next version.
Wdyt?

pkg/controller/infinispan/infinispan_controller.go
rigazilla: @dmvolod I was actually thinking to implement an automatic route/ingress selection like this:
https://github.com/application-stacks/runtime-component-operator/blob/d3bb9c7f29936169fd1e9092e255895980522ce1/pkg/controller/runtimecomponent/runtimecomponent_controller.go#L654

rigazilla: how does it sound?

dmvolod: Oh, @rigazilla , it would be nice to have this option. We can use just one expose type for this, like External or keep Route and remove Ingress and implement automatic choose as you suggested.

dmvolod:
@rigazilla please also pay attention, that deploy/crds/infinispan.org_infinispans_crd.yaml file openAPIV3Schema need to be changed related the new Expose struct

pkg/controller/infinispan/infinispan_controller.go
rigazilla: @dmvolod should works as we discussed above now. I've used `Route` as generic exposition type

pkg/controller/infinispan/infinispan_controller.go
rigazilla: ok, done

pkg/controller/infinispan/infinispan_controller.go
rigazilla: seems that typemeta is redundant, I've removed it

pkg/controller/infinispan/infinispan_controller.go
rigazilla: ok, done

pkg/controller/infinispan/infinispan_controller.go
rigazilla: yeah, since we already have port and host field I think it's ok to configure them also for route if provided by the user

dmvolod: It looks like service port, not custom defined and actual if multiple service ports are available. In out case it's just single but may be multiple in the future.

rigazilla: well, actually: `host` is ok. `port` is probably beyond a basic configuration for route/ingressa

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, cool

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Excellent implementation, thanks @rigazilla

rigazilla: thank you!

pkg/controller/infinispan/infinispan_controller.go
rigazilla: yeah, serviceExternal just populate a struct. I've cleanup the if statement

pkg/controller/infinispan/infinispan_controller.go
rigazilla: add host configuration

test/e2e/main_test.go
rigazilla: I would like to track this on a separate issue

rigazilla:
> @rigazilla please also pay attention, that deploy/crds/infinispan.org_infinispans_crd.yaml file openAPIV3Schema need to be changed related the new Expose struct

I've updated the crds

deploy/cr/expose/cr_ingress.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: infinispan.org/v1
+kind: Infinispan
+metadata:
+  name: example-infinispan
+spec:
+  replicas: 2
+  expose:
+    type: Ingress
```
dmvolod, line 8: @rigazilla please change to Route as you've removed this expose type

rigazilla: yeah thanks @dmvolod, I've removed it, since seems a duplicate of route.yaml

test/e2e/main_test.go
dmvolod: Yes, sure, separate test from core dev PR is prefer way

rigazilla: https://github.com/infinispan/infinispan-operator/issues/410

deploy/cr/expose/cr_ingress.yaml
dmvolod: Yes, sure
@rigazilla also we need to update doc/design.adoc with actual model. I will try to take care about it.

rigazilla: thanks!

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -79,6 +79,33 @@ type InfinispanLoggingSpec struct {
...
+	// every node, in addition to 'ClusterIP' type.
+	ExposeTypeNodePort ExposeType = "NodePort"
+
+	// ExposeTypeLoadBalancer means a service will be exposed via an
+	// external load balancer (if the cloud provider supports it), in addition
+	// to 'NodePort' type.
+	ExposeTypeLoadBalancer ExposeType = "LoadBalancer"
+
+	// ExposeTypeRoute means the service will be exposed via
+	// Route (Openshift resource)
```
dmvolod, line 97: @rigazilla looks like we need to say that is not only OCP Route, but k8s Ingress also. Or we would like to point in the docs.

rigazilla: right @dmvolod, I've changed the commit

dmvolod: Thanks, @rigazilla . Looks good.

PR #409, opened 2020-05-26
update nodePort example value

oraNod:
from @Crumby: "This isn't a good example value. From the OpenShift docs on NodePort:
NodePorts are in the 30000-32767 range by default"


rigazilla:
merged, thanks!

PR #414, opened 2020-06-03
Integration tests README.md

Crumby:
Provides README file on integration tests execution.

### DIFF
--- a/test-integration/README.md
+++ b/test-integration/README.md
@@ -1,0 +1,26 @@
+# Integration tests
+Brief description on how to run integration tests.
+
+## Prerequisities
+* OpenShift 4
+* Installed Operator
+  * Operator has to watch different namespace then one it's installed in
+
+## Running the tests
+* `mvn clean verify` to run all the tests
+* `mvn clean verify -Dit.test=${testName}` to run particular test
+
+### Params
+Include following parameters as part of your test execution command to provide target OpenShift instance details, eg.:
+* `-Dxtf.openshift.namespace=infinispan-suite`
+* `-Dxtf.openshift.url=https://api.my.openshift.com:6443`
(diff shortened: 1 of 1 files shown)


test-integration/README.md
```diff
@@ -0,0 +1,26 @@
+# Integration tests
+Brief description on how to run integration tests.
+
+## Prerequisities
+* OpenShift 4
+* Installed Operator
+  * Operator has to watch different namespace then one it's installed in
```
rigazilla, line 7: @Crumby I've changed the rbac.yaml to work in this configuration, how about to add it to the PR?
https://gist.github.com/rigazilla/d196be1732fa8c88eb1599c3aac32dad

Crumby: I'm not exactly sure I understand. Do you want the install process described in the readme too?

rigazilla: Only the rbac setup since it differs from the operator example due to the different namespaces configuration

rigazilla: @Crumby I'm not sure I'm doing right: it seems that my rbac settings go lost between a test and the other and I start to see access error in the operator log: 
```
forbidden: User "system:serviceaccount:operator-test:infinispan-operator" cannot list resource "infinispans" in API group "infinispan.org" in the namespace "test"
```

Crumby: That's weird, shouldn't happen. The tests have `@CleanBeforeAll` annotation that wipes out all non-default resources in tested namespace before the tests starts and that's why operator needs to run in another one. So if your rbac settings are in the tested namespace then that would expalin the issue you see.

Try installing released operator in one namespace using OpenShifts web console and then change the Operator Group to point to a different one where the tests should run.

rigazilla:
merging for now, will open new issues if needed. 
thanks!

PR #415, opened 2020-06-03
Use premetive auth with nonsecured REST calls

Crumby:
Use preemptive authorization to prevent endless loop in authorization process.


rigazilla:
merged, thanks @Crumby

PR #416, opened 2020-06-04
Check expose type in tests

Crumby:
This PR contains two test fixes for E2E tests:
* Hard-coded `NodePort` value causes failures on AWS with LoadBalancer when tests are executed sequentially. Tests are being executed faster then OpenShift manages to free the node port upon service deletion which causes that following tests won't create an external service as the `NodePort` is taken.
* Fix `waitForExternalService` to try NodePort only if the external service is set to be `NodePort`. Otherwise it takes 7 minutes for the function to find out that service is exposed as LoadBalancer on AWS doubling time needed for test execution.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -519,7 +519,6 @@ func TestExternalService(t *testing.T) {
 func exposeServiceSpec() *ispnv1.ExposeSpec {
 	return &ispnv1.ExposeSpec{
 		Type:     exposeServiceType(),
-		NodePort: 30222,
 	}
 }
 
--- a/test/e2e/util/test_kubernetes.go
+++ b/test/e2e/util/test_kubernetes.go
@@ -323,24 +323,26 @@ func (k TestKubernetes) WaitForExternalService(routeName string, timeout time.Du
 		err = k.Kubernetes.Client.Get(context.TODO(), ns, route)
 		ExpectNoError(err)
 
-		// depending on the k8s cluster setting, service is sometime available
-		// via Ingress address sometime via node port. So trying both the methods
-		// Try node port first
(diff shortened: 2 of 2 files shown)


Crumby:
@rigazilla Do you have an idea why CI fails for this one? I've checked the logs and it seems to correctly enter the `if` statement and resolve the node's port, yet it fails....

rigazilla:
@dmvolod , @Crumby `kind` needs to know at startup which port is used by the `nodePort` service. Configuration is in `kind-config.yaml`.

I don't know if it's possible to arrange a configuration that work with node port selected by OCP, maybe the simplest solution is to have different test behaviour selected via envvar.

Crumby:
@rigazilla Ah, got it. So lets say `HARDCODED_NODEPORT=true|false` (true by default) in `exposeServiceSpec` that returns either one or another?

dmvolod:
@Crumby , we can also try to check automatically as @rigazilla implemented for Route exposing if we are running on OCP or in travis k8s kind.

rigazilla:
> @Crumby , we can also try to check automatically as @rigazilla implemented for Route exposing if we are running on OCP or in travis k8s kind.

Yeah thanks @dmvolod .

@Crumby there's a generic platform check that we can import here:
https://github.com/application-stacks/runtime-component-operator/blob/c7eaf97c155eaf0a9cd3be9e0b1fff1fdc1fab16/pkg/utils/reconciler.go#L421

Crumby:
@rigazilla @dmvolod General check might not be complete bulletproof solution but it'll spare us an env for now.

I'm trying to thing about another platform that might be used to run the tests against. CRC (might prefer hardcoded even though it's OpenShift) or non-minikube kubernetes that might suffer from hardcoded port same way the OpenShift on AWS does....

Crumby:
Closing this one as outdated at this moment.

PR #421, opened 2020-06-05
porting fix for scheme selection and node IPs

rigazilla:
This PR ports from 1.1.x the protocol scheme selection for external service tests.
Plus a fix on how node address is selected for testing exposed NodePort.

### DIFF
--- a/pkg/controller/infinispan/util/kubernetes.go
+++ b/pkg/controller/infinispan/util/kubernetes.go
@@ -199,6 +199,26 @@ func (k Kubernetes) PublicIP() string {
 	return u.Hostname()
 }
 
+// GetNodesHost return the addresses of the k8s nodes
+func (k Kubernetes) GetNodesHost() []string {
+	nodes := &v1.NodeList{}
+	k.Client.List(context.TODO(), nodes)
+	nodeHosts := make([]string, 0, nodes.Size())
+	for _, n := range nodes.Items {
+		for _, v := range n.Status.Addresses {
+			if v.Type == v1.NodeHostName {
+				// Fallback using hostname if internal IP address is not available
+				nodeHosts = append(nodeHosts, v.Address)
+			}
+			if v.Type == v1.NodeInternalIP {
+				nodeHosts = append(nodeHosts, v.Address)
+				break
+			}
+		}
+	}
+	return nodeHosts
+}
+
 // FindKubeConfig returns local Kubernetes configuration
 func FindKubeConfig() string {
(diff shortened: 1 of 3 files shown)


pkg/controller/infinispan/util/kubernetes.go
```diff
@@ -199,6 +199,26 @@ func (k Kubernetes) PublicIP() string {
...
+			}
+			if v.Type == v1.NodeInternalIP {
+				nodeHosts[i] = v.Address
+				break
+			}
+		}
+	}
+	return nodeHosts
+}
+
```
rigazilla, line 221: this seems to fix #416 also. No need to specify a fixed nodeport any more

pkg/controller/infinispan/util/kubernetes.go
```diff
@@ -199,6 +199,26 @@ func (k Kubernetes) PublicIP() string {
 	return u.Hostname()
 }
 
+// GetNodesHost return the addresses of the k8s nodes
+func (k Kubernetes) GetNodesHost() []string {
```
dmvolod, line 203: @rigazilla why we need a slice if we always access the first item? Or it just for future use?

pkg/controller/infinispan/util/kubernetes.go
```diff
@@ -199,6 +199,26 @@ func (k Kubernetes) PublicIP() string {
...
+				// Fallback using hostname if internal IP address is not available
+				nodeHosts[i] = v.Address
+			}
+			if v.Type == v1.NodeInternalIP {
+				nodeHosts[i] = v.Address
+				break
+			}
+		}
+	}
+	return nodeHosts
```
dmvolod, line 219: @rigazilla looks like we need to trunc slice to prevent empty values or choose another solution

pkg/controller/infinispan/util/kubernetes.go
rigazilla: @dmvolod yeah, maybe it can be used somewhere else

pkg/controller/infinispan/util/kubernetes.go
rigazilla: I see what you mean, let me clean up

rigazilla: @dmvolod I've changed the impl. If it's ok I'll port also on #416

dmvolod: Thanks, @rigazilla, looks better

rigazilla:
thanks for the review @dmvolod !

PR #424, opened 2020-06-09
423 xsite coord

rigazilla:
This PR select only coordinators as endpoints for the -xsite service.
You can check it with:

oc describe svc example-infinispan-xsite

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -162,7 +162,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	// because remote site host:port combinations need to be injected into Infinispan.
 	xsite, err := r.computeXSite(infinispan, reqLogger)
 	if err != nil {
-		reqLogger.Error(err, "Error in computeXSite", "Infinispan.Namespace")
+		reqLogger.Error(err, "Error in computeXSite", "Infinispan.Namespace", infinispan.Namespace)
 		return reconcile.Result{}, err
 	}
 
@@ -430,7 +430,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 
 	// List the pods for this infinispan's deployment
 	podList := &corev1.PodList{}
-	labelSelector := labels.SelectorFromSet(ispncom.LabelsResource(infinispan.Name, ""))
+	labelSelector := labels.SelectorFromSet(ispncom.LabelsResource(infinispan.Name, "infinispan-pod"))
 	listOps := &client.ListOptions{Namespace: infinispan.Namespace, LabelSelector: labelSelector}
 	err = r.client.List(context.TODO(), podList, listOps)
 	if err != nil {
@@ -679,6 +679,16 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		// Spec updated - return and requeue
 		return reconcile.Result{Requeue: true}, nil
 	}
+
+	// If x-site enable configure the coordinator pods to be selected by the x-site service
+	if infinispan.HasSites() {
+		found := r.applyLabelsToCoordinatorsPod(podList, infinispan, cluster, reqLogger)
+		if !found {
+			// If a coordinator is not found then requeue early
+			return reconcile.Result{Requeue: true}, nil
+		}
+	}
+
 	// Update the Infinispan status with the pod status
 	// Wait until all pods have ips assigned
 	// Without those ips, it's not possible to execute next calls
@@ -718,6 +728,10 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		reqLogger.Info("notClusterFormed")
 		return reconcile.Result{Requeue: true, RequeueAfter: 15 * time.Second}, nil
 	}
+
+	// Below the code for a wellFormed cluster
+
+	// Create default cache if it doesn't exists.
 	if infinispan.IsCache() && !existsCacheServiceDefaultCache(podList.Items[0].Name, infinispan, cluster) {
 		reqLogger.Info("createDefaultCache")
 		err := createCacheServiceDefaultCache(podList.Items[0].Name, infinispan, cluster, reqLogger)
@@ -727,18 +741,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
-	// Check if pods container runs the right image
-	for _, item := range podList.Items {
-		if len(item.Spec.Containers) == 1 {
-			if !strings.HasSuffix(item.Spec.Containers[0].Image, infinispan.Spec.Image) {
-				// TODO: invent a reconciliation policy if images doesn't match
-				// Attention: this is something that can conflict with StatefulSet rolling upgrade
-				reqLogger.Info("Pod " + item.Name + " runs wrong image " + item.Spec.Containers[0].Image + " != " + infinispan.Spec.Image)
-			}
-		}
-	}
-
-	return reconcile.Result{}, nil
+	return reconcile.Result{}, err
 }
 
 func updateSecurity(infinispan *infinispanv1.Infinispan, client client.Client, reqLogger logr.Logger) error {
@@ -994,7 +997,7 @@ func (r *ReconcileInfinispan) computeXSite(infinispan *infinispanv1.Infinispan, 
 			return nil, err
 		}
 
-		localSiteHost, localSitePort, err := getLocalSiteServiceHostPort(siteService, infinispan, logger)
+		localSiteHost, localSitePort, err := getCrossSiteServiceHostPort(siteService, kubernetes, logger)
 		if err != nil {
 			logger.Error(err, "error retrieving local x-site service information")
 			return nil, err
@@ -1042,45 +1045,6 @@ func findRemoteLocations(localSiteName string, infinispan *infinispanv1.Infinisp
 	}
 
 	return
-}
-
-func getLocalSiteServiceHostPort(service *corev1.Service, infinispan *infinispanv1.Infinispan, logger logr.Logger) (string, int32, error) {
-	switch serviceType := service.Spec.Type; serviceType {
-	case corev1.ServiceTypeNodePort:
-		externalIP, err := infinispan.FindNodePortExternalIP()
-		if err != nil {
-			return "", 0, err
-		}
-
-		// If configuring NodePort, expect external IPs to be configured
-		return externalIP, kubernetes.GetNodePort(service), nil
-	case corev1.ServiceTypeLoadBalancer:
-		return getLoadBalancerServiceHostPort(service, logger)
-	default:
-		return "", 0, fmt.Errorf("unsupported service type '%v'", serviceType)
-	}
-}
-
-func getLoadBalancerServiceHostPort(service *corev1.Service, logger logr.Logger) (string, int32, error) {
-	port := service.Spec.Ports[0].Port
-
-	// If configuring load balancer, look for external ingress
-	if len(service.Status.LoadBalancer.Ingress) > 0 {
-		ingress := service.Status.LoadBalancer.Ingress[0]
-		if ingress.IP != "" {
-			return ingress.IP, port, nil
-		}
-		if ingress.Hostname != "" {
-			// Resolve load balancer host
-			ip, err := ispncom.LookupHost(ingress.Hostname, logger)
-
-			// Load balancer gets created asynchronously,
-			// so it might take time for the status to be updated.
-			return ip, port, err
-		}
-	}
-
-	return "", port, nil
 }
 
 // deploymentForInfinispan returns an infinispan Deployment object
@@ -1586,26 +1550,35 @@ func (r *ReconcileInfinispan) getOrCreateSiteService(siteServiceName string, inf
 
 func (r *ReconcileInfinispan) createSiteService(siteServiceName string, infinispan *infinispanv1.Infinispan, logger logr.Logger) (*corev1.Service, error) {
 	lsPodSelector := ispncom.LabelsResource(infinispan.ObjectMeta.Name, "infinispan-pod")
+	lsPodSelector["coordinator"] = "true"
 	lsService := ispncom.LabelsResource(infinispan.ObjectMeta.Name, "infinispan-service-xsite")
-	exposeSpec := infinispan.Spec.Service.Sites.Local.Expose
-	exposeSpec.Selector = lsPodSelector
 
-	switch exposeSpec.Type {
-	case corev1.ServiceTypeNodePort:
+	exposeSpec := corev1.ServiceSpec{}
+	exposeConf := infinispan.Spec.Service.Sites.Local.Expose
+	exposeSpec.Selector = lsPodSelector
+	nodePort := int32(0)
+	if exposeConf.NodePort != 0 {
+		nodePort = exposeConf.NodePort
+	}
+	switch exposeConf.Type {
+	case infinispanv1.ExposeTypeNodePort:
+		exposeSpec.Type = corev1.ServiceTypeNodePort
 		exposeSpec.Ports = []corev1.ServicePort{
 			{
-				Port:     7900,
-				NodePort: 32660,
+				Port:       7900,
+				NodePort:   nodePort,
+				TargetPort: intstr.IntOrString{IntVal: 7900},
 			},
 		}
-	case corev1.ServiceTypeLoadBalancer:
+	case infinispanv1.ExposeTypeLoadBalancer:
+		exposeSpec.Type = corev1.ServiceTypeLoadBalancer
 		exposeSpec.Ports = []corev1.ServicePort{
 			{
-				Port: 7900,
+				Port:       7900,
+				TargetPort: intstr.IntOrString{IntVal: 7900},
 			},
 		}
 	}
-
 	logger.Info("create exposed site service", "configuration", exposeSpec)
 
 	siteService := corev1.Service{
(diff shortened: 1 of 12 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -663,6 +663,16 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		// Spec updated - return and requeue
 		return reconcile.Result{Requeue: true}, nil
 	}
+
+	// If x-site enable configure the coordinator pods to be selected by the x-site service
+	if infinispan.HasSites() {
+		found, _ := r.applyLabelsToCoordinatorsPod(podList, infinispan, cluster, reqLogger)
+		if !found {
+			// If a coordinator is not found reque early
+			return reconcile.Result{Requeue: true}, err
```
dmvolod, line 672: @rigazilla this will return empty err or as previous call skips err get

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1743,18 +1716,67 @@ func appendKubernetesRemoteLocation(xsite *ispnutil.XSite, infinispan *infinispa
...
+
+	//select workers only. AFAIK, only the workers have the proxy to the pods
+	labelSelector := labels.SelectorFromSet(map[string]string{"node-role.kubernetes.io/worker": ""})
+
+	listOps := &client.ListOptions{
+		LabelSelector: labelSelector,
+	}
+	err := k.Client.List(context.TODO(), workerList, listOps)
+	if err != nil {
+		return "", service.Spec.Ports[0].NodePort, err
```
dmvolod, line 1748: @pruivo why we need to return port value if err is occurs?

pkg/controller/infinispan/infinispan_controller.go
rigazilla: yep thanks. I don't think we need to handle the error outside the `applyLabelsToCoordinatorsPod` atm. I'll change also the function return.

documentation/asciidoc/topics/proc_creating_sa_tokens.adoc
```diff
@@ -30,6 +30,13 @@ the view role to the service account on **LON** as follows:
 $ oc policy add-role-to-user view system:serviceaccount:my-xsite:lon
 ----
 +
+. Add cluster-reader role to the service accounts (if using NodePor to expose the service)
```
dmvolod, line 33: NodePort misprint

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1743,18 +1716,67 @@ func appendKubernetesRemoteLocation(xsite *ispnutil.XSite, infinispan *infinispa
...
+	logger.Info("Found NodePort endpoint", "Host", host, "Port", port)
+	return host, port, nil
+}
+
+func getLoadBalancerServiceHostPort(service *corev1.Service, logger logr.Logger) (string, int32, error) {
+	port := service.Spec.Ports[0].Port
+
+	// If configuring load balancer, look for external ingress
+	if len(service.Status.LoadBalancer.Ingress) > 0 {
+		ingress := service.Status.LoadBalancer.Ingress[0]
```
dmvolod, line 1763: @pruivo may be we need to check that Ingress[0] is present to prevent _Index out of range_?

documentation/asciidoc/topics/proc_creating_sa_tokens.adoc
oraNod: I noticed this earlier but didn't comment as I plan on making a pass at xsite docs after things are ready on the eng side. @rigazilla could you please notify when things are ready and I'll add docs to this PR or in a separate PR if that's more efficient? thanks.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1766,3 +1788,37 @@ func getServingCertsMode(remoteKubernetes *ispnutil.Kubernetes) string {
...
+
+func (r *ReconcileInfinispan) applyLabelsToCoordinatorsPod(podList *corev1.PodList, infinispan *infinispanv1.Infinispan, cluster ispnutil.ClusterInterface, logger logr.Logger) (bool, error) {
+	pass, err := cluster.GetPassword(consts.DefaultOperatorUser, infinispan.GetSecretName(), infinispan.GetNamespace())
+	coordinatorFound := false
+	if err != nil {
+		return coordinatorFound, err
+	}
+	protocol := string(infinispan.GetEndpointScheme())
+	for _, item := range podList.Items {
+		cacheManagerInfo, err := cluster.GetCacheManagerInfo(consts.DefaultOperatorUser, pass, "default", item.Name, item.Namespace, protocol)
```
dmvolod, line 1800: @rigazilla it would be nice to place "default" CacheManager to the constants. Do we need to have it as parameter? Is it possible to another manager rather than "default"

pkg/controller/infinispan/infinispan_controller.go
pruivo: no idea. First time coding in Go. I thought the first 2 args didn't matter.

pkg/controller/infinispan/infinispan_controller.go
pruivo: isn't `if len(service.Status.LoadBalancer.Ingress) > 0 {` enough?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1766,3 +1788,37 @@ func getServingCertsMode(remoteKubernetes *ispnutil.Kubernetes) string {
...
+		if err == nil {
+			lab, ok := item.Labels["coordinator"]
+			if cacheManagerInfo["coordinator"].(bool) {
+				coordinatorFound = true
+				if !ok || lab != "true" {
+					item.Labels["coordinator"] = "true"
+					err = r.client.Update(context.TODO(), &item)
```
dmvolod, line 1807: @rigazilla looks like we need to check errors here. Is'n it?

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, we need to handle this error in case of update problems inside applyLabelsToCoordinatorsPod

pkg/controller/infinispan/infinispan_controller.go
dmvolod: You can just return default value, i.e. 0 in case of error instead of service.Spec.Ports[0].NodePort

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Oh, sorry, point on incorrect line

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1759,18 +1732,67 @@ func appendKubernetesRemoteLocation(xsite *ispnutil.XSite, infinispan *infinispa
...
+	//The NodePort can be found in the service
+	port := service.Spec.Ports[0].NodePort
+
+	//The IPs must be fetch. Some cases, the API server (which handles REST requests) isn't the same as the worker
+	//So, we get the workers list. It needs some permissions cluster-reader permission
+	//oc create clusterrolebinding <name> -n ${NAMESPACE} --clusterrole=cluster-reader --serviceaccount=${NAMESPACE}:<account-name>
+	workerList := &corev1.NodeList{}
+
+	//select workers only. AFAIK, only the workers have the proxy to the pods
+	labelSelector := labels.SelectorFromSet(map[string]string{"node-role.kubernetes.io/worker": ""})
```
dmvolod, line 1757: @pruivo do we also need to check the Status of the Node. Is node in any status and state in enough?

pkg/controller/infinispan/infinispan_controller.go
rigazilla: mmm right, to keep things I would handle errors in the function and use only requeue as recovery strategy

pkg/controller/infinispan/infinispan_controller.go
pruivo: I wasn't worried about it for 2 reasons:
1. every blog post I read about NodePort says it isn't recommended in production. 
2. This is used for cross-site replication and our image doesn't support more than an <ip:port> pair per site. So, if the node crashes, it is going to fail and the Infinispan deployment must be deleted and re-applied on all sites.

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I don't know.
@pruivo you know if we have to care about CacheManager != "default"?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1782,3 +1804,37 @@ func getServingCertsMode(remoteKubernetes *ispnutil.Kubernetes) string {
 	}
 	return ""
 }
+
+func (r *ReconcileInfinispan) applyLabelsToCoordinatorsPod(podList *corev1.PodList, infinispan *infinispanv1.Infinispan, cluster ispnutil.ClusterInterface, logger logr.Logger) bool {
```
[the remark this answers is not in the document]
rigazilla: any changes to pods should generate an event for the reconciler and that should update the label if needed.
I've tested it killing the cordinator pod.

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, @pruivo true about NodePort
But you are listing all worker Nodes and takes the first one, but it can be in NotReady state by origin. Is it correct? You are not validating state of the node.

pkg/controller/infinispan/infinispan_controller.go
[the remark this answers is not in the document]
dmvolod: Good, thanks.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -684,7 +684,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	if infinispan.HasSites() {
 		found := r.applyLabelsToCoordinatorsPod(podList, infinispan, cluster, reqLogger)
 		if !found {
-			// If a coordinator is not found reque early
+			// If a coordinator is not found then reque early
```
dmvolod, line 687: Misprint in requeue

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1782,3 +1804,37 @@ func getServingCertsMode(remoteKubernetes *ispnutil.Kubernetes) string {
...
+	for _, item := range podList.Items {
+		cacheManagerInfo, err := cluster.GetCacheManagerInfo(consts.DefaultOperatorUser, pass, "default", item.Name, item.Namespace, protocol)
+		if err == nil {
+			lab, ok := item.Labels["coordinator"]
+			if cacheManagerInfo["coordinator"].(bool) {
+				if !ok || lab != "true" {
+					item.Labels["coordinator"] = "true"
+					err = r.client.Update(context.TODO(), &item)
+				}
+				coordinatorFound = (err == nil)
```
dmvolod, line 1824: Feedback from GoLand IDE: Redundant parentheses

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1759,18 +1732,67 @@ func appendKubernetesRemoteLocation(xsite *ispnutil.XSite, infinispan *infinispa
...
-func getRemoteSiteServiceHostPort(service *corev1.Service, remoteKubernetes *ispnutil.Kubernetes, logger logr.Logger) (string, int32, error) {
+func getCrossSiteServiceHostPort(service *corev1.Service, kubernetes *ispnutil.Kubernetes, logger logr.Logger) (string, int32, error) {
 	switch serviceType := service.Spec.Type; serviceType {
 	case corev1.ServiceTypeNodePort:
 		// If configuring NodePort, expect external IPs to be configured
-		return remoteKubernetes.PublicIP(), remoteKubernetes.GetNodePort(service), nil
+		return getNodePortServiceHostPort(service, kubernetes, logger)
 	case corev1.ServiceTypeLoadBalancer:
 		return getLoadBalancerServiceHostPort(service, logger)
 	default:
```
dmvolod, line 1742: @rigazilla is there any plans to support Route/Ingress or it's not possible value?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1586,26 +1550,35 @@ func (r *ReconcileInfinispan) getOrCreateSiteService(siteServiceName string, inf
...
-	exposeSpec := infinispan.Spec.Service.Sites.Local.Expose
-	exposeSpec.Selector = lsPodSelector
 
-	switch exposeSpec.Type {
-	case corev1.ServiceTypeNodePort:
+	exposeSpec := corev1.ServiceSpec{}
+	exposeConf := infinispan.Spec.Service.Sites.Local.Expose
+	exposeSpec.Selector = lsPodSelector
+	nodePort := int32(0)
+	if exposeConf.NodePort != 0 {
```
dmvolod, line 1560: @rigazilla should check that exposeConf.NodePort greater that 0

pkg/controller/infinispan/util/configuration.go
```diff
@@ -58,7 +60,9 @@ func CreateInfinispanConfiguration(name string, xsite *XSite, loggingCategories
 		ClusterName: name,
 		JGroups:     jgroups,
 	}
-
+	if os.Getenv("JGROUPS_DIAGNOSTICS") == "TRUE" {
```
dmvolod, line 63: @rigazilla looks like we should uppercase output value, as **TRUE** equals **true**

dmvolod: Another one, should this variable be at the operator level or cluster site? A little bit confusing.

pkg/controller/infinispan/infinispan_controller.go
pruivo: @dmvolod I've updated in my PR: https://github.com/infinispan/infinispan-operator/pull/441
@rigazilla do you want to merge my PR separately or cherry-pick the commit to yours?

pkg/controller/infinispan/util/configuration.go
pruivo: it would be better if we can set it in the Infinispan CRD.

dmvolod: Yes, true, @pruivo but we need to add a separate field for options like this (not just for single options but for future use also) or somehow to use InfinispanLoggingSpec.Categories.

pkg/controller/infinispan/infinispan_controller.go
rigazilla: Not atm, I'll create a constant for `default` for now

pkg/controller/infinispan/infinispan_controller.go
rigazilla: If it ok to you, I think we can go faster with a cherry pick.

pkg/controller/infinispan/infinispan_controller.go
rigazilla: Yeah, I would keep them just for human readability

pkg/controller/infinispan/infinispan_controller.go
rigazilla: Not atm, but I think it makes sense for a next release

pkg/controller/infinispan/infinispan_controller.go
pruivo: ok. it works for me.

pkg/controller/infinispan/util/configuration.go
rigazilla: We added this configuration envvar for developing and it's not something required atm, but we can design a feature issue for the next release.
Does diagnostic expose any sensitive data? Security issues?

dmvolod: @rigazilla wdyt about this?

dmvolod: @rigazilla support team can also use this for diagnostic. Let's keep it "as is" but use uppercase for easy setup.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1782,3 +1804,37 @@ func getServingCertsMode(remoteKubernetes *ispnutil.Kubernetes) string {
 	}
 	return ""
 }
+
+func (r *ReconcileInfinispan) applyLabelsToCoordinatorsPod(podList *corev1.PodList, infinispan *infinispanv1.Infinispan, cluster ispnutil.ClusterInterface, logger logr.Logger) bool {
+	pass, err := cluster.GetPassword(consts.DefaultOperatorUser, infinispan.GetSecretName(), infinispan.GetNamespace())
```
dmvolod, line 1818: @rigazilla , looks like we will never know when err will return from cluster.GetPassword

rigazilla: yep, let me add a log

pkg/controller/infinispan/util/configuration.go
dmvolod: @rigazilla please add uppercase func for os.Getenv

rigazilla: uh, I though I did, let me check

rigazilla: yeah @dmvolod it's there, GH is showing an outdated snippet

pkg/controller/infinispan/util/configuration.go
```diff
@@ -58,7 +61,9 @@ func CreateInfinispanConfiguration(name string, xsite *XSite, loggingCategories
 		ClusterName: name,
 		JGroups:     jgroups,
 	}
-
+	if strings.ToUpper(os.Getenv("JGROUPS_DIAGNOSTICS")) == "TRUE" {
```
dmvolod, line 64: Oh, sure, @rigazilla , missing this. Thanks

rigazilla:
@pruivo , @oraNod, if all it's ok please approve this. I'm  going to rebase and merge

pruivo:
:+1: @rigazilla

rigazilla:
merged, thanks all!

PR #426, opened 2020-06-09
Fixed exposed service update

rigazilla:
Update the infinispan CR with exposed service details before update

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -287,10 +287,12 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 					reqLogger.Error(err, "failed to create external Service", "Service", ser)
 					return reconcile.Result{}, err
 				}
-				err = r.client.Update(context.TODO(), infinispan)
-				if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -287,10 +287,12 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 					reqLogger.Error(err, "failed to create external Service", "Service", ser)
 					return reconcile.Result{}, err
 				}
-				err = r.client.Update(context.TODO(), infinispan)
-				if err != nil {
-					reqLogger.Error(err, "Failed to update Infinispan with Service", "Service", ser)
-					return reconcile.Result{}, err
+				if len(externalService.Spec.Ports) > 0 {
+					infinispan.Spec.Expose.NodePort = externalService.Spec.Ports[0].NodePort
+					err = r.client.Update(context.TODO(), infinispan)
```
dmvolod, line 292: @rigazilla, do we need to get actual externalService record after create or it will return to the variable back?
As for as I remember, some of the external services, like LoadBalancer are created asynchronously as contains of set of the steps on the underlying platform, i.e. AWS, etc.

rigazilla:
merged, thanks

PR #429, opened 2020-06-15
ISPN-10734 autoscale docs

oraNod:
https://issues.redhat.com/browse/ISPN-10734


documentation/asciidoc/topics/proc_configuring_autoscaling.adoc
```diff
@@ -0,0 +1,30 @@
...
+    maxReplicas: 5 <3>
+    minMemUsagePercent: 30 <4>
+    minReplicas: 2 <5>
+----
+
+<1> Enables automatic scaling.
+<2> Configures the maximum threshold, as a percentage, for memory usage on each node. When {ispn_operator} detects that any node in the cluster reaches the threshold, it creates a new node if possible. If {ispn_operator} cannot create a new node then it performs eviction when memory usage reaches 100 percent.
+<3> Defines the maximum number of replicas for each cache entry across the cluster.
+<4> Configures the minimum threshold, as a percentage, for memory usage across the cluster. When {ispn_operator} detects that memory usage falls below the minimum, it shuts down nodes.
+<5> Defines the minimum number of replicas for each cache entry across the cluster.
```
oraNod, line 30: @rigazilla are there any defaults for autoscale? can you just add "autoscale" to the CR without specifying thresholds?

I'd also like to expand on the max and min replica fields here. it isn't quite clear to me how these affect autoscale.

documentation/asciidoc/topics/ref_default_cache_service_config.adoc
```diff
@@ -1,11 +1,20 @@
...
+                   mode="SYNC" <2>
+                   owners="1"> <3>
+   <memory>
+    <off-heap eviction="MEMORY" <4>
+              strategy="REMOVE"/> <5>
+   </memory>
+   <partition-handling when-split="ALLOW_READ_WRITES" <6>
+                       merge-policy="REMOVE_ALL"/> <7>
+</distributed-cache>
 ----
```
oraNod, line 18: @rigazilla should we update this to reflect the new memory configuration in 11?

documentation/asciidoc/topics/proc_configuring_autoscaling.adoc
```diff
@@ -0,0 +1,30 @@
...
+.Procedure
+
+. Add the `spec.autoscale` resource to your **Infinispan** CR to enable automatic scaling.
+. Configure memory usage thresholds and number of replicas to control when {ispn_operator} automatically scales nodes up or down.
+. Apply the changes.
+
+----
+spec:
+  ...
+  autoscale: <1>
```
dmvolod, line 19: @rigazilla as idea, maybe we can also add enable flag to the  _autoscale_ section as it contains several options. When user will try to playground with this options, it would be nice to easy enable/disable without add/remove spec section.

documentation/asciidoc/topics/con_services.adoc
```diff
@@ -0,0 +1,34 @@
...
+By default, {cacheservice} nodes:
+
+* Synchronously distribute data across the cluster to ensure consistency.
+* Maintain single copies of cache entries to reduce size.
+* Store cache entries off-heap and use eviction for JVM efficiency.
+* Ensure data consistency with a default partition handling configuration.
+
+[IMPORTANT]
+====
+Restarting {cacheservice} nodes removes all data in the cache. Restarts occur
```
dmvolod, line 22: @oraNod looks like not any update on the **Infinispan** CR will occurs restarts.

oraNod: Thanks, Dmitry. This note was a bit of a last minute request from the support team and it was kind of an awkward fit. I've reworked it to include as part of the shutdown/restart section.

documentation/asciidoc/topics/proc_configuring_service_type.adoc
```diff
@@ -0,0 +1,33 @@
...
+    type: DataGrid
+----
+
+[NOTE]
+====
+You cannot change `.spec.service.type` after you create {brandname} clusters.
+
+For example, you create a cluster of {cacheservice} nodes but then want to
+change to {datagridservice} nodes. In this case you must create a new cluster
+with {datagridservice} nodes in a different {k8s} namespace.
```
dmvolod, line 27: @oraNod we can also add 'delete and create from scratch option' not only different {k8s} namespace

documentation/asciidoc/topics/proc_restart.adoc
```diff
@@ -1,8 +1,13 @@
...
 = Restarting {brandname} Clusters
+Complete the following procedure to restart {brandname} clusters after shutdown.
 
-* Set the value of `spec.replicas` to the same number of nodes that were in the
-cluster before you shut it down.
+.Procedure
+
+* Set the value of `spec.replicas` to an appropriate number of nodes for your cluster.
+
+For {datagridservice} nodes, the value of `spec.replicas` must be the same
```
dmvolod, line 9: Looks like we should also notify customers, that they are not needed to remember number of replicas and can get it from status.replicasWantedAtRestart

documentation/asciidoc/topics/proc_shut_down_gracefully.adoc
```diff
@@ -1,5 +1,15 @@
 [id='shutting_down-{context}']
-= Gracefully Shutting Down {brandname} Clusters
+= Shutting Down {brandname} Clusters
+Shutting down {cacheservice} nodes removes all data in the cache. For
+{datagridservice} nodes, you should configure the storage size for
+{datagridservice} nodes to ensure that the persistent volume can hold all your
+data.
+
+If the available container storage is less than the amount of memory available
+to {datagridservice} nodes, {brandname} writes an exception to logs and data
```
dmvolod, line 9: Operator will write a following message to the log: **WARNING: persistent volume size is less than memory size. Graceful shutdown may not work.** We are planning to send events like this to the event queue instead of logs.

documentation/asciidoc/topics/ref_cacheservice_resources.adoc
```diff
@@ -0,0 +1,46 @@
...
+      org.infinispan: trace
+      org.jgroups: trace
+  expose: <10>
+    type: LoadBalancer
+----
+
+<1> Names the {brandname} cluster.
+<2> Specifies the number of nodes in the cluster.
+<3> Creates {cacheservice} clusters.
+<4> Configures the number of replicas for each entry across the cluster.
```
dmvolod, line 40: @oraNod maybe add _cache_ between  _each_ and _entry_ Wdyt?

documentation/asciidoc/topics/proc_restart.adoc
oraNod: @dmvolod is `status.replicasWantedAtRestart` available from `kubectl get infinispan -o yaml`?

dmvolod: Yes, @oraNod sure
kubectl get infinispan cr-name -o=jsonpath='{.status.replicasWantedAtRestart}'

dmvolod: @rigazilla if we need to have replicas = replicasWantedAtRestart in some cases, can we set this option automatically without user notation as implemented now?

documentation/asciidoc/topics/ref_default_cache_service_config.adoc
rigazilla: @oraNod sounds good. We also need to update the `owners` attribute which now 2 by default if not specified in `infinispan.spec.service.replicationFactor`

documentation/asciidoc/topics/proc_restart.adoc
rigazilla: @dmvolod yeah sound very reasonable

documentation/asciidoc/topics/proc_configuring_autoscaling.adoc
rigazilla: @dmvolod sounds a good improvement, maybe we can add a `disable` flag so it's not showed by default

documentation/asciidoc/topics/ref_default_cache_service_config.adoc
oraNod: @rigazilla hopefully we're good now. Does `max-size="<maximum_size_in_bytes>"` look good? IIRC the operator calculates the maximum size when it creates pods.

rigazilla: @oraNod yep you're right

oraNod: great!

documentation/asciidoc/topics/proc_restart.adoc
oraNod: @dmvolod and @rigazilla do we need to create a separate issue to set `replicasWantedAtRestart` automatically?

documentation/asciidoc/topics/proc_configuring_autoscaling.adoc
oraNod: @dmvolod and @rigazilla need a separate issue for the `disable` flag too?

documentation/asciidoc/topics/proc_restart.adoc
rigazilla: @oraNod, I think a separate issue is better. so we don't mix too many things here

documentation/asciidoc/topics/proc_configuring_autoscaling.adoc
dmvolod: Yes, @rigazilla disable flag is much better
@oraNod maybe we can add option to docs now and implement it in the 2.0.0 milestone later

documentation/asciidoc/topics/proc_restart.adoc
oraNod: https://github.com/infinispan/infinispan-operator/issues/434

documentation/asciidoc/topics/proc_configuring_autoscaling.adoc
oraNod: https://github.com/infinispan/infinispan-operator/issues/435

oraNod: @dmvolod I think first comes the implementation then comes the doc

dmvolod:
@rigazilla we're missing question from @oraNod related to the autoscale default values

rigazilla:
> @rigazilla we're missing question from @oraNod related to the autoscale default values
@dmvolod we resolved via chat, but yeah I'll leave a trace also here

documentation/asciidoc/topics/proc_configuring_autoscaling.adoc
rigazilla: `maxReplicas` and `minReplicas` define the max/min number of pods the autoscaler is allowed to start. They aren't related to the owners/replicationFactor which is a different thing.

PR #436, opened 2020-06-16
ISPN-12016 cache store config

oraNod:
https://issues.redhat.com/browse/ISPN-12016


oraNod:
thanks for the reviews @rigazilla and @dmvolod !

Issue #437, opened 2020-06-16
Operator needs rights to create Route/Ingress


rigazilla:

dmvolod:
@rigazilla looks like we can close this issue as you are fixed it in #438
P.S. You can easy add keywords to commit/PR for automatically issue closing if one PR equals one issue
https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#:~:text=Closing%20an%20issue%20in%20the%20same%20repository,merged%20into%20the%20default%20branch.

dmvolod:
I'm closing this issue, as rights have be fixed

PR #438, opened 2020-06-16
Added rights to create Route/Ingress

rigazilla:
Fix for #437

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -41,6 +41,19 @@ rules:
   verbs:
   - '*'
 - apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
+  - routes/custom-host
+  verbs:
+  - '*'
+- apiGroups:
+  - networking.k8s.io
+  resources:
+  - ingresses
+  verbs:
+  - '*'
+- apiGroups:
   - monitoring.coreos.com
   resources:
   - servicemonitors
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -21,7 +21,7 @@ import (
 	"gopkg.in/yaml.v2"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
-	extv1 "k8s.io/api/extensions/v1beta1"
+	networkingv1beta1 "k8s.io/api/networking/v1beta1"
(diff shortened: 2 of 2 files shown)


deploy/role.yaml
```diff
@@ -40,6 +40,18 @@ rules:
   - statefulsets/finalizers
   verbs:
   - '*'
+- apiGroups:
+  - "route.openshift.io"
+  resources:
+  - routes
```
dmvolod, line 46: @rigazilla looks like we also need following resource
- routes/custom-host

deploy/role.yaml
```diff
@@ -40,6 +40,18 @@ rules:
...
+  verbs:
+  - '*'
+- apiGroups:
+  - "extensions"
```
dmvolod, line 50: @rigazilla is "extension" correct apiGroup. It's probably must be networking.k8s.io?
https://kubernetes.io/docs/concepts/services-networking/ingress/

deploy/role.yaml
rigazilla: right, user can also set the host. Thanks @dmvolod !

deploy/role.yaml
rigazilla: ... and right :) 
I've used an on group for Ingress. I'll change also in the code.
Thanks again @dmvolod !

deploy/role.yaml
```diff
@@ -40,6 +40,19 @@ rules:
...
   - '*'
+- apiGroups:
+  - "route.openshift.io"
+  resources:
+  - routes
+  - routes/custom-host
+  verbs:
+  - '*'
+- apiGroups:
+  - "networking.k8s.io"
```
dmvolod, line 51: @rigazilla just last points, do we need double quotes for apiGroups and can we restrict verbs instead of grant all access?

rigazilla: > @rigazilla just last points, do we need double quotes for apiGroups?

yeah they are useless

> and can we restrict verbs instead of grant all access?

that would be correct, but we're basically allowing all verbs for all the other resources. How about planning a fine tuning on all the roles/verbs in a separate issue?

dmvolod: Yes, @rigazilla it would be nice to do with separate issue as looks like we have much more rights that needed

rigazilla:
merged, and thanks @dmvolod for the review!

PR #441, opened 2020-06-17
ISPN-12026 Fetch the correct IP:port when NodePort is used

pruivo:
https://issues.redhat.com/browse/ISPN-12026

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -430,7 +430,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 
 	// List the pods for this infinispan's deployment
 	podList := &corev1.PodList{}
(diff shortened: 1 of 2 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1759,18 +1720,76 @@ func appendKubernetesRemoteLocation(xsite *ispnutil.XSite, infinispan *infinispa
...
+	err := k.Client.List(context.TODO(), workerList, listOps)
+	if err != nil {
+		return "", 0, err
+	}
+
+	for _, node := range workerList.Items {
+		//host := k.PublicIP() //returns REST API endpoint. not good.
+		//iterate over the all the nodes and return the first ready.
+		nodeStatus := node.Status
+		for _, nodeCondition := range nodeStatus.Conditions {
```
dmvolod, line 1756: Yes, good implementation, @pruivo 
Just FYI: fields selector can be more elegant solution for k8s...but not supported in all cases

pruivo:
already included in https://github.com/infinispan/infinispan-operator/pull/424

PR #442, opened 2020-06-19
Check array size before access item

rigazilla:
Graceful shutdown doesn't work if a pod has no ready container.

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -457,7 +457,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 				// send a graceful shutdown to the first ready pod
 				// if there's no ready pod we're in trouble
 				for _, pod := range podList.Items {
-					if len(pod.Status.ContainerStatuses) > 0 && pod.Status.ContainerStatuses[0].Ready {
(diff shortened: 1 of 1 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -482,7 +482,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 				// cluster has a `stopping` wait for all the pods becomes unready
 				reqLogger.Info("Waiting that all the pods become unready")
 				for _, pod := range podList.Items {
-					if pod.Status.ContainerStatuses[0].Ready {
+					if len(pod.Status.ContainerStatuses) > 0 && pod.Status.ContainerStatuses[0].Ready {
```
dmvolod, line 485: @rigazilla we already have IsPodReady() func in common package. Can we reuse it for all checks or create a new common func as checks pefromce more than once?

rigazilla: right @dmvolod! done

rigazilla:
> Thanks, @rigazilla
> We should also validate all places where we are checking pods/containers state to use common method as just in our current situation pod=container

yep, we need to reserve some time to do a code assessment.

rigazilla:
merged, thanks!

PR #443, opened 2020-06-19
New metadata file generated with operator-sdk

rigazilla:
metadata for 2.0.0.
Next step should be adopt `operator-sdk` tools


rigazilla:
I think we can merge this, since for a series of unexpected things it's already published on operator hub

rigazilla:
merged

PR #452, opened 2020-06-30
Fix route or ingress name to be created

### LINKED ISSUE #449: Route and Ingress are creating with one name but validating with another one
Route and Ingress are creating with cluster name (infinispan.Name), but validating with infinispan.GetServiceExternalName()

dmvolod:
Fixes #449


rigazilla:
merged, thanks @dmvolod !

Issue #471, opened 2020-07-17
CacheService: default cache eviction size has wrong set with low memory


rigazilla:
Eviction size computation gives unpredictable result (unsigned subtraction overflow) if pod has low memory limit.

I this case the operator should either returns an error from the reconcile loop or starts infinispan without eviction size limit.

dmvolod:
@rigazilla it also get unpredictable results, as all calculations are performing against node, where podList.Items[0] is scheduling. This means in case of pod re-scheduling or any others moving or decreasing number of nodes. We should define an algorithm which is not related on whole node start options. For example, user defined values or calculated with formulas, and if we can fit  with these values, go next, if not, try to reduce amount of requested memory and try to fit again.

rigazilla:
@dmvolod if it's ok to you, let's discuss new ideas for cache sizing in #470.

In this PR I would like to just fix current behaviour: I would go for the option to return an error if memory is not enough.

dmvolod:
Oh, yes, @rigazilla moving to the discussion issue

dmvolod:
@rigazilla should we close this issue?

rigazilla:
Now the operator blocks the cluster startup if pods are configured with insufficient memory

PR #474, opened 2020-07-20
Abort cluster startup on insufficient pod memory #471

rigazilla:
This aborts infinispan deployment if a CacheService is requested and pods have to little memory

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -157,6 +157,21 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	// Apply defaults if not already set
 	infinispan.ApplyDefaults()
 
+	// Perform all the possible preliminary checks before go on
+	recResult, err := infinispan.PreliminaryChecks()
+	if err != nil {
+		if infinispan.SetCondition("preliminaryChecksFailed", "true", err.Error()) {
+			err1 := r.client.Status().Update(context.TODO(), infinispan)
+			reqLogger.Error(err1, "Could not update error conditions")
+		}
+		return *recResult, err
+	} else {
+		if infinispan.RemoveCondition("preliminaryChecksFailed") {
+			err1 := r.client.Status().Update(context.TODO(), infinispan)
+			reqLogger.Error(err1, "Could not update error conditions")
+		}
+	}
+
 	// Check x-site configuration first.
 	// Must be done before creating any Infinispan resources,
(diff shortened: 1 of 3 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -157,6 +157,21 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	// Apply defaults if not already set
 	infinispan.ApplyDefaults()
 
+	// Perform all the possible preliminary checks before go on
+	recResult, err := infinispan.PreliminaryChecks()
+	if err != nil {
+		if infinispan.SetCondition("PreliminaryChecksFailed", "true", err.Error()) {
```
dmvolod, line 163: @rigazilla what do you think about to introduce conditions enum in infinispan type?

rigazilla: @dmvolod yeah, I usually tent to formalize things as late as possible, but It's probably time to give a shape to the bunch of string conditions. It would be great to have a separate discussion on this and maybe labels/selectors, to check if we're doing something that can be easily integrated with `oc wait` and maybe other `oc` commands.

dmvolod: Yes, agree with you, @rigazilla 
Other things in this PR good for me

dmvolod: @rigazilla , looks like this PreliminaryChecksFailed need to start from lowercase or not?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -157,6 +157,21 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+	// Perform all the possible preliminary checks before go on
+	recResult, err := infinispan.PreliminaryChecks()
+	if err != nil {
+		if infinispan.SetCondition("preliminaryChecksFailed", "true", err.Error()) {
+			err1 := r.client.Status().Update(context.TODO(), infinispan)
+			reqLogger.Error(err1, "Could not update error conditions")
+		}
+		return *recResult, err
+	} else {
+		if infinispan.RemoveCondition("PreliminaryChecksFailed") {
```
dmvolod, line 169: @rigazilla , looks like this PreliminaryChecksFailed need to start from lowercase also or not?

rigazilla: oh yes!

rigazilla:
merged, @dmvolod thanks for your review!

PR #478, opened 2020-07-22
Review run-local deploy scripts and tests running for OCP and K8S

### LINKED ISSUE #476: Review run-local deploy scripts running for OCP and K8S

dmvolod:
Fixes #476 

It seems to RBAC doesn't make sense for local run and travis tests.
It's not needed for accounts with cluster admin rights and not enough for non-privileged account.
If we need to run with less privileged account, we need to add scripts for admins and specific RoleBindings and ClusterRoleBindings which will subjects to user.

### DIFF
--- a/test/e2e/util/test_kubernetes.go
+++ b/test/e2e/util/test_kubernetes.go
@@ -205,60 +205,6 @@ func (k TestKubernetes) GracefulRestartInfinispan(infinispan *ispnv1.Infinispan,
 	ExpectNoError(err)
 }
 
-// CreateOrUpdateRole creates or updates a Role in the given namespace
-func (k TestKubernetes) CreateOrUpdateRole(role *rbacv1.Role, namespace string) {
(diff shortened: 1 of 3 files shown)


test/e2e/util/test_kubernetes.go
```diff
@@ -469,41 +469,12 @@ func (k TestKubernetes) DeleteSecret(secret *v1.Secret) {
 
 // RunOperator runs an operator on a Kubernetes cluster
 func (k TestKubernetes) RunOperator(namespace string) chan struct{} {
-	k.installRBAC(namespace)
```
rigazilla, line 472: Seems a good cleanup to me. @Crumby do you need the installRBAC func for your tests?

dmvolod: @rigazilla let's merge this PR on master
in case of some tests will fails for current release, 2.0.x will be suitable for testing, and for master we will revert this and just fix RBAC creation for scripts only

rigazilla:
merged, thanks @dmvolod !

Issue #484, opened 2020-07-24
Infinispan operator breaks the OpenShift catalog page


dgutride:
This file - https://github.com/infinispan/infinispan-operator/blob/master/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml contains a custom resource definition that is missing a display name for the cache:

https://github.com/infinispan/infinispan-operator/blob/master/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml#L226

Fixing it within the operator after installed corrects the problem with the catalog page - recommend adding a displayName property to this section.

PR #485, opened 2020-07-24
Add displayName for Infinispan Cache

dgutride:
Closes https://github.com/infinispan/infinispan-operator/issues/484

The catalog page in OpenShift would fail to open when the operator is installed - this adds a necessary field for the page to use.


rigazilla:
merged, thanks @dgutride !

Issue #448, opened 2020-06-30
Sites structure displays an empty value if it not defined


dmvolod:
```
spec:
  container:
    cpu: 1000m
    memory: 1Gi
  image: infinispan/server:10.1.8.Final
  logging:
    categories:
      org.infinispan: trace
      org.jgroups: debug
  replicas: 1
  security:
    endpointEncryption:
      certSecretName: test-datagrid-cert-secret
      certServiceName: service.beta.openshift.io
      type: service
    endpointSecretName: test-datagrid-secret
  service:
    container:
      storage: 1Gi
    sites:
      local:
        expose: {}
        name: ""
    type: DataGrid
status:
```

Config also set with xsite empties 
```
data:
  infinispan.yaml: |
    clusterName: test-datagrid
    jgroups:
      transport: tcp
      dnsPing:
        query: test-datagrid-ping.ispn-test.svc.cluster.local
      diagnostics: false
    keystore:
      path: /opt/infinispan/server/conf/keystore
      password: password
      alias: server
      crtPath: /etc/encrypt
    xsite:
      address: ""
      name: ""
      port: 0
      backups: []
    logging:
      categories:
        org.infinispan: trace
        org.jgroups: debug
```

dmvolod:
I'm closing this issue as it has been fixed now

PR #496, opened 2020-07-29
Regen CRD's with operator-sdk

dmvolod:


dmvolod:
@rigazilla after correct schema generating, the CRD schema validation has failed due to #448
We should avoid empty values on Sites structure if it not defined

rigazilla:
I've changed the spec.service.sites to pointer, so the `optional` semantic can work.
We must do this for all the optional struct in the spec

PR #502, opened 2020-07-31
Regen install bundle

dmvolod:


rigazilla:
merged, thanks @dmvolod !

PR #503, opened 2020-07-31
Regen install bundle for 2.0.x

dmvolod:


pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -38,7 +38,7 @@ const (
 	// Additional caches can be created, but only as copies of default cache.
 	ServiceTypeCache ServiceType = "Cache"
 
-	// Deploys Infinispan to act like a data grid.
+	// Deploys Infinispan to act like a Data Grid.
```
rigazilla, line 41: do we need it uppercase?
```suggestion
	// Deploys Infinispan to act like a data grid.
```

dmvolod: Yes, sure, it more type than name in this case

dmvolod:
@rigazilla looks like we should also regen OLM code. Is it automated or not right now?
I didn't see any CRD update in the OLM folder form the `deploy` CRD's.

rigazilla:
> @rigazilla looks like we should also regen OLM code. Is it automated or not right now?
> I didn't see any CRD update in the OLM folder form the `deploy` CRD's.

I'm updating it manually atm.
There was a partial automation process in place, but it's has become obsolete very soon.
We need something new with support for the new bundle format

dmvolod:
merged, thanks, @rigazilla

PR #508, opened 2020-08-04
Cache CR's becomes orphan after cluster deletion

### LINKED ISSUE #450: Cache CR's becomes orphan after cluster deletion
Looks like we need to update Cache CR with Infinispan CR owner reference after Cache CR processing to prevent orphans after Infinispan CR deletion. We should avoid to use finalizers for this use case.

dmvolod:
Fixes #450

### DIFF
--- a/pkg/controller/cache/cache_controller.go
+++ b/pkg/controller/cache/cache_controller.go
@@ -3,24 +3,24 @@ package cache
 import (
 	"context"
 	"fmt"
-	"time"
-
-	"github.com/infinispan/infinispan-operator/pkg/controller/utils/cache"
-	"github.com/infinispan/infinispan-operator/version"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/go-logr/logr"
 	infinispanv1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
 	infinispanv2alpha1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v2alpha1"
+	"github.com/infinispan/infinispan-operator/pkg/controller/constants"
 	ispnutil "github.com/infinispan/infinispan-operator/pkg/controller/infinispan/util"
+	"github.com/infinispan/infinispan-operator/pkg/controller/utils/cache"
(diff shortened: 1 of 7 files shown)


dmvolod:
@rigazilla please review, I tried to fix #509, but it still not resolved

deploy/cr/auth/cr_minimal_with_auth.yaml
```diff
@@ -4,5 +4,7 @@ metadata:
   name: example-infinispan
 spec:
   replicas: 2
+  service:
+    type: DataGrid
```
rigazilla, line 8: better to create a new file under `deploy/cr/datagrid` instead of changing this

dmvolod: Yes, sure, will create a new one with DataGrid type

rigazilla:
I think we can merge this first. I'll rebase my #515

dmvolod:
> I think we can merge this first. I'll rebase my #515

No, @rigazilla , please merge #515 before, it's not a problem to rebase this PR later

pkg/controller/cache/cache_controller.go
```diff
@@ -176,13 +177,23 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
 				}
 			}
 		}
+
+		ok, err := common.IsOwner(ispnInstance, instance, r.scheme)
```
rigazilla, line 181: It seems that the `SetControllerReference` does some similar checks. Do we need to add our owns?

dmvolod: Yes, `SetControllerReference` check for already set owner, but this validation prevents a redundant CR update

pkg/controller/cache/cache_controller.go
```diff
@@ -176,13 +177,21 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
 				}
 			}
 		}
+
+		if !metav1.IsControlledBy(instance, ispnInstance) {
```
dmvolod, line 181: Changed to `apimachinery` implementation instead of custom one

rigazilla:
merged, thanks @dmvolod !

PR #513, opened 2020-08-05
Validate unhandled errors

dmvolod:


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1227,9 +1232,17 @@ func (r *ReconcileInfinispan) reconcileGracefulShutdown(ispn *infinispanv1.Infin
...
+						logger.Error(err, "failed to get pod", "Pod.Name", pod.Name)
+						return &reconcile.Result{}, err
+					}
 					runtimePod.Spec.RestartPolicy = corev1.RestartPolicyNever
-					r.client.Update(context.TODO(), runtimePod)
+					err = r.client.Update(context.TODO(), runtimePod)
+					if err != nil {
+						logger.Error(err, "failed to update pod", "Pod.Name", pod.Name)
+						return &reconcile.Result{}, err
+					}
```
rigazilla, line 1245: If it's ok to you, this for loop can be remove.
The rest of this PR looks great!

dmvolod: Maybe we need to find other way to do this, I will try to remove and some research if it really needed

dmvolod: The pods are controlled by StatefulSet and can't be updated according this KB https://access.redhat.com/solutions/4800611
The StatefulSet also can't to change `RestartPolicy` other than `Always`.
Let's remove this code, @rigazilla and you probably extend this logic in your PR #514

dmvolod:
Thanks for review, @rigazilla , merged
I'm keeping two unsquashed commits for tracking these changes

PR #514, opened 2020-08-05
remove useless if and update

rigazilla:
removed useless if (always true)
removed redundant  infinispan CR update
added error check on statefulset update

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -1246,21 +1246,19 @@ func (r *ReconcileInfinispan) reconcileGracefulShutdown(ispn *infinispanv1.Infin
 			}
 
 			// If here all the pods are unready, set statefulset replicas and ispn.replicas to 0
-			if *statefulSet.Spec.Replicas != 0 {
-				ispn.Status.ReplicasWantedAtRestart = *statefulSet.Spec.Replicas
-				err := r.client.Status().Update(context.TODO(), ispn)
-				if err != nil {
-					logger.Error(err, "failed to update Infinispan status")
-					return &reconcile.Result{}, err
-				}
-				updateStatus = true
-				zeroReplicas := int32(0)
-				statefulSet.Spec.Replicas = &zeroReplicas
-				err = r.client.Update(context.TODO(), statefulSet)
-				if err != nil {
-					logger.Error(err, "failed to update StatefulSet", "StatefulSet.Name", statefulSet.Name)
-					return &reconcile.Result{}, err
-				}
+			ispn.Status.ReplicasWantedAtRestart = *statefulSet.Spec.Replicas
+			err := r.client.Status().Update(context.TODO(), ispn)
+			if err != nil {
+				logger.Error(err, "failed to update Infinispan status")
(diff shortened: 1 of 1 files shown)


dmvolod:
@rigazilla I also open PR #513 update related to unhandled errors. Please have a look. You can use code from it as a large number of places to be validate

dmvolod:
Please also note, that handling errors in pod updates always fails
https://travis-ci.org/github/infinispan/infinispan-operator/jobs/715198324

rigazilla:
> Please also note, that handling errors in pod updates always fails
> https://travis-ci.org/github/infinispan/infinispan-operator/jobs/715198324

this is surprising :) 
I would say we can remove that set

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1246,21 +1246,19 @@ func (r *ReconcileInfinispan) reconcileGracefulShutdown(ispn *infinispanv1.Infin
...
-				updateStatus = true
-				zeroReplicas := int32(0)
-				statefulSet.Spec.Replicas = &zeroReplicas
-				err = r.client.Update(context.TODO(), statefulSet)
-				if err != nil {
-					logger.Error(err, "failed to update StatefulSet", "StatefulSet.Name", statefulSet.Name)
-					return &reconcile.Result{}, err
-				}
+			ispn.Status.ReplicasWantedAtRestart = *statefulSet.Spec.Replicas
+			err := r.client.Status().Update(context.TODO(), ispn)
```
dmvolod, line 1250: `updateStatus` flag indicates that we need to update Status later in one place. Do we really need to update Status there

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1246,21 +1246,19 @@ func (r *ReconcileInfinispan) reconcileGracefulShutdown(ispn *infinispanv1.Infin
...
+			updateStatus = true
+			zeroReplicas := int32(0)
+			statefulSet.Spec.Replicas = &zeroReplicas
+			err = r.client.Update(context.TODO(), statefulSet)
+			if err != nil {
+				logger.Error(err, "failed to update StatefulSet", "StatefulSet.Name", statefulSet.Name)
+				return &reconcile.Result{}, err
 			}
 		}
 		if statefulSet.Status.CurrentReplicas == 0 {
```
dmvolod, line 1264: Next lines also contains code, which optimizer can ignore in some cases and didn't set a proper conditions
```
if statefulSet.Status.CurrentReplicas == 0 {
			updateStatus = updateStatus || ispn.SetCondition("gracefulShutdown", metav1.ConditionTrue, "")
			updateStatus = updateStatus || ispn.SetCondition("stopping", metav1.ConditionFalse, "")
}
```

pkg/controller/infinispan/infinispan_controller.go
rigazilla: We need to do it before zeroing the Replicas in the StatefulSet, after that the replicas needed at restart is lost if an error occurs in the Infinispan.Status update.
I've removed the `updateStatus=true` that is not needed

rigazilla:
merged, thanks!

Issue #509, opened 2020-08-04
Cache CR reconciles multiple times once created


dmvolod:
```
{"level":"info","ts":1596562591.8567762,"logger":"controller_cache","msg":"Reconciling Cache","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562591.857102,"logger":"controller_cache","msg":"Identify the target cluster","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562592.2456076,"logger":"controller_cache","msg":"Cache already exists","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562592.2457194,"logger":"controller_cache","msg":"Update CR with owner reference info","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562592.4072902,"logger":"controller_cache","msg":"Update CR status with connection info","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562592.4608173,"logger":"controller_cache","msg":"Reconciling Cache","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562592.4609013,"logger":"controller_cache","msg":"Identify the target cluster","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562592.825822,"logger":"controller_cache","msg":"Cache already exists","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562592.8262186,"logger":"controller_cache","msg":"Update CR status with connection info","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562593.8792214,"logger":"controller_cache","msg":"Reconciling Cache","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562593.8793037,"logger":"controller_cache","msg":"Identify the target cluster","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
{"level":"info","ts":1596562594.261027,"logger":"controller_cache","msg":"Cache already exists","Request.Namespace":"infinispan-operator","Request.Name":"example-cache"}
[...]

PR #515, opened 2020-08-06
Reduce events on status changes #509

rigazilla:
This reduce to 2 the reconcile calls on cache creation

### DIFF
--- a/pkg/controller/cache/cache_controller.go
+++ b/pkg/controller/cache/cache_controller.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/infinispan/infinispan-operator/pkg/controller/utils/cache"
+	"github.com/infinispan/infinispan-operator/version"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


pkg/controller/cache/cache_controller.go
```diff
@@ -206,11 +198,12 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
 		instance.Status.ServiceName = serviceList.Items[0].Name
 		statusUpdate = true
 	}
-	statusUpdate = statusUpdate || instance.SetCondition("Ready", metav1.ConditionTrue, "")
+	conditionUpdate := instance.SetCondition("Ready", metav1.ConditionTrue, "")
```
dmvolod, line 201: Oh, code optimizator prevents second expression to calculate it first time :)

dmvolod:
But looks like we will add additional call with SetControllerReference update

pkg/controller/cache/cache_controller.go
rigazilla: yep.
Well... I could probably just swap the `||` operands order

rigazilla:
merged, thanks!

PR #519, opened 2020-08-07
Allow initContainer image to be specified

### LINKED ISSUE #518: MAKE_DATADIR_WRITABLE initContainer image should be configurable
In trying to implement the workaround for #392 I noticed that the `initContainer` image is [hardcoded](https://github.com/infinispan/infinispan-operator/blob/817588187a58ba58286af39aa626a5789cf6c0db/pkg/controller/infinispan/infinispan_controller.go#L770) to `busybox`.

Unfortunately I run in an air gapped environment and need to be able to change the image to reference an internal registry. Something similar to [`DEFAULT_IMAGE`](https://github.com/infinispan/infinispan-operator/blob/51e06b80c75de84e46c4037441ebdee86bab9b68/deploy/operator.yaml#L36) would be appreciated.

donovanmuller:
Using the `INITCONTAINER_IMAGE` environment variable you can now specify a custom image for the initContainer that is used if `MAKE_DATADIR_WRITABLE=true`

Closes #518

### DIFF
--- a/pkg/controller/constants/constants.go
+++ b/pkg/controller/constants/constants.go
@@ -12,6 +12,9 @@ var (
 	// DefaultImageName is used if a specific image name is not provided
 	DefaultImageName = common.GetEnvWithDefault("DEFAULT_IMAGE", "infinispan/server:latest")
 
+	// InitContainerImageName allows a custom initContainer image to be used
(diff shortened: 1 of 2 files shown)


dmvolod:
@donovanmuller please squash commits

rigazilla:
merged, thanks @donovanmuller !

PR #520, opened 2020-08-07
Able to use server-native image

### LINKED ISSUE #473: Able to use server-native image
Operator is not able to run native image due following JDK options are not supported on it
- MaxRAM
- UseSerialGC
- MinHeapFreeRatio
- MaxHeapFreeRatio

Image: quay.io/infinispan/server-native:11.0.1.Final

We should detect/define that operator is running native image and don't set these options by default
@rigazilla wdyt?

Thanks the colleague of mine, @dsevost which is finding this during Quarkus demo preparation
https://github.com/e2e-openshift-demos/quarkus-demo/blob/b-step-03_4/jdg/datagrid-operator-setup.sh#L58

dmvolod:
Fixes #473 
Waiting until #532 is not merged and backported to master

### DIFF
--- a/test/e2e/constants/constants.go
+++ b/test/e2e/constants/constants.go
@@ -4,7 +4,7 @@ import (
 	"strings"
 	"time"
 
-	constants "github.com/infinispan/infinispan-operator/pkg/controller/constants"
+	"github.com/infinispan/infinispan-operator/pkg/controller/constants"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -36,6 +36,7 @@ var (
 	OperatorUpgradeStage = strings.ToUpper(constants.GetEnvWithDefault("OPERATOR_UPGRADE_STAGE", OperatorUpgradeStageNone))
 	CleanupInfinispan    = strings.ToUpper(constants.GetEnvWithDefault("CLEANUP_INFINISPAN_ON_FINISH", "true"))
 	ExpectedImage	     = constants.GetEnvWithDefault("EXPECTED_IMAGE", "infinispan/server:12.0")
+	NativeImageName      = constants.GetEnvWithDefault("NATIVE_IMAGE", "infinispan/server-native:12.0")
 	ExposeServiceType    = constants.GetEnvWithDefault("EXPOSE_SERVICE_TYPE", "NodePort")
 
 	OperatorUpgradeStateFlow = []string{"upgrade", "stopping", "wellFormed"}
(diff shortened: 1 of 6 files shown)


test/e2e/main_test.go
```diff
@@ -307,8 +307,9 @@ func TestCacheService(t *testing.T) {
 	t.Parallel()
 	spec := DefaultSpec.DeepCopy()
 	name := strcase.ToKebab(t.Name())
-
-	spec.ObjectMeta.Name = name
+	
+	spec.Name = name
+	spec.Spec.Image = pointer.StringPtr(tconst.NativeImageName)
```
ryanemerson, line 312: Do we want the native image to always be used by `TestCacheService`?

It would be good if we could parametise the tests so that all of them are executed with both the normal and native image.

test/e2e/constants/constants.go
```diff
@@ -28,6 +28,7 @@ var (
 	RunLocalOperator  = strings.ToUpper(constants.GetEnvWithDefault("RUN_LOCAL_OPERATOR", "true"))
 	RunSaOperator     = strings.ToUpper(constants.GetEnvWithDefault("RUN_SA_OPERATOR", "false"))
 	ImageName         = constants.GetEnvWithDefault("IMAGE", "infinispan/server:12.0")
+	NativeImageName   = constants.GetEnvWithDefault("NATIVE_IMAGE", "quay.io/infinispan/server-native:12.0")
```
ryanemerson, line 31: We should either use `Quay.io` or docker hub for both images. I don't mind which, but we should be consistent with `ImageName` and `NativeImageName`.

ryanemerson (changes requested):

test/e2e/constants/constants.go
dmvolod: Yes, sure, @ryanemerson . What is docker hub native image location?

test/e2e/main_test.go
dmvolod: Yes, it could be. But every test iteration will add additional time to CI execution. If it's ok, I will add this option.

test/e2e/constants/constants.go
ryanemerson: `infinispan/server-native:12.0`

test/e2e/main_test.go
ryanemerson: I think the extra test coverage would be worth it, @rigazilla any objections?

dmvolod: @rigazilla I'm waiting you feedback, do we need to run test TestCacheService with standard and native images or just with native. Could you please review this PR.

rigazilla: @dmvolod I agree with @ryanemerson that ideally we should run the testsuite agaist all the community released images.
Testsuite changes probably need an update, due to #614 ?

dmvolod: Yes will modify test and rebase on top of latest fix

dmvolod:
Something happens wrong with CI, try to close this PR and open it again

PR #521, opened 2020-08-07
Flag for disabling autoscale

### LINKED ISSUE #435: Add disable flag to the 'autoscale' section
@rigazilla as idea, maybe we can also add enable flag to the  _autoscale_ section as it contains several options. When user will try to playground with this options, it would be nice to easy enable/disable without add/remove spec section.

_Originally posted by @dmvolod in https://github.com/infinispan/infinispan-operator/pull/429_

rigazilla:
This resolves #435 
a bool field `disabled` has been added to the `autoscale` struct

### DIFF
--- a/deploy/operator-install.yaml
+++ b/deploy/operator-install.yaml
@@ -174,6 +174,8 @@ spec:
             autoscale:
               description: Autoscale describe autoscaling configuration for the cluster
               properties:
+                disabled:
+                  type: boolean
                 maxMemUsagePercent:
                   type: integer
                 maxReplicas:
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -410,7 +410,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	// All pods ready start autoscaler if needed
-	if infinispan.Spec.Autoscale != nil {
+	if infinispan.Spec.Autoscale != nil && infinispan.IsCache() {
 		addAutoscalingEquipment(types.NamespacedName{Namespace: infinispan.Namespace, Name: infinispan.Name}, r)
 	}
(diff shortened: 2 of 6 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -410,7 +410,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	// All pods ready start autoscaler if needed
-	if infinispan.Spec.Autoscale != nil {
+	if infinispan.Spec.Autoscale != nil && infinispan.Spec.Service.Type == infinispanv1.ServiceTypeCache {
```
dmvolod, line 413: We have an IsDataGrid() for easy validation

deploy/operator-install.yaml
```diff
@@ -174,6 +174,8 @@ spec:
             autoscale:
               description: Autoscale describe autoscaling configuration for the cluster
               properties:
+                disabled:
```
dmvolod, line 177: Looks like code-gen need to be run again

rigazilla: what's missing here? `make codegen` shouldn't be enough?

dmvolod: I'm not sure, but some comments can be missed

dmvolod: Yes, please run `make codegen` again

rigazilla: `make codegen` has nothing more to add. seems we're ok

Crumby:
@rigazilla I'm probably late on this one but this design doesn't fit k8s in my opinion. Feature being `disabled` is simply done by not-being defined. `disabled` field is in this case redundant and only present for lazy users. Also if this feature has disable field, shouldn't other features have one eg. xsite?

Place where `disabled` would make sense are features enabled by default, eg.: encryption, authentication.

Issue #498, opened 2020-07-30
Change to pointers all the optional structs in .Spec


rigazilla:
In the Infinispan CR optional fields represented by complex struct must be changed to pointer otherwise `optional` and `omitempty` attributes doesn't work

dmvolod:
It seems to be related to the #448
// +optional are related to the OpenAPI 3.0 schema generation
https://github.com/kubernetes-sigs/kubebuilder/issues/479

dmvolod:
Seems to already implemented for all structures. Closed

PR #522, opened 2020-08-07
changed to pointer some structs #498

rigazilla:
Most of the optional fields are actually defaulted by the operator, I've turned into operator only `spec.service.endpointEncryption` and `spec.logging`. Also defaulted `spec.service.container.storage`


pkg/apis/infinispan/v1/types_util.go
```diff
@@ -90,6 +90,9 @@ func (ispn *Infinispan) ApplyDefaults() {
 		cpuLimitString := comutil.ToMilliDecimalQuantity(constants.DefaultCPULimit)
 		ispn.Spec.Container.CPU = cpuLimitString.String()
 	}
+	if ispn.Spec.Service.Container.Storage == "" {
+		ispn.Spec.Service.Container.Storage = consts.DefaultPVSize.String()
```
dmvolod, line 94: The storage size is make sense for DataGrid only in current impl. Do we need to set defaults for all cases.
Otherwise, we also validate that `m.Spec.Service.Container.Storage != ""` and this is doesn't make sense now.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -787,7 +787,7 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 
 func setupServiceForEncryption(m *infinispanv1.Infinispan, ser *corev1.Service) {
 	ee := m.Spec.Security.EndpointEncryption
-	if strings.EqualFold(ee.Type, "Service") {
+	if ee != nil && strings.EqualFold(ee.Type, "Service") {
```
dmvolod, line 790: @rigazilla wdyth about to add getter for `ispn.Spec.Security.EndpointEncryption.Type` including strings.EqualFold conversion, as we have a large number of places to validate it.
Also, good point to align with enum definition maybe in this or separate PR.

rigazilla: how about apply Fold conversion in the ApplyDefault?
Or define compare func like IsService()?
Not a big problem cause we're more or less just turning letters uppercase but I don't like getters that do things under the hood.

rigazilla: probably fix data in the apply defaults is preferable

dmvolod: Yes, agree with you about Fold conversion in the ApplyDefault?
And we can also add function like IsSecurityService for comparison.

deploy/cr/cache/auth_secret.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 stringData:
   username: operator 
-  password: changemetoo
+  password: ja6O5oMmsWjzF7s2
```
dmvolod, line 4: Oh, looks like your test password :)

deploy/olm-catalog/infinispan.org_caches_crd.yaml
```diff
@@ -1,20 +1,20 @@
 apiVersion: apiextensions.k8s.io/v1beta1
```
dmvolod, line 1: The wrong file overwrite!

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -14,15 +14,26 @@ type InfinispanAuthInfo struct {
 
 // InfinispanSecurity info for the user application connection
 type InfinispanSecurity struct {
-	EndpointSecretName string             `json:"endpointSecretName,optional,omitempty"`
-	EndpointEncryption EndpointEncryption `json:"endpointEncryption,optional,omitempty"`
+	EndpointSecretName string              `json:"endpointSecretName,optional,omitempty"`
+	EndpointEncryption *EndpointEncryption `json:"endpointEncryption,optional,omitempty"`
 }
 
+// CertificateSourceType specifies all the possible sources for the encryption certificate
```
dmvolod, line 21: // +kubebuilder:validation:Enum=Service;Secret

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -234,18 +249,28 @@ func (ispn *Infinispan) IsUpgradeNeeded(logger logr.Logger) bool {
...
+		ispn.Spec.Security.EndpointEncryption.CertSecretName == "") {
+		if ispn.Spec.Security.EndpointEncryption == nil {
+			ispn.Spec.Security.EndpointEncryption = &EndpointEncryption{}
+		}
 		reqLogger.Info("Serving certificate service present. Configuring into CRD")
-		if ee.Type == "" {
-			ee.Type = "Service"
-			ee.CertServiceName = "service.beta.openshift.io"
+		if ispn.Spec.Security.EndpointEncryption.Type == "" {
+			ispn.Spec.Security.EndpointEncryption.Type = "Service"
```
dmvolod, line 263: Constant should use there

test/e2e/main_test.go
```diff
@@ -766,7 +766,7 @@ func getSchemaForRest(ispn *ispnv1.Infinispan) string {
 		panic(err.Error())
 	}
 	kubernetes.Kubernetes.Client.Get(context.TODO(), types.NamespacedName{Namespace: ispn.Namespace, Name: ispn.Name}, &curr)
-	if curr.Spec.Security.EndpointEncryption.Type != "" {
+	if curr.Spec.Security.EndpointEncryption != nil && curr.Spec.Security.EndpointEncryption.Type != "" {
```
dmvolod, line 769: We can introduce method like IsEndpointEncryptionTypeDefined

rigazilla: ok, looks nice

pkg/apis/infinispan/v1/infinispan_types.go
dmvolod: Oh, looks like this will not backward compatible

rigazilla: mmm yeah that's true

rigazilla:
merged, thanks for the review @dmvolod !

Issue #385, opened 2020-05-05
Integration with Skupper


rigazilla:
Skupper (https://skupper.io/) is a k8s component that provides intercluster communication.
Investigate if it can be useful to semplify cross site replication configuration.

dmvolod:
I think we can try to utilize both solutions, but first step `install and configure Skupper or Submariner` will be so huge
However next steps will be more easy

pruivo:
at least, the submariner has an operator: https://operatorhub.io/operator/submariner

not sure how good it is neither what it does xD

dmvolod:
> not sure how good it is neither what it does xD

Yeah, it utilize `sigs.k8s.io/mcs-api` and easy to expose cross-site services

dmvolod:
But seems to need access to the remote cluster Kubernetes API. In this case there is no value, except granting access on the k8s cluster level, not individual Infinispan 
https://submariner.io/getting-started/architecture/

PR #524, opened 2020-08-08
Include all nodes with label "node-role.kubernetes.io/worker=true"

### LINKED ISSUE #523: Worker label selector for getting NodePort host/port does not return any nodes
When configuring cross site replication and using a `NodePort` expose type on a Kubernetes cluster:

```yaml
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: infinispan
spec:
  replicas: 2
  security:
    endpointSecretName: xxx
  service:
    type: DataGrid
    expose:
      type: NodePort 
    sites:
      local:
        name: xxx 
        expose:
          type: NodePort 
...
```

the label selector for worker nodes does not return any nodes with the current empty label value on a RKE Kubernetes 1.15.11 cluster, despite their being ready worker nodes.  https://github.com/infinispan/infinispan-operator/blob/21d35fe983b504649dd773c23e74af9abac37696/pkg/controller/utils/infinispan/configuration/xsite.go#L173

donovanmuller:
Without a `node-role.kubernetes.io/worker` label value of `true`, no worker nodes are returned.

Tested on a Kubernetes 1.15 cluster. Unfortunately I don't have access to a OpenShift cluster to test if this behaviour is consistent but it should be.

Resolves #523

### DIFF
--- a/deploy/clusterrole.yaml
+++ b/deploy/clusterrole.yaml
@@ -7,13 +7,10 @@ rules:
   - ""
   resources:
   - nodes
-  - nodes/status
-  - nodes/metrics
-  - nodes/spec
-  - nodes/stats
   verbs:
   - get
   - list
+  - watch
 - apiGroups:
   - apiextensions.k8s.io
   resources:
--- a/pkg/controller/utils/infinispan/configuration/xsite.go
+++ b/pkg/controller/utils/infinispan/configuration/xsite.go
@@ -12,6 +12,7 @@ import (
 	"github.com/infinispan/infinispan-operator/pkg/controller/utils/common"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	restclient "k8s.io/client-go/rest"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -169,15 +170,22 @@ func getNodePortServiceHostPort(service *corev1.Service, k *util.Kubernetes, log
 	//oc create clusterrolebinding <name> -n ${NAMESPACE} --clusterrole=cluster-reader --serviceaccount=${NAMESPACE}:<account-name>
 	workerList := &corev1.NodeList{}
 
-	//select workers only. AFAIK, only the workers have the proxy to the pods
-	labelSelector := labels.SelectorFromSet(map[string]string{"node-role.kubernetes.io/worker": ""})
-
-	listOps := &client.ListOptions{
-		LabelSelector: labelSelector,
-	}
-	err := k.Client.List(context.TODO(), workerList, listOps)
+	//select workers first
+	req, err := labels.NewRequirement("node-role.kubernetes.io/worker", selection.Exists, nil)
 	if err != nil {
 		return "", 0, err
+	}
+	listOps := &client.ListOptions{
+		LabelSelector: labels.NewSelector().Add(*req),
+	}
+	err = k.Client.List(context.TODO(), workerList, listOps)
+
+	if err != nil || len(workerList.Items) == 0 {
+		// Fallback selecting everything
+		err = k.Client.List(context.TODO(), workerList, nil)
+		if err != nil {
+			return "", 0, err
+		}
 	}
 
 	for _, node := range workerList.Items {
--- a/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
@@ -172,10 +172,6 @@ spec:
               - ""
             resources:
               - nodes
-              - nodes/status
-              - nodes/metrics
-              - nodes/spec
-              - nodes/stats
             verbs:
               - get
               - list
--- a/deploy/operator-install.yaml
+++ b/deploy/operator-install.yaml
@@ -90,13 +90,10 @@ rules:
   - ""
   resources:
   - nodes
-  - nodes/status
-  - nodes/metrics
-  - nodes/spec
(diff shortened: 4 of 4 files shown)


dmvolod:
Thanks, @donovanmuller , I will check on OCP, but like like you're right

dmvolod:
@donovanmuller I've validated and suddenly this is will not working on OCP per this documentation
https://docs.openshift.com/container-platform/4.1/nodes/scheduling/nodes-scheduler-node-selectors.html

@rigazilla looks like we should to start to identify are we running on OCP or K8S in runtime

pkg/controller/utils/infinispan/configuration/xsite.go
```diff
@@ -170,7 +170,7 @@ func getNodePortServiceHostPort(service *corev1.Service, k *util.Kubernetes, log
 	workerList := &corev1.NodeList{}
 
 	//select workers only. AFAIK, only the workers have the proxy to the pods
-	labelSelector := labels.SelectorFromSet(map[string]string{"node-role.kubernetes.io/worker": ""})
+	labelSelector := labels.SelectorFromSet(map[string]string{"node-role.kubernetes.io/worker": "true"})
```
rigazilla, line 173: I'm wondering if we really need this selector at all. It seems that also "master" nodes expose NodePort service

dmvolod: This need to be tested

rigazilla: I've tested on OCP and minikube and all the nodes serve NodePort request

dmvolod: Oh, good, let's remove this label selector. We also need to backport this PR in 2.0.x also

dmvolod: @rigazilla you should also validate, that you don't have a cluster where `worker` and `master` node are combine on one host.

rigazilla: yeah:
OCP has 1 master, 2 worker
minikube only 1 master   (node-role worker)

looking at the NodePort docu, I didn't find any reference to the master/worker distinction

dmvolod: Yes, sure, @rigazilla 
I've also validated, it's true that NodePort is exposed on all nodes. But this will work for happy-path only. What happens, if we choose a `master` node host, make configuration and then that node will go down? Nothing changes in Pods and operator reconciliation happens, it's just a `master` , no Pods scheduling, etc. I think we need to choose a `worker` node with at least one scheduled Infinispan cluster Pod. 
P.S. It's just an idea, and this unhappy path need to be validated, but the precise choice get for us a more chances for success.

rigazilla: @dmvolod I can see the problem you want to address, but I dont think there's an easy solution atm: even if the reconciliation loop happens on the cluster where node goes down we need to fix the configuration on the other cluster.
also I'm not sure we can change the jgroups configuration at runtime.
I think that a proper solution to this problem would be to have more than 1 coordinator per cluster or maybe rely on Skupper recovery features. #385

rigazilla:
A possible solution that preferes worker nodes over master could be:
1. fix the selector (must work for k8s and OCP) to pickup one node from all the workers
2. if step 1 founds no nodes pickup one node from all the possible cluster nodes

pkg/controller/utils/infinispan/configuration/xsite.go
```diff
@@ -169,16 +170,23 @@ func getNodePortServiceHostPort(service *corev1.Service, k *util.Kubernetes, log
...
+	listOps := &client.ListOptions{
+		LabelSelector: labels.NewSelector().Add(*req),
+	}
+	err = k.Client.List(context.TODO(), workerList, listOps)
+
+	if err != nil || len(workerList.Items) == 0 {
+		// Fallback selecting everything
+		err := k.Client.List(context.TODO(), workerList, nil)
+		if err != nil || len(workerList.Items) == 0 {
+			return "", 0, err
```
dmvolod, line 187: This will return empty err if no nodes is found. Some error handling in this path doesn't make sense

dmvolod:
@rigazilla please pay attention looks like the ClusterRole no so correct, I've validated that enough records for listing nodes are
```
- apiGroups:
  - ""
  resources:
  - nodes
  verbs:
  - get
  - list
  - watch
```

rigazilla:
> @rigazilla please pay attention looks like the ClusterRole no so correct, I've validated that enough records for listing nodes are
> 
> ```
> - apiGroups:
>   - ""
>   resources:
>   - nodes
>   verbs:
>   - get
>   - list
>   - watch
> ```
> 
> Also I don't see a purpose to ask `customresourcedefinitions` on cluster level. Maybe just for OLM? But before xiste support the operator was working fine from the OperatorHub.

`customsourcedefinitions` is to enable serviceCA lookup for serving certificate service.

I would fix `nodes` rights on master if now are more than enough

deploy/clusterrole.yaml
```diff
@@ -7,10 +7,6 @@ rules:
   - ""
   resources:
   - nodes
-  - nodes/status
-  - nodes/metrics
-  - nodes/spec
-  - nodes/stats
   verbs:
   - get
   - list
```
dmvolod, line 12:
```suggestion
  - list
  - watch
```
Looks like deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml contains correct verbs, but clusterrole.yaml not

rigazilla: added `watch`... let's see if we are converging :)

rigazilla:
merged. thanks all!

PR #526, opened 2020-08-08
Set values on existing xsite

### LINKED ISSUE #525: Cross site details are blank
Despite valid cross site details and successfully finding local and remote site services, the cross site details are missing from the generated `infinispan.yaml` in the `<name>-configuration` ConfigMap. So although the `Infinispan` resource configured with `sites` does reconcile successfully, there is no cross site configuration.

donovanmuller:
Instead of creating a new XSite struct, set values on the existing reference.

Creating a new `XSite` struct and reassigning `xsite` to that https://github.com/infinispan/infinispan-operator/blob/21d35fe983b504649dd773c23e74af9abac37696/pkg/controller/utils/infinispan/configuration/xsite.go#L42-L46 was resulting in `xsite` in `infinispan_controller.go` https://github.com/infinispan/infinispan-operator/blob/21d35fe983b504649dd773c23e74af9abac37696/pkg/controller/infinispan/infinispan_controller.go#L179 never being assigned the values.

Resolves #525

### DIFF
--- a/pkg/controller/utils/infinispan/configuration/xsite.go
+++ b/pkg/controller/utils/infinispan/configuration/xsite.go
@@ -18,7 +18,7 @@ import (
 )
 
 // ComputeXSite compute the xsite struct for cross site function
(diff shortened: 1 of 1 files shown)


rigazilla:
good shot @donovanmuller !
merged, thanks!

Issue #491, opened 2020-07-27
Move some default values outside the operator code


dmvolod:
For flexible configuration it's would be nice to move some predefined values, which can be possible to change without operator code modification to the ConfigMap in yaml or any suitable format and be globally for whole operator.
The possible parameters are:
* Java memory parameters for server instance
* Defaults for users and caches
* Default server image name for whole cluster (instead of current env var location)
* Probably, delays interval

@rigazilla wdyt about it?

PR #532, opened 2020-08-11
Multiarch/531

### LINKED ISSUE #531: Operator should support multiarch images
The right operand image should be picked up by the operator basing on the build architecture.
See: https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/image_manager.go

rigazilla:
This allows image selection between the following GOARCH architectures:

-  ppc64, ppc64le, s390x, s390  --> will run the image named in RELATED_IMAGE_OPENJ9 env var
- all the remaining GOARCH --> runs RELATED_IMAGE_OPENJDK env var

Resolves #531 , I've applied a very simplified version of the code referred in the issue

### DIFF
--- a/pkg/controller/utils/common/image.go
+++ b/pkg/controller/utils/common/image.go
@@ -1,0 +1,33 @@
+package common
+
+import "runtime"
+
+const (
+	// OperandImageOpenJDK envvar name for OpenJDK operand image
+	OperandImageOpenJDK = "RELATED_IMAGE_OPENJDK"
+	// OperandImageOpenJ9 envvar name for OpenJ9 operand image
+	OperandImageOpenJ9 = "RELATED_IMAGE_OPENJ9"
+
+	// DefaultOperandImageOpenJDK default image for OpenJDK stack
+	DefaultOperandImageOpenJDK = "infinispan/server:latest"
+	// DefaultOperandImageOpenJ9 default image for OpenJ9 stack
+	DefaultOperandImageOpenJ9 = "infinispan-for-openj9/server:latest"
+)
+
+// GetDefaultInfinispanJavaImage returns default Infinispan Java image depends of the runtime architecture
+func GetDefaultInfinispanJavaImage() string {
+	// Full list of archs might be found here:
+	// https://github.com/golang/go/blob/release-branch.go1.10/src/go/build/syslist.go#L8
+	switch arch := runtime.GOARCH; arch {
+	case "ppc64", "ppc64le", "s390x", "s390":
+		return GetEnvWithDefault(OperandImageOpenJ9, DefaultOperandImageOpenJ9)
+	default:
+		// Try new name RELATED_IMAGE_OPENJDK first
+		retVal := GetEnvWithDefault(OperandImageOpenJDK, "")
+		if retVal != "" {
+			return retVal
+		}
+		// If RELATED_IMAGE_OPENJDK is not defined fallback on old name for backward comp
+		return GetEnvWithDefault("DEFAULT_IMAGE", DefaultOperandImageOpenJDK)
+	}
+}
--- a/pkg/controller/constants/constants.go
+++ b/pkg/controller/constants/constants.go
@@ -9,7 +9,7 @@ import (
 
 var (
 	// DefaultImageName is used if a specific image name is not provided
-	DefaultImageName = common.GetEnvWithDefault("DEFAULT_IMAGE", "infinispan/server:latest")
+	DefaultImageName = common.GetDefaultInfinispanJavaImage()
 
 	// DefaultMemorySize string with default size for memory
 	DefaultMemorySize = resource.MustParse("512Mi")
@@ -17,7 +17,7 @@ var (
 	// DefaultPVSize default size for persistent volume
 	DefaultPVSize = resource.MustParse("1Gi")
 
-	// DefaultCPUSize string with default size for CPU
+	// DefaultCPULimit string with default size for CPU
 	DefaultCPULimit int64 = 500
 
 	DeploymentAnnotations = map[string]string{


pkg/controller/utils/common/common.go
```diff
@@ -45,3 +46,37 @@ func LookupHost(host string, logger logr.Logger) (string, error) {
 	logger.Info("host resolved", "host", host, "addresses", addresses)
 	return host, nil
 }
+
+const (
+	// OperandImageOpenJDK envvar name for OpenJDK operand image
+	OperandImageOpenJDK = "RELATED_IMAGE_OPENJDK"
```
dmvolod, line 52: What is motivation for implementing a new env variable?
I'm prefer to move from env variables to the one configuration file in the ConfigMap as describer in #491

pkg/controller/utils/common/common.go
```diff
@@ -45,3 +46,37 @@ func LookupHost(host string, logger logr.Logger) (string, error) {
...
+	// OperandImageOpenJ9 envvar name for OpenJ9 operand image
+	OperandImageOpenJ9 = "RELATED_IMAGE_OPENJ9"
```
dmvolod, line 54: OpenJ9 and OpenJDK looks like can't coexists in one operator deployment. What is the purpose for this implementation? Looks like one env is enough for this option.
The init container and Infinispan container env vars are separate, but OpenJ9 and OpenJDK are not.

pkg/controller/utils/common/common.go
```diff
@@ -3,6 +3,7 @@ package common
 import (
```
dmvolod, line 3: It's a good point to use seaparate image.go file or/and separate package for this functions, as we will add more functions in #520

pkg/controller/utils/common/common.go
rigazilla: this because some build process needs those variables (RELATED_IMAGE_*) in the CSV by convention. i.e. they're used to enable installation in disconnected environments

pkg/controller/utils/common/common.go
rigazilla: see above, this is to be friendly with some build processes that need the list of all RELATED_IMAGES in the CSV

rigazilla:
@osnipso this PR should work with the following changes in the operator metadata CSV file.

```
@@ -207,8 +207,10 @@ spec:
                       fieldPath: metadata.name
                 - name: OPERATOR_NAME
                   value: infinispan-operator
-                - name: DEFAULT_IMAGE
-                  value: registry-proxy.engineering.redhat.com/rh-osbs/datagrid-datagrid-8.1:latest
+                - name: RELATED_IMAGE_OPENJDK
+                  value: my-openjdk-image-name:tag (or digest SHA256)
+                - name: RELATED_IMAGE_OPENJ9
+                  value: my-open-j9-image-name:tag (or digest SHA256)
                 image: registry-proxy.engineering.redhat.com/rh-osbs/datagrid-8-rhel8-operator:latest
                 name: infinispan-operator
                 ports:
```

pkg/controller/utils/common/common.go
dmvolod: Oh, thanks, this is more incompatible with DEFAULT_IMAGE, maybe we need to brake DEFAULT_IMAGE using in the master?

pkg/controller/utils/common/image.go
```diff
@@ -0,0 +1,37 @@
...
+	default:
+		// Try new name RELATED_IMAGE_OPENJDK first
+		retVal := GetEnvWithDefault(OperandImageOpenJDK, "")
+		if retVal != "" {
+			return retVal
+		}
+		// If RELATED_IMAGE_OPENJDK is not defined fallback on old name for backward comp
+		return GetEnvWithDefault("DEFAULT_IMAGE", DefaultOperandImageOpenJDK)
+	}
+	panic("Unknown architecture")
```
dmvolod, line 36: Not sure that is need to invoke

pkg/controller/utils/common/image.go
```diff
@@ -0,0 +1,37 @@
...
+	OperandImageOpenJ9 = "RELATED_IMAGE_OPENJ9"
+
+	// DefaultOperandImageOpenJDK default image for OpenJDK stack
+	DefaultOperandImageOpenJDK = "infinispan/server:latest"
+	// DefaultOperandImageOpenJ9 default image for OpenJ9 stack
+	DefaultOperandImageOpenJ9 = "infinispan-for-openj9/server:latest"
+)
+
+// GetDefaultInfinispanImageForCurrentAchitecture returns ...
+func GetDefaultInfinispanImageForCurrentArchitecture() string {
```
dmvolod, line 18:
```suggestion
func GetDefaultInfinispanJavaImage string {
```
and we will reserve `GetDefaultInfinispanNativeImage` for native in the future

rigazilla:
I've move the CSV update for OLM on the branch

pkg/controller/utils/common/image.go
```diff
@@ -0,0 +1,33 @@
...
+	// OperandImageOpenJ9 envvar name for OpenJ9 operand image
+	OperandImageOpenJ9 = "RELATED_IMAGE_OPENJ9"
+
+	// DefaultOperandImageOpenJDK default image for OpenJDK stack
+	DefaultOperandImageOpenJDK = "infinispan/server:latest"
+	// DefaultOperandImageOpenJ9 default image for OpenJ9 stack
+	DefaultOperandImageOpenJ9 = "infinispan-for-openj9/server:latest"
+)
+
+// GetDefaultInfinispanImageForCurrentAchitecture returns ...
```
dmvolod, line 17:
```suggestion
// GetDefaultInfinispanJavaImage returns default Infinispan Java image depends of the runtime architecture
```

dmvolod:
Thanks for review, @osnipso , I could merge, but looks like this is not finished by @rigazilla yet

dmvolod (approved):

PR #534, opened 2020-08-13
Update push-okd4 with cluster roles

dmvolod:
This helps to validate Roles and ClusterRoles on local install without remote docker registry
ClusterRoleBinding on-flight update support


dmvolod:
Merging this as it not related to main operator code and just for testing

PR #537, opened 2020-08-14
role.yaml rights assessment

### LINKED ISSUE #440: role.yaml rights assessment
Operator currently runs with more rights than necessary.
A fine tuning would be good to enhance security.

dmvolod:
Fixes #440

### DIFF
--- a/pkg/controller/infinispan/xsite.go
+++ b/pkg/controller/infinispan/xsite.go
@@ -48,7 +48,7 @@ func ComputeXSite(infinispan *ispnv1.Infinispan, kubernetes *kube.Kubernetes, se
 
 		remoteLocations := findRemoteLocations(xsite.Name, infinispan)
 		for _, remoteLocation := range remoteLocations {
-			err := AppendRemoteLocation(infinispan, &remoteLocation, kubernetes, logger, xsite)
+			err := appendRemoteLocation(infinispan, &remoteLocation, kubernetes, logger, xsite)
 			if err != nil {
 				return err
 			}
@@ -59,7 +59,7 @@ func ComputeXSite(infinispan *ispnv1.Infinispan, kubernetes *kube.Kubernetes, se
 	return nil
 }
 
-func ApplyLabelsToCoordinatorsPod(podList *corev1.PodList, infinispan *ispnv1.Infinispan, cluster ispn.ClusterInterface, client client.Client, logger logr.Logger) bool {
+func applyLabelsToCoordinatorsPod(podList *corev1.PodList, cluster ispn.ClusterInterface, client client.Client, logger logr.Logger) bool {
 	coordinatorFound := false
 	for _, item := range podList.Items {
 		cacheManagerInfo, err := cluster.GetCacheManagerInfo(consts.DefaultCacheManagerName, item.Name)
@@ -88,7 +88,7 @@ func ApplyLabelsToCoordinatorsPod(podList *corev1.PodList, infinispan *ispnv1.In
 	return coordinatorFound
 }
 
-func AppendRemoteLocation(infinispan *ispnv1.Infinispan, remoteLocation *ispnv1.InfinispanSiteLocationSpec, kubernetes *kube.Kubernetes,
+func appendRemoteLocation(infinispan *ispnv1.Infinispan, remoteLocation *ispnv1.InfinispanSiteLocationSpec, kubernetes *kube.Kubernetes,
 	logger logr.Logger, xsite *config.XSite) error {
 	restConfig, err := getRemoteSiteRESTConfig(infinispan, remoteLocation, kubernetes, logger)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


rigazilla:
We should have a way to run testsuite as an `infinispan-operator` account, to test this kind of PR. I'm looking into

dmvolod:
> We should have a way to run testsuite as an `infinispan-operator` account, to test this kind of PR. I'm looking into

Yes, it's possible but need some manual work. make push-okd4 deploys operator with account support and tests can be run with RUN_LOCAL=false
I did the same steps during tunung this rules

rigazilla:
I have errors with this test procedure:
```
oc new-project namespace-for-testing
make test RUN_LOCAL_OPERATOR=false
history 
```

dmvolod:
> history

Looking to it. Something wrong with registry detection.

dmvolod:
> I have errors with this test procedure:
> 
> ```
> make push-okd4
> oc new-project namespace-for-testing
> make test RUN_LOCAL_OPERATOR=false
> history 
> ```

Oh, looks like we need more documented on this procedure.
The namespace need to be created before, as push-okd4 creates an ImageStream there. Also following env variables need to be set 
export RUN_LOCAL_OPERATOR=false
export TESTING_NAMESPACE=
export PROJECT_NAME=
export OC_USER=
export EXPOSE_SERVICE_TYPE=LoadBalancer

rigazilla:
I've been able to run `make test` as an `infinispan-operator` account. 
Code needs some small changes for this, I'm wondering if we should put some effort in this and run the test as `infinispan-operator` also in CI.

rigazilla:
Also I'm surprised that `persistentvolumeclaims` access is not necessary... probably because is PV binding is driven by the pod machinery (?)

dmvolod:
> Also I'm surprised that `persistentvolumeclaims` access is not necessary... probably because is PV binding is driven by the pod machinery (?) 

We don't need it right now, but will need in the current PR's. I'm planning to add these rights

dmvolod:
@rigazilla what do you think to add some synthetic tests for cover all required Roles and ClusterRoles by ServiceAccount?

rigazilla:
> @rigazilla what do you think about to add some synthetic tests for cover all required Roles and ClusterRoles by ServiceAccount?

sounds good, I think they could save us from regression

ryanemerson (approved):

dmvolod:
> sounds good, I think they could save us from regression

I add call of the infinispan.GetNodePortServiceHostPort here, StorageClass lookup will be in the corresponding PR
Let me know, if you can see any other rights not covered by integration tests.
Hope, @ryanemerson will cover Pod's CRUD rights with the backup/restore tests.

pkg/controller/infinispan/xsite.go
```diff
@@ -178,7 +178,7 @@ func getNodePortServiceHostPort(service *corev1.Service, k *kube.Kubernetes, log
 
 	if err != nil || len(workerList.Items) == 0 {
 		// Fallback selecting everything
-		err = k.Client.List(context.TODO(), workerList, nil)
+		err = k.Client.List(context.TODO(), workerList, &client.ListOptions{})
```
dmvolod, line 181: @rigazilla please pay attention, that xsite regression have found and need to be backported to 2.0.x, as I'm not sure that we need to backport whole roles assessment to the 2.0.x

rigazilla: @dmvolod good spot! It's probably better to have a separate issue/bug/pr for this and a 2.0.x backport

dmvolod: > good spot! It's probably better to have a separate issue/bug/pr for this and a 2.0.x backport

Yes, fixed in #588

test/e2e/main_test.go
```diff
@@ -112,6 +115,12 @@ func TestNodeStartup(t *testing.T) {
 	waitForPodsOrFail(spec, 1)
 }
 
+// Run some functions for testing rights not covered by integration tests
```
rigazilla, line 118: Tests don't run in the `infinispan-operator` account, only the operator does.
We need another kubernetes object to verify roles directly from test.
I'm adding a commit with a fix

`testKube.Kubernetes` runs in the caller context (`kind-kind` in CI), to test the `infinispan-operator` role from we need another kubernetes object.
Let me add a commit

dmvolod:
Sorry, @rigazilla , I occasionally delete your commit with force psuh

rigazilla:
> Sorry, @rigazilla , I accidentally delete your commit with force push

no worries, committing again

test/e2e/main_test.go
dmvolod: Thanks @rigazilla . Looks very good.

rigazilla (approved):

rigazilla:
merged, thanks @dmvolod

PR #541, opened 2020-08-20
Specify StorageClassName for PVC

### LINKED ISSUE #388: Specify PVC StorageClassName
The cluster I am deploying to does not define a default storage class.  I would like to be able to specify which storage class to use as part of the CRD.

dmvolod:
Fixes #388
#550 need to be merged before
#537 need to be merged before and than add StorageClass query synthetic test

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -774,7 +774,7 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 
 		if m.IsDataGrid() {
 			var pvErr error
-			pvSize, pvErr = resource.ParseQuantity(*m.Spec.Service.Container.Storage)
+			pvSize, pvErr = resource.ParseQuantity(m.StorageSize())
 			if pvErr != nil {
 				return nil, pvErr
 			}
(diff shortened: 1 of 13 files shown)


ryanemerson (approved):

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -802,6 +802,13 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 		controllerutil.SetControllerReference(m, pvc, r.scheme)
 		pvc.OwnerReferences[0].BlockOwnerDeletion = pointer.BoolPtr(false)
 		dep.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{*pvc}
+		// Set a storage class if it specified
+		if storageClassName := m.StorageClassName(); storageClassName != "" {
+			if _, err = kube.FindStorageClass(storageClassName, r.client); err != nil {
+				return nil, err
+			}
+			pvc.Spec.StorageClassName = &storageClassName
+		}
```
rigazilla, line 811:
```suggestion
		// Set a storage class if it specified
		if storageClassName := m.StorageClassName(); storageClassName != "" {
			if _, err = kube.FindStorageClass(storageClassName, r.client); err != nil {
				return nil, err
			}
			pvc.Spec.StorageClassName = &storageClassName
		}
		dep.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{*pvc}
```
`StorageClassName` needs to be assigned before`pvc` is copied into `dep.Spec.VolumeClaimTemplates`

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -802,6 +802,13 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 		controllerutil.SetControllerReference(m, pvc, r.scheme)
 		pvc.OwnerReferences[0].BlockOwnerDeletion = pointer.BoolPtr(false)
 		dep.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{*pvc}
+		// Set a storage class if it specified
+		if storageClassName := m.StorageClassName(); storageClassName != "" {
+			if _, err = kube.FindStorageClass(storageClassName, r.client); err != nil {
```
rigazilla, line 807: Why we need our own error handling? Can't we just leave the pod `Pending`?

rigazilla (changes requested):

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, in this case it's better to use own handler rather to delegate to the system level. This will create a PVC with desired StorageClass and if it will never create, never finished, But in current implementation there is a chance to fix error without recreating from scratch.

rigazilla (approved):

dmvolod:
@ryanemerson , @rigazilla thanks for review

dmvolod:
Merged

PR #550, opened 2020-08-30
Add ephemeral storage support

### LINKED ISSUE #463: Is it possible to use the operator without using a PVC?
In the current CRD of Infinispan it is required to add the parameter for the container storage ([see CRD](https://github.com/infinispan/infinispan-operator/blob/master/deploy/crds/infinispan.org_infinispans_crd.yaml#L123)).
Is there a way to deploy it without the persistent storage if we don't need it for the distributed cache (e.g. by setting it to 0Gi)?
Or is this setting required nevertheless to store operator specific internal data?

dmvolod:
Fixes #463


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -760,66 +763,72 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
...
+	} else {
+		volumes := &dep.Spec.Template.Spec.Volumes
+		ephemeralVolume := corev1.Volume{
+			Name: m.Name,
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		}
+		*volumes = append(*volumes, ephemeralVolume)
+		reqLogger.Info("WARNING: An ephemeral storage is using. All data will be lost after restarting the cluster nodes.")
```
ryanemerson, line 828: " An ephemeral storage is using" doesn't really make sense. How about "WARNING: Ephemeral storage configured. All data will be lost on cluster shutdown and restart."

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -760,66 +763,72 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
...
-			Name:    "chmod-pv",
-			Command: []string{"sh", "-c", "chmod -R g+w /opt/infinispan/server/data"},
-			VolumeMounts: []corev1.VolumeMount{{
-				Name:      m.ObjectMeta.Name,
-				MountPath: "/opt/infinispan/server/data",
-			}},
-		}}
+		blockOwnerDeletion := false
+		controllerutil.SetControllerReference(m, pvc, r.scheme)
+		pvc.OwnerReferences[0].BlockOwnerDeletion = &blockOwnerDeletion
```
ryanemerson, line 803: No need for `blockOwnerDeletion` var, we can just set inline.

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -94,8 +94,9 @@ func (ispn *Infinispan) ApplyDefaults() {
 		if ispn.Spec.Service.Container == nil {
 			ispn.Spec.Service.Container = &InfinispanServiceContainerSpec{}
 		}
-		if ispn.Spec.Service.Container.Storage == "" {
-			ispn.Spec.Service.Container.Storage = consts.DefaultPVSize.String()
+		if ispn.Spec.Service.Container.Storage == nil {
+			def := consts.DefaultPVSize.String()
+			ispn.Spec.Service.Container.Storage = &def
```
ryanemerson, line 99: No need for the variable, we can just do this in a one-liner.

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -43,7 +43,8 @@ type EndpointEncryption struct {
 
 // InfinispanServiceContainerSpec resource requirements specific for service
 type InfinispanServiceContainerSpec struct {
-	Storage string `json:"storage"`
+	Storage          *string `json:"storage,optional,omitempty"`
+	EphemeralStorage *bool   `json:"ephemeralStorage,optional,omitempty"`
```
ryanemerson, line 47: As `Storage` is redundant if `EphemeralStorage` is set and because `Storage` is simply a string, we could allow users to just configure `Spec.Service.Containe.Storage=ephemeral`.

WDYT @rigazilla ?

dmvolod: I'm not sure that's good idea. I would like to have on additional Storage container with size, ephemeral and class fields. But this will require additional migration steps on upgrade.

ryanemerson: +1 to having a struct for configuring storage would be a lot more flexible, but like you say I'm not sure how we can handle this migration.

rigazilla: honestly I would go for reusing the `Storage` field. I know the semantic become a little confused this way, but do we have a use case that needs a more structured data model atm?

dmvolod: We would like to have an additional structure as next PR related to storage class will add next field.
Also Storage field is a little bit confusing as it's a size but implicitly defined.
We can reuse this field but it will brake convention.

dmvolod: Also processing single field update will be nightmare and can't be validated with OpenAPI schema. For example ephemeralStorage must be immutable, but stage size can be extended depends of the provided storage class.

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Thanks, fixed

pkg/apis/infinispan/v1/types_util.go
dmvolod: Fixed

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Thanks, fixed

pkg/apis/infinispan/v1/infinispan_types.go
rigazilla: ok @dmvolod, it seems reasonable, let's keep the `ephemeralStorage` field.

rigazilla (approved):

ryanemerson:
Thanks @dmvolod

Issue #549, opened 2020-08-27
Unable to delete Infinispan after cluster upgrade from 1.1.x to 2.0.x


dmvolod:
This is regression of the #413
During upgrade we need to do following steps:
- Remove the finalizer if it defined (we don't use them anymore)
- Update already created PVC's with the Infinispan CR owner reference to avoid orphaned PVC's (in the new versions the PVC's in the VolumeClaimTemplates use Infinispan CR as owner reference)

PR #552, opened 2020-08-30
Unable to delete Infinispan after cluster upgrade from 1.1.x to 2.0.x

dmvolod:
Implementing #549 for version 2.0.x

### DIFF
--- a/pkg/controller/infinispan/util/kubernetes.go
+++ b/pkg/controller/infinispan/util/kubernetes.go
@@ -256,9 +256,9 @@ func (k Kubernetes) HasServiceCAsCRDResource() bool {
 	return status >= 200 && status < 299
 }
 
-// GetK8sResources returns the list of the type list of resources associated to the cluster
-func (k Kubernetes) GetK8sResources(name, namespace string, list runtime.Object) error {
-	labelSelector := labels.SelectorFromSet(common.LabelsResource(name, ""))
+// ResourcesList returns a typed list of resource associated with the cluster
+func (k Kubernetes) ResourcesList(name, namespace, resourceType string, list runtime.Object) error {
+	labelSelector := labels.SelectorFromSet(common.LabelsResource(name, resourceType))
 	listOps := &client.ListOptions{Namespace: namespace, LabelSelector: labelSelector}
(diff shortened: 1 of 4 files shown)


dmvolod:
@rigazilla should we forwardport this PR to the master? 
It depends of the upgrade workflow definition for the 2.0.x and further versions. Which strategy you select for upgrades? 
https://docs.openshift.com/container-platform/4.4/operators/understanding_olm/olm-understanding-olm.html

pkg/controller/infinispan/util/kubernetes.go
```diff
@@ -256,9 +256,9 @@ func (k Kubernetes) HasServiceCAsCRDResource() bool {
 	return status >= 200 && status < 299
 }
 
-// GetK8sResources returns the list of the type list of resources associated to the cluster
-func (k Kubernetes) GetK8sResources(name, namespace string, list runtime.Object) error {
-	labelSelector := labels.SelectorFromSet(common.LabelsResource(name, ""))
+// ResourcesTypeList returns a typed list of resource associated with the cluster
+func (k Kubernetes) ResourcesTypeList(name, namespace, resourceType string, list runtime.Object) error {
```
rigazilla, line 260: maybe `ResourcesList`is enough?

rigazilla (commented):
> @rigazilla should we forwardport this PR to the master?
> It depends of the upgrade workflow definition for the 2.0.x and further versions. Which strategy you are selecting for upgrades?
> https://docs.openshift.com/container-platform/4.4/operators/understanding_olm/olm-understanding-olm.html

Channels are organized in z-stream: https://docs.openshift.com/container-platform/4.4/operators/understanding_olm/olm-understanding-olm.html#olm-upgrades-z-stream_olm-understanding-olm
anyway, if there are no drawbacks, I would say to port this PR also on master, so we don't need to set assumption in the code on how OLM channels looks like

dmvolod:
Thanks for review, @rigazilla 
Fixed and will port it to master

rigazilla (approved):

rigazilla:
merged, thanks @dmvolod !

PR #564, opened 2020-09-08
Port: Unable to delete Infinispan after cluster upgrade from 1.1.x to 2.0.x

### LINKED ISSUE #549: Unable to delete Infinispan after cluster upgrade from 1.1.x to 2.0.x

dmvolod:
Fixes #549
Port #552 to master


rigazilla (approved):

ryanemerson:
Thanks @dmvolod

Issue #566, opened 2020-09-09
Tests should evaluates for cluster is wellFormed


dmvolod:
Tests should validate that Infinispan.Status.Conditions.Type is wellFormed rather than looking for Pod's status
In some cases, Pod's can be running but some services, routes and cluster communications are not

This is actual for two and more nodes cluster, one node cluster is always well formed. But in most 
- TestContainerCPUUpdateWithTwoReplicas

But in some cases `WaitForInfinispanPods` is not enough for the condition to be satisfied in test. Also `wellFormed` state means that Infinispan CR will never update by operator itself in current test case.

dmvolod:
Fixed

Issue #560, opened 2020-09-04
Automates cluster upgrade test


dmvolod:
Automates cluster upgrade test not in CI, just for manual testing
The test must accept source and target version parameters, checkout source branch or tag, deploys operator version, create a cluster and than upgrade it to the target version.
#575 need to be resolved before
#576 need to be resolved before
#579 need to be resolved before

Issue #542, opened 2020-08-21
Infinispan CR doesn't display cluster status.conditions if can't access Pod's IP


dmvolod:
The 'status.conditions` is empty and doesn't display that Pod's are not accessed, in case of incorrect listening IP, etc.

Also `infinispan.Status.StatefulSetName` updates too late after creation. It means if Pod's IP is not accessed, this filed will never be updated even created. We need to set init reconcile state on start. 
https://github.com/infinispan/infinispan-operator/blob/4ffef65c94100090497ea90ab8911afba1051241/pkg/controller/infinispan/infinispan_controller.go#L428

Also `WellFormed` condition is still set to `True` when Pod's become unavailable. Need  to reset this condition ASAP.

PR #567, opened 2020-09-09
Create CurlClient abstraction to centralise curl http logic

ryanemerson:


test/e2e/main_test.go
```diff
@@ -389,40 +402,37 @@ func genericTestForGracefulShutdown(clusterName string, modifier func(*ispnv1.In
...
 	ispn := ispnv1.Infinispan{}
-	kubernetes.Kubernetes.Client.Get(context.TODO(), types.NamespacedName{Namespace: spec.Namespace, Name: spec.Name}, &ispn)
+	testKube.Kubernetes.Client.Get(context.TODO(), types.NamespacedName{Namespace: spec.Namespace, Name: spec.Name}, &ispn)
 	protocol := getSchemaForRest(&ispn)
-	pods := kubernetes.GetPods(spec.Name, tconst.Namespace)
-	podName := pods[0].Name
-
-	pass, err := cluster.Kubernetes.GetPassword(cconsts.DefaultOperatorUser, spec.GetSecretName(), tconst.Namespace)
-	tutils.ExpectNoError(err)
+	pods := testKube.GetPods("app=infinispan-pod", tconst.Namespace)
```
dmvolod, line 428:
```suggestion
	pods := testKube.GetPods(spec.Name, tconst.Namespace)
```

dmvolod (changes requested):

pkg/controller/infinispan/autoscale.go
```diff
@@ -80,15 +74,22 @@ func autoscalerLoop(clusterNsn types.NamespacedName, r *ReconcileInfinispan) {
...
+		if err != nil {
+			continue
+		}
+
+		cluster := ispnutil.NewCluster(user, pass, podList.Items[0].Namespace, ispn.GetEndpointScheme(), kubernetes)
 		for _, pItem := range podList.Items {
 			// time.Sleep(time.Duration(10000/len(podList.Items)) * time.Millisecond)
 			if metricMinPodNum == 0 {
-				metricMinPodNum, err = getMetricMinPodNum(consts.DefaultOperatorUser, pass, string(protocol), pItem)
+				metricMinPodNum, err = getMetricMinPodNum(pItem, cluster)
```
dmvolod, line 87: Looks like we can arrange with common pod exec functions just providing pItem,Name instead of transferring whole `corev1.Pod` record

pkg/controller/infinispan/autoscale.go
```diff
@@ -80,15 +74,22 @@ func autoscalerLoop(clusterNsn types.NamespacedName, r *ReconcileInfinispan) {
 		podList := &corev1.PodList{}
 		kubernetes.GetK8sResources(ispn.ObjectMeta.Name, ispn.ObjectMeta.Namespace, podList)
 
+		user := constants.DefaultOperatorUser
+		pass, err := kubernetes.GetPassword(user, ispn.Spec.Security.EndpointSecretName, ispn.Namespace)
+		if err != nil {
+			continue
+		}
+
+		cluster := ispnutil.NewCluster(user, pass, podList.Items[0].Namespace, ispn.GetEndpointScheme(), kubernetes)
```
dmvolod, line 83:
```suggestion
		cluster := ispnutil.NewCluster(user, pass, ispn.Namespace, ispn.GetEndpointScheme(), kubernetes)
```

pkg/controller/infinispan/autoscale.go
```diff
@@ -80,15 +74,22 @@ func autoscalerLoop(clusterNsn types.NamespacedName, r *ReconcileInfinispan) {
...
 			// time.Sleep(time.Duration(10000/len(podList.Items)) * time.Millisecond)
 			if metricMinPodNum == 0 {
-				metricMinPodNum, err = getMetricMinPodNum(consts.DefaultOperatorUser, pass, string(protocol), pItem)
+				metricMinPodNum, err = getMetricMinPodNum(pItem, cluster)
 				if err != nil {
 					log.Error(err, "Unable to get metricMinPodNum for pod", "podName", pItem.Name)
 				}
 			}
-			err = getMetricDataMemoryPercentUsage(&metricDataMemoryPercentUsed, consts.DefaultOperatorUser, pass, string(protocol), pItem)
+			err = getMetricDataMemoryPercentUsage(&metricDataMemoryPercentUsed, pItem, cluster)
```
dmvolod, line 92: The same as for the `getMetricMinPodNum`

ryanemerson:
Updated. Thanks for the review @dmvolod

dmvolod (approved):

rigazilla (commented):
Generally LGTM but I'm thinking that this PR add the assumption that all the pods must be started by this version or later, because of the POD_IP variable.
Wdyt?

dmvolod:
> Generally LGTM but I'm thinking that this PR add the assumption that all the pods must be started by this version or later, because of the POD_IP variable.
> Wdyt?

@rigazilla I'm not sure, that we will cause a problems, as all pod's and statefullset's are deleting during operator version upgrade until we are not implementing rolling update support on upgrade.

rigazilla:
> > Generally LGTM but I'm thinking that this PR add the assumption that all the pods must be started by this version or later, because of the POD_IP variable.
> > Wdyt?
> 
> @rigazilla I'm not sure, that we will cause a problems until we are not implementing rolling update support on upgrade, as all pod's and statefullset's are deleting during operator version upgrade.

mmm maybe the problem is bigger than expected: we need a check that this version can drive the old cluster shutdown during the upgrade

dmvolod:
> > > Generally LGTM but I'm thinking that this PR add the assumption that all the pods must be started by this version or later, because of the POD_IP variable.
> > > Wdyt?
> > 
> > @rigazilla I'm not sure, that we will cause a problems until we are not implementing rolling update support on upgrade, as all pod's and statefullset's are deleting during operator version upgrade.
> 
> mmm maybe the problem is bigger than expected: we need a check that this version can drive the old cluster shutdown during the upgrade

Yes, you are right
We can do check on operator start and patch statefulset, but this is little bit complex upgrade
That the reason why I'm creating issue for automate test upgrade. The manual test is slow and complex

dmvolod:
@ryanemerson ,@rigazilla can we send shutdown signal to cluster invoking just in any one node?

rigazilla:
> @ryanemerson ,@rigazilla can we send shutdown signal to cluster invoking just in any single node?

yep, it should work this way

dmvolod:
> > @ryanemerson ,@rigazilla can we send shutdown signal to cluster invoking just in any single node?
> 
> yep, it should work this way

Good, thanks. In this case we could modify just shutdown logic to interact via internal service to keep backward compatible.
@rigazilla wdyt about it?

ryanemerson:
Another option is I can initialise the POD_IP var if it's not already set on the host as part of the "exec" command.

dmvolod:
> Another option is I can initialise the POD_IP var if it's not already set on the host as part of the "exec" command.

We should avoid using direct IP access and use k8s entities as much as possible. If we need to access Pod IP address via service, `headless` service can be used

rigazilla:
> > > @ryanemerson ,@rigazilla can we send shutdown signal to cluster invoking just in any single node?
> > 
> > 
> > yep, it should work this way
> 
> Good, thanks. In this case we could modify just shutdown logic to interact via internal service to keep backward compatible.
> @rigazilla wdyt about it?

@dmvolod, sounds really good to me.

dmvolod (changes requested):
@ryanemerson do you have a plans to change shutdown logic in this PR or will submit a separate one?

ryanemerson:
Turns out that there was never a need for `POD_IP` anway. As we already have the headless service, it's possible to execute the following `curl http://<pod-name>:11222/rest/v2`

> @ryanemerson do you have a plans to change shutdown logic in this PR or will submit a separate one?

How does the shutdown logic currently work? If it just uses a rest call, there shouldn't be a need to change anything else with ^ change.

dmvolod:
> Turns out that there was never a need for `POD_IP` anway. As we already have the headless service, it's possible to execute the following `curl http://<pod-name>:11222/rest/v2`. So I have removed the env variable and updated the curl url.

Not certainly in that way, it can access pod 

> How does the shutdown logic currently work? If it just uses a rest call, there shouldn't be a need to change anything else with ^ change.

Yes, with new changes looks like no more changes needed, but right now we don't have an automatic tests for this. Need to be tested manually. I submitted an issue for automation this kind of tests #560

dmvolod:
@ryanemerson I run this pr locally and looks like hitting #542
The reason why tests passing successfully is described in #566 and they are limiting to Pod's IP interaction rather than validate Infinispan Status

dmvolod:
> > Turns out that there was never a need for `POD_IP` anway. As we already have the headless service, it's possible to execute the following `curl http://<pod-name>:11222/rest/v2`. So I have removed the env variable and updated the curl url.
> 
> Not certainly in that way, it can access pod
> 
> > How does the shutdown logic currently work? If it just uses a rest call, there shouldn't be a need to change anything else with ^ change.
> 
> Yes, with new changes looks like no more changes needed, but right now we don't have an automatic tests for this. Need to be tested manually. I submitted an issue for automation this kind of tests #560

Service is a load balancing across Pods, it's possible to use Pod's IP, but can't use custom selected Pod.

dmvolod:
> Turns out that there was never a need for `POD_IP` anway. As we already have the headless service, it's possible to execute the following `curl http://<pod-name>:11222/rest/v2`. So I have removed the env variable and updated the curl url.

Service is a load balancing across Pods, it's possible to use Pod's IP, but can't use custom selected Pod.

dmvolod:
> @ryanemerson I run this pr locally and looks like hitting #542
> The reason why tests passing successfully is described in #566 and they are limiting to Pod's IP interaction rather than validate Infinispan Status

Ops, now all is working fine. This is strange, will check again

dmvolod (approved):

rigazilla (approved):
lgtm

rigazilla:
merged, thanks @ryanemerson !

Issue #568, opened 2020-09-11
Package Restructure


ryanemerson:
The current codebase has evolved over time as we became more familiar with both `go` and the `operator-sdk`. Initially the code was written to service a single controller, however we will soon have multiple CRs and controllers e.g. Backup, Cache, Infinispan and Restore. In order to better facilitate code re-use and ensure loose-coupling of components, I think we need to revisit the structure of `pkg/` and come up with an agreed notion of best practices.

[This talk](https://www.youtube.com/watch?v=MzTcsI6tn-0) that @rigazilla recommended is the basis of this proposal.

As a starting point, I propose utilising the following structure:

```
pkg
├── apis
│   ├── addtoscheme_infinispan_v1.go
│   ├── addtoscheme_infinispan_v2alpha1.go
│   ├── apis.go
│   └── infinispan
│       ├── group.go
│       ├── v1
│       │   ├── doc.go
│       │   ├── infinispan_types.go
│       │   ├── register.go
│       │   ├── types_util.go
│       │   ├── zz_generated.deepcopy.go
│       │   └── zz_generated.openapi.go
│       └── v2alpha1
│           ├── cache_types.go
│           ├── doc.go
│           ├── register.go
│           ├── types_util.go
│           ├── zz_generated.deepcopy.go
│           └── zz_generated.openapi.go
├── controller [1]
│   ├── add_cache.go
│   ├── add_infinispan.go
│   ├── backup
│   ├── cache
│       ├── cache_controller.go
│       └── default.go
│   ├── controller.go
│   ├── infinispan [2]
│   │   ├── autoscale.go
│   │   ├── infinispan_controller_fake.go
│   │   ├── infinispan_controller.go
│   │   ├── infinispan_controller_test.go
│   │   └── xsite.go
│   └── restore
├── infinispan [3]
│   ├── caches.go # All logic related to manipulating caches
│   ├── cluster.go # All cluster related types, CacheManagerInfo etc
│   ├── constants.go # Resource Urls etc
│   ├── client
│   │   └── http
│   │       ├── client.go # Contains HttpClient and HttpConfig interface
│   │       └── curl
│   │           └── curl.go # implementation
│   └── configuration
│       ├── image.go # Previously controller/utils/infinispan/configuration/
│       ├── service_site.go
│       └── xsite.go
├── launcher
│   ├── launcher.go
│   └── main.go
├── kubernetes [4] 
    ├── client.go
    ├── labels.go
    └── pods.go
```

[...]

PR #570, opened 2020-09-14
Support for multiarch #531

rigazilla:
Port on master of #532

### DIFF
--- a/pkg/controller/constants/constants.go
+++ b/pkg/controller/constants/constants.go
@@ -10,7 +10,7 @@ import (
 
 var (
 	// DefaultImageName is used if a specific image name is not provided
-	DefaultImageName = GetEnvWithDefault("DEFAULT_IMAGE", "infinispan/server:latest")
+	DefaultImageName = GetDefaultInfinispanJavaImage()
 
 	// InitContainerImageName allows a custom initContainer image to be used
 	InitContainerImageName = GetEnvWithDefault("INITCONTAINER_IMAGE", "busybox")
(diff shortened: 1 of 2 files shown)


dmvolod (approved):

rigazilla:
Rebasing after #568.
I'm moving `image.go` file in the `pkg/controller/constants`, this to avoid depend cycle.

dmvolod (approved):

pkg/controller/constants/constants.go
```diff
@@ -10,7 +10,7 @@ import (
 
 var (
 	// DefaultImageName is used if a specific image name is not provided
-	DefaultImageName = GetEnvWithDefault("DEFAULT_IMAGE", "infinispan/server:latest")
+	DefaultImageName = GetDefaultInfinispanJavaImage()
 
 	// InitContainerImageName allows a custom initContainer image to be used
 	InitContainerImageName = GetEnvWithDefault("INITCONTAINER_IMAGE", "busybox")
```
dmvolod, line 16: Is default init container image for os390 has the same name as for x86?

rigazilla: @dmvolod good point! This needs a check

rigazilla: busybox supports only s390x and ppc64le, I've narrowed down the arch for openJ9

dmvolod: @rigazilla do we need to introduce variables for init container in `RELATED_IMAGE_` style or it has the same name?

rigazilla: It's a multiarch image, good for all the arch.
I would keep hardcoded for now (the real improvement to this would be to find a proper solution to the read only PV problem)

rigazilla:
merged, thanks!

PR #578, opened 2020-09-17
Fix run scripts and add TestOperatorUpgrade function for 1.1.x

### LINKED ISSUE #560: Automates cluster upgrade test

dmvolod:
Preparation for fix #560

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -24,60 +24,74 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-const TestTimeout = 5 * time.Minute
-const SinglePodTimeout = 5 * time.Minute
-const RouteTimeout = 240 * time.Second
-const DefaultPollPeriod = 1 * time.Second
+const (
+	TestTimeout          = 5 * time.Minute
+	SinglePodTimeout     = 5 * time.Minute
+	RouteTimeout         = 240 * time.Second
+	DefaultPollPeriod    = 1 * time.Second
 
-var CPU = getEnvWithDefault("INFINISPAN_CPU", "500m")
-var Memory = getEnvWithDefault("INFINISPAN_MEMORY", "512Mi")
-var Namespace = getEnvWithDefault("TESTING_NAMESPACE", "namespace-for-testing")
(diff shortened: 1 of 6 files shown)


ryanemerson:
@rigazilla Do we still want fixes for `1.1.x`? I assumed that it's not necessary now that `2.0.x` is what's being consumed downstream.

dmvolod:
> @rigazilla Do we still want fixes for `1.1.x`? I assumed that it's not necessary now that `2.0.x` is what's being consumed downstream.

It's just for upgrade testing purposes

ryanemerson:
Sure, but if we're no longer supporting this branch then there is no need for this fix or the follow up. I just wanted to clarify this so that you don't waste anytime creating unnecessary PRs

dmvolod:
> Sure, but if we're no longer supporting this branch then there is no need for this fix or the follow up. I just wanted to clarify this so that you don't waste any time creating unnecessary PRs

I hope it will last PR for the 1.1.x :smile: 
It supported and maintained yet, please see related article
https://access.redhat.com/articles/4933551
8.0 is 1.1.x branch
8.1 is 2.0.x branch
If someone customer or client happens migration problem from 1.1.x or 8.0 it would be nice to have common automation tool for upgrade task rather than waste time for manual testing for problem detection. Many of the issues come from the automated tests and can't handle manually.

test/e2e/main_test.go
```diff
@@ -24,60 +24,74 @@ import (
...
+	SinglePodTimeout     = 5 * time.Minute
+	RouteTimeout         = 240 * time.Second
+	DefaultPollPeriod    = 1 * time.Second
+
+	OperatorUpgradeStageNone = "NONE"
+	OperatorUpgradeStageFrom = "FROM"
+	OperatorUpgradeStageTo   = "TO"
+)
+
+var (
```
ryanemerson, line 38: This was the case before, but is there any reason why these are vars and not constants? I haven't checked everyone, but it seems in most cases the values are never updated after creation.

ryanemerson:
Thanks @dmvolod

Issue #581, opened 2020-09-21
If Spec.Image is defined by default it will never been updated on operator Upgrade


dmvolod:
When Spec.Image set by default with following code
https://github.com/infinispan/infinispan-operator/blob/23bba88eb4edbe25ed6a869552183d5c2b67f866/pkg/apis/infinispan/v1/types_util.go#L81-L85
In the future the Image name will be constant and never been updated (DEFAULT_IMAGE in the Pod will be new, but in the StatefulSet will keep the old)
https://github.com/infinispan/infinispan-operator/blob/23bba88eb4edbe25ed6a869552183d5c2b67f866/pkg/controller/infinispan/infinispan_controller.go#L662
https://github.com/infinispan/infinispan-operator/blob/23bba88eb4edbe25ed6a869552183d5c2b67f866/pkg/controller/infinispan/infinispan_controller.go#L704-L705

This will occurs while upgrade from 2.0.x to the latest master branch

ryanemerson:
@dmvolod I'm probably missing something, but as `Spec.Image` is always set on Infinispan CR creation by the operator, is it not possible for the new operator version to simply inspect existing CRs and see that the `Spec.Image` points to an older image? and if a new cluster rollout is required, the `Spec.Image` field can simply be updated by the new operator?

dmvolod:
> @dmvolod I'm probably missing something, but as `Spec.Image` is always set on Infinispan CR creation by the operator, is it not possible for the new operator version to simply inspect existing CRs and see that the `Spec.Image` points to an older image? and if a new cluster rollout is required, the `Spec.Image` field can simply be updated by the new operator?

Yes, @ryanemerson it can possible to update `Spec.Image` on each operator Upgrade but we can't separate this value defined by operator from the user provided. If we done this update the user defined Image will be lost. Maybe @rigazilla has another idea what we need to do in this case.

ryanemerson:
This raises a large issue for me, which is what level of support do we offer for custom images? Imo custom images should be used purely for debugging and this should be documented.

dmvolod:
> This raises a larger issue for me, which is what level of support do we offer for custom images? Imo custom images should be used purely for debugging and this should be documented.

I'm not sure that custom image is for debug purpose only. Customer can use it for running previous Infinispan version or native image choosing.

dmvolod:
Ok, good, agreed!
[...]

PR #582, opened 2020-09-21
Upgrade to Infinispan 12.x

### LINKED ISSUE #469: Move from 'latest' to the specific Infinispan server image tag name
As the operator starts to depend on specific Infinispan server features (logs changing via REST, host bindings, etc.) we should bind  an image name to the specific release, not infinispan/server:latest
Also, on dev stage, we should use latest dev Infinispan server rather than stable previous build

ryanemerson:
Fixes #469

### DIFF
--- a/pkg/infinispan/cluster.go
+++ b/pkg/infinispan/cluster.go
@@ -80,8 +80,7 @@ func (c Cluster) GetClusterSize(podName string) (int, error) {
 
 // GracefulShutdown performs clean cluster shutdown
 func (c Cluster) GracefulShutdown(podName string) error {
-	// TODO can this be HEAD?
-	_, err, reason := c.Client.Get(podName, consts.ServerHTTPClusterStop, nil)
+	_, err, reason := c.Client.Post(podName, consts.ServerHTTPClusterStop, "", nil)
 	if err != nil {
 		return fmt.Errorf("unexpected error during graceful shutdown, stderr: %v, err: %v", reason, err)
 	}
--- a/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
@@ -71,7 +71,7 @@ spec:
   replaces: infinispan-operator.v9.9.8
   relatedImages:
     - name: infinispan-server
-      image: infinispan/server:latest
+      image: infinispan/server:12.0
   labels:
     operated-by: infinispan-operator
   selector:
@@ -219,7 +219,7 @@ spec:
                       - name: OPERATOR_NAME
                         value: "infinispan-operator"
                       - name: DEFAULT_IMAGE
-                        value: "infinispan/server:latest"
+                        value: "infinispan/server:12.0"
   customresourcedefinitions:
     owned:
       - description: Cache is the Schema for the caches API
--- a/deploy/operator-install.yaml
+++ b/deploy/operator-install.yaml
@@ -604,4 +604,4 @@ spec:
             - name: OPERATOR_NAME
               value: "infinispan-operator"
             - name: DEFAULT_IMAGE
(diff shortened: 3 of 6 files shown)


rigazilla:
@ryanemerson it seems that shutdown request for 12 must be a POST:
```
curl -i --insecure --http1.1 -u operator:A99zampta7Z7S575 "https://test-check-data-survive-to-shutdown-0:11222/rest/v2/cluster?action=stop"
HTTP/1.1 405 Method Not Allowed
content-length: 11
```
```
curl -X POST -v -i --insecure --http1.1 -u operator:A99zampta7Z7S575 "https://test-check-data-survive-to-shutdown-0:11222/rest/v2/cluster?action=stop"
HTTP/1.1 204 No Content
```

ryanemerson:
@rigazilla Good spot! Thanks

pkg/infinispan/cluster.go
```diff
@@ -80,8 +80,7 @@ func (c Cluster) GetClusterSize(podName string) (int, error) {
 
 // GracefulShutdown performs clean cluster shutdown
 func (c Cluster) GracefulShutdown(podName string) error {
-	// TODO can this be HEAD?
-	_, err, reason := c.Client.Get(podName, consts.ServerHTTPClusterStop, nil)
+	_, err, reason := c.Client.Post(podName, consts.ServerHTTPClusterStop, "", nil)
```
dmvolod, line 83: @ryanemerson have you validated the upgrade path?
For 2.0.x -> master due to the #581 issue Infinispan image never been updated, but for 1.1.x it could handle an error, an old image may not support a new HTTP verb

ryanemerson: > @ryanemerson have you validated the upgrade path? The scripts are ready, you can try to use, but I'm waiting on

No. I have never looked at these tests tbh and don't know how they work. Do we have any docs/notes anywhere?

> For 2.0.x -> master due to the #581 issue Infinispan image never been updated, but for 1.1.x it could handle an error, an old image may not support the new HTTP verb

POST actions for graceful shutdown were added in Infinispan 11 [ISPN-11967](https://issues.redhat.com/browse/ISPN-11967) and the corresponding GET methods removed in 12. So it should be possible to upgrade from the `2.0.x` operator to `master`, e.g. Infinispan 11->12

dmvolod: Yes, @ryanemerson , we should be create a backward compatible operator code, as anyone can customize image or use another version rather that default
Just two ideas:
- Use CLI inside container instead of REST call for graceful shutdown only. This needs to be verified
- Invoke with POST verb and if this call returns 405 error code try to call with GET

ryanemerson: There has to be a limit on what Infinispan versions the operator can manage, we can't maintain backwards compatibility indefinitely as it's not flexible enough. @rigazilla What guarantees do we provide for upgrading? e.g. 10.1.x -> 12.x or just 11.x -> 12.x? We need to determine this, then we can implement a suitable solution.

dmvolod:
Also, as idea, we should create a script for updating Infinispan server version once branch switched to next release

pkg/infinispan/cluster.go
rigazilla: I think we can suppose upgrades to be incremental: i.e. 10.x -> 11.x -> 12.x.
Operator catalog works that way, so I don't think we need to support 10.x -> 12.x.

ryanemerson: This PR should be Ok to merge then

rigazilla (approved):

dmvolod (approved):

ryanemerson:
The `-Dinfinispan.cluster.stack=image-tcp` hack is no longer necessary with the `Dev04` image, so I just updated the commit to remove this. If the CI is green, I'll merge.

PR #585, opened 2020-09-22
Running operator as `infinispan-operator` service account instead of admin on CI

rigazilla:
Running operator as `infinispan-operator` service account instead of admin on CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,8 +46,12 @@ jobs:
         - export GO111MODULE=on
       script:
         - go mod vendor
-        - make test PARALLEL_COUNT=2
+        - export TESTING_CONTEXT=`kubectl --insecure-skip-tls-verify config current-context`
+        - ./build/prepare-sa-test.sh
+        - RUN_SA_OPERATOR=TRUE make test PARALLEL_COUNT=2
+        - kubectl config use-context $TESTING_CONTEXT
       after_failure:
+        - kubectl config use-context $TESTING_CONTEXT
         - kubectl get events --all-namespaces
         - kubectl cluster-info
         - df -h
--- a/build/prepare-sa-test.sh
+++ b/build/prepare-sa-test.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+TESTING_NAMESPACE=${TESTING_NAMESPACE-namespace-for-testing}
+
+kubectl delete namespace $TESTING_NAMESPACE --wait=true
+kubectl create namespace $TESTING_NAMESPACE
+kubectl apply -f deploy/service_account.yaml -n $TESTING_NAMESPACE
+
+KUBE_DEPLOY_SECRET_NAME=`kubectl -n $TESTING_NAMESPACE get sa infinispan-operator -o jsonpath='{.secrets[-1].name}'`
+KUBE_API_EP=`kubectl get ep kubernetes -n default -o jsonpath='{.subsets[0].addresses[0].ip}'`
+KUBE_API_TOKEN=`kubectl -n $TESTING_NAMESPACE get secret $KUBE_DEPLOY_SECRET_NAME -o jsonpath='{.data.token}'|base64 --decode`
+
+kubectl apply -f deploy/role.yaml -n $TESTING_NAMESPACE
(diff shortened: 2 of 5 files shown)


build/prepare-sa-test.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+kubectl delete namespace namespace-for-testing
```
dmvolod, line 3: Wait for delete need to be added as delete project/namespace is async procedure
Not problem for CI as we don't have a project in kind created from scratch or this need to be removed

dmvolod, line 3: Have you try to make namespace configurable?

.travis.yml
```diff
@@ -46,7 +46,8 @@ jobs:
         - export GO111MODULE=on
       script:
         - go mod vendor
-        - make test PARALLEL_COUNT=2
+        - ./build/prepare-sa-test.sh $KUBECONFIG  ./sakubeconfig
```
dmvolod, line 49: Looks like no parameters are using inside script

dmvolod:
Excellent script, @rigazilla 
If we will cover all paths with the tests, the validation of the Roles and ClusterRoles will be automatically handled by CI

build/prepare-sa-test.sh
rigazilla: Do you have any specific application in mind?
It could be an improvement anyway, let me see if there are impacts in the code.

rigazilla: done

.travis.yml
rigazilla: fixed

build/prepare-sa-test.sh
rigazilla: done

dmvolod (approved):

dmvolod:
@rigazilla have you plan to backport this to 2.0.x or it doesn't make sense?

dmvolod:
@rigazilla seems to CRD's are created twice

clusterrolebinding.rbac.authorization.k8s.io/infinispan-operator created
**customresourcedefinition.apiextensions.k8s.io/infinispans.infinispan.org created
customresourcedefinition.apiextensions.k8s.io/caches.infinispan.org created**
Cluster "k8s" set.
User "user-for-testing" set.
Context "k8s" created.
Switched to context "k8s".
The command "./build/prepare-sa-test.sh" exited with 0.
87.72s$ RUN_SA_OPERATOR=TRUE make test PARALLEL_COUNT=2
./build/build.sh linux
building infinispan-operator... version 3f1881e
build/run-tests.sh /home/travis/kind-kube-config.yaml
Using KUBECONFIG '/home/travis/kind-kube-config.yaml'
kubernetes.go: TEXTING_CONTEST: Create CRD infinispans.infinispan.org
**Wait for CRD to created
CRD established
Create CRD caches.infinispan.org
Wait for CRD to created
CRD established**
=== RUN   TestSimple

.travis.yml
```diff
@@ -46,8 +46,13 @@ jobs:
         - export GO111MODULE=on
       script:
         - go mod vendor
-        - make test PARALLEL_COUNT=2
+        - kubectl --insecure-skip-tls-verify config current-context
+        - export TESTING_CONTEXT=`kubectl --insecure-skip-tls-verify config current-context`
+        - ./build/prepare-sa-test.sh
+        - RUN_SA_OPERATOR=TRUE make test PARALLEL_COUNT=2
```
dmvolod, line 51: We can also save time a bit and call ./build/run-tests.sh directly as make test compile whole operator code which is not needed at this step. Compile was done on unit-test step already

rigazilla: I would keep the two stages independent, compile should work in incremental way and just do nothing on the second call

rigazilla:
ok should be ok now: operator works as `infinispan-operator` and tests run in the caller context

rigazilla:
> @rigazilla have you plan to backport this to 2.0.x or it doesn't make sense?

I can't see any drawbacks, I'll try a porting on 2.0.x

rigazilla:
merged, thanks for the review @dmvolod !

PR #586, opened 2020-09-23
If Spec.Image is defined by default it will never been updated on operator Upgrade

### LINKED ISSUE #581: If Spec.Image is defined by default it will never been updated on operator Upgrade

dmvolod:
Fixes #581


rigazilla (approved):
Great spot, @dmvolod.
I tried to find a way to save custom image, but I think we can't have anything better, #581 really mess up things and _real_ custom images are undetectable.

ryanemerson:
Thanks @dmvolod

PR #591, opened 2020-09-30
ISPN-12372 docs for storageclass

oraNod:
https://issues.redhat.com/browse/ISPN-12372

### DIFF
--- a/documentation/asciidoc/topics/ref_container_storage.adoc
+++ b/documentation/asciidoc/topics/ref_container_storage.adoc
@@ -1,8 +1,8 @@
 [id='storage_resources-{context}']
 = Storage Resources
 By default, {ispn_operator} allocates `1Gi` for storage for both
-{cacheservice} and {datagridservice} nodes. You can configure storage size for
-{datagridservice} nodes but not {cacheservice} nodes.
+{cacheservice} and {datagridservice} nodes. You can configure storage resources
+for {datagridservice} nodes but not {cacheservice} nodes.
 
 [source,options="nowrap",subs=attributes+]
 ----
@@ -10,6 +10,8 @@ include::yaml_annotated/cr_container_storage.yaml[]
 ----
 
 <1> Configures the storage size for {datagridservice} nodes.
+<2> Defines whether storage is ephemeral or permanent. Set the value to `true` to use ephemeral storage, which means all data in storage is deleted when clusters shut down or restart. The default value is `false`, which means storage is permanent.
+<3> Specifies the name of a StorageClass object to use for the persistent volume claim. If you include this field, you must specify an existing storage class as the value. If you do not include this field, the persistent volume claim uses the storage class that has the `storageclass.kubernetes.io/is-default-class` annotation set to `true`.
 
 .Persistent Volume Claims
 
--- a/documentation/asciidoc/topics/ref_datagridservice_resources.adoc
+++ b/documentation/asciidoc/topics/ref_datagridservice_resources.adoc
@@ -11,8 +11,10 @@ spec:
   replicas: 6 <2>
   service:
     type: DataGrid <3>
-    container:
-      storage: 2Gi <4>
+    container: <4>
+      storage: 2Gi
+      ephemeralStorage: false
+      storageClassName: my-storage-class
     sites: <5>
       local:
       name: NYC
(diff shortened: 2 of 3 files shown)


documentation/asciidoc/topics/ref_container_storage.adoc
```diff
@@ -1,15 +1,17 @@
...
+for {datagridservice} nodes but not {cacheservice} nodes.
 
 [source,options="nowrap",subs=attributes+]
 ----
 include::yaml_annotated/cr_container_storage.yaml[]
 ----
 
 <1> Configures the storage size for {datagridservice} nodes.
+<2> Defines whether storage is ephemeral or not. Set the value to `true` to use ephemeral storage, which means all data in storage is deleted when clusters shut down or restart. The default value is `false`.
+<3> Specifies the name of a StorageClass object to use for the persistent volume claim.
```
oraNod, line 14: @dmvolod Is there any default for the StorageClass? is it standard? what happens if the user omits this or leaves a null value?

documentation/asciidoc/topics/ref_container_storage.adoc
```diff
@@ -1,15 +1,17 @@
...
+{cacheservice} and {datagridservice} nodes. You can configure storage resources
+for {datagridservice} nodes but not {cacheservice} nodes.
 
 [source,options="nowrap",subs=attributes+]
 ----
 include::yaml_annotated/cr_container_storage.yaml[]
 ----
 
 <1> Configures the storage size for {datagridservice} nodes.
+<2> Defines whether storage is ephemeral or not. Set the value to `true` to use ephemeral storage, which means all data in storage is deleted when clusters shut down or restart. The default value is `false`.
```
oraNod, line 13: @dmvolod default is false, yeah? so in previous versions data in the PVC is permanent - hence "persistent volume claim"?

oraNod (commented):
@dmvolod In docs we also state "Persistent volume claims use the ReadWriteOnce (RWO) access mode." - Is this still true by default? If users specify a different StorageClass can they use a different access mode?

documentation/asciidoc/topics/ref_container_storage.adoc
dmvolod: Yes, true. Infinispan by default will always use a permanent PVC

documentation/asciidoc/topics/ref_container_storage.adoc
dmvolod: @oraNod the default Storage Class name depends of the platform where k8s/ocp/okd is deploying.
https://kubernetes.io/docs/concepts/storage/storage-classes
If user set empty or not exists Storage Class name operator will print an error that it's not exists.
If Storage Class name is not provided, the PVC will use storage class which has an annotation storageclass.kubernetes.io/is-default-class set to true.
https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/

documentation/asciidoc/topics/ref_datagridservice_resources.adoc
```diff
@@ -11,9 +11,14 @@ spec:
   replicas: 6 <2>
   service:
     type: DataGrid <3>
-    container:
-      storage: 2Gi <4>
-    sites: <5>
+    container: <4>
+      extraJvmOpts: "-XX:NativeMemoryTracking=summary"
```
dmvolod, line 15: @oraNod looks like extraJvmOpts, cpu and memory and storage parameters are from different container levels

dmvolod:
> In docs we also state "Persistent volume claims use the ReadWriteOnce (RWO) access mode." - Is this still true by default? If users specify a different StorageClass can they use a different access mode?

yes, true. We don't need another access mode rather than ReadWriteOnce (RWO) as it hard-coded in the operator.

documentation/asciidoc/topics/ref_datagridservice_resources.adoc
oraNod: @dmvolod thanks, for some reason I thought `spec.container` and `spec.service.container` were dupes. I guess that's what happens when you have too many meetings in one day. fixed.

dmvolod: No, but I'm also confusing `container's` sometimes. But not sure that's possible to change this structure dramatically...

dmvolod (approved):

ryanemerson (approved):

PR #593, opened 2020-10-01
Namespace should be created if not found in test (fix for 2.0.x)

dmvolod:

PR #594, opened 2020-10-01
Namespace should be created if not found in test

dmvolod:


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -799,7 +799,7 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 		},
 			Spec: corev1.PersistentVolumeClaimSpec{
 				AccessModes: []corev1.PersistentVolumeAccessMode{
-					"ReadWriteOnce",
+					corev1.ReadWriteOnce,
```
rigazilla, line 802: Is this intended?
Change is ok by me, better to align #593 also?

dmvolod: Yes, @rigazilla. I know that would be better to create a separate issue for memories and fix all non-related things, but didn't this yet.
Will do for #593 also

test/e2e/k8s/kubernetes.go
```diff
@@ -411,6 +410,7 @@ func (k TestKubernetes) DeleteSecret(secret *v1.Secret) {
 
 // RunOperator runs an operator on a Kubernetes cluster
 func (k TestKubernetes) RunOperator(namespace string) chan struct{} {
+	k.NewNamespace(namespace)
```
ryanemerson, line 413: Isn't this required due to the changes introduced with [tconst.RunSaOperator](https://github.com/infinispan/infinispan-operator/blob/master/test/e2e/main_test.go#L89-L94) ?

If we're going to create the namespace here, we should remove the call from ^ if statement.

dmvolod: Yeah, @ryanemerson . The problem looks like not with NewNamespace invoke, but seams to be with RunSaOperator defaults

test/e2e/constants/constants.go
```diff
@@ -26,7 +26,7 @@ var (
 	Memory            = constants.GetEnvWithDefault("INFINISPAN_MEMORY", "512Mi")
 	Namespace         = strings.ToLower(constants.GetEnvWithDefault("TESTING_NAMESPACE", "namespace-for-testing"))
 	RunLocalOperator  = strings.ToUpper(constants.GetEnvWithDefault("RUN_LOCAL_OPERATOR", "true"))
-	RunSaOperator     = strings.ToUpper(constants.GetEnvWithDefault("RUN_SA_OPERATOR", "true"))
+	RunSaOperator     = strings.ToUpper(constants.GetEnvWithDefault("RUN_SA_OPERATOR", "false"))
```
dmvolod, line 29: @rigazilla please pay attention, that problem is that RunSaOperator was always set to true, instead of false

rigazilla: uh yes @dmvolod , wrong default. my bad

rigazilla (approved):

dmvolod:
Thanks for review, @ryanemerson and @rigazilla 
Merged

PR #602, opened 2020-10-08
Integration test updates for 2.0.x

Crumby:
* Finally brings tests updates to 1.1.x branch to 2.0.x
* Updates tests for changes introduced in 2.0.x release
* Adds several tests for new features: replicationFactor and autoscaling


test-integration/test-server/src/main/java/org/infinispan/operator/remote/auth/hotrod/HotRodServlet.java
```diff
@@ -29,14 +29,18 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
 
          ConfigurationBuilder builder = new ConfigurationBuilder();
          builder.addServer().host(serviceName).port(11222);
-         builder.security().authentication().realm("default").serverName("infinispan").username(username).password(password).enable();
+
+         if(username != null || password != null) {
```
dmvolod, line 33: Is it really should by OR, not AND?

test-integration/operator-tests/src/test/java/org/infinispan/operator/AdvancedSetupAIT.java
```diff
@@ -0,0 +1,194 @@
...
+ *     categories:
+ *       x.y.z: trace
+ *       t.u.v: debug
+ *   security:
+ *     endpointEncryption:
+ *        type: secret
+ *        certSecretName: encryption-secret
+ *     endpointSecretName: connect-secret
+ *   service:
+ *     type: datagrid
```
dmvolod, line 55: Not related to code, just description side
```suggestion
 *     type: DataGrid
```

test-integration/operator-tests/src/test/java/org/infinispan/operator/MinimalSetupIT.java
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * Should result in the same as:
+ *
+ * spec:
+ *   replicas: 2
+ *   container:
+ *     cpu: 500m
+ *     memory: 512Mi
+ *   service:
+ *     type: cache
```
dmvolod, line 50: Not related to code, just description side
```suggestion
 *     type: Cache
```

test-integration/test-server/src/main/java/org/infinispan/operator/remote/auth/hotrod/HotRodServlet.java
Crumby: It doesn't really matter in this particular code as the endpoint is triggered from the tests with either both setup or none (so the result will be the same). If we wanted to test what happens if you setup only one of them then OR would be valid as the other would be set to `null`. But that's not in the scope of the Operator tests.

test-integration/operator-tests/src/test/java/org/infinispan/operator/MinimalSetupIT.java
```diff
@@ -0,0 +1,193 @@
...
+   }
+
+   /**
+    * Ensure that all resource created by Operator in this tests are deleted.
+    */
+   @AfterAll
+   static void undeploy() throws IOException {
+      infinispan.delete();
+
+      BooleanSupplier bs = () -> openShift.apps().statefulSets().withName(appName).get() == null &&
```
dmvolod, line 95: This looks like very common for each cluster deletion and can be placed in separate class which will be extended by all tests.

Crumby: Right. I wouldn't say inheritance is the proper solution in here as the expected deletions differ slightly from deployment to deployment and I tried to avoid test inheritance to keep the code as clear as possible. But moving it into common utility class and composing it into `@AfterAll` call would definitely help.

dmvolod (approved):

rigazilla:
merged, thanks @Crumby !

PR #608, opened 2020-10-13
ISPN-12422 missing attributes for service monitor

oraNod:
https://issues.redhat.com/browse/ISPN-12422 

Needs backport to 2.0.x


documentation/asciidoc/topics/proc_setting_up_prometheus.adoc
```diff
@@ -29,9 +29,11 @@ include::yaml_annotated/service_monitor.yaml[]
 +
```
dmvolod, line 29: @oraNod please pay attention, that option "Install the Prometheus Operator" is actual on k8s platform only.
OCP has installed Prometheus in openshift-monitoring namespace and using a separate Prometheus Operator installation in not supported. In the case of OCP the following article can be used for custom service monitoring
https://access.redhat.com/documentation/en-us/openshift_container_platform/4.5/html/monitoring/monitoring-your-own-services#enabling-monitoring-of-your-own-services_monitoring-your-own-services

rigazilla (approved):
Apart from the Prometheus stuff, PR lgtm

dmvolod:
Can't validate this solution: both versions don't work on my OCP, OLM installs broken Prometheus operator, techPreviewUserWorkload fails on upgrade.
Will think that it's correct CR for monitoring

dmvolod (approved):

documentation/asciidoc/topics/proc_setting_up_prometheus.adoc
oraNod: thanks @dmvolod I've reworded from "Install the Prometheus Operator" to "Set up a Prometheus stack on your k8s/OpenShift cluster." unfortunately it's beyond the scope of our docs to describe prometheus install via OLM or any other method. I have linked to the OCP docs that you've pointed out too, even though that solution is tech preview.

oraNod:
> Can't validate this solution: both versions don't work on my OCP, OLM installs broken Prometheus operator, techPreviewUserWorkload fails on upgrade.
> Will think that it's correct CR for monitoring

I've been able to validate the service monitor CR and have checked it against a CR that a RH developer created too. for our docs I think we just need to ensure that the service monitor CR is correct.

PR #610, opened 2020-10-13
Remove TestSimple test

Crumby:
Removes TestSimple:
* Test isn't related to the Operator and doesn't provide any added value.
* If run as non-admin user the test fail as regular user doesn't have access to cluster nodes resource

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -102,12 +102,6 @@ func TestMain(m *testing.M) {
 	}
 }
 
-// Simple smoke test to check if the Kubernetes/OpenShift is alive
-func TestSimple(t *testing.T) {
-	fmt.Printf("%v\n", testKube.Nodes())
-	fmt.Printf("%s\n", testKube.Kubernetes.PublicIP())
-}
-
 // Test operator and cluster version upgrade flow
 func TestOperatorUpgrade(t *testing.T) {
 	spec := DefaultSpec.DeepCopy()


Crumby:
@dmvolod Verification that testing environment is up and ready should be before the tests are executed. If that isn't true then the tests will fail no matter whether this one is there or not. Either during main or right away with `TestNodeStartup`

dmvolod:
Yes, sure
Let's move this calls to the TestMain in case of the RUN_SA_OPERATOR env variable is set to true

Crumby:
Do you want to keep `fmt.Printf("%v\n", testKube.Nodes())`? Test themselves probably shouldn't be executed as admin either but rather as a user with extended rights to manipulate Infinispan CRs... (That's another issue though)

dmvolod:
> Test themselves probably shouldn't be executed as admin either but rather as a user with extended rights to manipulate Infinispan CRs... (That's another issue though)

Yes, correct. Right now operator runs not under admin and with serviceAccount implemented by @rigazilla in #585
It would be nice to use serviceAccount for other test operations but it will be separate one, not infinispan-operator I think.

dmvolod:
> fmt.Printf("%v\n", testKube.Nodes())

I think it could be useful for CI only, but I don't know why it was added to tests. Maybe @rigazilla can help with this code.

rigazilla:
@dmvolod , @Crumby  I think we used it in the early development stages to check if the OKD cluster was running correctly.
I can't see any drawback in removing it now.

dmvolod:
> @dmvolod , @Crumby I think we used it in the early development stages to check if the OKD cluster was running correctly.
> I can't see any drawback in removing it now.

Good to know this. Let's remove this test

dmvolod (approved):

rigazilla:
@Crumby  please rebase

Crumby:
@rigazilla Done

rigazilla:
merged, thank you!

PR #613, opened 2020-10-14
Fix typos in descriptions

osnipso:
They don't affect the functioning of the image, but are an annoyance and need to be fixed anyway.
This probably should be cherry-picked into all active branches: master, 1.1.x, 2.0.x.1, etc.


ryanemerson (approved):
Thanks @osnipso, can you create a PR for master as well.

PR #614, opened 2020-10-15
Drop Image field from test deployments

Crumby:
Operator already has server image mapping defined and that image should be used for test deployments by default. Current settings where each deployments has `Image` field set up leads or could lead to potential issues:
* Default image of the Operator forgets to be updated  and tests won't run with the server that will be used to create clusters by default
* Server image mapping breaks in the code but tests won't recognize this as they set the Image
* Alternative Operator that uses alternative server runs with Infinispan server instead

Possible test extensions:
* Verification that cluster is deployed with expected server (basically done by `TestOperatorUpgrade`)
* Verification that `Image` field is really used when (Unit tests seems to be ideal in this case)

### DIFF
--- a/pkg/controller/constants/image.go
+++ b/pkg/controller/constants/image.go
@@ -9,9 +9,9 @@ const (
 	OperandImageOpenJ9 = "RELATED_IMAGE_OPENJ9"
 
 	// DefaultOperandImageOpenJDK default image for OpenJDK stack
-	DefaultOperandImageOpenJDK = "infinispan/server:latest"
+	DefaultOperandImageOpenJDK = "infinispan/server:12.0"
 	// DefaultOperandImageOpenJ9 default image for OpenJ9 stack
-	DefaultOperandImageOpenJ9 = "infinispan-for-openj9/server:latest"
+	DefaultOperandImageOpenJ9 = "infinispan-for-openj9/server:12.0"
 )
 
 // GetDefaultInfinispanJavaImage returns default Infinispan Java image depends of the runtime architecture
--- a/test/e2e/constants/constants.go
+++ b/test/e2e/constants/constants.go
@@ -35,7 +35,7 @@ var (
 	RunSaOperator        = strings.ToUpper(constants.GetEnvWithDefault("RUN_SA_OPERATOR", "false"))
 	OperatorUpgradeStage = strings.ToUpper(constants.GetEnvWithDefault("OPERATOR_UPGRADE_STAGE", OperatorUpgradeStageNone))
 	CleanupInfinispan    = strings.ToUpper(constants.GetEnvWithDefault("CLEANUP_INFINISPAN_ON_FINISH", "true"))
-	ImageName            = constants.GetEnvWithDefault("IMAGE", "infinispan/server:12.0")
+	ExpectedImage	     = constants.GetEnvWithDefault("EXPECTED_IMAGE", "infinispan/server:12.0")
 	ExposeServiceType    = constants.GetEnvWithDefault("EXPOSE_SERVICE_TYPE", "NodePort")
 
 	OperatorUpgradeStateFlow = []string{"upgrade", "stopping", "wellFormed"}
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -31,7 +31,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
-	"k8s.io/utils/pointer"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
 
@@ -53,7 +52,6 @@ var DefaultSpec = ispnv1.Infinispan{
 			CPU:    tconst.CPU,
 			Memory: tconst.Memory,
 		},
-		Image:    pointer.StringPtr(tconst.ImageName),
 		Replicas: 1,
 		Expose:   exposeServiceSpec(),
 	},
@@ -65,7 +63,6 @@ var MinimalSpec = ispnv1.Infinispan{
 		Name: tconst.DefaultClusterName,
 	},
 	Spec: ispnv1.InfinispanSpec{
-		Image:    pointer.StringPtr(tconst.ImageName),
 		Replicas: 2,
 	},
 }
@@ -125,8 +122,8 @@ func TestOperatorUpgrade(t *testing.T) {
 		err := testKube.Kubernetes.ResourcesList(tconst.Namespace, ispnctrl.PodLabels(spec.Name), pods)
 		tutils.ExpectNoError(err)
 		for _, pod := range pods.Items {
-			if pod.Spec.Containers[0].Image != tconst.ImageName {
-				tutils.ExpectNoError(fmt.Errorf("upgraded image [%v] in Pod not equal desired cluster image [%v]", pod.Spec.Containers[0].Image, tconst.ImageName))
+			if pod.Spec.Containers[0].Image != tconst.ExpectedImage {
+				tutils.ExpectNoError(fmt.Errorf("upgraded image [%v] in Pod not equal desired cluster image [%v]", pod.Spec.Containers[0].Image, tconst.ExpectedImage))
 			}
 		}
 	default:
@@ -540,7 +537,6 @@ func TestExternalService(t *testing.T) {
 				CPU:    tconst.CPU,
 				Memory: tconst.Memory,
 			},
-			Image:    pointer.StringPtr(tconst.ImageName),
 			Replicas: 1,
 			Expose:   exposeServiceSpec(),
 		},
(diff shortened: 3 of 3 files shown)


dmvolod:
@Crumby in this case we should change default operator image for the master to the `infinispan/server:12.0` as with your changes tests will run against `infinispan/server:latest` but targeted to version 12.x
https://github.com/infinispan/infinispan-operator/blob/af542c4e456b42a962b1a07fe62f8559c23ec784/pkg/controller/constants/image.go#L12

Crumby:
@dmvolod Right. I'd go with former option as it's the only piece of code that refers `latest` instead of `12.0`. Does the same applies for OpenJ9 images? Or that image is just a placeholder? It doesn't seem to exist.

https://github.com/infinispan/infinispan-operator/blob/af542c4e456b42a962b1a07fe62f8559c23ec784/pkg/controller/constants/image.go#L14

@rigazilla Any thoughts?

rigazilla:
> @dmvolod Right. I'd go with former option as it's the only piece of code that refers `latest` instead of `12.0`. 

Shouldn't be better to set DEFAULT_IMAGE? This should be the most common use case since Catalog/Hub is becoming the primary installation mode and CSV is supposed to set that variable.


Does the same applies for OpenJ9 images? Or that image is just a placeholder? It doesn't seem to exist.
> 
> https://github.com/infinispan/infinispan-operator/blob/af542c4e456b42a962b1a07fe62f8559c23ec784/pkg/controller/constants/image.go#L14
> 
> @rigazilla Any thoughts?

It's just a placehoder for now, but it should be treated the same

Crumby:
@rigazilla I believe you are right. I haven't realized that if you start Operator locally the env isn't set up. Anyway, I tried to add:

`_ = os.Setenv("DEFAULT_IMAGE", tconst.ImageName)`

line directly under

https://github.com/infinispan/infinispan-operator/blob/af542c4e456b42a962b1a07fe62f8559c23ec784/test/e2e/k8s/kubernetes.go#L452

and the code doesn't catch up and still defaults to `latest`. The issue is that `DefaultImageName` in `Constants.go` is resolved on startup before the tests are executed. I'd like to avoid setting up the env before test execution as that would transfer the rqquirement to set the correct image to a user in case of local operator run.

dmvolod:
> I'd like to avoid setting up the env before test execution as that would transfer the rqquirement to set the correct image to a user in case of local operator run.

Yes, @Crumby good point. Let's remove `DefaultImageName` constant and will use `GetDefaultInfinispanJavaImage()` directly. It doesn't make sense for operator, but avoid set `DEFAULT_IMAGE` env variable in test script.

pkg/controller/constants/image.go
```diff
@@ -14,20 +14,31 @@ const (
...
+
 	// Full list of archs might be found here:
 	// https://github.com/golang/go/blob/release-branch.go1.10/src/go/build/syslist.go#L8
 	switch arch := runtime.GOARCH; arch {
 	case "ppc64le", "s390x":
-		return GetEnvWithDefault(OperandImageOpenJ9, DefaultOperandImageOpenJ9)
+		defaultImage = GetEnvWithDefault(OperandImageOpenJ9, DefaultOperandImageOpenJ9)
 	default:
 		// Try new name RELATED_IMAGE_OPENJDK first
 		retVal := GetEnvWithDefault(OperandImageOpenJDK, "")
```
dmvolod, line 36: We can use following code here and don't use if-then section later
```suggestion
		retVal := GetEnvWithDefault(OperandImageOpenJDK, GetEnvWithDefault("DEFAULT_IMAGE", DefaultOperandImageOpenJDK))
```

Crumby:
@dmvolod @rigazilla One thing I'm scratching my head with is whether `GetDefaultInfinispanJavaImage()` shouldn't fail hard if it doesn't find any image. Since having `DEFAUL_...` or `RELATED_...` image env is mandatory. Currently we won't recognize if the code haven't broke since it'll always default to latest.

Other thing that came to my mind is whether it wouldn't make more sense to separate Operator startup and test execution (managed through script). Test wouldn't have to no longer differentiate whether they need to start the operator or not and `DefaultImage` could stay as a constant (which probably makes a bit more sense but isn't achievable atm with Operator starting up after initialization). On the other hand I understand this could complicate upstream test execution a little (or not if the script would be well written).

dmvolod:
@Crumby probably I'm something missing, but don't see a situation when `DEFAUL_...` or `RELATED_...` image are not defined except `make run-local` which is suitable for development only. For tests you are defined `DEFAULT_IMAGE` env variable, but already installed operator via OLM or operator-install.yaml bundle will have these variables always.
As alternative way, we can create a script on release upstream to change image version in all possible places, including constants, operator deploy.yaml, tests, etc.

Crumby:
> but don't see a situation when DEFAUL_... or RELATED_... image are not defined

Exactly, `GetDefaultInfinispanJavaImage()` failing hard should be safeguard against regression in code. `infinispan/server:latest` shouldn't be defined as a default as the code shouldn't even get there.

> except make run-local

It shouldn't be hard to have a default env setup in the Makefile for this use-case.

Nevertheless this is probably out of scope of this PR

dmvolod:
> Exactly, `GetDefaultInfinispanJavaImage()` failing hard should be safeguard against regression in code. `infinispan/server:latest` shouldn't be defined as a default as the code shouldn't even get there.

Yes, fully agree.

> Nevertheless this is probably out of scope of this PR

Yes, and drop default image setup in code in case of `DEFAULT_IMAGE` env is not present also. We need to arrange this changes together to avoid brake of local running. Also need to validate places where are local operator running

rigazilla:
> @dmvolod @rigazilla One thing I'm scratching my head with is whether `GetDefaultInfinispanJavaImage()` shouldn't fail hard if it doesn't find any image. Since having `DEFAUL_...` or `RELATED_...` image env is mandatory. Currently we won't recognize if the code haven't broke since it'll always default to latest.

I agree to make DEFAULT_IMAGE mandatory defined before the operator

> Other thing that came to my mind is whether it wouldn't make more sense to separate Operator startup and test execution (managed through script). Test wouldn't have to no longer differentiate whether they need to start the operator or not and `DefaultImage` could stay as a constant (which probably makes a bit more sense but isn't achievable atm with Operator starting up after initialization). On the other hand I understand this could complicate upstream test execution a little (or not if the script would be well written).

I also agree in separate the env setup (operator startup) from the test execution, but I would keep the current startup code (maybe disabled by default) in the unit test, for lazy developers that wants to run easily the testsuite in the IDE :)

Crumby:
@dmvolod @rigazilla So is this PR good as it is atm? I'd like to make similar changes for 2.0.x branch if possible

pkg/controller/constants/image.go
```diff
@@ -14,20 +14,26 @@ const (
...
 	default:
-		// Try new name RELATED_IMAGE_OPENJDK first
-		retVal := GetEnvWithDefault(OperandImageOpenJDK, "")
-		if retVal != "" {
-			return retVal
-		}
-		// If RELATED_IMAGE_OPENJDK is not defined fallback on old name for backward comp
-		return GetEnvWithDefault("DEFAULT_IMAGE", DefaultOperandImageOpenJDK)
+		// Try new name RELATED_IMAGE_OPENJDK first, if not defined fallback on old name for backward comp
+		defaultImage = GetEnvWithDefault(OperandImageOpenJDK, GetEnvWithDefault("DEFAULT_IMAGE", DefaultOperandImageOpenJDK))
```
dmvolod, line 36: @Crumby it's doesn't make sense for production deployment, but we could arrange DefaultOperandImageOpenJDK with the current Infinispan working image, but not latest, i.e set it to the "infinispan/server:12.0".

If we decide to make `DEFAULT_IMAGE` or `RELATED_IMAGE_OPENJDK` madatory, we can call `panic` if they are not present but `GetDefaultInfinispanJavaImage()` need to be validated  on operator startup, not lazy.

@rigazilla wdyt?
Otherwise looks good.

rigazilla: > @Crumby it's doesn't make sense for production deployment, but we could arrange DefaultOperandImageOpenJDK with the current Infinispan working image, but not latest, i.e set it to the "infinispan/server:12.0".
> 

Yep, it should be 12 for master and 11 for 2.0.x

> If we decide to make `DEFAULT_IMAGE` or `RELATED_IMAGE_OPENJDK` madatory, we can call `panic` if they are not present but `GetDefaultInfinispanJavaImage()` need to be validated on operator startup, not lazy.
> 
> @rigazilla wdyt?
> Otherwise looks good.

@Crumby seems you saved the _mandatory_ solution for another issue, right?
lgtm for fixing the initial issue, though removing all the DefaultOperand* in the code and fail fast should be a safer solution imo

Crumby: It looks like we walked full circle at this point. So my suggestion is:

1) Drop the image fields and change the default from `latest` to `12` in this PR. This will cause that tests executed with Data Grid Operator will not default to Infinispan but to `DEFAULT_IMAGE` of the Operator. Tests running the operator locally will still default to Infinispan 12. Should be non-invasive.
2) Create an issue and another PR for `GetDefaultInfinispanJavaImage()` to panic if `DEFAULT_IMAGE` isn't defined as safeguard preventing regression. This change isn't trivial because it'll mean that in order to run the tests from IDE the variable `DEFAULT_IMAGE` will have to be defined before the tests are executed.

Does it make sense?

rigazilla (approved):

dmvolod (approved):

rigazilla:
merged, thanks @Crumby

Issue #597, opened 2020-10-05
Resources should utilise recommended Kubernetes labels


ryanemerson:
There are recommended labels that should be applied to resources the operator creates. We should add these to the various statefulsets, services etc that the operator creates:

[k8s Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

edeandrea:
In addition, we should allow the CRs to allow users to specify labels/annotations that would carry forward to resources that the operator creates, like these

https://docs.openshift.com/container-platform/4.5/applications/application_life_cycle_management/odc-viewing-application-composition-using-topology-view.html#odc-labels-and-annotations-used-for-topology-view_viewing-application-composition-using-topology-view

PR #596, opened 2020-10-02
ISPN-12222 Backup/Restore CRs

ryanemerson:
https://issues.redhat.com/browse/ISPN-12222

@rigazilla @dmvolod I'm issuing this as a draft for now so that you have an idea of what I have been working on. The following are currently still required:

- [x] Set zero-capacity=true on restore pod config. BackRestoreTest fails when true atm.
- [x] Infinispan Dev05 image for required server changes
- [x] Test for backup/restore CacheService
~~- [ ] Test for backup to existing VolumeSource~~
- [ ] Test for custom zero pod resources
- [ ] Docs

I noticed that because the Backup and Restore Spec use the k8 VolumeSource struct directly the CRD yaml automatically pulls in all of the VolumeSource descriptions. Can we prevent this? Or is it even an issue as it's generated?

PR #615, opened 2020-10-16
Add Direct XSite Host/Port Configuration Option

### LINKED ISSUE #604: Operator Requires Kubernetes API Access to Remote Sites
The current `Infinispan` CRD configuration for x-site replication requires remote backup sites to be specified with their kubernetes API details. My understanding is that the operator uses this to actually query the remote kubernetes API and check for the respective external service, and then lookup the host and port.

Is it possible that a configuration option could be added that would support specifying the remote backup sites `host`/`port` details directly, instead of requiring access to the kubernetes API?
Not only does this access require populating credential secrets for each remote backup site in every other kubernetes cluster, but it also requires network access to the kubernetes API from the other clusters.
In the infrastructure environment I'm working in, these network connections are currently blocked.
Allowing and authorizing these connections would have significant configuration and security downsides; so I'd rather not go down that road if at all possible.

Is there anything I'm missing with respect to the requirement for cross cluster kubernetes API access?

timothymathison:
Fixes #604 
Adds the option to directly specify `host` and `port` for each `spec.service.sites.locations` entry in the `CRD`, and forgo the `url`/`secretName` options which required access to the remote's kubernetes API.

Also partially addresses #597 and supports configuration of annotations for external and xsite `Service`s.

**Checklist:**
- [x] Modify `CRD` and xsite logic for explicit `host`/`port`
- [x] ~Implement remote site `post`/`port` validation,~ depending on discussion
- [x] Add `external` and `site` service annotation configurations.
- [x] Expose xsite port `7900` used for JBoss relay.
- [x] Update documentation


deploy/crds/infinispan.org_caches_crd.yaml
```diff
@@ -2,8 +2,6 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: caches.infinispan.org
-  labels:
```
dmvolod, line 5: @timothymathison please run `make codegen` for full code generation

build/crds-gen.sh
```diff
@@ -20,7 +20,7 @@ installOperatorSDK() {
     rm -rf "${OPERATOR_SDK_BUNDLE}"
     exit 1
   fi
-  chmod +x operator-sdk-"${OPERATOR_SDK_VERSION}"-x86_64-linux-gnu && sudo mkdir -p /usr/local/bin/ && sudo cp "${OPERATOR_SDK_BUNDLE}" /usr/local/bin/operator-sdk && rm "${OPERATOR_SDK_BUNDLE}"
```
dmvolod, line 23: Could you please add `arch` command call instead of hardcoded `x86_64` architecture to conform all possible Operator SDK OS installations

deploy/crds/infinispan.org_caches_crd.yaml
timothymathison: > @timothymathison please run `make codegen` for full code generation

Oddly, running `make crdsgen` (which is the first part of `make codegen`) is what removed this label. Is that desired?

dmvolod: codegen will add required labels

pkg/controller/infinispan/xsite.go
```diff
@@ -48,7 +48,16 @@ func ComputeXSite(infinispan *ispnv1.Infinispan, kubernetes *kube.Kubernetes, se
 
 		remoteLocations := findRemoteLocations(xsite.Name, infinispan)
 		for _, remoteLocation := range remoteLocations {
-			err := appendRemoteLocation(infinispan, &remoteLocation, kubernetes, logger, xsite)
+			var err error
+			if remoteLocation.Host != "" {
```
dmvolod, line 52: @timothymathison what do you think about remote host validation. Once added, no chance to replace it without whole Infinispan CR remove/add operation. Incorrect host will display inside Infinispan server logs, not o the operator side. Not sure that's possible to validate whole host:port combination but need to be validated.

timothymathison: Yeah I'm not sure what the validation would look like exactly. We could simply validate that the host can be resolved ([as is currently done](https://github.com/infinispan/infinispan-operator/blob/af542c4e456b42a962b1a07fe62f8559c23ec784/pkg/controller/infinispan/xsite.go#L217)) or that `host`/`port` can be connected to at the socket layer.
However, with multiple sites this may create a bit of a race condition: all the sites would need to be created at exactly the same time so that the site services all exist by the time each operator reaches this logic. At least that's my understanding?

timothymathison: >Once added, no chance to replace it without whole Infinispan CR remove/add operation.

Does infinispan not respect later updates to the `infinispan.yaml` configuration file?

dmvolod: > Does infinispan not respect later updates to the `infinispan.yaml` configuration file?

Looks like yes, but need to be validated if it will reflect on the cluster or not.

dmvolod: > Yeah I'm not sure what the validation would look like exactly. We could simply validate that the host can be resolved ([as is currently done](https://github.com/infinispan/infinispan-operator/blob/af542c4e456b42a962b1a07fe62f8559c23ec784/pkg/controller/infinispan/xsite.go#L217)) or that `host`/`port` can be connected to at the socket layer.
> However, with multiple sites this may create a bit of a race condition: all the sites would need to be created at exactly the same time so that the site services all exist by the time each operator reaches this logic. At least that's my understanding?

Yes, sure. But right now there is condition, that all remote server locations must be active. Probably, there's no need to add validation, but we will know about fail server connection too late from the server log only.
@pruivo what do you think about this?

deploy/crds/infinispan.org_caches_crd.yaml
timothymathison: @dmvolod, after running `make codegen` the `label` is still removed.

pkg/controller/infinispan/service_site.go
```diff
@@ -49,16 +49,21 @@ func SiteService(siteServiceName string, infinispan *ispnv1.Infinispan, scheme *
...
-			Namespace: infinispan.ObjectMeta.Namespace,
-			Annotations: map[string]string{
-				"service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "tcp",
-			},
-			Labels: lsService,
+	objectMeta := metav1.ObjectMeta{
+		Name:      siteServiceName,
+		Namespace: infinispan.ObjectMeta.Namespace,
+		Annotations: map[string]string{
+			"service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "tcp",
```
timothymathison, line 56: This was hard-coded from before; I don't really know what it was needed for?
Does it need to be hard-coded anymore if `annotations` can be specified from the `CRD`?

deploy/crds/infinispan.org_caches_crd.yaml
dmvolod: I'm not sure, maybe sed imp in crd-gen.sh works diffidence for Linux and Mac OS?

build/kubernetes/run-local.sh
```diff
@@ -0,0 +1,11 @@
...
+
+NAMESPACE=${1-infinispan}
+
+echo "Using kubeconfig: ${KUBECONFIG}"
+echo "Running for cluster: $(kubectl config current-context)"
+
+kubectl apply -f deploy/crds/infinispan.org_infinispans_crd.yaml
+kubectl apply -f deploy/crds/infinispan.org_caches_crd.yaml
+kubectl wait --for condition=established crd infinispans.infinispan.org --timeout=60s
+WATCH_NAMESPACE="${NAMESPACE}" ./build/_output/bin/infinispan-operator -kubeconfig "${KUBECONFIG}"
```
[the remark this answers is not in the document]
dmvolod: I'm not using minikube/k8s, but seems to be no specific for run/run-local between plain k8s and minikube.
It would be nice to split and don't duplicate k8s/minikube commons, like run, run-local and clean that's not specific for minikube and minukube start/stop/delete scipts.
@rigazilla wdyt about it?

pkg/controller/infinispan/service_site.go
dmvolod: It specific to run xsite mode on AWS, not sure that we should move this annotation to the customer site as it's now possible as we should minimize the work which end user need to be done manually.
We can easily add this predefined annotation to the customer provided.

pkg/controller/infinispan/xsite.go
```diff
@@ -136,6 +145,18 @@ func appendKubernetesRemoteLocation(infinispan *ispnv1.Infinispan, remoteLocatio
 		"host", host,
 		"port", port,
 	)
+	remoteLocation.Host = host
+	remoteLocation.Port = port
+	return appendBackupSite(logger, remoteLocation, xsite)
+}
+
+func appendBackupSite(logger logr.Logger, remoteLocation *ispnv1.InfinispanSiteLocationSpec, xsite *config.XSite) error {
```
dmvolod, line 153: logger logr.Logger is redundant here and not using
```suggestion
func appendBackupSite(remoteLocation *ispnv1.InfinispanSiteLocationSpec, xsite *config.XSite) error {
```

pkg/controller/infinispan/xsite.go
pruivo: > But right now there is condition, that all remote server locations must be active.

That's a limitation in JGroups since it requires the IP/port of the remote site before starting. In the current code, the operator provides that info and it needs the other site to be up to extract the service IP/port. 

Since we are "moving" the logic from the operator to the user, IMO it is fine not to wait for the other site to be up. Of course, the user needs to check if the xsite bridge is up or not and, if the user makes a mistake, it has to change the config and restart the cluster.

build/kubernetes/run-local.sh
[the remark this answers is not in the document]
dmvolod: Also would be fine to have common scripts for both ocp/k8s as difference between `oc` and `kubectl` utility is very limited. Just for our scripts it's maybe `oc new-project` that's equivalent `kubectl create ns` nothing more. Otherwise all commands are the same...

pkg/controller/infinispan/xsite.go
dmvolod: Please also pay attention, that according to docs
```
To perform cross-site replication, Infinispan Operator requires Infinispan clusters to have the same name and run in matching namespaces.
```
https://infinispan.org/infinispan-operator/master/operator.html#backup_sites
which will not validated at the customer host/port provided case. The initial JGroups connection looks like will work, but in case of fails will not work. This need to be validated.

pruivo: The way I see is, if hardcoded host/port, you are giving up on the operator to manage xsite. 
All the docs related to xsite are invalidated in that case. You don't need to create permissions, accounts, share secrets, use the same namespace... actually, you don't need an openshift cluster in the other site. You can do xsite between opensift cluster and bare metal.

pkg/controller/infinispan/xsite.go
timothymathison: > logger logr.Logger is redundant here and not using

Ah forgot to remove that 👍 I was considering adding a log statement in this function

pkg/controller/infinispan/service_site.go
timothymathison: > It specific to run xsite mode on AWS, not sure that we should move this annotation to the customer site as it's now possible as we should minimize the work which end user need to be done manually.
> We can easily add this predefined annotation to the customer provided.

Okay, i'll leave it here. If the user needs to override `annotations` then they will need to include this one as well.

build/kubernetes/run-local.sh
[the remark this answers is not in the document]
timothymathison: Are you suggesting that `oc` commands not be used for `run` or `run-local`, and `kubectl` be used instead in a common script?
There is also not an equivalent command for `oc login` that I know of, but the docs could instruct the user to run that separately and then populate the `KUBECONFIG` environment variable.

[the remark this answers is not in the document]
timothymathison: I think it would definitely be doable to extract the unique commands to the docs, and put everything else inside common `run` and `run-local` scripts which use `kubectl`.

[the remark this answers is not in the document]
dmvolod: First, let's remove all kubernetes/minikube things from this PR and arrange this in separate one. We can create a new issue for this. I can do this, but can't test as using OCP in AWS only and don't have alternative one.
@timothymathison you are welcome to do this.

Second, it would be nice to separate minikube things from plain k8s. For example, keep just minikube related scripts in minikube folder and make targets named as minikube just for them. Place kubernetes related scripts such as test, run-local, etc. in kubernetes folder and change make targets for kubernetes only not related to minikube.
As maximum target it would be nice to have common ocp/k8s scripts for run-local, run, etc. which will run with just parameter depends of the make target which will define where is this script running, on ocp or k8s independently deployment platform for k8s, such as minikube, kind or plain k8s on baremetal.
Hope, I explain correctly, sorry for possible grammatical mistakes, I'm not native :smile:

deploy/crds/infinispan.org_caches_crd.yaml
dmvolod: @timothymathison could you please validate, it seems to this code doesn't work correctly on MacOS
```
sed -i -e "/name: infinispans.infinispan.org/a \  labels:\n    name: infinispan-operator" deploy/crds/infinispan.org_infinispans_crd.yaml
```

timothymathison: @dmvolod Yes it was an issue with mac os `sed`. I fixed it by using `gsed` installed with `brew install gnu-sed`

dmvolod: Have you try with this solution without additional software install
https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

timothymathison: @dmvolod the problem I was encountering was actually described here:
https://stackoverflow.com/questions/40843994/extra-characters-after-at-the-end-of-a-command

I wasn't able to find a nice solution without adding literal newlines. I can use that solution and test it in linux if you want me to.

dmvolod: Yes, good point. I can test it on Linux, let me know if it needed.

timothymathison: @dmvolod I attempted the fix described in that stack overflow. But that breaks it for gnu `sed`. I don't think there is a simple solution to make the same command work with both versions of `sed`.
I've left it the way I had it (where it checks if `gsed` is installed); unless we want to use something besides `sed`.

timothymathison:
I've updated the documentation. Is there anything I've missed or additional **relevant** changes that need to be made?

deploy/crds/infinispan.org_caches_crd.yaml
dmvolod: Good, let's keep this solution. Thanks

dmvolod:
@timothymathison , looks like PR is ready for review.

rigazilla (changes requested):
Looks really good to me, @timothymathison .
I've added two comments on docs.

documentation/asciidoc/topics/ref_xsite_crd.adoc
```diff
@@ -40,6 +40,8 @@ ifdef::productized[]
 <5> Specifies a backup location that matches `.spec.service.sites.local.name`.
 <6> Specifies the URL of the {openshiftshort} API for the backup location.
 <7> Specifies the secret that contains the service account token for the backup site.
-<8> Logs error messages for the JGroups TCP protocol.
-<9> Logs fatal messages for the JGroups RELAY2 protocol.
+<8> Directly specifies the `host` routing to another {brandname} backup instance.
```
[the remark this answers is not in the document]
timothymathison: > can we show that <6><7> and <8><9> aren't in sequence but two different options?

Sure! I'm not very familiar with ascidoc, so formatting suggestions are welcome.

documentation/asciidoc/topics/yaml_annotated/cr_backup_site.yaml
```diff
@@ -12,9 +12,9 @@ spec:
         url: openshift://api.site-a.devcluster.openshift.com:6443 <6>
         secretName: lon-token <7>
       - name: NYC
-        url: openshift://api.site-b.devcluster.openshift.com:6443
-        secretName: nyc-token
+        host: infinispan-nyc.myhost.com <8>
```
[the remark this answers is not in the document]
timothymathison: We could have an entirely separate `yaml` example. However, I thought this might help to show that these options can actually be mixed and matched for different `sites`?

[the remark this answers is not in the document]
oraNod: We were chatting about keeping this content as a separate procedure to avoid confusion downstream. This asciidoc is the same source for the Data Grid Operator so we want to be sure the use cases are clear. What do you think about copying proc_configuring_sites.adoc to something like proc_manually_configuring_sites.adoc and adding all the details there?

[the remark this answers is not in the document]
timothymathison: Ah, I was not aware of the DataGrid Operator. I can do that if you want me to. Would both of those docs still be rendered from `assembly_xsite_replication` (just with different headings)?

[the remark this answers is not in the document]
oraNod: No worries. To be honest you probably shouldn't need to worry about downstream. The main thing here is that these changes introduce a new use case that we should make clear in the docs. That said I don't want to create too much overhead for contributors when it comes to the docs.

I've added these changes to our release plan. If you like, we can keep docs in this PR as-is and I can add a separate docs PR to break things out and apply style edits, etc. That will probably make your life a bit easier (and also mean docs don't hold up this PR too much). 

If you would like to break things out into a new procedure, that's cool too. And, yes, in that case you'd want to include the procedure that you create in `assembly_xsite_replication`.

documentation/asciidoc/topics/ref_xsite_crd.adoc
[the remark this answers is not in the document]
rigazilla: I would use <6a><7a> and <6b><7b> if asciidoc doesn't complain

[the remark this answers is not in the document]
oraNod: @rigazilla unfortunately that' wouldn't work. asciidoc would treat <6a> as text rather than a callout. 

The more I think about it, the more I feel like we need two new user stories, which means two new assemblies. One user story for the k8s api approach, which means the operator looks up the connections and controls them. Another user story for the direct xsite connection. That way we can clearly show they are two distinct options and explain the pros and cons.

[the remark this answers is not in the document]
ryanemerson: I agree. A separate user story for direct xsite would allow us to emphasize that this approach can be useful for communicating with non-k8 clusters as well.

documentation/asciidoc/topics/yaml_annotated/cr_backup_site.yaml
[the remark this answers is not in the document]
timothymathison: Either way is fine with me. If someone else that better understands the documentation structure can do it, that's preferable to me. However, I would be willing to make at least some of the changes, if that would lead to this getting released sooner?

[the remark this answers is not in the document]
oraNod: OK. I can handle all the heavy lifting for the documentation on this one. Like I say, I don't want to hold things up too much or create a lot of overhead with docs. We can make things really easy and just merge docs as-is with this PR if you think all the relevant details are included. I can work with that and break content into separate user stories as we've discussed.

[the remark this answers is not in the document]
oraNod: I hope I'm not adding too much churn to your PR but what do you think about reverting doc changes and adding simple procedure in asciidoc or markdown or whatever you're comfortable with? 

Similar to what Ryan added in his backup/restore PR: https://github.com/infinispan/infinispan-operator/pull/596/files#diff-0dc4c153215f1f66772c25de52306c668896b46d0ef7097fb447643687dba2b1 

After your PR gets merged I can work with the source file to create a user story. If this sounds good then I think we can take the same approach for future contributions. Thanks for helping us figure this out!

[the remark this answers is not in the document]
timothymathison: Okay, I'll write an `md` file,  since that's what I'm familiar with, under the `docs/` directory.

build/crds-gen.sh
```diff
@@ -43,9 +45,17 @@ fi
 
 echo "Generating CRDs for API's..."
 
+allsed() {
```
dmvolod, line 48: @timothymathison you can probably invoke gsed on MacOS OS type and fails in not found, as if `gsed` not installed and `sed` installed script will work incorrectly.

timothymathison: @dmvolod, let me know what you think of it now.

dmvolod: We should validate `gsed` presence before operator-sdk generate to avoid unpredictable changes

dmvolod: @timothymathison looks good, thanks

documentation/asciidoc/topics/yaml_annotated/cr_backup_site.yaml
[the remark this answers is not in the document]
timothymathison: @oraNod, I've added a `docs/xsite.md` file with examples for both procedures.
I'm thinking I should also go ahead and update `design.adoc` to reflect the new `CRD` definitions.

timothymathison:
I'm not sure why, but CI even though it was previously passing (with all the current code changes included) is now failing.
It seems to be an issue with the graceful shutdown timing out. I've actually noticed that graceful shutdown never completes when actually running the operator. But I'm not sure how that could be related to my changes.

dmvolod:
> I'm not sure why, but even though it was previously passing, (with all the current code changes included) CI is now failing.
> It seems to be an issue with the graceful shutdown timing out. I've actually noticed that graceful shutdown never completes when actually running the operator. But I'm not sure how that could be related to my changes.

It's something going wrong with native image test due new image release. I will check.

dmvolod:
@timothymathison please rebase on top the master, test has bee fixed

timothymathison:
I've rebased twice to address recent conflicts.
@oraNod, @rigazilla, can I get another review and/or approval?

oraNod (approved):
Thanks @timothymathison! doc/xsite.md looks good to me. I can work with that as a source to create a user story for the main doc set. Really appreciate your flexibility and patience with adding docs.

doc/xsite.md
```diff
@@ -0,0 +1,104 @@
+# Cross-Site Replication/Backup Configurations
+
+There are two configuration approaches available for setting up cross-site replication between multiple **Infinispan** clusters.
+
+1. Auto configuration managed by the **Infinispan Operator** via Kubernetes API communication between clusters.
+1. Manual connection configurations provided via the `Cluster` `CRD`.
```
rigazilla, line 6:
```suggestion
1. Manual connection configurations provided via the `Infinispan CR`.
```

rigazilla (approved):
Just a minor note, otherwise looks great to me.

dmvolod:
@timothymathison looks good, thanks, I will try to test this configuration option later
please also squash commits into the single

dmvolod (approved):

pruivo (approved):

timothymathison:
I don't have permission to merge this, can someone else?

dmvolod:
@timothymathison we are ready to merge, please squash commits before we

timothymathison:
> @timothymathison we are ready to merge, please squash commits before

@dmvolod, can you use thee _squash merge_ option?

dmvolod:
> @dmvolod, can you use the _squash merge_ option?

It could be possible, but not enabled inside this repo, don't know why. Probably @rigazilla knows. I can do this manually but in separate PR.

rigazilla:
> > @dmvolod, can you use the _squash merge_ option?
> 
> It could be possible, but not enabled inside this repo, don't know why. Probably @rigazilla knows. I can do this manually but in separate PR.

No idea, maybe because the author is not preserved? I think we can enable it anyway

rigazilla:
squash merge enabled

dmvolod:
> squash merge enabled

Thanks, Vittorio!

timothymathison:
Are we still waiting on something here?

@dmvolod if you think it would really be better, I can go ahead and squash all my commits? Then you can merge it?

dmvolod:
Yes, @timothymathison I can merge when will get access my laptop, or somebody from community can help with merge. @rigazilla , @ryanemerson cloud you please to help squash and merge this PR.

dmvolod:
Merged
Thanks, @timothymathison for implementation
Thanks, @rigazilla , @pruivo and @oraNod for review

timothymathison:
Thanks @dmvolod!

How soon  can I expect a new release that includes these changes?

dmvolod:
@timothymathison I don't have this info, the operator master builds on every commit and you can use latest master release https://quay.io/repository/infinispan/operator?tab=tags
@rigazilla what is the next upstream release based on top of master?

rigazilla:
> @timothymathison I don't have this info, the operator master builds on every commit and you can use latest master release https://quay.io/repository/infinispan/operator?tab=tags
> @rigazilla what is the next upstream release based on top of master?

Yeah it's probably time to have a 2.1.0 for Infinispan 12, I'll prepare the Hub stuff in the next few days

timothymathison:
@rigazilla, any update on publishing this change to operatorhub?

rigazilla:
I start work on it:
https://github.com/infinispan/infinispan-operator/issues/657

PR #623, opened 2020-10-27
[2.0.x] JDG-3643 Missing name for default services port causes prometheus to …

ryanemerson:
https://issues.redhat.com/browse/JDG-3643


rigazilla:
merged

PR #625, opened 2020-10-28
Able to use server-native image (master)

### LINKED ISSUE #473: Able to use server-native image

dmvolod:
Fixes #473

### DIFF
--- a/pkg/apis/infinispan/v1/types_util.go
+++ b/pkg/apis/infinispan/v1/types_util.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/infinispan/infinispan-operator/pkg/controller/constants"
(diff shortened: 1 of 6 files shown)


pkg/apis/infinispan/v1/types_util.go
```diff
@@ -13,6 +14,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+type ImageType string
+
+const (
+	// Container image based on JDK
+	ImageTypeJava ImageType = "Java"
```
ryanemerson, line 21: Nitpick, but should we call this "JVM"? Both the normal and native images are written in Java, this just determines the runtime.

ryanemerson (commented):
Overall LGTM, just one small thing.

pkg/apis/infinispan/v1/types_util.go
dmvolod: Yes, good point, @ryanemerson . Will fix.

ryanemerson (approved):

rigazilla (approved):

dmvolod:
Thanks for review, @rigazilla and @ryanemerson 
Merged

PR #633, opened 2020-11-03
ConfigMap should reflect Infinispan CR data changes

### LINKED ISSUE #632: ConfigMap should reflect Infinispan CR data changes
Currently there is no way to update infinispan.yaml configuration, when did mistake, for example in xsite configuration or anything else. Implementing #604 makes xsite host:port configuration non validated and currently not able to change these parameters in case of error or misprint. We should add this option for ConfigMap.

dmvolod:
Fixes #632

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -2,6 +2,8 @@ package infinispan
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -398,8 +400,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 
 	// Here where to reconcile with spec updates that reflect into
 	// changes to statefulset.spec.container.
-
-	res, err = r.reconcileContainerConf(infinispan, statefulSet, reqLogger)
+	res, err = r.reconcileContainerConf(infinispan, statefulSet, configMap, reqLogger)
 	if res != nil {
 		return *res, err
 	}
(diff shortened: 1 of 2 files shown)


ryanemerson (approved):

rigazilla (approved):
Looks good to me.

I'm a little bit worried about how it's simple to trigger a cluster restart... but it's a general concern this is not strictly related to this PR.

dmvolod:
Yes, @rigazilla agree with you. We can wait for backup/restore discussion as some assumptions in implementation can brake this logic. Also logger config should remove from the configuration and can be done via REST API now rather than static yaml config.
We need to add docs notes which Infinispan CR could reflect a cluster restart.

rigazilla:
maybe we can run a k8s rolling upgrade in this case (and other cases), since nodes have all the same Infinispan version

dmvolod:
> maybe we can run a k8s rolling upgrade in this case (and other cases), since nodes have all the same Infinispan version

It seems to be enabled and using by default in our code
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates
https://github.com/infinispan/infinispan-operator/blob/69c8f16f9f2720527ec4c8c6216a9f5f2656361a/pkg/controller/infinispan/infinispan_controller.go#L715

rigazilla:
> It seems to be enabled and using by default in our code. 

Ouch, yeah sorry I was mixing things up with upgrade procedure.


>We should add and utilize non-stop test steps in two-nodes CPU update (TestContainerCPUUpdateWithTwoReplicas)
> https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates
> 
> https://github.com/infinispan/infinispan-operator/blob/69c8f16f9f2720527ec4c8c6216a9f5f2656361a/pkg/controller/infinispan/infinispan_controller.go#L715

yep we should test update reliability

dmvolod:
Thanks @rigazilla and @ryanemerson for review, merged.

PR #641, opened 2020-11-11
Remove crdgen target and fix scripts

### LINKED ISSUE #629: Remove separate crdgen make target
Now a separate `make crdgen` target doesn't make sense as code generation more complex task than just CRD
We should keep `make codegen` target only for CRD and full code generation.

dmvolod:
Fixes #629


ryanemerson:
Thanks @dmvolod

PR #653, opened 2020-11-16
Add OCP/OKD requirements to Backup/Restore test and code cleanup

dmvolod:


test/e2e/backup_restore_test.go
```diff
@@ -53,6 +53,10 @@ func testBackupRestore(clusterSpec clusterSpec) func(*testing.T) {
 
 		// 3. Backup the cluster's content
 		backupSpec := &v2.Backup{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "infinispan.org/v2alpha1",
+				Kind:       "Backup",
+			},
```
ryanemerson, line 59: Why is adding the `TypeMeta` necessary?

pkg/controller/zerocapacity/zero_controller.go
```diff
@@ -217,31 +216,23 @@ func (z *Controller) initializeResources(request reconcile.Request, instance Res
 	}
 
 	if err := ensureClusterStability(infinispan); err != nil {
-		return reconcile.Result{}, fmt.Errorf("Infinispan not stable: %w", err)
-	}
-
-	statefulset := &appsv1.StatefulSet{}
```
ryanemerson, line 223: Good spot. I forgot to remove this when I updated how the zero pod was configured.

test/e2e/backup_restore_test.go
dmvolod: GVK requires on OCP/OKD for Update. Probably not in this case, but we should don't forget about it
Please have a look at this case and code
https://github.com/infinispan/infinispan-operator/blob/f44347dfd4c5484425a75b11ddde07f0b5fda941/test/e2e/main_test.go#L706-L708

pkg/controller/zerocapacity/zero_controller.go
```diff
@@ -217,31 +216,23 @@ func (z *Controller) initializeResources(request reconcile.Request, instance Res
...
-	if err := z.Client.Get(ctx, clusterKey, statefulset); err != nil {
-		z.Log.Info(fmt.Sprintf("Unable to load Infinispan StatefulSet '%s': %w", clusterName, err))
-		if errors.IsNotFound(err) {
-			return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCluster}, nil
-		}
-		return reconcile.Result{}, err
+		z.Log.Info("Infinispan not ready: %s", err.Error())
```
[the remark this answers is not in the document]
ryanemerson: LGTM

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -248,6 +248,17 @@ func (ispn *Infinispan) NotClusterFormed(pods, replicas int) bool {
 	return notFormed || notEnoughMembers
 }
 
+func (ispn *Infinispan) EnsureClusterStability() error {
```
ryanemerson, line 251: I don't think `EnsureClusterStability` belongs in `type_util` as the logic is specific to the zero controller and it's perception of whether the cluster is stable enough for the zero pod to be created.

dmvolod: @ryanemerson if you can see, I switched IsWellFormed function to use EnsureClusterStability as you implementation more correct for cluster validation than just check single condition

ryanemerson: O right, I missed that. In the case let's keep it. 

`IsWellFormed` is only used by the cache controller, maybe we should remove `IsWellFormed` and just update the call to:  
```go
if err := ispnInstance.EnsureClusterStability(); err != nil {
...
}
```

dmvolod: I've found IsWellFormed in `autoscale` and test also. Are you agree to change all possible places with `EnsureClusterStability`. If yes, I will do.

ryanemerson: Sorry my search was on an outdated `master` .. if it has a few uses I don't mind keeping it, but whichever you prefer :+1:

pkg/controller/zerocapacity/zero_controller.go
```diff
@@ -419,17 +398,14 @@ func (z *Controller) zeroPodSpec(name, namespace string, configMap *corev1.Confi
 	}
 
 	if zeroSpec.Volume.UpdatePermissions {
-		pod.Spec.InitContainers = []corev1.Container{
-			ispnCtrl.ChmodInitContainer("backup-chmod-pv", name, zeroSpec.Volume.MountPath),
-		}
+		ispnCtrl.AddVolumeChmodInitContainer("backup-chmod-pv", name, zeroSpec.Volume.MountPath, &pod.Spec)
 	}
 
-	ispnCtrl.AddServerDataChmodInitContainer(dataVolName, &pod.Spec)
```
dmvolod, line 427: @ryanemerson do we really need to add chmod init container for zero pod data volume as it's EmptyDir, not real volume
Also OCP tests were fails as backup-chmod-pv adds without lookup MAKE_DATADIR_WRITABLE env variable.
wdyt about this?

ryanemerson: > do we really need to add chmod init container for zero pod data volume as it's EmptyDir, not real volume

Possibly not. I made this change when I was investigating the CI failures, so I think we can remove it.



> Also OCP tests were fails as backup-chmod-pv adds without lookup MAKE_DATADIR_WRITABLE env variable.
> wdyt about this?

Right, makes sense to check the env var then. I think minikube will fail without the initContainer, but I can just set `MAKE_DATADIR_WRITABLE` in that case.

dmvolod: Thanks, already fixed.

ryanemerson:
Thanks @dmvolod

PR #652, opened 2020-11-16
Align Status.Conditions.Type and Status with the k8s commons

dmvolod:
Preparation for #281

PR #654, opened 2020-11-17
Add anti-affinity to StatefulSet PodSpec to ensure distribution acros…

ryanemerson:
This ensures that the pod scheduler prioritises the creation of pods across available nodes. The benefit of this is increased availability in the event of node failure(s).

@rigazilla @dmvolod I went with `PreferredDuringSchedulingIgnoredDuringExecution` opposed to `RequiredDuringSchedulingIgnoredDuringExecution` as I felt that was too restrictive, it also prevents solutions such as `minikube`[1] being used for testing. 

[1] minikube does have multi-node support, but it's experimental and my brief attempts to get it to work ended in failure.

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity

Fixes ##651

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -665,13 +665,32 @@ func (r *ReconcileInfinispan) scheduleUpgradeIfNeeded(infinispan *infinispanv1.I
 	return nil, nil
 }
 
+func podAffinity(i *infinispanv1.Infinispan, matchLabels map[string]string) *corev1.Affinity {
+	// The user hasn't configured Affinity, so we utilise the default strategy of preferring pods are deployed on distinct nodes
+	if i.Spec.Affinity == nil {
+		return &corev1.Affinity{
+			PodAntiAffinity: &corev1.PodAntiAffinity{
+				PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{{
+					Weight: 100,
+					PodAffinityTerm: corev1.PodAffinityTerm{
+						LabelSelector: &metav1.LabelSelector{
+							MatchLabels: matchLabels,
+						},
+						TopologyKey: "kubernetes.io/hostname",
+					},
+				}},
+			},
+		}
+	}
+	return i.Spec.Affinity
+}
+
 // deploymentForInfinispan returns an infinispan Deployment object
 func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispan, secret *corev1.Secret, configMap *corev1.ConfigMap) (*appsv1.StatefulSet, error) {
 	reqLogger := log.WithValues("Request.Namespace", m.Namespace, "Request.Name", m.Name)
 	lsPod := PodLabels(m.ObjectMeta.Name)
 
 	memory := resource.MustParse(m.Spec.Container.Memory)
-
 	replicas := m.Spec.Replicas
 	dep := &appsv1.StatefulSet{
 		TypeMeta: metav1.TypeMeta{
@@ -696,6 +715,7 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 					Annotations: map[string]string{"updateDate": time.Now().String()},
 				},
 				Spec: corev1.PodSpec{
+					Affinity: podAffinity(m, lsPod),
 					Containers: []corev1.Container{{
 						Image:          m.ImageName(),
 						Name:           "infinispan",
--- a/pkg/apis/infinispan/v1/infinispan_types.go
+++ b/pkg/apis/infinispan/v1/infinispan_types.go
@@ -4,6 +4,7 @@ package v1
 // NOTE: json tags are required. Any new fields you add must have json tags for the fields to be serialized.
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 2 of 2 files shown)


ryanemerson:
I'll create a backport PR once this has been merged.

dmvolod:
@ryanemerson I don't know why exactly, need to identify, but after fix in PR #652 test become not stable. Before this fix, `EnsureClusterStability` was not working as expected, but now, test validates for `waitForPodsOrFail` but sometimes didn't pass `WellFormed` validation that's strange. Probably some stale data is getting in zero controller but need to be investigated.

dmvolod:
It seems to be #653 fixed tests instability

dmvolod:
> [1] minikube does have multi-node support, but it's experimental and my brief attempts to get it to work ended in failure.

@ryanemerson  that's strange as Kind also has only one control-plane node and tests are passed. What is error on minukube you have got?

ryanemerson:
The issue was with minikube launching with mult-node enabled, not our code

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -696,6 +696,19 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 					Annotations: map[string]string{"updateDate": time.Now().String()},
 				},
 				Spec: corev1.PodSpec{
+					Affinity: &corev1.Affinity{
```
dmvolod, line 699: @ryanemerson wdyt about make this configurable. Fox example not for each CR, i.e. globally for whole operator

ryanemerson: Do you have a use-case in mind where it's desirable for pods to not be spread across distinct nodes? Co-locating pods on the same node offers zero fault-tolerance of node failures. If a cluster exclusively consists of multiple pods on the same node, it would actually be more beneficial to have a single infinispan pod with great cpu/memory resources as this avoids the overhead of our cluster commands and the ~400mb server footprint required on each pod.

dmvolod: Yeah, right. The single use-case I can see is the testing, demonstration and POC's with low resources envs. Otherwise this is good implementation and using in production ready envs.

ryanemerson: I still think we should be fine for testing etc as we use `PreferredDuringSchedulingIgnoredDuringExecution` opposed to `RequiredDuringSchedulingIgnoredDuringExecution`. The latter would cause issues with those environments, but with Preferred it's still possible for a single node to house all pods if that's the only option.

dmvolod (approved):

ryanemerson:
Having further considered the problem, I think that it's necessary to allow configurable affinity strategies as different Infinispan use-cases can require very different (anti-)affinity rules.

I propose that we add `corev1.Affinity` to the Infinispan spec so that users can provide custom strategies. Example of different possible strategies:

- Prefer|Require pods are deployed across availability zones
- Prefer|Require pods are deployed across nodes within the same availability zone
- Prefer|Require pods are colated on nodes containing a specific app

So the Spec becomes:
```go
type InfinispanSpec struct {
	Replicas  int32                   `json:"replicas"`
	Image     *string                 `json:"image,optional,omitempty"`
	Security  InfinispanSecurity      `json:"security,optional,omitempty"`
	Container InfinispanContainerSpec `json:"container,optional,omitempty"`
	Service   InfinispanServiceSpec   `json:"service,optional,omitempty"`
	Logging   *InfinispanLoggingSpec  `json:"logging,optional,omitempty"`
	Expose    *ExposeSpec             `json:"expose,optional,omitempty"`
	Autoscale *Autoscale              `json:"autoscale,optional,omitempty"`
	Affinity  *corev1.Affinity        `json:"affinity,optional,omitempty"`
}
```

If `Affinity == nil`, then we can set the following as a default to ensure that distributing  pods across nodes is preferred:

```
						PodAntiAffinity: &corev1.PodAntiAffinity{
							PreferredDuringSchedulingIgnoredDuringExecution: []corev1.WeightedPodAffinityTerm{{
								Weight: 100,
								PodAffinityTerm: corev1.PodAffinityTerm{
									LabelSelector: &metav1.LabelSelector{
										MatchLabels: lsPod,
									},
									TopologyKey: "kubernetes.io/hostname",
								},
							}},
						},
```

rigazilla:
As long as we can provide good defaults, it's ok to me to offer more configurability to the user

dmvolod:
> ```go
> Affinity  *corev1.Affinity        `json:"affinity,optional,omitempty"`
> ```

Yeah. Good option, we can think to place this structure deeply inside other structures or keep it as is.

Crumby:
> I propose that we add corev1.Affinity to the Infinispan spec so that users can provide custom strategies.

Crumby:
> I propose that we add corev1.Affinity to the Infinispan spec so that users can provide custom strategies.

That probably makes sense.

> If Affinity == nil, then we can set the following as a default to ensure that distributing pods across nodes is preferred

What if I don't want any affinity settings? From the docs above:
> _Inter-pod affinity and anti-affinity require substantial amount of processing which can slow down scheduling in large clusters significantly. We do not recommend using them in clusters larger than several hundred nodes._

ryanemerson:
> What if I don't want any affinity settings? From the docs above:

The more I have thought about this, I don't think it makes sense to ever run Infinispan without Affinity settings even if scheduling is more expensive. Without basic Affinity rules we can't guarantee availability in the event of a single failure. 

> To simplify the writing of affinity terms, we could automatically populate the `LabelSelector` field to the infinispan pod labels if it is missing from the Infinispan spec.

I have omitted this for now as I believe it's valid to define Affinity without label selectors, so automatically this could cause unexpected behaviour for the user.

rigazilla:
merged, thanks!

rigazilla:
@ryanemerson I've maybe merged too early. Don't you need to update the crds and install also?

ryanemerson:
@rigazilla Whoops, sorry about that! https://github.com/infinispan/infinispan-operator/pull/662

PR #655, opened 2020-11-17
ISPN-12455 backup restore docs

oraNod:
https://issues.redhat.com/browse/ISPN-12455


documentation/asciidoc/topics/proc_backing_up_cluster.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+.Prerequisites
+
+* Create a {brandname} cluster of {datagridservice} nodes.
+* Create at least one cache and add some entries.
+
+[IMPORTANT]
+====
+The backup operation preserves the state of {brandname} clusters.
+To prevent data loss, you should ensure there are no active client connections to the cluster you want to back up.
```
ryanemerson, line 13: It might be better to state something like "{brandname} backups do not provide [snapshot isolation](https://en.wikipedia.org/wiki/Snapshot_isolation), entries are stored in the archive as they are read meaning that any write operations peformed between a backup starting and finishing, may not be included in the archive. Therefore, if a backup of the exact container content is required, it's necessary to ensure ensure no active client connections exist when the backup is initiated".

Having active connections during the backup won't result in data loss from the running cluster, it just means that the backup may include an original entry value of `k=1`, whereas the value in the cluster could be updated to `k=2` during the backup process and this will only be reflected in the live cluster as the entry has already been stored in the backup archive.

documentation/asciidoc/topics/proc_restoring_cluster.adoc
```diff
@@ -0,0 +1,51 @@
+[id='restoring_clusters-{context}']
+= Restoring {brandname} Clusters
+Restore {brandname} cluster state from backup archives.
+When you restore cluster state, {ispn_operator} applies all caches and data from the backup archive on to the target cluster.
```
ryanemerson, line 4: It's also possible to selectively restore content from a backup via:

```yaml
spec:
  resources:
    caches:
      - some-cache
      - and-another-one
    counters:
      - counter-name
```

documentation/asciidoc/topics/proc_restoring_cluster.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+.Prerequisites
+
+* Create a {brandname} cluster of {datagridservice} nodes where you want to restore state.
++
+[IMPORTANT]
+====
+The operation to restore {brandname} overwrites state on the target cluster.
+To prevent data loss, you should ensure there are no active client connections to the cluster where you want to restore state.
```
ryanemerson, line 13: Similar to the comments about data loss for backups, however it's probably more important in the Restore case as data restored from an archive could overwrite a client modification `k=2` that is received during the restore, but before `k=1` is processed during the restore.

documentation/asciidoc/topics/proc_restoring_cluster.adoc
```diff
@@ -0,0 +1,51 @@
...
+====
+
+* Create a backup archive on a source cluster.
```
ryanemerson, line 16: This has to be a Backup CR

documentation/asciidoc/topics/proc_restoring_cluster.adoc
```diff
@@ -0,0 +1,51 @@
...
+* Ensure the backup archive is available on the persistent volume claim under: +
```
ryanemerson, line 17: This is handled by the Backup CR, so the user doesn't have to worry about this

documentation/asciidoc/topics/proc_restoring_cluster.adoc
```diff
@@ -0,0 +1,51 @@
...
+For example, create `my-restore.yaml` with the following:
++
+[source,options="nowrap",subs=attributes+]
+----
+include::yaml_annotated/cr_restore.yaml[]
+----
++
+<1> Specifies a **Restore** CR.
+<2> Provides a unique name for the **Restore** CR.
+<3> Specifies the name of the backup file to restore.
```
ryanemerson, line 33: This is the name of the Backup CR

documentation/asciidoc/topics/proc_restoring_cluster.adoc
```diff
@@ -0,0 +1,51 @@
...
+<4> Specifies the name of the {brandname} cluster that you want to restore from the backup.
```
ryanemerson, line 34: This is the name of the Infinispan CR

documentation/asciidoc/topics/proc_backing_up_cluster.adoc
```diff
@@ -0,0 +1,53 @@
+[id='backing_up_clusters-{context}']
+= Backing Up {brandname} Clusters
+Create backup files that stores the state of {brandname} clusters to persistent volume claims.
```
dmvolod, line 3: Probably stores not in PVC, but in PV. PVC is a request for PV
https://kubernetes.io/docs/concepts/storage/persistent-volumes/

documentation/asciidoc/topics/proc_backing_up_cluster.adoc
```diff
@@ -0,0 +1,53 @@
...
+
+.Prerequisites
+
+* Create a {brandname} cluster of {datagridservice} nodes.
+* Create at least one cache and add some entries.
```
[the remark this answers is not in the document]
oraNod: Yeah I was wondering if we should go with the full "content that the Cache Manager controls" or something. I'll take a stab at it but was trying to avoid being wordy and ambiguous.

documentation/asciidoc/topics/proc_restoring_cluster.adoc
oraNod: @ryanemerson these are all the available options under `spec.resources` right?

ryanemerson: Bingo!

documentation/asciidoc/topics/proc_backing_up_cluster.adoc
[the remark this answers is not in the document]
oraNod: nice. I was actually thinking about this.

documentation/asciidoc/topics/proc_restoring_cluster.adoc
oraNod: thx. am kicking myself a little for not looking there in the first place. one of my questions, that I didn't ask, was about selective restore. I think those resources are pretty self-explanatory so I didn't add callouts for each one. hopefully the examples make them sufficiently clear too.

dmvolod (approved):

documentation/asciidoc/topics/yaml_annotated/cr_backup.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: infinispan.org/v2alpha1
+kind: Backup <1>
+metadata:
+  name: my-backup <2>
+spec:
+  cluster: source-cluster <3>
```
ryanemerson, line 6: It's possible to selectively backup resources as well, maybe we should add `spec.resources` like in the Restore example?

oraNod: Thanks, Ryan. I've added `spec.resources` as a separate step for both backup and restore procedures. Hopefully it looks alright that way.

ryanemerson:
Great work @oraNod. Can you squash the commits and then I will merge.

ryanemerson:
Thanks @oraNod

Issue #657, opened 2020-11-19
Publish 2.1.0 on hub/catalog from master branch


rigazilla:
- [ ] regenerate csv file with `operator-sdk generate csv`
- [ ] add human-written extra info
- [ ] change DEFAULT_IMAGE to RELATED_IMAGE_OPENJDK
- [ ] tag master with 2.1.0.Alpha1

PR #659, opened 2020-11-20
updated files to be operator-sdk friendly #657

rigazilla:
- removed duplicated crd files
- removed useless info in csv
- versioned csv in the place where operator-sdk create it
- changed DEFAULT_IMAGE to RELATE_IMAGE_OPENJDK


deploy/role.yaml
```diff
@@ -31,7 +31,6 @@ rules:
 - apiGroups:
   - apps
   resources:
-  - replicasets
```
dmvolod, line 34: replicasets is required. You will get following error during operator startup without this permission  

```
{"level":"info","ts":1605865859.5327392,"logger":"cmd","msg":"Could not create metrics Service","error":"failed to initialize service object for metrics: replicasets.apps \"infinispan-operator-85c5568b5c\" is forbidden: User \"system:serviceaccount:infinispan-operator:infinispan-operator\" cannot get resource \"replicasets\" in API group \"apps\" in the namespace \"infinispan-operator\""}
```

dmvolod (changes requested):

deploy/operator.yaml
```diff
@@ -35,5 +35,5 @@ spec:
                   fieldPath: metadata.name
             - name: OPERATOR_NAME
               value: "infinispan-operator"
-            - name: DEFAULT_IMAGE
+            - name: RELATED_IMAGE_OPENJDK
```
dmvolod, line 38: Regen install-bundle also requires for this changes

deploy/role.yaml
rigazilla: ouch, I though it was a relic from the first version. thanks @dmvolod

dmvolod: No, @rigazilla I validated all possible rights during role cleanup

deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -133,10 +131,8 @@ spec:
           - apiGroups:
               - apps
             resources:
-              - deployments
```
dmvolod, line 136: @rigazilla this should be aligned with role.yaml

rigazilla: yeah sure

rigazilla: mmm, actually the `permissions` is generated by the operator-sdk in the right place from role.yaml.
I would remove it from here so we don't have to align duplicated text

rigazilla:
@dmvolod I've updated the csv, please have another look

dmvolod:
Looks good, thanks @rigazilla 
But CI fails again. It seems to be fails before but we never did WellFormed validation for two-nodes cluster as Backup/Restore controller does right now.

test/e2e/constants/constants.go
```diff
@@ -36,8 +36,8 @@ var (
 	RunSaOperator        = strings.ToUpper(constants.GetEnvWithDefault("RUN_SA_OPERATOR", "false"))
 	OperatorUpgradeStage = strings.ToUpper(constants.GetEnvWithDefault("OPERATOR_UPGRADE_STAGE", OperatorUpgradeStageNone))
 	CleanupInfinispan    = strings.ToUpper(constants.GetEnvWithDefault("CLEANUP_INFINISPAN_ON_FINISH", "true"))
-	ExpectedImage        = constants.GetEnvWithDefault("EXPECTED_IMAGE", "infinispan/server:12.0")
-	NativeImageName      = constants.GetEnvWithDefault("NATIVE_IMAGE", "infinispan/server-native:12.0")
+	ExpectedImage        = constants.GetEnvWithDefault("EXPECTED_IMAGE", "quay.io/infinispan/server:12.0")
```
dmvolod, line 39: @rigazilla do we need to use images from quay.io in all places, i.e. deployment, test, etc?

rigazilla: @dmvolod images are on docker and quay, not a requirement but quay should be preferable.
Actually this is just a test if CI is happier with quay, I've remembered that time ago I've got  a popup performance warning from Travis UI about pulling images from docker

dmvolod: Oh, @rigazilla yes, but we should update default Infinispan image for operator also to prevent this problem
Will try to test with quay based images

rigazilla: yeah, let's see if CI is happy with quay. If yes we can align all the code to pull from there

rigazilla: hey this PR turned to green...

dmvolod: Yeah, but tests still using images from docker hub. This means that CI is not stable anymore
```
{},"ready":false,"restartCount":0,"image":"docker.io/infinispan/server:12.0","imageID":"docker.io/infinispan/server@sha256:8696e77b3b866a938b205a65d5b4122b2d22d34769efc2fb87addf5fa0a52819","containerID":"containerd://7a5e416285a3b20e31c5740b0fc914dad7489fd0c05021e0caef0577b05113be","started":true}]}
```

PR #665, opened 2020-11-30
wellFormed condition isn't present until cluster is formed for the first time

dmvolod:


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1003,8 +1014,6 @@ func (r *ReconcileInfinispan) computeConfigMap(xsite *config.XSite, m *infinispa
 
 	loggingCategories := m.GetLogCategoriesForConfigMap()
 	config := config.CreateInfinispanConfiguration(name, loggingCategories, namespace, xsite)
-	// Explicitly set the numner of lock owners in order for zero-capacity nodes to be able to utilise clustered locks
-	config.Infinispan.Locks.Owners = m.Spec.Replicas
```
dmvolod, line 1007: @ryanemerson we can't remove this option from Infinispan cluster config, as this cause following exception while zero pod starting
```
10:10:36,564 ERROR (non-blocking-thread--p2-t2) [org.infinispan.topology.LocalTopologyManagerImpl] ISPN000230: Failed to start rebalance for cache org.infinispan.LOCKS java.util.concurrent.CompletionException: java.lang.ClassCastException: class org.infinispan.distribution.ch.impl.ReplicatedConsistentHash cannot be cast to class org.infinispan.distribution.ch.impl.DefaultConsistentHash (org.infinispan.distribution.ch.impl.ReplicatedConsistentHash and org.infinispan.distribution.ch.impl.DefaultConsistentHash are in unnamed module of loader java.net.URLClassLoader @6d03e736)
	at java.base/java.util.concurrent.CompletableFuture.encodeRelay(CompletableFuture.java:367)
	at java.base/java.util.concurrent.CompletableFuture.completeRelay(CompletableFuture.java:376)
	at java.base/java.util.concurrent.CompletableFuture$UniRelay.tryFire(CompletableFuture.java:1019)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:506)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2088)
	at org.infinispan.util.concurrent.ActionSequencer$SequenceEntry.accept(ActionSequencer.java:213)
	at org.infinispan.util.concurrent.ActionSequencer$SequenceEntry.accept(ActionSequencer.java:179)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:859)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenCompleteStage(CompletableFuture.java:883)
	at java.base/java.util.concurrent.CompletableFuture.whenComplete(CompletableFuture.java:2251)
	at java.base/java.util.concurrent.CompletableFuture.whenComplete(CompletableFuture.java:143)
	at org.infinispan.util.concurrent.ActionSequencer$SequenceEntry.run(ActionSequencer.java:227)
	at org.infinispan.util.concurrent.ActionSequencer$SequenceEntry.apply(ActionSequencer.java:219)
	at org.infinispan.util.concurrent.ActionSequencer$SequenceEntry.apply(ActionSequencer.java:179)
	at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:930)
	at java.base/java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:907)
	at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.lang.ClassCastException: class org.infinispan.distribution.ch.impl.ReplicatedConsistentHash cannot be cast to class org.infinispan.distribution.ch.impl.DefaultConsistentHash (org.infinispan.distribution.ch.impl.ReplicatedConsistentHash and org.infinispan.distribution.ch.impl.DefaultConsistentHash are in unnamed module of loader java.net.URLClassLoader @6d03e736)
	at org.infinispan.distribution.ch.impl.SyncConsistentHashFactory.union(SyncConsistentHashFactory.java:58)
	at org.infinispan.topology.LocalTopologyManagerImpl.doHandleRebalance(LocalTopologyManagerImpl.java:562)
	at org.infinispan.topology.LocalTopologyManagerImpl.lambda$handleRebalance$16(LocalTopologyManagerImpl.java:531)
	at org.infinispan.topology.LocalTopologyManagerImpl.lambda$orderOnCache$24(LocalTopologyManagerImpl.java:736)
	at org.infinispan.util.concurrent.ActionSequencer.safeNonBlockingCall(ActionSequencer.java:57)
	at org.infinispan.util.concurrent.ActionSequencer.access$400(ActionSequencer.java:32)
	at org.infinispan.util.concurrent.ActionSequencer$SequenceEntry.run(ActionSequencer.java:226)
	... 8 more
```
But keeping this config on the Infinispan cluster side, every change in `spec.relicas` cause whole cluster rolling update instead of silent add/remove nodes.

ryanemerson: Right. Setting the number of owners for the locks is actually a workaround ... I will discuss with Dan about getting a fix for the above `ClassCastException` so that we can avoid setting this value in the first place.

dmvolod: Thanks, please keep me informed.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -410,6 +410,19 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+		err = r.client.Get(context.TODO(), types.NamespacedName{Namespace: infinispan.Namespace, Name: infinispan.Name}, infinispan)
+		if err != nil {
+			reqLogger.Error(err, "failed to reload Infinispan status")
+			return reconcile.Result{}, err
+		}
+		if infinispan.SetCondition(infinispanv1.ConditionWellFormed, metav1.ConditionUnknown, "Pods are not ready") || infinispan.Status.StatefulSetName != statefulSet.Name {
+			infinispan.Status.StatefulSetName = statefulSet.Name
+			err := r.client.Status().Update(context.TODO(), infinispan)
+			if err != nil {
+				reqLogger.Error(err, "failed to update Infinispan status")
```
ryanemerson, line 422: Should we create a function the handle the updating of the status?

A quick search in `infinispan_controller` shows that before this change there's already 9 places where we log "failed to update Infinispan status"

dmvolod: Yeah, I was discussed this case with @rigazilla before and he was not agree with creating common function for update.

rigazilla: Yep, my impression is that we are just wrapping the error log in a function without saving a lot. 

When we discussed this the code was more unstable than today, I was also worried to maintain the flexibility to change the error handling for a single situation.

Now I just have a slight preference for inline code because I consider it more readable than a function.
No problem in adding a function anyway, if you think it worth

ryanemerson: I think we should reconsider. Adding something simple like the following to `infinispan_controller` centralises the logic:

```go
func (r *ReconcileInfinispan) update(ispn *infinispanv1.Infinispan, logger logr.Logger) error {
	if err := r.client.Status().Update(context.TODO(), ispn); err != nil {
		logger.Error(err, "Failed to update Infinispan Status")
		return err
	}
	return nil
}
```

We can then do something like the following when returning from the various reconcile functions:

```
return &reconcile.Result{}, update(ispn, logger)
```

rigazilla: does this works in a general way? Sometimes code returns only in case of error.

dmvolod: Thanks, @rigazilla and @ryanemerson sure.
As Status update flow is always static
- get latest Infinispan CR
- verify that update is needed
- set values
- update
will try to implement common function for all possible cases

dmvolod:
Added common UpdateStatus functions. @ryanemerson , @rigazilla please review again

rigazilla (approved):
ok by me

rigazilla:
Does it worth to port the update refactoring on 2.0.x also, to keep branches close?

dmvolod:
> Does it worth to port the update refactoring on 2.0.x also, to keep branches close?

Probably yes, I will backport this PR to 2.0.x

ryanemerson:
Thanks @dmvolod

PR #670, opened 2020-12-03
[2.0.x] Align Status.Conditions.Type and Status with the k8s commons

dmvolod:


dmvolod:
Close, will create a new PR

PR #673, opened 2020-12-04
Fix StatusUpdate regression from #665

dmvolod:


pkg/apis/infinispan/v1/types_util.go
```diff
@@ -324,11 +324,13 @@ func (ispn *Infinispan) UpdateStatus(client client.Client, logger logr.Logger, s
 		if statusUpdate != nil {
 			statusUpdate()
 		}
-		err := client.Status().Update(context.TODO(), ispn)
+		ispn.Status.DeepCopyInto(&infinispan.Status)
```
ryanemerson, line 327: Why are we explicitly calling `DeepCopyInto` here? Can't we just do `infinispan.status = ispn.status` and the value will be copied.

dmvolod: `DeepCopy` do the same if `struc` don't have an elements we need special operation, like arrays, slices, pointers, etc.

dmvolod: And I don't want to be care about all of this

dmvolod:
@ryanemerson CI now green but output still has some sort of errors during update as part of StatusUpdate regression
I will plan to fix them as part of the #674

pkg/apis/infinispan/v1/types_util.go
rigazilla: ok by me

rigazilla (approved):

ryanemerson:
Thanks @dmvolod

rigazilla:
https://github.com/infinispan/infinispan-operator/blob/44890d0e7c97772c0bf31baa85f4722c31d6134e/pkg/apis/infinispan/v1/types_util.go#L323

Little detail
I haven't noticied before, but this line can be simpilfied

dmvolod:
> https://github.com/infinispan/infinispan-operator/blob/44890d0e7c97772c0bf31baa85f4722c31d6134e/pkg/apis/infinispan/v1/types_util.go#L323
> 
> Little detail
> I haven't noticied before, but this line can be simpilfied

Probably yes, @rigazilla please suggest chnages

rigazilla:
```
	if statusValidate == nil || (statusValidate != nil && statusValidate()) {
		if statusUpdate != nil {
			statusUpdate()
		}
```
I think that this is equivalent, due to how go evaluates boolean expr
```
	if statusValidate == nil  || statusValidate() {
		if statusUpdate != nil {
			statusUpdate()
		}
```

dmvolod:
Cool, exactly, thanks, @rigazilla

PR #675, opened 2020-12-04
StatusUpdate, updateSecurity and spec update functions review

dmvolod:


pkg/apis/infinispan/v1/types_util.go
```diff
@@ -313,23 +319,50 @@ func (ispn *Infinispan) IsEncryptionCertSourceDefined() bool {
...
-type StatusUpdateFn func()
-type StatusValidateFn func() bool
+type UpdateFn func()
+type ValidateFn func() bool
+
+func (ispn *Infinispan) Update(client clnt.Client, logger logr.Logger, validate ValidateFn, update UpdateFn) error {
+	infinispan := &Infinispan{}
+	err := client.Get(context.TODO(), types.NamespacedName{Namespace: ispn.Namespace, Name: ispn.Name}, infinispan)
+	if err != nil {
+		logger.Error(err, "failed to reload Infinispan")
```
ryanemerson, line 329: No need to log errors in the `Update` method as we always return the `err` instance in the infinispan_controller if it's not null. Logging the error twice just adds verbosity

ryanemerson: The same applies to the `UpdateStatus` method.

ryanemerson (changes requested):

pkg/apis/infinispan/v1/types_util.go
dmvolod: Yeah, agree with you @ryanemerson 
But the problem describe in #676 says, that second error trace (from the controller-runtime) fully informal and don't display stack trace and lace of the error
I would prefer keep this "as is" and discuss and implement this at the your issue #690

pkg/controller/zerocapacity/zero_controller.go
```diff
@@ -401,7 +401,7 @@ func (z *Controller) zeroPodSpec(name, namespace string, configMap *corev1.Confi
 
 func (z *Controller) configureServer(name, namespace string, infinispan *v1.Infinispan, instance Resource) (*corev1.ConfigMap, error) {
 	clusterConfig := &corev1.ConfigMap{}
-	clusterConfigName := ispnCtrl.ServerConfigMapName(instance.Cluster())
+	clusterConfigName := infinispan.GetConfigName()
```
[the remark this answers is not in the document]
dmvolod: Yes, sure, will do

pkg/apis/infinispan/v1/types_util.go
ryanemerson: > I would prefer keep this "as is" and discuss and implement this at the your issue #690

Makes sense. We can ensure we're consistent with our approach that way.

ryanemerson:
One final thought @dmvolod. Does it make sense for the `Update` and `UpdateStatus` to be in `type_utils`? Imo this logic should reside in the controller as it's mutating state. Obviously we would have to update the signatures to not extend `Infinispan`, but the advantage is that it's no longer necessary to pass the client and controller per invocation, for example we could something like:

```go
func (r *ReconcileInfinispan) UpdateStatus(ispn *ispn.Infinispan, statusValidate StatusValidateFn, statusUpdate StatusUpdateFn) error {
	infinispan := &ispn.Infinispan{}
	err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: ispn.Namespace, Name: ispn.Name}, infinispan)
        ...
}
```

dmvolod:
> One final thought @dmvolod. Does it make sense for the `Update` and `UpdateStatus` to be in `type_utils`? Imo this logic should reside in the controller as it's mutating state. Obviously we would have to update the signatures to not extend `Infinispan`, but the advantage is that it's no longer necessary to pass the client and controller per invocation, for example we could something like:

Yeah, thanks @ryanemerson make sense. Will do
I would prefer to have a struct like in Backup/Restore controllers combining reconcile CR and k8s client in one place, but this need to be very big refactoring. We need to split controllers before it.

ryanemerson (approved):

dmvolod:
@rigazilla please have a look, but it doesn't make sense a lot in case of #684 refactoring

rigazilla:
> @rigazilla please have a look, but this PR doesn't make sense a lot in case of #684 refactoring

yeah @dmvolod, tbh I would focus on #684 first and rebase this on top later

dmvolod:
> yeah @dmvolod, tbh I would focus on #684 first and rebase this on top later

Yeah, @rigazilla but I suddenly made the inverse dependencies. This would be nice to merge #675 first, but #684 need to be changing after it

rigazilla:
> > yeah @dmvolod, tbh I would focus on #684 first and rebase this on top later
> 
> Yeah, @rigazilla but I suddenly made the inverse dependencies. This would be nice to merge #675 first, but #684 need to be changing after it

then LGTM :)

rigazilla (approved):

ryanemerson:
Thanks @dmvolod

Issue #573, opened 2020-09-15
Organize constants and env variables usage


dmvolod:
Add more parameters as constants and consolidate env variable interaction in one place
Validate lowercase and uppercase flags and values usage (test and core operator code)

PR #684, opened 2020-12-09
Operator don't care about autogenerated resources state

dmvolod:
The main purpose of this PR was adding support for recovering auto-generated resources such as ConfigMap, Service, Route/Ingress and Secret in case of update/delete them outside of controller.
But dealing this logic inside single infinispan_controller is very complex task, as we need to handle events from those resources in one place and create a large number of resource specific logic.
This will be also helpful for future zero-downtime upgrade functionality as current single Infinispan CR controller very complex and new function will add more complexity and will contains a large number of if-then-else combinations included inside each others.
The idea just create a separate controllers for each resource type or group of the resources, i.e.

- [x] config_controller for Infinispan cluster configuration ConfigMap
- [x] service_controller for all types of the services (internal, external and ping) as well as for Service/Route/Ingress exposition
- [x] secret_controller for operator managed secrets (auto-generated if not present).

Also separate controllers will add parallel processing in some cases instead of current sequential processing, except wait for resources in main controller

Probably, we can use `kustomize` resource creation when we will update to the Operator SDK 1.x and later, but only few number of our resources can be created with templates easily (not required special logic inside and plain structure without complex conditions and dependencies). For example, [quay-operator](https://github.com/quay/quay-operator) which creates just few deployments (Clair, PostgreSQL, Redis, etc.) and can't reflect changes in CR has large number of custom code on top of the `kustomize`

It seems to be updates of the Infinispan CR became much more stable now


pkg/controller/infinispan/resources/resource_controller.go
```diff
@@ -0,0 +1,157 @@
...
+	return reflect.TypeOf(rt.ObjectType).Elem().Name()
+}
+
+type Reconciler interface {
+	// The k8 structs being handled by this controller
+	Types() []*ReconcileType
+	// Events for the struct handled by this controller
+	EventsPredicate() predicate.Predicate
+	// Create a new instance of Infinispan wrapping the actual k8s type
+	InfinispanInstance(infinispan *v1.Infinispan, ctrl *Controller, kube *kube.Kubernetes, log logr.Logger) Resource
```
ryanemerson, line 58: How about `ResourceInstance` as we're returning an implementation of the `Resource` interface?

pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -0,0 +1,110 @@
...
+		},
+	}
+}
+
+func Add(mgr manager.Manager) error {
+	return resources.CreateController(ControllerName, &reconcileSecret{mgr.GetClient()}, mgr)
+}
+
+func (s *secretResource) Process() (reconcile.Result, error) {
+	if s.infinispan.IsGeneratedSecret() {
```
ryanemerson, line 77: If we invert the condition we can return early an avoid the entire method being indendented in the if.
```go
`if !s.infinispan.IsGeneratedSecret() {
}
// Continue to process
... 
```

pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -0,0 +1,110 @@
...
+func (s secretResource) computeSecret(identities []byte) *corev1.Secret {
+	return &corev1.Secret{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Secret",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      s.infinispan.GetSecretName(),
+			Namespace: s.infinispan.Namespace,
+			Labels:    infinispan.LabelsResource(s.infinispan.Name, "infinispan-secret-identities"),
```
ryanemerson, line 104: We can make "infinispan-secret-identities" a const.

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -0,0 +1,170 @@
...
+
+	err := infinispan.ConfigureServerEncryption(c.infinispan, &config, c.client)
+	if err != nil {
+		return nil, err
+	}
+	configYaml, err := config.Yaml()
+	if err != nil {
+		return nil, err
+	}
+	lsConfigMap := infinispan.LabelsResource(name, "infinispan-configmap-configuration")
```
ryanemerson, line 128: We can make "infinispan-configmap-configuration"" a constant.

pkg/controller/infinispan/resources/service/xsite_service.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	ispnv1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
+	consts "github.com/infinispan/infinispan-operator/pkg/controller/constants"
+	"github.com/infinispan/infinispan-operator/pkg/controller/infinispan"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+func computeSiteService(ispn *ispnv1.Infinispan) *corev1.Service {
```
ryanemerson, line 12: Should we just incorporate this as part of `server_controller.go` as it's only a singe method?

pkg/controller/infinispan/resources/resource_controller.go
```diff
@@ -0,0 +1,157 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+//
```
ryanemerson, line 29: Empty comment

ryanemerson (commented):
@dmvolod A few minor points, but in general I like this approach a lot :+1:

pkg/controller/infinispan/resources/resource_controller.go
dmvolod: Yes, agreed. Will change

pkg/controller/infinispan/resources/service/xsite_service.go
dmvolod: Yes, thanks, will fix

pkg/controller/infinispan/resources/resource_controller.go
dmvolod: Thanks, fixed

pkg/controller/infinispan/resources/secret/secret_controller.go
dmvolod: Let's keep this "as is" and plant to review all constants usage in separate submitted issue #573

pkg/controller/infinispan/resources/config/config_controller.go
dmvolod: Let's keep "as is" and plan to review all constants usage in separate submitted issue #573

pkg/controller/infinispan/resources/secret/secret_controller.go
dmvolod: Agree, will fix

rigazilla (commented):
I like the refactoring.
I tried a run on my PC seems that *-generated-secret is not deleted after the infinispan CR is deleted. Maybe there's something to check?

dmvolod:
> I tried a run on my PC seems that *-generated-secret is not deleted after the infinispan CR is deleted. Maybe there's something to check?

Oh, thanks @rigazilla . I missed SetOwnerReference for Secret during refactoring

rigazilla (approved):

ryanemerson:
@dmvolod LGTM. Can you squash the two commits and then I'll merge

dmvolod:
Thanks, @ryanemerson and @rigazilla for review
I've found some issues related to drop Secret and update ConfigMap. Working on it. Will update soon

dmvolod:
We also need to merge #675 before

dmvolod:
Some logic was changed.
@ryanemerson , @rigazilla please have a look again.
Thanks

rigazilla (approved):
I've deployed some clusters and randomly deleted secrets/services and route: seems to work quite well. Great!

dmvolod:
PR is ready for final review and merge
Thanks, @ryanemerson and @rigazilla

ryanemerson (approved):

dmvolod:
Thanks @ryanemerson and @rigazilla for review
Merged into master

PR #693, opened 2020-12-16
ISPN-12579: Able to choose operator documentation version at the site

dmvolod:
Preparation for website fix

### DIFF
--- a/documentation/asciidoc/css/css.css
+++ b/documentation/asciidoc/css/css.css
@@ -1,0 +1,27 @@
+#toc {
+    overflow-x: hidden;
+}
+#tocsearch {
+    font-family: 'FontAwesome';
+    width: 80%;
+}
+#toctreecollapse, #toctreeexpand {
+    padding: 0 0.3em;
+}
+#toctree {
+    font-size: 0.9em;
+}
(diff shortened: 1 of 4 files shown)


oraNod:
@dmvolod Looks really good to me but some observations. 

1. We need to backport to the 1.1.x and 2.0.x branches so they show up in the version selector. 
2. Need to add versions.xml to docs/infinispan-operator. For the core documentation that XML file is built with https://github.com/infinispan/infinispan.github.io/blob/develop/_plugins/fetch_docs.rb 
3. Also need to modify fetch_docs.rb to place the CSS and JS files in the right place. 

It might be worth separating the operator docs part of fetch_docs.rb. It might be cleaner to have fetch_operator_docs.rb that gets called after fetch_docs.rb

image

dmvolod:
@oraNod 2 and 3 already done before (while implementing) please review
https://github.com/infinispan/infinispan.github.io/pull/164
Yes, backport need to be implemented

oraNod (approved):
LGTM

oraNod:
Thanks @dmvolod Can you please create the backports?

PR #695, opened 2020-12-17
added multinamespace support

### LINKED ISSUE #104: Support operator being installed in multi namespace
It should be possible to install the operator in multiple namespaces.

rigazilla:
This fixes #104 and allows multinamespace and global install modes for the operator

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,10 @@ jobs:
       script:
         - go mod vendor
         - RUN_SA_OPERATOR=TRUE make test PARALLEL_COUNT=2
+          # Run multinamespace tests as admin
+          # skip role settings check for now
+        - kubectl config use-context $TESTING_CONTEXT
+        - make multinamespace-test PARALLEL_COUNT=2
       after_failure:
         - cat debug.log
         - kubectl config use-context $TESTING_CONTEXT
--- a/pkg/launcher/launcher.go
+++ b/pkg/launcher/launcher.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 
 	"github.com/infinispan/infinispan-operator/pkg/apis"
 	"github.com/infinispan/infinispan-operator/pkg/controller"
@@ -23,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp" // blank import (?)
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
@@ -85,11 +87,23 @@ func Launch(params Parameters) {
 		os.Exit(1)
 	}
 
-	// Create a new Cmd to provide shared dependencies and start components
-	mgr, err := manager.New(cfg, manager.Options{
+	// Set default manager options
+	options := manager.Options{
 		Namespace:          namespace,
 		MetricsBindAddress: fmt.Sprintf("%s:%d", metricsHost, metricsPort),
-	})
+	}
+
+	// Add support for MultiNamespace set in WATCH_NAMESPACE (e.g ns1,ns2)
+	// Note that this is not intended to be used for excluding namespaces, this is better done via a Predicate
+	// Also note that you may face performance issues when using this with a high number of namespaces.
+	// More Info: https://godoc.org/github.com/kubernetes-sigs/controller-runtime/pkg/cache#MultiNamespacedCacheBuilder
+	if strings.Contains(namespace, ",") {
+		options.Namespace = ""
+		options.NewCache = cache.MultiNamespacedCacheBuilder(strings.Split(namespace, ","))
+	}
+
+	// Create a new manager to provide shared dependencies and start components
+	mgr, err := manager.New(cfg, options)
 	if err != nil {
 		log.Error(err, "")
 		os.Exit(1)
@@ -122,7 +136,11 @@ func Launch(params Parameters) {
 	}
 
 	// Add the Metrics Service
-	addMetrics(ctx, cfg, namespace)
+	if operatorNs, err := k8sutil.GetOperatorNamespace(); err == nil {
+		addMetrics(ctx, cfg, operatorNs)
+	} else {
+		log.Info("Could not get OperatorNamespace. Not in cluster? Skipping addMetrics")
+	}
 
 	log.Info("Starting the Cmd.")
 
--- a/test/e2e/multinamespace/multinamespace_test.go
+++ b/test/e2e/multinamespace/multinamespace_test.go
@@ -1,0 +1,142 @@
+package multinamespace
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
+	ispnv1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
+	cconsts "github.com/infinispan/infinispan-operator/pkg/controller/constants"
+	ispnctrl "github.com/infinispan/infinispan-operator/pkg/controller/infinispan"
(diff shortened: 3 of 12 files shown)


pkg/launcher/launcher.go
```diff
@@ -122,8 +136,10 @@ func Launch(params Parameters) {
 	}
 
 	// Add the Metrics Service
-	addMetrics(ctx, cfg, namespace)
+	if !strings.Contains(namespace, ",") {
```
dmvolod, line 139: @rigazilla looks like we should invoke `serveCRMetrics` explicitly otherwise.

rigazilla: @dmvolod `namespace` is used as a namespace later in the [code](https://github.com/infinispan/infinispan-operator/blob/3688b94a4ae0e06d94ba1d5853cc4475e96dafcb/pkg/launcher/launcher.go#L183-L184), passing the `namespace` value as is, doesn't work for multinamespace.
I didn't spent a lot of time on it since this call has been removed in [1.0.0](https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.0.0/#removed-packages-pkgkube-metrics-and-pkgmetrics), I've just minimized the changes for now.

I don't know, maybe we can just replace it with the namespace where the operator is actually running?

dmvolod: Yeah, sure @rigazilla . This could be `k8sutil.GetOperatorNamespace`
I agree with you, we need to move to the Operator SDK `1.x.x` soon

pkg/launcher/launcher.go
```diff
@@ -136,10 +136,12 @@ func Launch(params Parameters) {
 	}
 
 	// Add the Metrics Service
-	if !strings.Contains(namespace, ",") {
-		addMetrics(ctx, cfg, namespace)
-
+	if operatorNs, err := k8sutil.GetOperatorNamespace(); err == nil {
+		addMetrics(ctx, cfg, operatorNs)
+	} else {
+		log.Info("Could not get OperatorNamespace. Not in cluster? Skipping addMetrics")
```
dmvolod, line 142: Seems to be `serveCRMetrics` do the same

rigazilla: What do you me @dmvolod?

dmvolod: We can modify code like this:
```
if !strings.Contains(namespace, ",") {
	addMetrics(ctx, cfg, namespace)
} else {
	serveCRMetrics(cfg)
}
```
wdyt?

rigazilla: @dmvolod I don't know. We're skipping all addMetrics code after line:
https://github.com/infinispan/infinispan-operator/blob/e055d4d751202485ed3fde908c0cfb6449c43cfe/pkg/launcher/launcher.go#L170

is that ok?

dmvolod: Yeah, probably true
Need to be test with different options

ryanemerson:
@rigazilla Can we add a test for this? Something like:

1. Install the operator in a namespace `infinispan-operator`
2. Create a Infinispan CRs with `spec.replicas=1`  in namespaces `infinispan-cluster-1` and `infinispan-cluster-2`
3. Verify that a Infinispan pod is up and ready in each of the clusters.

rigazilla:
> @rigazilla Can we add a test for this? Something like:
> 
>     1. Install the operator in a namespace `infinispan-operator`
> 
>     2. Create a Infinispan CRs with `spec.replicas=1`  in namespaces `infinispan-cluster-1` and `infinispan-cluster-2`
> 
>     3. Verify that a Infinispan pod is up and ready in each of the clusters.

I think this will need some changes on the testsuite, since it works on a single namespace atm. Let me see...

ryanemerson:
> I think this will need some changes on the testsuite, since it works on a single namespace atm. Let me see...

I was thinking it would be best in it's own file, similar to `backup_restore_test.go`. That way we don't have to worry about adapting the whole testsuite. It would also help with @dmvolod  suggestion of separating tests into core and non-core #723

rigazilla:
> > I think this will need some changes on the testsuite, since it works on a single namespace atm. Let me see...
> 
> I was thinking it would be best in it's own file, similar to `backup_restore_test.go`. That way we don't have to worry about adapting the whole testsuite. It would also help with @dmvolod suggestion of separating tests into core and non-core #723

Since testing multiple namespaces requires a different setup procedure, we needs a new test package.

dmvolod:
> Since testing multiple namespaces requires a different setup procedure, we needs a new test package.

@rigazilla what do you think about using docker image build and `kind load docker-image` implementing real operator deployment for this kind of test?

rigazilla:
> > Since testing multiple namespaces requires a different setup procedure, we needs a new test package.
> 
> @rigazilla what do you think about using docker image build and `kind load docker-image` implementing real operator deployment for this kind of test? In this solution it will be easier to make validation of roles and permission

Sounds like a good idea, but I would treat this as a separate issue, also I would wait to see the impact of travis migration before adding new tools to the pipeline.

.travis.yml
```diff
@@ -51,6 +51,10 @@ jobs:
       script:
         - go mod vendor
         - RUN_SA_OPERATOR=TRUE make test PARALLEL_COUNT=2
+          # Run multinamespace tests as admin
+          # skip role settings check for now
+        - kubectl config use-context $TESTING_CONTEXT
+        - make multinamespace-test PARALLEL_COUNT=2
```
dmvolod, line 57:
```suggestion
        - make multinamespace-test
```
PARALLEL_COUNT is not needed for this test, no t.Parallel() or t.Run(...) calls there. But can be keep as not adds additional problems

dmvolod (approved):

pkg/launcher/launcher.go
ryanemerson: So is it not possible to have operator metrics when using multi-namespace? Not a blocker obviously, but that doesn't sound ideal.

test/e2e/multinamespace/multinamespace_test.go
```diff
@@ -0,0 +1,142 @@
...
+	namespaces := strings.Split(nsAsString, ",")
+	if "TRUE" == tconst.RunLocalOperator {
+		for _, namespace := range namespaces {
+			testKube.DeleteNamespace(namespace)
+		}
+		testKube.DeleteCRD("infinispans.infinispan.org")
+		testKube.DeleteCRD("caches.infinispan.org")
+		testKube.DeleteCRD("backup.infinispan.org")
+		testKube.DeleteCRD("restore.infinispan.org")
+		for _, namespace := range namespaces {
```
ryanemerson, line 52: Can't we do the deletion and creation of namespaces in the same loop?

test/e2e/multinamespace/multinamespace_test.go
```diff
@@ -0,0 +1,142 @@
...
+		wg.Add(1)
+		go func() {
+			waitForPodsOrFail(spec, 1)
+			wg.Done()
+		}()
+	}
+	wg.Wait()
+}
+
+func waitForPodsOrFail(spec *ispnv1.Infinispan, num int) {
```
ryanemerson, line 86: All of the functions below this appear to be the same as in `main_test.go`. We should centralise these in a single .go file and import them in `main_test.go` and `multinamespace_test.go`.

ryanemerson (changes requested):

test/e2e/multinamespace/multinamespace_test.go
rigazilla: I just wanted to keep cleanup separated from setup.

test/e2e/multinamespace/multinamespace_test.go
rigazilla: I tried to centralize those methods in one of the utils package but I got problems with cycling deps.
But yeah probably I can leave them in the test package in separated file. let me try

rigazilla: @ryanemerson methods also rely on package variable `testKube` so I need to change method signature and all the calls around.
Since it's not as simple as moving methods, I would prefer to refact and resolve the cyclic deps in a separate pr.

pkg/launcher/launcher.go
rigazilla: @ryanemerson I've checked with a manual deploy: current implementation is ok, metrics service is deployed in the namespace where the operator is running

PR #699, opened 2020-12-18
Fix update regression from #684

dmvolod:


rigazilla:
merged thanks @dmvolod !

Issue #702, opened 2020-12-18
Able to setup XSite for different namespaces


dmvolod:
I don't know what is the limitation to have setup XSite sites with different namespaces just adding 'namespace` as optional parameter into the `InfinispanSiteLocationSpec` structure. It wold be flexible configuration as much as possible

Final proposal:
- add `namespace` and `clusterName` options to the `InfinispanSiteLocationSpec` struct
- if `InfinispanSiteLocationSpec.namespace` not defined, namespace is the same as current cluster namespace
- if `InfinispanSiteLocationSpec.clusterName` not defined, name is the same as current cluster name

pruivo:
It makes sense to me, at least for demos / POC / testing.
It would be nice to have the cluster name (not sure if it is the correct name, but I mean the `metadata.name`)  as well.

dmvolod:
> It would be nice to have the cluster name (not sure if it is the correct name, but I mean the `metadata.name`) as well.

@pruivo how we can utilize this name in Cross-Site setup?

pruivo:
using your PR (https://github.com/infinispan/infinispan-operator/pull/706) as an example:

```
    sites:
      local:
        name: SiteB
        expose:
          type: None
      locations:
        - name: SiteA
          url: <openshift REST endoint>
          project-name: example-clustera
        - name: SiteB
          url: <openshift REST endoint>
          project-name: example-clusterb
```

The operator would query opensift to get the cross-site service IP.

dmvolod:
Oh, yes. Infinispan CR name can be also different

PR #706, opened 2020-12-23
Able to configure xsite inside single cluster

### LINKED ISSUE #635: Able to configure xsite inside single cluster
This is doesn't make sense for production while ocp/k8s cluster placed in single DC, but make sense for POC's to demonstrate xsite ocp/k8s functions in single cluster or for production cross DC replicated clusters to avoid split-brain for the Infinispan cluster internode  communication.
We should an add additional processing if `spec.service.sites.local.expose.type` is not specified and just create a general service for inter-cluster communication.

dmvolod:
Fixes #635


ryanemerson:
@dmvolod I have missed the context for this change, when is this behaviour required? Do we have a GH issue for this?

dmvolod:
> @dmvolod I have missed the context for this change, when is this behaviour required? Do we have a GH issue for this?

Oh, @ryanemerson . Missed GH action in commit. Paste inside comment on top of the PR.

ryanemerson (approved):

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -656,6 +656,7 @@ spec:
                   - NodePort
                   - LoadBalancer
                   - Route
+                  - None
```
oraNod, line 659: @dmvolod Would "Internal" be more meaningful than "None"?

oraNod (commented):
I'll wait until Pedro Ruivo gives his review, as he is the master of cross-site. IMHO I wouldn't like to see another configuration option. For POC I should think the manual xsite config should be OK. 

Reading the issue I have a question about "production cross DC replicated clusters to avoid split-brain for the Infinispan cluster internode communication" - Maybe I still have holiday brain but how does this change avoid the split-brain case? Can you please explain this a little more?

deploy/crds/infinispan.org_infinispans_crd.yaml
dmvolod: Yes, correct @oraNod I was looking for more relevant meaning.
Probably, we need to split expose types for xsite and cluster services, as some values are not allowed for xiste, some for service and these setup could confuse customers.

oraNod: Isn't there already a default "internal" service that the Operator creates? Could that be used to achieve the same functionality that this PR offers?

dmvolod: Yes and no, the ports and names of these services are different. We could try to utilize them but this adds additional complexity.
Another option is skip expose type for internal service. In this `case spec.expose.type` need to make optional.

oraNod: Yeah. Maybe I need more time for my brain to kick in after PTO but either way it seems like the changes in this PR introduce complexity. Is the functionality gained worth it? Seems like we already have the manual xsite config. Could you add the internal IP of the cluster and get the same?

dmvolod: > Yeah. Maybe I need more time for my brain to kick in after PTO but either way it seems like the changes in this PR introduce complexity. Is the functionality gained worth it? Seems like we already have the manual xsite config. Could you add the internal IP of the cluster and get the same?

Sometimes yes, sometimes no. For example on AWS or Azure you don't know about node IP's if you don't control whole env explicitly. This gives much more flexibility to customers.

pruivo:
@dmvolod to make it clear for me, this is for cross-site between 2 clusters in the same openshift namespace right?

deploy/crds/infinispan.org_infinispans_crd.yaml
pruivo: I would avoid "None" since it seems that nothing is exposed. 
I probably choose "ClusterIP" too keep it consistent with openshift names, because we already have:
```
ispnv1.ExposeTypeNodePort => corev1.ServiceTypeNodePort
ispnv1.ExposeTypeLoadBalancer => corev1.ServiceTypeLoadBalancer
```

deploy/cr/xsite/static-internal/xsite_a.yaml
```diff
@@ -0,0 +1,20 @@
...
+  service:
+    type: DataGrid
+    sites:
+      local:
+        name: SiteA
+        expose:
+          type: None
+      locations:
+      - name: SiteA
+        host: example-clustera-site
```
pruivo, line 18: where host `example-clustera-site` comes from? It looks like `metadata.name` + `-site`. Will it resolve to some IP?

dmvolod:
> @dmvolod to make it clear for me, this is for cross-site between 2 clusters in the same openshift namespace right?

@pruivo if network policy allows service communication between namespaces, it's also possible to setup cross-site between clusters in separate namespaces

Could you please also look at this open issue to provide feedback about adding `namespace` parameter and able to setup cross-site between clusters in separate k8s/ocp and separate namespaces #702

deploy/cr/xsite/static-internal/xsite_a.yaml
dmvolod: This is the k8s Service name resolved to internal IP.
If we will setup cross-site between separate namespaces, the service will include namespace, i.e. `example-clustera-site.namespace-a.svc.cluster.local`

deploy/crds/infinispan.org_infinispans_crd.yaml
dmvolod: > I probably choose "ClusterIP" too keep it consistent with openshift names, because we already have:

Good, make sense. I will try to think about splitting expose.type for Cross-Site and REST/HR endpoints, as Router doesn't make sense for Cross-Site, but ClusterIP service doesn't make sense for REST/HR and always created.

pruivo: `ClusterIP` for REST/HR could make sense if the application is running in the same namespace, like EAP session externalization with RHDG :) 
I agree to keep `ClusterIP` away from REST/HR

dmvolod: Yeah, agree.
Another reason of splitting that customers confusing Route option for cross-site which is not possible

pruivo (approved):

ryanemerson (approved):

oraNod (approved):
@dmvolod Please take a look at the commit I added and let me know if we need any changes. TBH I don't have time to fully test this so I'm kind of doing a bit of guesswork.

dmvolod:
> @dmvolod Please take a look at the commit I added and let me know if we need any changes. TBH I don't have time to fully test this so I'm kind of doing a bit of guesswork.

Thanks, @oraNod . LGTM, it's enough for basic internal xsite cluster setup.

ryanemerson:
@dmvolod Can you squash the commits and then I think this is ready to merge. @rigazilla any thoughts?

dmvolod:
@ryanemerson squashing is available on GitHub now. You can click to the merge button down and select Squash&Merge option

PR #707, opened 2020-12-23
GetEndpointScheme() should return lowercase string

dmvolod:


pkg/apis/infinispan/v1/types_util.go
```diff
@@ -205,8 +205,16 @@ func (ispn *Infinispan) GetSiteServiceName() string {
...
+func (ispn *Infinispan) GetEndpointScheme() string{
 	if ispn.IsEncryptionCertSourceDefined() {
+		return strings.ToLower(string(corev1.URISchemeHTTPS))
+	}
+	return strings.ToLower(string(corev1.URISchemeHTTP))
+}
+
+// GetEndpointSchemeURI returns the protocol URI scheme used by the Infinispan cluster
+func (ispn *Infinispan) GetEndpointSchemeURI() corev1.URIScheme {
+	if ispn.GetEndpointScheme() == strings.ToLower(string(corev1.URISchemeHTTPS)) {
```
ryanemerson, line 217: This can be simplified to:

```go
// GetEndpointScheme returns the protocol scheme used by the Infinispan cluster
func (ispn *Infinispan) GetEndpointScheme() string {
	return strings.ToLower(string(ispn.GetEndpointSchemeURI()))
}

// GetEndpointSchemeURI returns the protocol URI scheme used by the Infinispan cluster
func (ispn *Infinispan) GetEndpointSchemeURI() corev1.URIScheme {
	if ispn.IsEncryptionCertSourceDefined() {
		return corev1.URISchemeHTTPS
	}
	return corev1.URISchemeHTTP
}
```

ryanemerson: Also, how about `GetEndpointURIScheme`? That way we're reusing the convention established by the k8s corev1 package.

dmvolod: Yeah, sure. Thanks, @ryanemerson

ryanemerson:
Thanks @dmvolod

PR #697, opened 2020-12-17
[2.0.x] Port 482, 542, 595, 674, 536 and part of the 281

dmvolod:

Issue #694, opened 2020-12-16
Skipping tests in Travis CI for documentation and other not related files doesn't work properly


dmvolod:
Skipping tests in Travis CI for documentation and other not related files doesn't work properly
Now it's implemented as call `exit` inside `before_install` steps inside travis file and not working as expected since some Travis CI release

dmvolod:
Fixed

PR #709, opened 2020-12-23
Fix empty build checking

rigazilla:
Seems that the check on empty builds doesn't work any more on branch build.

This is a hack

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,15 @@ os: linux
 
 before_install:
   - |
-    set -e
-    git fetch origin $TRAVIS_BRANCH
-    git diff --name-only FETCH_HEAD | grep -qvE '(\.md$)|(^(doc|documentation|manual|olm|ci|deploy/cr/|deploy/olm-catalog|build/cekit|build/jenkins|build/minikube|test-integration|.gitignore))/' || {
-      echo "Only docs, manual steps, test-integration,CI and OLM files were updated, stopping build process."
-      travis_terminate 0
-      exit 1
-    }
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ] ; then
+       set -e
+       git fetch origin $TRAVIS_BRANCH
+       git diff --name-only FETCH_HEAD | grep -qvE '(\.md$)|(^(doc|documentation|manual|olm|ci|deploy/cr/|deploy/olm-catalog|build/cekit|build/jenkins|build/minikube|test-integration|.gitignore))/' || {
+         echo "Only docs, manual steps, test-integration,CI and OLM files were updated, stopping build process."
+         travis_terminate 0
+         exit 1
+        }
+    fi
 
 jobs:
   include:


.travis.yml
```diff
@@ -8,11 +8,13 @@ os: linux
...
-    git fetch origin $TRAVIS_BRANCH
-    git diff --name-only FETCH_HEAD | grep -qvE '(\.md$)|(^(doc|documentation|manual|olm|ci|deploy/cr/|deploy/olm-catalog|build/cekit|build/jenkins|build/minikube|test-integration|.gitignore))/' || {
-      echo "Only docs, manual steps, test-integration,CI and OLM files were updated, stopping build process."
-      exit
-    }
+    if [ "$TRAVIS_PULL_REQUEST" != "false" ] ; then
+       git fetch origin $TRAVIS_BRANCH
+       git diff --name-only FETCH_HEAD | grep -qvE '(\.md$)|(^(doc|documentation|manual|olm|ci|deploy/cr/|deploy/olm-catalog|build/cekit|build/jenkins|build/minikube|test-integration|.gitignore))/' || {
+         echo "Only docs, manual steps, test-integration,CI and OLM files were updated, stopping build process."
+         exit
```
dmvolod, line 15: @rigazilla unfortunately trick with exit inside `before_install` doesn't work correctly in Travis CI anymore. We should find another way for this case
I filled issue #694

rigazilla: Yep, that's why I've added the logic to exit only on PR branch.
Anyway I've noticed that the `exit` statement breaks the build only for the master branch.
See build on 2.0.x:
https://travis-ci.org/github/infinispan/infinispan-operator/jobs/753160871
don't know why, I'm looking into

rigazilla: @dmvolod  removing these 2 lines:
https://github.com/infinispan/infinispan-operator/blob/aa20bfbdcf360e9733718682b5846120ae76c901/.travis.yml#L21-L22
things go better.
They seems to make the `Build system information` fail somewhere. Do we need to specify the architecture?

dmvolod: Currently arm64 machines have much less provisioning time. Which errors occurs on this platform?

rigazilla: @dmvolod, I don't know exactly what happens, but the first step `Worker information` shows incomplete info and error (no version and unable to resolve localhost), also docker server info are missing. The overall result is that the `exit` statement hangs instead of exit.

dmvolod: @rigazilla oh, we could try to use another Linux distribution search for arm64 hack. Will try to find

dmvolod: @rigazilla seems to be hangs with exit fixed in #728

dmvolod (approved):

dmvolod:
I will port this PR as part of the #697

rigazilla:
thanks for the review @dmvolod. merged

PR #715, opened 2020-12-25
Port and use controllerutil.CreateOrPatch for Infinispan CR update

dmvolod:


pkg/kubernetes/controllerutil.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+		if err := c.Create(ctx, obj); err != nil {
+			return controllerutil.OperationResultNone, err
+		}
+		return controllerutil.OperationResultCreated, nil
+	}
+
+	if err := c.Get(ctx, key, obj); err != nil {
+		return controllerutil.OperationResultNone, err
+	}
```
rigazilla, line 53: This differs from the original version. Do we need to reload the object?

rigazilla (commented):
If I've understood correctly the PATCH method doesn't check if an object has been changed by someone else in the while. This seems a sensible change from the current implementation. Do we need to worry about it?

pkg/kubernetes/controllerutil.go
dmvolod: It seems to Get always reload object and than mutate it with function. I don't know why we need to reload object every time.

rigazilla: I can't understand, but I was probably unclear.
I mean why do we need an additional c.Get() call, which is not present in the original implementation ?
https://github.com/kubernetes-sigs/controller-runtime/blob/dea6dee6c3e1ae2a2c0555f37ab28266105a1206/pkg/controller/controllerutil/controllerutil.go#L249

dmvolod: Oh, @rigazilla, are you talking about our first c.Get inside Infinspan controller? The reason why we need it, it's an non standard NotFound handle to prevent Create it in case of delete

rigazilla: I'm not sure, I'm referring to [these lines of code](https://github.com/infinispan/infinispan-operator/blob/c7da3615fb724b0892c619f77187f5cbf04c5059/pkg/kubernetes/controllerutil.go#L51-L53) of the `CreateOrPatch` in this PR which introduces an additional c.Get() which is not present in [the original version](https://github.com/kubernetes-sigs/controller-runtime/blob/1d63e429f706bd9ad81a1e47642745ed20e76bd1/pkg/controller/controllerutil/controllerutil.go#L250)

dmvolod: Oh, @rigazilla thanks a lot
This was missing from some tests. I will fix and validate 2.0.x backport.also

dmvolod: Fixed. Please review

rigazilla (commented):
Code looks good to me, but I didn't get feedbacks to my comment above:
https://github.com/infinispan/infinispan-operator/pull/715#pullrequestreview-563369672

 Are we ok to change from UPDATE to PATCH, which doesn't provide the optimistic lock mechanism? (i.e. see [this post](https://blog.atomist.com/kubernetes-apply-replace-patch/))

dmvolod:
> Are we ok to change from UPDATE to PATCH, which doesn't provide the optimistic lock mechanism? (i.e. see [this post](https://blog.atomist.com/kubernetes-apply-replace-patch/))

@rigazilla I'm not sure, that we will have a problems for:
- `status` update as only one controller can do this currently
- `spec` update as our previous implementation did the same: reload data from CR, update it. Also only one controller update one Infinispan CR in current implemnation

rigazilla:
> > Are we ok to change from UPDATE to PATCH, which doesn't provide the optimistic lock mechanism? (i.e. see [this post](https://blog.atomist.com/kubernetes-apply-replace-patch/))
> 
> @rigazilla I'm not sure, that we will have a problems for:
> 
>     * `status` update as only one controller can do this currently
> 
>     * `spec` update as our previous implementation did the same: reload data from CR, update it. Also only one controller update one Infinispan CR in current implementation
> 
>     * also we have very short reconcile loop and could had the same in previous implementation

yeah @dmvolod, seems reasonable to me.

rigazilla (approved):

test/e2e/backup_restore_test.go
```diff
@@ -197,13 +197,13 @@ func assertNumEntries(cacheName, pod string, numEntries int, client ispnclient.H
...
-	if int(numRead) != numEntries {
-		panic(fmt.Sprintf("Expected %d cache entries but received %d", numEntries, numRead))
-	}
+		body, err := ioutil.ReadAll(rsp.Body)
+		tutils.ExpectNoError(err)
+		numRead, err := strconv.ParseInt(string(body), 10, 64)
+		tutils.ExpectNoError(err)
+		if int(numRead) != numEntries {
+			panic(fmt.Sprintf("Expected %d cache entries but received %d", numEntries, numRead))
+		}
```
ryanemerson, line 206: Unnecessary formatting changes

ryanemerson (commented):
Overall LGTM @dmvolod. Can you rebase on master and remove the Dev07 image commit so that CR1 is used. It should be good to merge then.

test/e2e/backup_restore_test.go
dmvolod: Yes, @ryanemerson but this was adding by IDE. Will revert.

ryanemerson:
Thanks @dmvolod

PR #725, opened 2021-01-10
Validate StatusCode in GetClusterMembers

dmvolod:


dmvolod:
Close, this will be fixed in #727

Issue #716, opened 2020-12-25
Infinispan cluster can't WellFormed using 12.0.0.CR1 image on k8s/TravisCI


dmvolod:
OCP works fine with 12.0.0.CR1 image
Rolling back TravisCI back to 12.0.0.Dev07 temporary

ryanemerson:
Travis has been failing consistently  on master since before CR1 was released. Is the issue really with the CR1 distribution?

dmvolod:
Yes, @ryanemerson . It could be failed sometimes, but now every build is failed
Fixing version to 12.0.0.Dev07 is resolving fails build. Stable failing become 12 days ago, when CR1 was released

ryanemerson:
Stack trace from failed cluster
```
11:59:41,220 INFO  (jgroups-7,test-backup-restore-data-grid-source-0-8893) [org.infinispan.CLUSTER] ISPN100001: Node test-backup-restore-data-grid-source-1-39638 left the cluster
11:59:41,231 WARN  (non-blocking-thread--p2-t2) [org.infinispan.topology.ClusterTopologyManagerImpl] ISPN000196: Failed to recover cluster state after the current node became the coordinator (or after merge), will retry org.infinispan.remoting.RemoteException: ISPN000217: Received exception from test-backup-restore-data-grid-source-0-8893, see cause for remote stack trace
    at org.infinispan.remoting.transport.ResponseCollectors.wrapRemoteException(ResponseCollectors.java:25)
    at org.infinispan.topology.ClusterTopologyManagerImpl$CacheStatusResponseCollector.addException(ClusterTopologyManagerImpl.java:363)
    at org.infinispan.topology.ClusterTopologyManagerImpl$CacheStatusResponseCollector.addException(ClusterTopologyManagerImpl.java:317)
    at org.infinispan.remoting.transport.ValidResponseCollector.addResponse(ValidResponseCollector.java:29)
    at org.infinispan.topology.TopologyManagementHelper.lambda$addLocalResult$2(TopologyManagementHelper.java:129)
    at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:930)
    at java.base/java.util.concurrent.CompletableFuture.uniHandleStage(CompletableFuture.java:946)
    at java.base/java.util.concurrent.CompletableFuture.handle(CompletableFuture.java:2266)
    at java.base/java.util.concurrent.CompletableFuture.handle(CompletableFuture.java:143)
[...]

PR #727, opened 2021-01-11
#716 Utilise digest authentication in curl commands

ryanemerson:
It's no longer possible to utilise BASIC authentication in our curl commands as the image only allows DIGEST authentication by default. Prior to CR1 this was not respected by the server, however this has been fixed by [ISPN-12576](https://issues.redhat.com/browse/ISPN-12576), therefore it's necessary to utilise digest authentication in our curl commands.

### DIFF
--- a/pkg/infinispan/client/http/curl/curl.go
+++ b/pkg/infinispan/client/http/curl/curl.go
@@ -51,7 +51,7 @@ func (c *CurlClient) Post(podName, path, payload string, headers map[string]stri
 func (c *CurlClient) executeCurlCommand(podName string, path string, headers map[string]string, args ...string) (*http.Response, error, string) {
 	httpURL := fmt.Sprintf("%s://%s:%d/%s", c.config.Protocol, podName, consts.InfinispanPort, path)
 	user := fmt.Sprintf("-u %v:%v", c.credentials.Username, c.credentials.Password)
-	curl := fmt.Sprintf("curl -i --insecure --http1.1 %s %s %s %s", user, headerString(headers), strings.Join(args, " "), httpURL)
+	curl := fmt.Sprintf("curl -i --insecure --digest --http1.1 %s %s %s %s", user, headerString(headers), strings.Join(args, " "), httpURL)
 
 	execOut, execErr, err := c.Kubernetes.ExecWithOptions(
 		kube.ExecOptions{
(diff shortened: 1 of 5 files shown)


pkg/infinispan/client/http/curl/curl.go
```diff
@@ -51,7 +51,7 @@ func (c *CurlClient) Post(podName, path, payload string, headers map[string]stri
 func (c *CurlClient) executeCurlCommand(podName string, path string, headers map[string]string, args ...string) (*http.Response, error, string) {
 	httpURL := fmt.Sprintf("%s://%s:%d/%s", c.config.Protocol, podName, consts.InfinispanPort, path)
 	user := fmt.Sprintf("-u %v:%v", c.credentials.Username, c.credentials.Password)
-	curl := fmt.Sprintf("curl -i --insecure --http1.1 %s %s %s %s", user, headerString(headers), strings.Join(args, " "), httpURL)
+	curl := fmt.Sprintf("curl -i --insecure --digest --http1.1 %s %s %s %s", user, headerString(headers), strings.Join(args, " "), httpURL)
```
dmvolod, line 54: @ryanemerson is this working for 11.x images? I'm worries about upgrade case

pkg/infinispan/client/http/curl/curl.go
```diff
@@ -64,7 +64,17 @@ func (c *CurlClient) executeCurlCommand(podName string, path string, headers map
...
 	}
 
-	rsp, err := http.ReadResponse(bufio.NewReader(&execOut), nil)
+	reader := bufio.NewReader(&execOut)
+	rsp, err := http.ReadResponse(reader, nil)
+	if err != nil {
+		return nil, err, ""
+	}
+
+	if rsp.StatusCode != http.StatusUnauthorized {
```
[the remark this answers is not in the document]
ryanemerson: I think we should handle the no auth case on a separate path as the `curl` command will also be different. For example:

```go
func (c *CurlClient) executeCurlCommand(podName string, path string, headers map[string]string, args ...string) (*http.Response, error, string) {
	httpURL := fmt.Sprintf("%s://%s:%d/%s", c.config.Protocol, podName, consts.InfinispanPort, path)
        	httpURL := fmt.Sprintf("%s://%s:%d/%s", c.config.Protocol, podName, consts.InfinispanPort, path)
	if c.credentials != nil {
		c.executeDigest(httpURL, headers)
	} else {
		c.execute(httpURL, headers)
	}
}
```

pkg/infinispan/client/http/curl/curl.go
ryanemerson: Yep. Digest was supported and enabled by default in 11.x. I have manually ran the curl commands against the docker image and it works as expected.

I haven't tried the pre CR1 images, however they should work for the same reason. Regardless CR1 images were Dev releases so we provide no guarantees of compatibility.

ryanemerson:
I've added an additional commit which introduces a very basic http lib with digest support and updated our tests. The upshot of this is that it cleans up the code a bit so we don't have to pass user/pass and protocol strings between all of the methods.

dmvolod (approved):

rigazilla (approved):

rigazilla:
merged, thanks @ryanemerson !

ryanemerson:
Thanks for taking the time to diagnose why CR1 was failing @dmvolod, much appreciated.

PR #734, opened 2021-01-13
Doc title should include Infinispan Operator version

### LINKED ISSUE #724: Doc title should include Infinispan Operator version

dmvolod:


oraNod (approved):
@dmvolod Thanks, Dmitry. This looks good to me. I wonder though if it would make sense to include the corresponding Infinispan version here too. 

What if we insert a line below the title like "{ispn_operator} {ispn_operator_version} works with {brandname} Server {ispn_version}" 

Although I suppose it is possible to specify the image version in the CR for community so maybe that doesn't make sense. What do you think?

dmvolod:
> What if we insert a line below the title like "{ispn_operator} {ispn_operator_version} works with {brandname} Server {ispn_version}"
> 
> Although I suppose it is possible to specify the image version in the CR for community so maybe that doesn't make sense. What do you think?

Good idea, I thought about it, @oraNod but didn't find a correct place. Thanks, will add.

dmvolod:
@oraNod do we need to add version for the downstream product-attributes.adoc?

oraNod:
> @oraNod do we need to add version for the downstream product-attributes.adoc?

@dmvolod no, we can keep this in community only. the `operator.asciidoc` title file is not used downstream. 

it's also a little easier downstream because the corresponding DG version is pretty obvious in the docs. https://access.redhat.com/documentation/en-us/red_hat_data_grid/8.1/html/running_data_grid_on_openshift/index

documentation/asciidoc/titles/operator.asciidoc
```diff
@@ -14,9 +14,11 @@
 include::{topics}/attributes/community-attributes.adoc[]
 
 [id='infinispan_operator']
-= {ispn_operator} Guide
+= {ispn_operator} {ispn_operator_version} Guide
 
 The {ispn_operator} provides operational intelligence and reduces management
 complexity for deploying {brandname} on {k8s} and {openshift}.
 
+{ispn_operator} {ispn_operator_version} works with {brandname} Server {ispn_version}
```
oraNod, line 22:
```suggestion
{ispn_operator} {ispn_operator_version} corresponds to {brandname} {ispn_version}.
```

oraNod:
thanks @dmvolod

PR #737, opened 2021-01-13
Update Spec.Expose.Host for LoadBalancer expose type

dmvolod:


ryanemerson:
Code LGTM.

@dmvolod One general point about PRs/commits. There's no benefit of creating an issue and linking it in the commit if the issue doesn't have a description and the title is the same as the commit message. This is just adding additional steps to the process for the sake of it and you may as well just create a PR with a commit and no issue linked.

The ideal is to create an issue with a description explaining what the problem is and why a change is necessary, it's then possible when navigating the git history to return to the GH issue if additional information is required.

dmvolod:
Oh, yes, @ryanemerson sure.
Generally I providing description in the issue, but on this case thought that the title message is enough. Will write description in the issue.

ryanemerson:
Thanks for the PR and updating the issue @dmvolod

PR #738, opened 2021-01-13
Avoid empty xsite and other config fields if not defined and required

dmvolod:


pkg/controller/infinispan/resources/config/xsite.go
```diff
@@ -20,51 +20,51 @@ import (
...
 // ComputeXSite compute the xsite struct for cross site function
-func ComputeXSite(infinispan *ispnv1.Infinispan, kubernetes *kube.Kubernetes, service *corev1.Service, logger logr.Logger, xsite *config.XSite) error {
-	if infinispan.HasSites() {
-		siteServiceName := infinispan.GetSiteServiceName()
-		localSiteHost, localSitePort, err := getCrossSiteServiceHostPort(service, kubernetes, logger)
-		if err != nil {
-			logger.Error(err, "error retrieving local x-site service information")
-			return err
-		}
+func ComputeXSite(infinispan *ispnv1.Infinispan, kubernetes *kube.Kubernetes, service *corev1.Service, logger logr.Logger) (error, *config.XSite) {
```
ryanemerson, line 23: Can you update the return type to be `(*config.XSite, error)` as it's more consistent with how we have handled errors elsewhere.

pkg/controller/infinispan/resources/config/xsite.go
```diff
@@ -20,51 +20,51 @@ import (
...
+	siteServiceName := infinispan.GetSiteServiceName()
+	localSiteHost, localSitePort, err := getCrossSiteServiceHostPort(service, kubernetes, logger)
+	if err != nil {
+		logger.Error(err, "error retrieving local x-site service information")
+		return err, nil
```
ryanemerson, line 28: Is it necessary to log this here and then again in `Process()`?

I think it would be better to return an update error here, then only log once in `Process()`. E.g.:

```
	if err != nil {
		return fmt.Errorf("Error retrieving local x-site service information: %w", err), nil
	}
```

pkg/controller/infinispan/resources/config/xsite.go
```diff
@@ -20,51 +20,51 @@ import (
...
+	}
+
+	if localSiteHost == "" {
+		msg := "local x-site service host not yet available"
+		logger.Info(msg)
```
ryanemerson, line 33: Redudant info log as the error will be logged by `Process()`. Can just be:

```
return fmt.Errorf("Local x-site service host not yet available"), nil
```

pkg/controller/infinispan/resources/config/xsite.go
```diff
@@ -20,51 +20,51 @@ import (
...
+		return fmt.Errorf(msg), nil
+	}
+
+	logger.Info("local site service",
+		"service name", siteServiceName,
+		"host", localSiteHost,
+		"port", localSitePort,
+	)
```
ryanemerson, line 41: Shouldn't this be debug logging?

pkg/controller/infinispan/resources/config/xsite.go
```diff
@@ -20,51 +20,51 @@ import (
...
+		var err error
+		if remoteLocation.Host != "" {
+			err = appendBackupSite(&remoteLocation, xsite)
+		} else if remoteLocation.URL != "" { // lookup remote service via kubernetes api
+			err = appendRemoteLocation(infinispan, &remoteLocation, kubernetes, logger, xsite)
+		} else {
+			msg := fmt.Sprintf("invalid xsite location, remote name: %s does not define host or url", remoteLocation.Name)
 			logger.Info(msg)
-			return fmt.Errorf(msg)
+			err = fmt.Errorf(msg)
```
ryanemerson, line 59: Redundant info logging, we can just return an updated error instance and let `Process()` handle it.

pkg/controller/infinispan/resources/config/xsite.go
```diff
@@ -20,51 +20,51 @@ import (
...
+		Port:    localSitePort,
+	}
 
-		if localSiteHost == "" {
-			msg := "local x-site service host not yet available"
+	remoteLocations := findRemoteLocations(xsite.Name, infinispan)
+	for _, remoteLocation := range remoteLocations {
+		var err error
+		if remoteLocation.Host != "" {
+			err = appendBackupSite(&remoteLocation, xsite)
```
ryanemerson, line 53: `appendBackupSite` appears to always return nil.

pkg/controller/infinispan/resources/config/xsite.go
```diff
@@ -20,51 +20,51 @@ import (
...
-			}
+		if err != nil {
+			return err, nil
 		}
-
-		logger.Info("x-site configured", "configuration", xsite)
 	}
-	return nil
+
+	logger.Info("x-site configured", "configuration", xsite)
```
ryanemerson, line 66: I think this is more applicable as trace/debug logging.

ryanemerson (changes requested):

pkg/controller/infinispan/resources/config/xsite.go
dmvolod: Agreed, @ryanemerson 
Saw, that xsite and other components logging need to be improved, let's do this in separate issue you opened

pkg/controller/infinispan/resources/config/xsite.go
dmvolod: Thanks, will fix

pkg/controller/infinispan/resources/config/xsite.go
dmvolod: Yes, correct, will fix it

dmvolod:
@ryanemerson I fixed all overall issues not related to debug/log/info. Please review again.

rigazilla (approved):
lgtm. It ok by me to fix logs here or in a separate PR

dmvolod:
> lgtm. It ok by me to fix logs here or in a separate PR

It would be nice to make all operator changes in separate log refactoring PR

ryanemerson:
Thanks @dmvolod.

Issue #617, opened 2020-10-20
User must be allowed to disable encryption even if OCP/K8S provides certificates


rigazilla:
User must be able to start a cluster without TLS via Infinispan CR.

Proposed solution:

add a new value 'None' for spec.security.endpointEncryption.type, that disable TLS.

CR fragment example:
```
...
     spec:
        security:
           endpointEncryption:
                type: None
...
```

PR #742, opened 2021-01-14
Issue 617 disable encryption docs

oraNod:
Docs for #https://github.com/infinispan/infinispan-operator/issues/617

### DIFF
--- a/documentation/asciidoc/topics/proc_disabling_encryption.adoc
+++ b/documentation/asciidoc/topics/proc_disabling_encryption.adoc
@@ -1,0 +1,22 @@
+[id='disabling_encryption-{context}']
+= Disabling Encryption
+You can disable encryption so clients do not need TLS certificates to establish connections with {brandname}.
+
+[IMPORTANT]
+====
+{brandname} does not recommend disabling encryption in production environments where endpoints are exposed outside the {k8s} cluster via `spec.expose.type`.
+====
+
+.Procedure
+
+* Set `None` as the value for the `spec.security.endpointEncryption.type` field in your `Infinispan` CR and then apply the changes.
++
+[source,options="nowrap",subs=attributes+]
+----
+spec:
+  ...
+  security:
+    endpointEncryption:
+            type: None <1>
+----
+<1> Disables encryption for {brandname} endpoints.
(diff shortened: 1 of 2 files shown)


documentation/asciidoc/topics/proc_disabling_encryption.adoc
```diff
@@ -0,0 +1,18 @@
+[id='disabling_encryption-{context}']
+= Disabling Encryption
+You can disable encryption in test and development environments so clients do not need TLS certificates to establish connections with {brandname}.
+For production environments, you should always use encryption to secure connections and restrict access to data.
```
ryanemerson, line 4: I think this is a bit strong. We've added the ability to disable TLS specifically because people do want it in production envrionments. More specifically for when clients utilising the endpoints reside in the same k8s cluster as the Infinispan server(s).

ryanemerson: I think it might be worth adding a note saying something like:

"We do not recommend disabling encryption in production enviroments where the Infinispan endpoints are exposed outside of the k8s cluster via `spec.Expose.Type`"

oraNod: @ryanemerson thanks, dogg. updated and squashed.

ryanemerson (approved):

documentation/asciidoc/topics/proc_disabling_encryption.adoc
```diff
@@ -0,0 +1,22 @@
+[id='disabling_encryption-{context}']
+= Disabling Encryption
+You can disable encryption so clients do not need TLS certificates to establish connections with {brandname}.
```
rigazilla, line 3: maybe it worth to move this section near the `Openshift Service Certificates`? I mean: on plain k8s, connection is unencrypted by default unless certs are explicitly provided

oraNod: sounds good @rigazilla I'll update the docs now

rigazilla (approved):

ryanemerson:
Thanks @oraNod

Issue #726, opened 2021-01-11
Validate StatusCode for REST calls in ClusterInterface methods


dmvolod:
Following methods need to be validate expected `rsp.StatusCode` before decode content to avoid unexpected behavior
- CacheNames
- GetMetrics
- GetCacheManagerInfo
- GracefulShutdown

Special function need to be created to validate expected `StatusCode` values

ryanemerson:
@dmvolod Can you add some details to the description of the issue. The title on it's own is pretty vague.

dmvolod:
Fixed, close issue

PR #745, opened 2021-01-18
Implement dynamic logging management with REST API

### LINKED ISSUE #640: Implement dynamic logging management with REST API
The steps are:

- [x] Initial logging configuration should keep in the infinispan.yaml (this make possible to add logging without waiting server to be fully available)
- [x] Avoid infinispan.yaml update in case of the `spec.logging.categories` change
- [x] Invoke REST API while `spec.logging.categories` present. Compares loggers GET REST call and logging categories configuration inside Infinispan CR

dmvolod:
Fixes #640 #726

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -323,7 +323,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 
 	if infinispan.IsUpgradeNeeded(reqLogger) {
 		reqLogger.Info("Upgrade needed")
-		err = r.destroyResources(infinispan, reqLogger)
+		err = r.destroyResources(infinispan)
 		if err != nil {
 			reqLogger.Error(err, "failed to delete resources before upgrade")
 			return reconcile.Result{}, err
@@ -424,6 +424,10 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	}
 
 	// Below the code for a wellFormed cluster
+	err = configureLoggers(podList, cluster, infinispan)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
 
 	// Create default cache if it doesn't exists.
 	if infinispan.IsCache() {
@@ -445,6 +449,26 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	return reconcile.Result{}, err
 }
 
+func configureLoggers(pods *corev1.PodList, cluster ispn.ClusterInterface, infinispan *infinispanv1.Infinispan) error {
+	if infinispan.Spec.Logging != nil && len(infinispan.Spec.Logging.Categories) > 0 {
+		for _, pod := range pods.Items {
+			serverLoggers, err := cluster.GetLoggers(pod.Name)
+			if err != nil {
+				return err
+			}
+			for category, level := range infinispan.Spec.Logging.Categories {
+				serverLevel, ok := serverLoggers[category]
+				if !(ok && strings.ToLower(string(level)) == strings.ToLower(serverLevel)) {
+					if err := cluster.SetLogger(pod.Name, category, string(level)); err != nil {
+						return err
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
 // LookupResource lookup for resource to be created by separate resource controller
 func LookupResource(name, namespace string, resource runtime.Object, client client.Client, logger logr.Logger) (*reconcile.Result, error) {
 	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: name}, resource)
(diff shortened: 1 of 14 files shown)


pkg/infinispan/cluster.go
```diff
@@ -275,3 +241,58 @@ func (c Cluster) GetCacheManagerInfo(cacheManagerName, podName string) (map[stri
...
+	}
+	return nil
+}
+
+func validateResponse(rsp *http.Response, reason string, err error, entity string, validCodes...int) error {
+	if err != nil {
+		return fmt.Errorf("unexpected error %s, stderr: %v, err: %v", entity, reason, err)
+	}
+
+	if len(validCodes) > 0 && rsp != nil {
```
ryanemerson, line 279: If we do:

```
	if (rsp == nil || len(validCodes) == 0) {
		return nil
	}
```

We can avoid additional nesting.

pkg/infinispan/cluster.go
```diff
@@ -275,3 +241,58 @@ func (c Cluster) GetCacheManagerInfo(cacheManagerName, podName string) (map[stri
...
+
+	defer rsp.Body.Close()
+	var loggers []Logger
+	if err := json.NewDecoder(rsp.Body).Decode(&loggers); err != nil {
+		return nil, fmt.Errorf("unable to decode: %v", err)
+	}
+	lm := make(map[string]string)
+	for _, logger := range loggers {
+		if logger.Name != "" {
+			lm[logger.Name] = logger.Level
```
ryanemerson, line 259: We can do `strings.ToLower(logger.Level)` here so that it's not necessary in the consuming methods.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -444,6 +448,26 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	return reconcile.Result{}, err
 }
 
+func configureLoggers(pods *corev1.PodList, cluster ispn.ClusterInterface, infinispan *infinispanv1.Infinispan) error {
+	if infinispan.Spec.Logging != nil && len(infinispan.Spec.Logging.Categories) > 0 {
```
ryanemerson, line 452: We can return early to avoid additional nesting:

```
	if infinispan.Spec.Logging == nil || len(infinispan.Spec.Logging.Categories) == 0 {
		return nil
	}
```

pkg/infinispan/cluster.go
```diff
@@ -50,6 +55,8 @@ type ClusterInterface interface {
 	CacheNames(podName string) ([]string, error)
 	GetMetrics(podName, postfix string) (*bytes.Buffer, error)
 	GetCacheManagerInfo(cacheManagerName, podName string) (map[string]interface{}, error)
+	GetLoggers(podName string) (map[string]string, error)
+	SetLogger(podName, loggerName, loggerLevel string) error
```
ryanemerson, line 59: Should we use `Logger` here for the parameters and return type?

dmvolod: Yes and no. It could be get additional a value, if pass/get objects are the same structure. But in our case we need to transform Logger to the plain text path in the request and back to the map[string]string in response

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, sure, will fix

pkg/infinispan/cluster.go
dmvolod: Yes, will fix

pkg/infinispan/cluster.go
dmvolod: Thanks, will fix

pkg/infinispan/cluster.go
```diff
@@ -275,3 +241,58 @@ func (c Cluster) GetCacheManagerInfo(cacheManagerName, podName string) (map[stri
...
+		for _, code := range validCodes {
+			if code == rsp.StatusCode {
+				return nil
+			}
+		}
+
+		defer rsp.Body.Close()
+		responseBody, err := ioutil.ReadAll(rsp.Body)
+		if err != nil {
+			fmt.Errorf("server side error %s. Unable to read response body, %v", entity, err)
```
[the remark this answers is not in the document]
dmvolod: Thanks, will fix

ryanemerson (approved):

rigazilla (approved):
Looks good. 
How about adding a setlogger/getlogger session in an existing test?

dmvolod:
> How about adding a setlogger/getlogger session in an existing test?

Yes, sure, thought about this but didn't decide what kind test to implement: real with getting pod logs or synthetic with rest API call.

I will plan to create a separate issue to consolidate all e2e tests we need to implement.

ryanemerson:
LGTM just needs a rebase and another CI run

ryanemerson:
Thanks @dmvolod

Issue #703, opened 2020-12-19
Rename "minikube" term to "kubernetes" in context of XSite configuration


dmvolod:
`minikube` is just a one of the many possible Kubernetes deployments and very specific term
It would be nice rename it to the `kubernetes` or `k8s` prefix in context of the XSite deployment configuration.
There may be issues with some of the current XSite deployments on k8s, but not sure that's so much.
It's better to align terminology ASAP

PR #768, opened 2021-01-27
Use xsite://host:port URI configuration for static cross-site config

dmvolod:
We should remove separate host and port fields inside spec.service.sites.locations and use xsite://host:port URI for the url field. This will be aligned code with original design v3 described in #108 that's we missed during #604 implementation and code review.

Current cross-site configuration

```
locations:
      - name: SiteA
        host: infinispan-sitea.myhost.com
        port: 7900
      - name: SiteB
        host: infinispan-siteb.myhost.com
        port: 7900
```

Planned cross-site configuration
```
locations:
      - name: SiteA
        url: xsite://infinispan-sitea.myhost.com:7900
      - name: SiteB
        url: xsite://infinispan-siteb.myhost.com:7900
```

This is much simple and flexible explicit configuration and we have a chance to fix this quickly before release.


deploy/operator-install.yaml
```diff
@@ -934,21 +934,18 @@ spec:
...
                             type: string
                           namespace:
                             type: string
-                          port:
-                            format: int32
-                            type: integer
                           secretName:
                             type: string
                           url:
+                            pattern: ^(kubernetes|openshift|xsite):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])(:[0-9]+)?$
```
rigazilla, line 944:
```suggestion
                            pattern: ^(minikube|openshift|xsite):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])(:[0-9]+)?$
```
`minikube` scheme is documented for kubernetes.
Do we really need a so strict checking for url?

dmvolod: No other values are possible. This will reduce number of customer mistakes.
Also proposed changes from `minikube` to `kuberntes` with issue #703 and PR #764

rigazilla: I would make this change backward compatible then, to avoid upgrade problem, can we accept both scheme for static conf?

dmvolod: Yeah, good idea to keep for backward compatibility. Will update both PR's
Now operator try to access remote site on every config change, but we would like to reflect ot only in case of locations URL or secret changes in the future

documentation/asciidoc/topics/ref_xsite_crd_automatic.adoc
```diff
@@ -16,9 +16,9 @@ ifndef::productized[]
 <4> Provides connection information for all backup locations.
 <5> Specifies a backup location that matches `.spec.service.sites.local.name`.
 <6> Specifies a backup location.
-//* Use `xsite://` if the backup location has a static hostname and port.
 * Use `minikube://` if the backup location is a Minikube instance.
```
pruivo, line 19: kubernetes instead of minikube?

documentation/asciidoc/topics/proc_configuring_sites_manually.adoc
```diff
@@ -1,16 +1,17 @@
 [id='configuring_sites-{context}']
-= Specifying {brandname} Cluster Hostnames for Cross-Site Replication
-Specify hostnames and ports for {brandname} clusters so they can establish connections and form cross-site views.
+= Specifying {brandname} Cluster static hosts and ports for Cross-Site Replication
```
oraNod, line 2:
```suggestion
= Specifying Static Hosts and Ports for {brandname} Clusters
```

documentation/asciidoc/topics/proc_configuring_sites_manually.adoc
```diff
@@ -1,16 +1,17 @@
...
 
-* Ensure that you have the correct hostnames and port numbers for each {brandname} cluster that you plan to configure as a backup location.
+* Ensure that you have the correct hosts and port numbers for each {brandname} cluster that you plan to configure as a backup location.
 
 .Procedure
 
 . Create an `Infinispan` CR for each {brandname} cluster.
 . Specify the name of the local site with `spec.service.sites.local.name`.
-. Provide the name, host, and port for each {brandname} cluster that acts as a backup location with `spec.service.sites.locations`.
+. Provide the name and static host and port URL for each {brandname} cluster that acts as a backup location with `spec.service.sites.locations`.
```
oraNod, line 13:
```suggestion
. Provide the name and static URL for each {brandname} cluster that acts as a backup location with `spec.service.sites.locations`.
```

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -751,21 +751,18 @@ spec:
...
                             type: string
                           namespace:
                             type: string
-                          port:
-                            format: int32
-                            type: integer
                           secretName:
                             type: string
                           url:
+                            pattern: ^(kubernetes|minikube|openshift|xsite):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])(:[0-9]+)?$
```
pruivo, line 761: `_` is no supported. is it intentional?
the following example won't match:
```
minikube://my_host.com
```

documentation/asciidoc/topics/proc_configuring_sites_manually.adoc
```diff
@@ -1,16 +1,17 @@
...
+. Port can be omitted and default `7900` value will be used.
```
oraNod, line 14:
```suggestion
```

oraNod: We say 7900 is the default in the reference section below. This line seems unnecessary imo.

documentation/asciidoc/topics/proc_configuring_xsite_within_clusters.adoc
```diff
@@ -7,7 +7,7 @@ For evaluation and demonstration purposes, you can configure {brandname} to back
 . Create an `Infinispan` CR for each {brandname} cluster.
 . Specify the name of the local site with `spec.service.sites.local.name`.
 . Set `ClusterIP` as the value of the `spec.service.sites.local.expose.type` field.
-. Provide the name and host for each {brandname} cluster that acts as a backup location with `spec.service.sites.locations`.
+. Provide the hostname in the URL for each {brandname} cluster that acts as a backup location with `spec.service.sites.locations`.
```
oraNod, line 10:
```suggestion
. Provide the {brandname} cluster hostname as the URL for each backup location with `spec.service.sites.locations`.
```

documentation/asciidoc/topics/proc_configuring_xsite_within_clusters.adoc
```diff
@@ -30,11 +30,11 @@ spec:
...
       locations:
         - name: SiteA
-          host: example-clustera-site <1>
+          url: xsite://example-clustera-site <1>
         - name: SiteB
-          host: example-clusterb-site
+          url: xsite://example-clusterb-site
 ----
-<1> The value of the `host` field is the {k8s} service name that resolves to an internal IP address.
+<1> The value of the hostname in the `URL` field is the {k8s} service name that resolves to an internal IP address.
```
oraNod, line 37:
```suggestion
<1> The value of the `url` field is the {k8s} service name that resolves to an internal IP address.
```

documentation/asciidoc/topics/ref_xsite_crd_manual.adoc
```diff
@@ -20,7 +20,6 @@ ifdef::productized[]
 endif::productized[]
 <4> Provides connection information for all backup locations.
 <5> Specifies a backup location that matches `.spec.service.sites.local.name`.
-<6> Specifies the hostname for the backup location.
-<7> Optionally specifies the port that the backup location uses for the JGroups RELAY2 protocol. The default is `7900`.
-<8> Logs error messages for the JGroups TCP protocol.
-<9> Logs fatal messages for the JGroups RELAY2 protocol.
+<6> Specifies the hostname and port URL for the backup location. The default port value is `7900`.
```
oraNod, line 23:
```suggestion
<6> Specifies the hostname and port URL for the backup location. The default port is `7900`.
```

oraNod (commented):
Thanks @dmvolod !

documentation/asciidoc/topics/ref_xsite_crd_automatic.adoc
dmvolod: Yes, as minikube just one of the many possible deployment of the k8s

deploy/crds/infinispan.org_infinispans_crd.yaml
dmvolod: Yes, thanks @pruivo
Will change regex for this case

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -91,11 +91,10 @@ type InfinispanSitesLocalSpec struct {
 
 type InfinispanSiteLocationSpec struct {
 	Name        string `json:"name"`
+	// +kubebuilder:validation:Pattern=`(^(kubernetes|minikube|openshift):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])*(:[0-9]+)+$)|(^(xsite):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])*(:[0-9]+)*$)`
+	URL         string `json:"url"`
 	Namespace   string `json:"namespace,optional,omitempty"`
 	ClusterName string `json:"clusterName,optional,omitempty"`
-	Host        string `json:"host,optional,omitempty"`
```
rigazilla, line 96: This also breaks backward compatibility. I would consider a transition phase: i.e. keep these fields and, if empty populate URL field with them.
wdyt?

documentation/asciidoc/topics/ref_xsite_crd_automatic.adoc
tristantarrant: why special-case minikube then ? Unless it requires special handling it should just fall under the `kubernetes` case.

tristantarrant:
I would prefer if we qualified the "xsite" scheme with infinispan. Using the URI rules, which allow the `+` character in schemes:
`infinispan+xsite://host:[port]`

pkg/apis/infinispan/v1/infinispan_types.go
dmvolod: Probably yes, but arguments to change this immediately are:
- Not yet productized in downstream
- Less use case for this options
- Less than 2 month in release on OperatorHub
- Brakes original design doc v3

ryanemerson: Was the release on OperatorHub marked as experimental/dev?

dmvolod: @ryanemerson we do not update stable and development channels for a long time.Latest stable version is 1.1.2 and development-preview is 2.0.4. This means that 2.1.x is not stable and can be changed.

rigazilla: No we don't have that distinction since we moved to x.y.z channel names.

We don't need to document the old use case any more, just provide support to avoid problems with upgrade

ryanemerson: Is there anyway on operatorhub to mark a stream as being `dev or experimental`? Even adding a README would suffice if there's nothing else.

This would make it so that the compatiblity guarantees with provide for users are explicit. In the same way that our use of `DevXX` tags vs `.Final` is.

> This means that 2.1.x is not stable and can be changed.

Where have we informed users of this?

dmvolod: > Where have we informed users of this?

It seems to no information and we don't support named channels.
At the end, we need some rules what we can brake in the changes and what not. Probably short release and not downstream can be changed

deploy/operator-install.yaml
dmvolod: Also upgrade fails now, it's time to add some breaking changes I think :+1:

pkg/apis/infinispan/v1/infinispan_types.go
ryanemerson: > At the end, we need some rules what we can brake in the changes and what not.

Totally agree. We need this flexibility for productive development, however we must communicate this with upstream users some how.


> Probably short release and not downstream can be changed

I'm not sure what you mean by "short release".

Channels which are used downstream have to have backwards compatibility like you say. But so does any upstream channel that is beyond a `x-y-0.Final` tag. 

Where I think we have failed with the `2.1.x` tag, is on operatorhub etc we have not documented to the user when the dev cycle ended and when the stable version of that channel started. It appears as if it's one continous stream of releases, which will cause confusion to a user if OLM upgrades their clusters now and it breaks due to incompatible CR spec changes.

I'm not sure why the "dev-preview" channel was abandoned as I wasn't focusing on the operator then. Why can't we always use the "dev-preview" channel for `master` releases and then only release to the `x.y.z` channels after the first `.Final` tag once things are stable?

dmvolod: > I'm not sure what you mean by "short release".

Sorry, minor of course.
Yes, agree with you. I think the serious brakes are not possible between `x.y.z.Final` releases

Crumby: We really shouldn't remove anything from V1 of the CRD. It's supposed to be stable. Operator channel/CSV version isn't only thing that's ensuring in the stack providing availability of a particular feature. (If you touch CRD you mess up with the users of the older versions. If you touch the code only, the CRD will imply that the functionality is there).

dmvolod: Yeah, @Crumby totally agreed. But this changes was introduced in intermediate release and not stable. Not included in downstream also. We should prevent to make changes like this, but missing on review stage comparing with design document

documentation/asciidoc/topics/ref_xsite_crd_automatic.adoc
dmvolod: Yes, @tristantarrant will change it to `kubernetes` in sepearte PR #764

pkg/apis/infinispan/v1/infinispan_types.go
ryanemerson: As discussed on Zulip, we should initially keep port+url and transform the CR when users use this field in the next upstream release. The `host+port` fields will then be removed on the subsequent upstream release to ensure that it does not affect downstream.

dmvolod: Yes, sure, @ryanemerson 
Kept host and port fileds and add URL generator on upgrade step

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -760,18 +760,22 @@ spec:
...
                           port:
+                            description: Deprecated and to be removed on subsequent
+                              release. Use .URL with infinispan+xsite schema instead.
                             format: int32
                             type: integer
                           secretName:
                             type: string
                           url:
-                            pattern: ^(kubernetes|minikube|openshift):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])(:[0-9]+)?$
+                            pattern: (^(kubernetes|minikube|openshift):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])*(:[0-9]+)+$)|(^(infinispan\+xsite):\/\/(([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9\-]*[a-z0-9])*(:[0-9]+)*$)
```
pruivo, line 778: can't we replace the [a-z0-9] with `\w`?
the infinispan+xsite pattern should end with `(:[0-9]+)?` and not `(:[0-9]+)*`

pruivo: I've been playing around and I think it can be simplified like this
`(^(kubernetes|minikube|openshift|infinispan\+xsite):\/\/(\w|\w[\w\-\.]*\w)(:[0-9]+)?$)`
See https://regex101.com/r/YSCQe4/1

dmvolod: It seems to be kubernetes://111 is valid, but this is not valid address
This complex regex validates correct IP address as well as hostname, also `kubernetes` must contains port, but `infinispan+xsite` not

dmvolod: Got it from this library collection https://regexr.com/
It contains more and less complex solutions

rigazilla:
should we go on with this? seems backward compatible now

ryanemerson (approved):

pruivo (approved):

rigazilla (approved):

rigazilla:
@oraNod if docs are ok I would go on and merge this

oraNod (approved):
@dmvolod and @rigazilla Sorry for the delayed response. I missed the message in my inbox yesterday. LGTM. Please go ahead and merge.

rigazilla:
merged, thanks @dmvolod, @oraNod  and all !

PR #769, opened 2021-01-27
Fix operator upgrade issue for #643

### LINKED ISSUE #643: Remove duplicate name from the cluster PVC
PVC name create in StatefulSet VolumeClaimTemplates inherit StatefulSet name and add this name to the volume again
```
18:09 $ oc get pvc
NAME                                      STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
example-infinispan-backup                 Bound    pvc-ba7279c9-d18c-460f-83e7-3e591d80016b   1Gi        RWO            gp2            5m4s
example-infinispan-example-infinispan-0   Bound    pvc-77166b4b-7edd-4205-b381-5408c90d5266   1Gi        RWO            gp2            9m7s
example-infinispan-example-infinispan-1   Bound    pvc-41ee4852-e2f2-44aa-b934-65dffa03a1a0   1Gi        RWO            gp2            8m5s  
```
Name should be static, not Infinispan CR name as its already present there

dmvolod:
fixes #643

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -636,6 +636,19 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 	reqLogger := log.WithValues("Request.Namespace", m.Namespace, "Request.Name", m.Name)
 	lsPod := PodLabels(m.Name)
 
+	pvcs := &corev1.PersistentVolumeClaimList{}
+	err := kubernetes.ResourcesList(m.Namespace, LabelsResource(m.Name, ""), pvcs)
+	if err != nil {
+		return nil, err
+	}
+	dataVolumeName := DataMountVolume
+	for _, pvc := range pvcs.Items {
+		if strings.HasPrefix(pvc.Name, fmt.Sprintf("%s-%s", m.Name, m.Name)) {
+			dataVolumeName = m.Name
+			break
+		}
+	}
+
 	memory := resource.MustParse(m.Spec.Container.Memory)
 	replicas := m.Spec.Replicas
 	dep := &appsv1.StatefulSet{
@@ -680,7 +693,7 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 							Name:      IdentitiesVolumeName,
 							MountPath: consts.ServerSecurityRoot,
 						}, {
-							Name:      DataMountVolume,
+							Name:      dataVolumeName,
 							MountPath: DataMountPath,
 						}},
 					}},
@@ -724,7 +737,7 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 		}
 
 		pvc := &corev1.PersistentVolumeClaim{ObjectMeta: metav1.ObjectMeta{
-			Name:      DataMountVolume,
+			Name:      dataVolumeName,
 			Namespace: m.Namespace,
 		},
 			Spec: corev1.PersistentVolumeClaimSpec{
@@ -750,11 +763,11 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 		}
 		dep.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{*pvc}
 
-		AddVolumeChmodInitContainer("data-chmod-pv", DataMountVolume, DataMountPath, &dep.Spec.Template.Spec)
+		AddVolumeChmodInitContainer("data-chmod-pv", dataVolumeName, DataMountPath, &dep.Spec.Template.Spec)
 	} else {
 		volumes := &dep.Spec.Template.Spec.Volumes
 		ephemeralVolume := corev1.Volume{
-			Name: DataMountVolume,
+			Name: dataVolumeName,
 			VolumeSource: corev1.VolumeSource{
 				EmptyDir: &corev1.EmptyDirVolumeSource{},
 			},
--- a/test/e2e/utils/kubernetes.go
+++ b/test/e2e/utils/kubernetes.go
@@ -91,7 +91,6 @@ func (k TestKubernetes) DeleteNamespace(namespace string) {
 	obj := &v1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      namespace,
-			Namespace: namespace,
 		},
 	}
 	err := k.Kubernetes.Client.Delete(context.TODO(), obj, DeleteOpts...)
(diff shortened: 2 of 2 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -636,6 +636,16 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 	reqLogger := log.WithValues("Request.Namespace", m.Namespace, "Request.Name", m.Name)
 	lsPod := PodLabels(m.Name)
 
+	pvcs := &corev1.PersistentVolumeClaimList{}
+	err := kubernetes.ResourcesList(m.Namespace, LabelsResource(m.Name, ""), pvcs)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	}
+	dataVolumeName := DataMountVolume
+	if len(pvcs.Items) > 0 && strings.HasPrefix(pvcs.Items[0].Name, fmt.Sprintf("%s-%s", m.Name, m.Name)) {
```
ryanemerson, line 645: This won't work if there are multiple PVCs in the namespace and the data PVC is not the first in the list.

IMO we should just revert the initial commit, as I don't think it solves an actual problem.

Prior to the commit the PVC is named `<CR name>-infinispan-<#pod>`. The duplication only occurs if the Infinispan CR is named "infinispan". For many users this won't be the case.

dmvolod: This is not duplication problem and can't be occurred as PVC Claim template construct name from the StatefulSet name and a volume name and pod id
```
18:09 $ oc get pvc
NAME                                      STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
example-infinispan-backup                 Bound    pvc-ba7279c9-d18c-460f-83e7-3e591d80016b   1Gi        RWO            gp2            5m4s
example-infinispan-example-infinispan-0   Bound    pvc-77166b4b-7edd-4205-b381-5408c90d5266   1Gi        RWO            gp2            9m7s
example-infinispan-example-infinispan-1   Bound    pvc-41ee4852-e2f2-44aa-b934-65dffa03a1a0   1Gi        RWO            gp2            8m5s  
```
The name is **example-infinispan-example-infinispan-0** - CR name twice plus Pod Id

ryanemerson: Argh my mistake. I didn't think about the pod name changing as well!

> This won't work if there are multiple PVCs in the namespace and the data PVC is not the first in the list.

Unless I'm missing something still, I think this is an issue with only checking `pvcs.Items[0]`

ryanemerson: Or does `LabelsResource` prevent this?

dmvolod: @ryanemerson during create/recreate StatefulSet process we should support following cases for the data volume:
- find previous created PVC for current Infinispan CR and validate that name template `cr-name`-`cr-name` - this is migration from the old operator version
- PVC not found or name not matched template - this is new StatefulSet creation or new operator versions with correct PVC template name 

We can't use labels only for this case, as `app` label contains just Infinispan CR name

I agree, that would be nice to write functional design like you did for Batch/Cache CR for all functional features before to avoid this situations in the future, but this is bit old implementation need to be fixed.

ryanemerson: > @ryanemerson during create/recreate StatefulSet process we should support following cases for the data volume:
> 
> * find previous created PVC for current Infinispan CR and validate that name template `cr-name`-`cr-name` - this is migration from the old operator version
> * PVC not found or name not matched template - this is new StatefulSet creation or new operator versions with correct PVC template name

:+1: 



> We can't use labels only for this case, as `app` label contains just Infinispan CR name

What I meant was does ` kubernetes.ResourcesList(m.Namespace, LabelsResource(m.Name, ""), pvcs)`, specifically the call to `LabelsResource(m.Name, ""),` prevent other PVCs in the namespace from being returned that would require us to iterate over the array instead of calling `[0]`. Looking more closely I believe it does as we don't apply the "infinispan_cr" label to the backup PVCs, so I think the current impl is good :slightly_smiling_face:

dmvolod:
We are hitting two cases in the Travis CI
- Job not terminated after first set of the test in case of the fail first one. @rigazilla could you please have a look at this case. It seems to be multinamespace brings this situation
- `Failed to connect to test-container-cpu-update-with-two-replicas-0 port 11222: Connection refused\n, err: command terminated with exit code 7"` seems to be Infinispan server rejects connections when Pods are ready

I will try to retest this PR

pkg/controller/infinispan/infinispan_controller.go
rigazilla: Should we ensure this assumption will remain true adding a test?

dmvolod: Yes, good idea, @rigazilla 
but upgrade test for Travis CI is not so easy impl, I thought about it, but need a time to do this
Now manual upgrade test form 2.0.x to 2.1.x fails, see discussion in Zulip

rigazilla: My idea was not strictly related to upgrade, I was thinking to a test that ensures that the resourceList will select the correct PVC also in the future. My concern is about future features that maybe add PVC, we forgot about this assumption and the code goes broken.

dmvolod: Oh, @rigazilla you means about upgrade with migration? Yes, correct, this will cause a brakes as name templates will not match with current situation. I will try to think about this, but, probably we should document all these assumptions inside the upgrade with migration issue as bullets for future implementation.

rigazilla: yeah @dmvolod, migration upgrade is a great example of what I mean

dmvolod: Correct, I think you will brake some naming conventions we have, during implementing migration upgrade. It's too difficult to predict all of them

ryanemerson: The potentially brittle part of this code is the call to `pvcs.Items[0]`. If we replace that with a loop of `pvcs.Items`, setting `dataVolumeName = m.Name` on the first match of `hasPrefix` then we should be good regardless of the number of pvcs returned, as there can only be one set of pvcs with that prefix and `infinispan_cr` labels.

dmvolod: Excellent point, @ryanemerson Will change.

ryanemerson (approved):
LGTM

rigazilla:
> We are hitting two cases in the Travis CI
> 
>     * Job not terminated after first sets of the test in case of the fail first one. @rigazilla could you please have a look at this case. It seems to be multinamespace brings this situation

I'm trying to catch up. @dmvolod is this still an issue?

dmvolod:
@rigazilla fixed this issue in #788
Please review

rigazilla (approved):

ryanemerson:
Thanks @dmvolod

PR #779, opened 2021-01-29
Remove CR1 image binding in test

dmvolod:
This was introduced in #727

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -50,7 +50,6 @@ var DefaultSpec = ispnv1.Infinispan{
 		},
 		Replicas: 1,
 		Expose:   exposeServiceSpec(),
-		Image:    pointer.StringPtr("quay.io/infinispan/server:12.0.0.CR1"),
 	},
 }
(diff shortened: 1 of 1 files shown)


ryanemerson:
Good spot @dmvolod. My mistake!

ryanemerson (approved):

dmvolod:
Ops, something going wrong with the latest image.
`Unexpected response code 400` in test

dmvolod:
@ryanemerson do you have an idea, why sometimes cache create via REST fails with HTTP 400 code

ryanemerson:
@dmvolod The 400 on the CI logs is very strange. The error is being thrown by the backup/restore tests, however those tests do not use the `DefaultSpec` var in `main_test.go` so they shouldn't be related.

I have pulled your change locally and I'm unable to reproduce.

dmvolod:
Thanks, @ryanemerson 
I will close this PR as also removed CR1 in #781 which never failed
Sims to be a problem with CI

Issue #791, opened 2021-02-02
Allow Endpoint Authentication to be disabled


ryanemerson:
It should be possible for endpoint authentication to be disabled via the Spec:

```yaml
spec:
  security:
    endpointAuthentication: true | false
```

dmvolod:
@ryanemerson wdyt about adding enable/disable Authentication option, not just disable on the start?

ryanemerson:
> wdyt about adding enable/disable Authentication option, not just disable on the start?

I'm not sure what you mean exactly. Do you mean enable/disable at runtime?

dmvolod:
> > wdyt about adding enable/disable Authentication option, not just disable on the start?
> 
> I'm not sure what you mean exactly. Do you mean enable/disable at runtime?

Yes

ryanemerson:
That's not possible with the current server design as authentication and security-realms need to be configured before the server starts.

dmvolod:
Yes, sure, we can change auth option in config and create/remove identities volume and start a StetafulSet rolling update for this

Issue #696, opened 2020-12-17
Authentication should be disabled via infinispan CR


rigazilla:
User should be allowed to disable authentication.
An additional field `infinispan.spec.security.authentication` should disable authentication if equals `none`.
Authentication must be enable by default.

PR #792, opened 2021-02-02
Allow Endpoint Authentication to be disabled #791

ryanemerson:
~~Don't merge until `12.0.1.Final` image is available.~~


dmvolod:
@ryanemerson secret controller should be skip Secret creation in this case

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -15,8 +15,9 @@ type InfinispanAuthInfo struct {
 
 // InfinispanSecurity info for the user application connection
 type InfinispanSecurity struct {
-	EndpointSecretName string              `json:"endpointSecretName,optional,omitempty"`
-	EndpointEncryption *EndpointEncryption `json:"endpointEncryption,optional,omitempty"`
+	EndpointAuthentication *bool               `json:"endpointAuthentication,optional,omitempty"`
```
dmvolod, line 18: It would be nice to have an enum for enable, disable or set another authentication type as described in #696 
This is will give flexibility with the future authentication options

ryanemerson: What other type of authentication options do you envisage being required that could be satisfied with a single field?

Bear in mind that once we move to the next CR version we will re-work `InfinispanSecuriy` anyway so that it's more flexible. For example, this was a potential option discussed on zulip:

```yaml
spec:
  security:
    authorization: ...
    authentication:
      enabled: true | false
      secretName: ...
    endpoints:
      encryption: ...
```

dmvolod: It can be an auth type like None, for disable and anything else for container embedded ocp/k8s.
Let's keep current proposal and change it in the next CRD version.

ryanemerson: Sounds good. I have removed `InfinispanAuthInfo` as it's a redundant struct.

test/e2e/backup-restore/backup_restore_test.go
```diff
@@ -32,16 +32,16 @@ func TestMain(m *testing.M) {
 }
 
 func TestBackupRestore(t *testing.T) {
-	t.Run(string(v1.ServiceTypeDataGrid), testBackupRestore(datagridService))
-	t.Run(string(v1.ServiceTypeCache), testBackupRestore(cacheService))
+	t.Run(string(v1.ServiceTypeDataGrid), testBackupRestore(datagridService, 2))
+	t.Run(string(v1.ServiceTypeDataGrid)+"NoAuth", testBackupRestore(datagridServiceNoAuth, 1))
```
dmvolod, line 36: Why we need extra Backup/Restore test without auth?

ryanemerson: It's to ensure that the zero capacity pod is configured correctly with/without identities and that the backup/restore controllers enable/disable auth as configured.

The `NoAuth` test only creates clusters with 1 pod, so it isn't as heavyweight as the normal backup tests. Similarly, I have only added the test for the DataGrid Service as the behaviour will be the same for the two service types.

dmvolod: Thanks, make sense.

rigazilla (approved):
lgtm!

ryanemerson:
@dmvolod Anything else to add? Otherwise I think we're good to merge

dmvolod (approved):

dmvolod:
Thanks @ryanemerson , merged.

Issue #784, opened 2021-02-01
CI test didn't stop execute tests if first step is fails


dmvolod:
There are two portions of tests are running on Travis CI now with
* RUN_SA_OPERATOR=TRUE make test PARALLEL_COUNT=2
* make multinamespace-test PARALLEL_COUNT=2

even if first test is failed, second continuing to execute instead of stop CI job execution with fail state 
This is default TravisCI behavior and should be changed https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase

dmvolod:
Close. Fixed for master and 2.0.x branches

PR #793, opened 2021-02-02
[2.0.x] Port Issue #784 and PR #798

dmvolod:


ryanemerson:
@dmvolod The #784 commit here seems to include a lot more changes then the equivalent in [master](https://github.com/infinispan/infinispan-operator/pull/788/commits/0e09f4aca1fe2d5d9a03ebbb8356659392029cfa). Why are these necessary?

Minor thing, but can you keep the commit names the same when you backport them. This makes it much easier to compare the git logs of the two branches and you still get a link back to the original issue due to the # in the message. Adding "port of #" helps in the PR description though :+1:

dmvolod:
Yes, @ryanemerson correct, also added to this PR commit related to default KUBECONFIG setup. The commit names are the same, but didn't add this to PR. Will rename PR. Will also try to create issues for all possible cases for better tracking.

dmvolod:
> The #784 commit here seems to include a lot more changes then the equivalent in [master](https://github.com/infinispan/infinispan-operator/pull/788/commits/0e09f4aca1fe2d5d9a03ebbb8356659392029cfa). Why are these necessary?

I was cherry-pick original commit and it was bring some unported changes. The 2.0.x branch is really old structure without global code refactoring. We should try to avoid make a big changes while porting just test infra and some default values

ryanemerson:
> I cherry-picked original commit and it brings some un-ported changes. The 2.0.x branch is really old structure without global code refactoring. We should try to avoid make a big changes while porting just test infra and some default values

The thing I don't understand is that the #784 commit is just a case of moving `export RUN_SA_OPERATOR=TRUE` to the `beforeScript` section and making it so the three commands are executed with `&&` ... I don't understand why this needs changes to anything other than just the `.travis.yml` file.

dmvolod:
> The thing I don't understand is that the #784 commit is just a case of moving `export RUN_SA_OPERATOR=TRUE` to the `beforeScript` section and making it so the three commands are executed with `&&` ... I don't understand why this needs changes to anything other than just the `.travis.yml` file as per the original commit.

Yes, correct, but I tried to align run-tests.sh scripts with master, removing separate run-multinamespace-tests.sh that's redundant
There are so many inconsistencies and temporary things in scripts, such as links to personal directories, clusterup links that's already retires and so on, and I'm trying to bring the project into a single form, which is not very easy...

ryanemerson:
> Yes, correct, but I tried to align run-tests.sh scripts with master, removing separate run-multinamespace-tests.sh that's redundant
> There are so many inconsistencies and temporary things in scripts, constants and build procedures, such as links to personal directories, clusterup links that's already retires and so on, and I'm trying to bring the project into a single form, which is not very easy...

All efforts which are appreciated :slightly_smiling_face:. If we try to make it so these changes are contained in an additional commit to the backported commit, I think it will make the intention of the changes clearer for fellow reviewers/maintainers as the reason for the changes can be explained in the commit message.

rigazilla (approved):

dmvolod:
@ryanemerson can we merge this or need to add something?

dmvolod:
Thanks @ryanemerson and @rigazilla for review.
Merged, this requires for back-porting #702

PR #795, opened 2021-02-03
[test-integration] Update with recent features/fixes

Crumby:
* Fixes expected `(W|w)ellFormed` condition
* Adds `None` type of encryption to the MinimalSetupIT
* Adds default AntiAffinity settings test
* Adds check that modified affinity configuration is propagated to StatefulSet

### DIFF
--- a/test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
+++ b/test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
@@ -68,7 +68,7 @@ public class Infinispan {
 
          List<Condition> conditions = infinispanObject.getStatus().getConditions();
          if (conditions != null) {
-            Condition wellFormed = conditions.stream().filter(c -> "wellFormed".equals(c.getType())).findFirst().orElse(null);
+            Condition wellFormed = conditions.stream().filter(c -> "WellFormed".equals(c.getType())).findFirst().orElse(null);
             return wellFormed != null && "True".equals(wellFormed.getStatus());
          } else {
             return false;
--- a/test-integration/operator-tests/src/test/java/org/infinispan/operator/AdvancedSetupBIT.java
+++ b/test-integration/operator-tests/src/test/java/org/infinispan/operator/AdvancedSetupBIT.java
@@ -3,9 +3,13 @@ package org.infinispan.operator;
 import java.io.IOException;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import io.fabric8.kubernetes.api.model.Pod;
 import org.apache.commons.lang3.RandomStringUtils;
 import org.apache.http.entity.ContentType;
 import org.assertj.core.api.Assertions;
@@ -33,12 +37,12 @@ import lombok.extern.slf4j.Slf4j;
  *
  * spec:
  *   autoscale:
- *     maxMemUsagePercent: 70
+ *     maxMemUsagePercent: 40
  *     maxReplicas: 5
- *     minMemUsagePercent: 30
+ *     minMemUsagePercent: 20
  *     minReplicas: 3
  *   expose:
- *     type: LoadBalancer
+ *     type: Route
  *   service:
  *     type: Cache
  *     replicationFactor: 3
@@ -60,6 +64,7 @@ class AdvancedSetupBIT {
    @BeforeAll
    static void deploy() throws Exception {
       appName = infinispan.getClusterName();
+      hostName = openShift.generateHostname(appName + "-external");
 
       infinispan.deploy();
       testServer.withSecret(appName + "-cert-secret").deploy();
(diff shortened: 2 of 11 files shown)


test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
```diff
@@ -68,7 +68,9 @@ public void waitFor() {
 
          List<Condition> conditions = infinispanObject.getStatus().getConditions();
          if (conditions != null) {
-            Condition wellFormed = conditions.stream().filter(c -> "wellFormed".equals(c.getType())).findFirst().orElse(null);
+            Condition wellFormed = conditions.stream()
+                    .filter(c -> "WellFormed".equals(c.getType()) || "wellFormed".equals(c.getType()))
```
dmvolod, line 72: @Crumby it would be nice to use equalsIgnoreCase for this to don't care about in the fiuture

test-integration/pom.xml
```diff
@@ -11,10 +11,10 @@
 
     <properties>
         <!-- Dependencies versions -->
-        <version.infinispan>12.0.0.Dev04</version.infinispan>
+        <version.infinispan>11.0.3.Final</version.infinispan>
```
dmvolod, line 14: Why Infinispan version need to be downgraded for master?

test-integration/test-server/pom.xml
```diff
@@ -11,7 +11,7 @@
     <packaging>war</packaging>
 
     <properties>
-        <version.infinispan.hotrod>12.0.0.Dev04</version.infinispan.hotrod>
+        <version.infinispan.hotrod>11.0.3.Final</version.infinispan.hotrod>
```
dmvolod, line 14: Can we consolidate all possible version properties inside parent pom?

ryanemerson, line 14: Shouldn't this be `12.0.0.Final` on `master`?

test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
Crumby: @dmvolod Actually I'm thinking of hardcoding only the `WellFormed` version. There are conventions that apply for the status and this is itself is way to discover a regression.

My original idea was to make the tests backward compatible but they contain new tests not valid for older releases and older version of the tests is available under corresponding repository tag. What do you think?

test-integration/pom.xml
Crumby: That's unintentional mistake created by working on tests for 2.0.x and checking out the directory to match. I'll fix that.

test-integration/test-server/pom.xml
Crumby: `test-server` module doesn't have parent as it's supposed to be build as a standalone WebServer application with rest triggered endpoints. The application is built in the OpenShift and the parent directory isn't loaded there.

test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
dmvolod: Good, @Crumby agreed

test-integration/test-server/pom.xml
dmvolod: Good, I see. It would be nice to keep versions in one place to avoid typos if possible.

Crumby:
@dmvolod @ryanemerson Updated. I've added several more changes on top of the original state above:
* Update Fabric8 and Xtf dependencies for next OCP version compatibility
* Change LoadBalancer to Route to enable run tests on environment where LoadBalancers aren't available
* Tune autoscaling setup a bit for quicker test execution

test-integration/operator-tests/src/test/java/org/infinispan/operator/AdvancedSetupBIT.java
```diff
@@ -75,11 +80,7 @@ static void deploy() throws Exception {
       log.info("Password: {}", pass);
 
       Https.doesUrlReturnOK("http://" + testServer.host() + "/ping").waitFor();
-
-      Service external = openShift.services().withName(appName + "-external").get();
-      hostName = external.getStatus().getLoadBalancer().getIngress().get(0).getHostname();
-
-      Https.doesUrlReturnCode("https://" + hostName + ":11222/console/welcome", 200).interval(TimeUnit.SECONDS, 10).waitFor();
+      Https.doesUrlReturnCode("https://" + hostName + "/console/welcome", 200).interval(TimeUnit.SECONDS, 10).waitFor();
```
dmvolod, line 83: It would be nice to test via unsecured health status rather than console availability.
`GET /rest/v2/cache-managers/default/health/status`

test-integration/operator-tests/src/test/java/org/infinispan/operator/MinimalSetupIT.java
```diff
@@ -81,7 +71,7 @@ static void deployInfinispanCluster() throws IOException {
       pass = developer.getPassword();
 
       Https.doesUrlReturnOK("http://" + testServerHost + "/ping").waitFor();
-      Https.doesUrlReturnCode("https://" + hostName, 200).waitFor();
+      Https.doesUrlReturnCode("http://" + hostName, 200).waitFor();
```
dmvolod, line 74: Test probe also should be a REST health status

test-integration/operator-tests/src/test/java/org/infinispan/operator/AdvancedSetupBIT.java
Crumby: That's already done above by checking `WellFormed` condition which should be equivalent [1]. This is more of a safeguard. Routes and mainly LoadBalancers can take some time to start redirecting traffic.

[1] https://github.com/infinispan/infinispan-operator/blob/master/test-integration/operator-tests/src/test/java/org/infinispan/operator/AdvancedSetupBIT.java#L68.

dmvolod (approved):

rigazilla:
merged, thanks!

PR #803, opened 2021-02-05
ISPN-12702 Create Grafana dashboards

wburns:
https://issues.redhat.com/browse/ISPN-12702

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,8 +7,9 @@ require (
 	github.com/go-openapi/spec v0.19.4
 	github.com/go-playground/validator/v10 v10.3.0
 	github.com/iancoleman/strcase v0.0.0-20190422225806-e506e3ef7365
+	github.com/integr8ly/grafana-operator/v3 v3.4.0
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
-	github.com/operator-framework/operator-sdk v0.15.2
+	github.com/operator-framework/operator-sdk v0.16.0
 	github.com/spf13/pflag v1.0.5
 	gopkg.in/yaml.v2 v2.2.4
 	k8s.io/api v0.0.0
--- a/pkg/infinispan/grafana/constants.go
+++ b/pkg/infinispan/grafana/constants.go
@@ -2,6 +2,9 @@ package grafana
 
 // Constants for a community Infinispan Grafana installation
 const (
-	ApplicationName = "infinispan"
-	MonitoringKey   = "middleware"
+	ApplicationNameVariable = "DASHBOARD_NAME"
+	MonitoringKeyVariable   = "DASHBOARD_MONITORING_KEY"
+
+	DefaultApplicationName = "infinispan"
+	DefaultMonitoringKey   = "middleware"
 )
--- a/pkg/infinispan/grafana/constants.go
+++ b/pkg/infinispan/grafana/constants.go
@@ -1,0 +1,7 @@
+package grafana
+
+// Constants for a community Infinispan Grafana installation
+const (
+	ApplicationName = "infinispan"
+	MonitoringKey   = "middleware"
+)
--- a/pkg/infinispan/grafana/dashboard_controller.go
+++ b/pkg/infinispan/grafana/dashboard_controller.go
@@ -3,14 +3,14 @@ package grafana
 import (
 	"context"
 
+	"github.com/infinispan/infinispan-operator/pkg/controller/constants"
 	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/discovery"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
 )
@@ -21,7 +21,7 @@ func AddToScheme(scheme *runtime.Scheme) error {
 	return grafanav1alpha1.AddToScheme(scheme)
 }
 
-func EnsureDashboardExported(context context.Context, namespace string, mgr manager.Manager) error {
+func Create(namespace string, mgr manager.Manager) error {
 	dc, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
 	if err != nil {
 		return err
@@ -33,52 +33,30 @@ func EnsureDashboardExported(context context.Context, namespace string, mgr mana
 		return nil
 	}
 
-	infinispanDashboard := &grafanav1alpha1.GrafanaDashboard{}
-
-	selector := client.ObjectKey{
-		Name:      ApplicationName,
-		Namespace: namespace,
-	}
-
 	// Using mgr.GetClient() does not work as we don't configure the cache for this scheme
 	controllerClient, err := client.New(mgr.GetConfig(), client.Options{})
 	if err != nil {
 		return err
 	}
 
-	log.Info("Checking if grafana dashboard already exists")
-	err = controllerClient.Get(context, selector, infinispanDashboard)
+	infinispanDashboard := dashboard(namespace)
 
-	if err == nil {
-		log.Info("Grafana dashboard already exists, not creating")
+	if _, err = controllerutil.CreateOrUpdate(context.Background(), controllerClient, infinispanDashboard, func() error {
 		return nil
-	}
-	// We verify that the error is not a NotFound - if it is NotFound we ignore so we can create the dashboard
-	if !meta.IsNoMatchError(err) && !errors.IsNotFound(err) {
-		log.Error(err, "Error encountered while retrieving grafana dashboard")
+	}); err != nil {
 		return err
 	}
 
-	infinispanDashboard = dashboard(namespace)
-
-	log.Info("Creating grafana dashboard in namespace %s\n", namespace)
-	err = controllerClient.Create(context, infinispanDashboard)
-	if err != nil {
-		log.Error(err, "Error encountered while creating grafana dashboard\n")
-		return err
-	}
-
-	log.Info("Successfully grafana dashboard")
 	return nil
 }
 
 func dashboard(namespace string) *grafanav1alpha1.GrafanaDashboard {
 	return &grafanav1alpha1.GrafanaDashboard{
 		ObjectMeta: v12.ObjectMeta{
-			Name:      ApplicationName,
+			Name:      constants.GetEnvWithDefault(ApplicationNameVariable, DefaultApplicationName),
 			Namespace: namespace,
 			Labels: map[string]string{
-				"monitoring-key": MonitoringKey,
+				"monitoring-key": constants.GetEnvWithDefault(MonitoringKeyVariable, DefaultMonitoringKey),
 				"app":            "grafana",
 			},
 		},
@@ -97,7 +75,7 @@ func dashboard(namespace string) *grafanav1alpha1.GrafanaDashboard {
 
 func dashboardSelector(namespace string) client.ObjectKey {
 	return client.ObjectKey{
-		Name:      ApplicationName,
+		Name:      constants.GetEnvWithDefault(ApplicationNameVariable, DefaultApplicationName),
 		Namespace: namespace,
 	}
 }
--- a/pkg/infinispan/grafana/dashboard_controller.go
+++ b/pkg/infinispan/grafana/dashboard_controller.go
@@ -4,8 +4,10 @@ import (
 	"context"
 
 	"github.com/infinispan/infinispan-operator/pkg/controller/constants"
+	"github.com/infinispan/infinispan-operator/pkg/controller/infinispan"
 	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
 	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	appsv1 "k8s.io/api/apps/v1"
 	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/discovery"
@@ -41,7 +43,27 @@ func Create(namespace string, mgr manager.Manager) error {
 
 	infinispanDashboard := dashboard(namespace)
 
+	operatorDeployment := &appsv1.Deployment{}
+	result, err := infinispan.LookupResource("infinispan-operator", namespace, operatorDeployment, controllerClient, log)
+
+	if err != nil {
+		log.Error(err, "There was a problem looking up the infinispan operator to add the dashboard as a secondary resource")
+		return err
+	}
+
+	if result != nil {
+		log.Info("Infinispan operator resource not initialized, not installing Grafana Dashboard")
+		return nil
+	}
+
+	// result and err should be analyzed
+	if err = controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, mgr.GetScheme()); err != nil {
+		log.Error(err, "Unable to set Grafana Dashboard controller reference to Infinispan operator - skipping creation")
+		return nil
+	}
+
 	if _, err = controllerutil.CreateOrUpdate(context.Background(), controllerClient, infinispanDashboard, func() error {
+		// Don't do any merge
 		return nil
 	}); err != nil {
 		return err
--- a/pkg/infinispan/grafana/dashboard_controller.go
+++ b/pkg/infinispan/grafana/dashboard_controller.go
@@ -1,0 +1,103 @@
+package grafana
+
+import (
+	"context"
+
+	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	v12 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/discovery"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("infinispan-grafana")
+
+func AddToScheme(scheme *runtime.Scheme) error {
+	return grafanav1alpha1.AddToScheme(scheme)
+}
+
+func EnsureDashboardExported(context context.Context, namespace string, mgr manager.Manager) error {
+	dc, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
+	if err != nil {
+		return err
+	}
+	// detect the GrafanaDashboard resource type resourceExists on the cluster
+	resourceExists, _ := k8sutil.ResourceExists(dc, grafanav1alpha1.SchemeGroupVersion.String(), grafanav1alpha1.GrafanaDashboardKind)
+	if !resourceExists {
+		log.Info("Grafana CRD not present - not installing dashboard CR")
+		return nil
+	}
+
+	infinispanDashboard := &grafanav1alpha1.GrafanaDashboard{}
+
+	selector := client.ObjectKey{
+		Name:      ApplicationName,
+		Namespace: namespace,
+	}
+
+	// Using mgr.GetClient() does not work as we don't configure the cache for this scheme
+	controllerClient, err := client.New(mgr.GetConfig(), client.Options{})
+	if err != nil {
+		return err
+	}
+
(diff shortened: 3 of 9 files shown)


wburns:
I have not added documentation, but what I have is sufficient for the operator. Let me know if I should do something different.

pkg/launcher/launcher.go
```diff
@@ -129,12 +130,24 @@ func Launch(params Parameters) {
...
+	}
+
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
 
+	// TODO: should I worry about the context here?
+	if err := grafana.EnsureDashboardExported(context.Background(), namespace, mgr); err != nil {
```
dmvolod, line 146: We didn't identify in what namespace we should create a Grafana CR during discussion at ZulipChat with @ryanemerson and @rigazilla . The `namespace` there is a watching namespace and can take following values
* single namespace - if operator installed and watching in one namespace
* multiple, comma separated values - if operator watching several namespaces
* empty - if operator watching entire cluster

In this case, we should install Grafana CR inside operatorNs or custom defined parameter namespace (via env var, for example). Also we should take care about potential Grafana CR orphan record, if operator will be uninstalled. The following scenario should be implemented: lookup Infinispan operator deployment itself the operator code and set it as controller with `controllerutil.SetControllerReference(...)`

go.mod
```diff
@@ -7,8 +7,9 @@ require (
 	github.com/go-openapi/spec v0.19.4
 	github.com/go-playground/validator/v10 v10.3.0
 	github.com/iancoleman/strcase v0.0.0-20190422225806-e506e3ef7365
+	github.com/integr8ly/grafana-operator/v3 v3.4.0
 	github.com/openshift/api v3.9.1-0.20190924102528-32369d4db2ad+incompatible
-	github.com/operator-framework/operator-sdk v0.15.2
+	github.com/operator-framework/operator-sdk v0.16.0
```
dmvolod, line 12: The Operator SDK should be updated in Makefile and than `make codegen` need to be run

oraNod:
> I have not added documentation, but what I have is sufficient for the operator. Let me know if I should do something different.

@wburns Feel free to add documentation in md, adoc, txt (whatever you feel like) to https://github.com/infinispan/infinispan-operator/tree/master/doc - I can work with that and incorporate it into the documentation. Of course you can add, edit to documentation as much as you like but the purpose of the `doc` folder is for you to be able to add docs without having to think about format or structure. Cheers.

pkg/launcher/launcher.go
```diff
@@ -129,12 +130,24 @@ func Launch(params Parameters) {
...
+		os.Exit(1)
+	}
+
 	// Setup all Controllers
 	if err := controller.AddToManager(mgr); err != nil {
 		log.Error(err, "")
 		os.Exit(1)
 	}
 
+	// TODO: should I worry about the context here?
```
ryanemerson, line 145: No need to worry about. Just use `ctx.Background()` in `EnsureDashboardExported` and remove the method parameter.

pkg/infinispan/grafana/constants.go
```diff
@@ -0,0 +1,7 @@
+package grafana
+
+// Constants for a community Infinispan Grafana installation
+const (
+	ApplicationName = "infinispan"
+	MonitoringKey   = "middleware"
+)
```
ryanemerson, line 7: Is this required in a separate `.go` file so that it can be easily replaced downstream?

pkg/infinispan/grafana/dashboard_controller.go
```diff
@@ -0,0 +1,103 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("infinispan-grafana")
+
+func AddToScheme(scheme *runtime.Scheme) error {
+	return grafanav1alpha1.AddToScheme(scheme)
+}
+
+func EnsureDashboardExported(context context.Context, namespace string, mgr manager.Manager) error {
```
ryanemerson, line 24: `CreateDashboard`?

ryanemerson: Or even just `Create`. That way you have `grafana.Create` when calling using the package name.

pkg/infinispan/grafana/dashboard_controller.go
```diff
@@ -0,0 +1,103 @@
...
+	log.Info("Creating grafana dashboard in namespace %s\n", namespace)
+	err = controllerClient.Create(context, infinispanDashboard)
+	if err != nil {
+		log.Error(err, "Error encountered while creating grafana dashboard\n")
+		return err
+	}
+
+	log.Info("Successfully grafana dashboard")
+	return nil
+}
```
ryanemerson, line 73: Following from  our Zulip conversation: `CreateOrPatch` can be used here. The "status" field that's expected in the struct is not an issue, as the `GrafanaDashboard` struct has the `Status GrafanaDashboardStatus` field. This is expected for CR structs and you will find it on most (all?) Kubernete CRs.

ryanemerson (changes requested):

pkg/infinispan/grafana/constants.go
rigazilla: Values that need downstream changes should be stored in the [csv](https://github.com/infinispan/infinispan-operator/blob/3c4554b970ae9b179f811e72c8e81028e51dcc33/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml#L134) file and get from the env (i.e. [here](https://github.com/infinispan/infinispan-operator/blob/3c4554b970ae9b179f811e72c8e81028e51dcc33/pkg/controller/constants/image.go#L26)).
All downstream customizations are done there atm.

pkg/launcher/launcher.go
wburns: > We didn't identify in what namespace we should create a Grafana CR during discussion at ZulipChat with @ryanemerson and @rigazilla . The `namespace` there is a watching namespace and can take following values
> 
> * single namespace - if operator installed and watching in one namespace
> * multiple, comma separated values - if operator watching several namespaces
> * empty - if operator watching entire cluster
> 
> In this case, we should install Grafana CR inside operatorNs or custom defined parameter namespace (via env var, for example).

Yes, I thought it was the namespace the operator was installed in. I was thinking of just using the namespace the operator was installed in though, since we are expecting 1 operator eventually and would map well with having 1 dashboard. I will poke around unless you can point me to that :)

> Also we should take care about potential Grafana CR orphan record, if operator will be uninstalled. The following scenario should be implemented: lookup Infinispan operator deployment itself the operator code and set it as controller with `controllerutil.SetControllerReference(...)`

Yes, I thought about this before but then forgot about implementing it. I will have to find where this is.

go.mod
wburns: Aha, didn't know this. Let me look into it.

pkg/launcher/launcher.go
wburns: Okay.

pkg/infinispan/grafana/constants.go
wburns: Alright, I will look at that. I just copied what keycloak had done.

pkg/infinispan/grafana/dashboard_controller.go
wburns: I wasn't able to use `CreateOrPatch` as the status field is a different type (complained about not being able to convert to uint64 iirc).

pkg/infinispan/grafana/dashboard_controller.go
wburns: Hrmm, okay

pkg/launcher/launcher.go
dmvolod: We have a function in the Infinispan controller code to lookup resources

pkg/infinispan/grafana/dashboard_controller.go
```diff
@@ -0,0 +1,103 @@
...
+	return grafanav1alpha1.AddToScheme(scheme)
+}
+
+func EnsureDashboardExported(context context.Context, namespace string, mgr manager.Manager) error {
+	dc, err := discovery.NewDiscoveryClientForConfig(mgr.GetConfig())
+	if err != nil {
+		return err
+	}
+	// detect the GrafanaDashboard resource type resourceExists on the cluster
+	resourceExists, _ := k8sutil.ResourceExists(dc, grafanav1alpha1.SchemeGroupVersion.String(), grafanav1alpha1.GrafanaDashboardKind)
```
dmvolod, line 30: Seems to be CRD present validation is not enough for correct working. We should ensure that Grafana operator watching expected namespace (where Grafana CR has created)

pkg/launcher/launcher.go
rigazilla: > Yes, I thought it was the namespace the operator was installed in. I was thinking of just using the namespace the operator was installed in though, since we are expecting 1 operator eventually and would map well with having 1 dashboard. I will poke around unless you can point me to that :)

I fear that install CR in the operator ns could be a problem for cluster scoped operator: this kind of operator runs in `openshift-operators` ns and I'm not sure we're allowed to install CR there

pkg/infinispan/grafana/dashboard_controller.go
wburns: The error is the following
```
panic: cannot deep copy uint64

goroutine 1 [running]:
k8s.io/apimachinery/pkg/runtime.DeepCopyJSONValue(0x1440f00, 0x2099020, 0xc000aada38, 0xc0003976e0)
	/home/wburns/RedHat/external/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/runtime/converter.go:474 +0x4e6
k8s.io/apimachinery/pkg/runtime.DeepCopyJSONValue(0x149ab20, 0xc000397470, 0x1, 0x1)
	/home/wburns/RedHat/external/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/runtime/converter.go:458 +0x157
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.NestedFieldCopy(0xc0003960c0, 0xc000aadc08, 0x1, 0x1, 0x0, 0x0, 0xa, 0x17e4300, 0xc0002f2780)
	/home/wburns/RedHat/external/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20191004115801-a2eda9f80ab8/pkg/apis/meta/v1/unstructured/helpers.go:43 +0xa6
github.com/infinispan/infinispan-operator/pkg/kubernetes.CreateOrPatch(0x180f3a0, 0xc000044140, 0x181d620, 0xc0003fe340, 0x17e4300, 0xc0002f2780, 0x0, 0x0, 0x0, 0xc00091fcb8, ...)
	/home/wburns/RedHat/infinispan-operator/pkg/kubernetes/controllerutil.go:63 +0x4bf
github.com/infinispan/infinispan-operator/pkg/infinispan/grafana.Create(0xc0000424d0, 0xa, 0x18245e0, 0xc00048bc80, 0x0, 0x0)
	/home/wburns/RedHat/infinispan-operator/pkg/infinispan/grafana/dashboard_controller.go:46 +0x1c6
github.com/infinispan/infinispan-operator/pkg/launcher.Launch(0x0)
	/home/wburns/RedHat/infinispan-operator/pkg/launcher/launcher.go:145 +0x4f3
main.main()
	/home/wburns/RedHat/infinispan-operator/cmd/manager/main.go:8 +0x2a
Process exiting with code: 2 signal: false
```

dmvolod: > Which happens if the dashboard is already present when using CreateOrPatch. From the line number it is attempted to parse out the JSON and a type doesn't match the expected one.

You should provide different code for Create and Patch via mutate function to update only reflected sections or skip all updates

dmvolod: Not in this code, this is for CreateOrPatch which Ryan suggested

wburns: That is the code I am talking about if you look at the stack trace. I replaced my code with `CreateOrPatch` and I get the error and the mutate function is never referenced.

```
github.com/infinispan/infinispan-operator/pkg/kubernetes.CreateOrPatch(0x180f3a0, 0xc000044140, 0x181d620, 0xc0003fe340, 0x17e4300, 0xc0002f2780, 0x0, 0x0, 0x0, 0xc00091fcb8, ...)
	/home/wburns/RedHat/infinispan-operator/pkg/kubernetes/controllerutil.go:63 +0x4bf
```

dmvolod: @wburns could you please to share updated code

wburns: @dmvolod I added an additional commit that uses `CreateAndPatch`. It works great if the dashboard isn't present, but it if is then it blows up with the above stack trace.

wburns: Here is a gist of the output from the command line as well and log for the commit I am running.
https://gist.github.com/wburns/a632a622f02673f147085f91ff471054

wburns: Changing it to `CreateOrUpdate` seems to work.

pkg/infinispan/grafana/dashboard_controller.go
wburns: Does this really matter? I would say just creating if the CRD is present should be sufficient. Once I figure out how to do the `SetControllerReference` it will go away if the Infinispan operator is remove as well.

wburns:
Updated, but I still have to look into SetControllerReference to remove the CR when Infinispan operator is removed. Any pointers would be appreciated :)

dmvolod:
> Updated, but I still have to look into SetControllerReference to remove the CR when Infinispan operator is removed. Any pointers would be appreciated :)

The code should looks like, but didn't test it
```
operatorDeployment := &appsv1.Deployment{}
_, err = infinispan.LookupResource("infinispan-operator", namespace, operatorDeployment, controllerClient, log)
controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, mgr.GetScheme())
```

wburns:
I have not added documentation, but what I have is sufficient for the operator. Let me know if I should do something different.

> > Updated, but I still have to look into SetControllerReference to remove the CR when Infinispan operator is removed. Any pointers would be appreciated :)
> 
> The code should looks like, but didn't test it
> 
> ```
> operatorDeployment := &appsv1.Deployment{}
> result, err := infinispan.LookupResource("infinispan-operator", namespace, operatorDeployment, controllerClient, log)
> // result and err should be analyzed
> controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, mgr.GetScheme())
> ```

So I added this but when I stop the operator the resource isn't removed. Can you double check my code what I might be doing wrong?

pkg/infinispan/grafana/dashboard_controller.go
```diff
@@ -0,0 +1,94 @@
...
+	operatorDeployment := &appsv1.Deployment{}
+	_, err = infinispan.LookupResource("infinispan-operator", namespace, operatorDeployment, controllerClient, log)
+
+	if err != nil {
+		log.Error(err, "There was a problem looking up the infinispan operator to add the dashboard as a secondary resource")
+		return err
+	}
+
+	// result and err should be analyzed
+	controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, mgr.GetScheme())
```
dmvolod, line 61: @wburns this should be invoke before `controllerutil.CreateOrUpdate`

pkg/infinispan/grafana/dashboard_controller.go
```diff
@@ -0,0 +1,94 @@
...
+	infinispanDashboard := dashboard(namespace)
+
+	if _, err = controllerutil.CreateOrUpdate(context.Background(), controllerClient, infinispanDashboard, func() error {
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	operatorDeployment := &appsv1.Deployment{}
+	_, err = infinispan.LookupResource("infinispan-operator", namespace, operatorDeployment, controllerClient, log)
```
dmvolod, line 53: Both `result` and `err` should be validated. If they are `nil` this means that Deployment is found
Also please pay attention that this logic should work even if Deployment not found (local run and test where Deployment entity is not present)

wburns: Yeah, I didn't validate as I did it afterwards and didn't make much sense. So I will add the validation when I flip the ordering as it makes sense then ;)

wburns: Updated to do the check before hand and added in return value checking. Need to still verify it works when there is a deployment though.

wburns:
Updated to handle dashboard removal when deployment is removed. Had to add in roles to view and remove the dashboard to our operator. Also can confirm when not using a deployment it will not install the dashboard.

Only problem outstanding as far as I can remember is which namespace to install the Grafana dashboard into. I am guessing we will need to add an env variable to configure where it is installed and possibly not install it if the environment variable is not set. I would love to discuss this next week though to come to the best solution.

pkg/infinispan/grafana/dashboard_controller.go
dmvolod: This is main Grafana CR and it presence and correct status means that Grafana correctly installed and configured inside desired namespace.

dmvolod:
What options can we suggest for storing and changing the namespace of the Grafana Dashboard:
- create a ConfigMap at the operator deployment level as described in the issue #491. This will solve the problem with the need to store this parameter in an environment variable. A variable environment is a good solution for standalone deployments, but is limited to use in an OLM-managed cluster. Customers cannot change the predefined values, which will lead to the need to configure Grafana only in a special namespace configured on the OLM level.
- create a separate CR, which can be called `InfinispanConfiguration` or `OperatorConfiguration`, or come up with another name. The parameters set at the level of this CR will propagate to the operator level. In this case, we will need to accept whether we need to implement a singleton configuration (handle only single CR with a specific name) or make the parameters inheritable in case of creating multiple CRs in different namespaces or different names

It ssems to be any others option will be limited to the OLM management clusters

rigazilla:
I would go for a ConfigMap solution implemented like [here](https://docs.openshift.com/container-platform/4.6/monitoring/enabling-monitoring-for-user-defined-projects.html#enabling-monitoring-for-user-defined-projects_enabling-monitoring-for-user-defined-projects)

- operator should watch configmap creation and
- when a configmap is created/update which looks like a infinispan-operator configuration (needs criteria)
- operator read the config and creates a grafana dashboard if specified by the user

this also allows on the fly changes if required and can be extended if we need more config options.

wdyt?

ryanemerson:
A ConfigMap per operator instance sounds like a good idea to me. In the future if additional operator configuration is required, we can simply add support for it in this ConfigMap.

dmvolod:
> * when a configmap is created/update which looks like a infinispan-operator configuration (needs criteria)

We need to create a ConfigMap controller which will filter watching events by specific name and namespace. I did the same for operator console prototyping but in case of Deployment.

rigazilla:
Updating the design:
- configMap for configuration will be specified in the CSV and installed via OLM
- ocp admin will install the operator
- ocp admin will be in charge of any change to the configuration
- operator should watch for configmap updates and do reconciliation
     - in the specific operator read the config and creates a grafana dashboard if request by the user

rigazilla:
ConfigMap created via OLM seems not a good solution, during upgrade OLM updates the configmap with the CSV value so all the user configuration will be lost: https://docs.openshift.com/container-platform/4.5/operators/understanding/olm-packaging-format.html#olm-bundle-format-manifests_olm-packaging-format

dmvolod:
> ConfigMap created via OLM seems not a good solution, during upgrade OLM updates the configmap with the CSV value so all the user configuration will be lost: https://docs.openshift.com/container-platform/4.5/operators/understanding/olm-packaging-format.html#olm-bundle-format-manifests_olm-packaging-format

We can start with empty object to avoid values changes on update and can look at the idea of the `InfinispanConfiguration` CRD. But both ideas need to be validated.

rigazilla:
> > ConfigMap created via OLM seems not a good solution, during upgrade OLM updates the configmap with the CSV value so all the user configuration will be lost: https://docs.openshift.com/container-platform/4.5/operators/understanding/olm-packaging-format.html#olm-bundle-format-manifests_olm-packaging-format
> 
> Are you sure that the content is lost? From the docs "If the name of the optional object is the same, OLM updates it in place."
> 

My understand is that configMap is updated with the CSV content, but I can test it.

> Alternatively, maybe a `InfinispanConfiguration` CRD is the way to go. Previously I favoured the `ConfigMap` due to simplicity, however a CRD provides a well defined spec fo the user to follow and will scale better if/when we add additional configuration options.

Don't know, imo configmap is more flexible at this stage.

dmvolod:
> My understand is that configMap is updated with the CSV content, but I can test it.
> 

We can have default and user data section and OLM will contains the only first and never update user data.

> > Alternatively, maybe a `InfinispanConfiguration` CRD is the way to go. Previously I favoured the `ConfigMap` due to simplicity, however a CRD provides a well defined spec fo the user to follow and will scale better if/when we add additional configuration options.
> 
> Don't know, imo configmap is more flexible at this stage.

+1

rigazilla:
I've run some test, the configmap is updated overwriting the whole spec.data section during operator upgrade so all the previous data gone lost.
I'm going to leave that part behind and suppose that the cm is created somehow if user needs special settings

dmvolod:
> I've run some test, the configmap is updated overwriting the whole spec.data section during operator upgrade so all the previous data gone lost.
> I'm going to leave that part behind and suppose that the cm is created somehow if user needs special settings

It's expected behaviour, but idea could be different, i.e. have data.default key and use will create it's own one called data.user with required values. Didn't drive into OLM code, but not sure that it will remove or update nonexistent key.

rigazilla:
non existent keys seem to be deleted

dmvolod:
> non existent keys seem to be deleted

Oh, not a good news, seems to be some unexpected or regressed behavior in OLM

rigazilla:
closing this. Work has been moved to #894

Issue #804, opened 2021-02-08
Floating 'Unexpected response code 404' in Backup\Restore test


dmvolod:
There is a floating error could occurs during Backup\Restore test when cache presence is validated after restore

dmvolod:
I've added additional diagnostics steps before cache presence validation, there is detected that only first call to the REST API just fails with `404 (Not Found)`. Following message is printing during debug
`Health response , status: 404` and than next validation passes test.

dmvolod:
Close, as #770 should fix this issue

dmvolod:
Not reproduces in latest master. Fixed

PR #807, opened 2021-02-09
infinispan.GracefulShutdown incorrectly handles REST stop response

dmvolod:


dmvolod:
The same problem described in issue #804

pkg/infinispan/cluster.go
```diff
@@ -88,7 +88,7 @@ func (c Cluster) GetClusterSize(podName string) (int, error) {
 // GracefulShutdown performs clean cluster shutdown
 func (c Cluster) GracefulShutdown(podName string) error {
 	rsp, err, reason := c.Client.Post(podName, consts.ServerHTTPClusterStop, "", nil)
-	return validateResponse(rsp, reason, err, "during graceful shutdown", http.StatusOK)
+	return validateResponse(rsp, reason, err, "during graceful shutdown", http.StatusNoContent)
```
rigazilla, line 91: maybe it's for a differen PR but...
how about a more relaxed check status in 200..299? 
I saying this because I've already seen this kind of changes and return code isn't strictly specified server side.

dmvolod: Sorry, @rigazilla , but I don't agree with you in this case. Return codes should be strictly specified for implementation as well as documentation also.
That's the reason why I've reviewed whole REST API in PR https://github.com/infinispan/infinispan/pull/9040 for correct return codes to be specified. If API will change we will never know about it with this kind of relaxations.

rigazilla: well in my opinion a generic success return code would be sufficient in most of the cases.
Anyway it's also ok with me if a well define interface is established server side.

dmvolod: Yes, agreed about most cases, but in some cases we need detailed code validation, not in range

rigazilla (approved):
ok, lgtm

dmvolod:
@ryanemerson could you please to look at this small fix and merge if possible

dmvolod:
Thanks for review, @rigazilla 
Merged into master

PR #770, opened 2021-01-27
Update Backup/Restore names

ryanemerson:
A bit late to the party, but without this the json for templates and tasks won't be recognised on the server.

PR #808, opened 2021-02-09
Print diagnostic message for issue #804

dmvolod:
@ryanemerson , @rigazilla I don't have any idea how to catch #804 as it can be appears sometimes
Adding this output to the test, try to merge and wait for the catch 
Any ideas will be appreciated


dmvolod:
Close, as #770 should fix this issue

Issue #650, opened 2020-11-12
Custom labels on Resources created by Operator


buehlmann:
Let the user give the possibility to add his own custom labels to resources created by the Infinispan operator.

alvarolop:
Hello,

I want to +1 this issue. I am going to explain the reason why I would really appreciate this feature. 

I am currently setting up several Infinispan clusters on Openshift and I am monitoring them using Prometheus (The one provided on the platform). Each Business Unit uses one or more Infinispan clusters and they want to monitor them all together. 

For us, the easiest way to filter pods is using the bu_code=mybu label which is present in all the pods of each BU. Unfortunately, labels of the Infinispan CR are not propagated to the Service nor Pods. The simplest way to propagate labels to the metrics ingested in Prometheus is having those labels on the pods or the services by configuring targetLabels or podTargetLabels [1]. 

[1] https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#servicemonitorspec

I think that this feature would increase the synergy between these two projects, so much.

Is this feature planned for the future?
Regards, 
Álvaro López

rigazilla:
I would go on implementing this use case:
- user adds it's own labels to Infinispan CR
- user set two new fields (list of label names as strings) `targetLabels` and `podTargetLabels`
- operator will propagate Infinispan CR labels in the `targetLabels` to all the managed services
- operator will propagate Infinispan CR labels in the `podTargetLabels` to all the managed pods

is this ok?

dmvolod:
@rigazilla is there any case when set of the labels are different between Infinispan CRs for Pods and services?

alvarolop:
Hello Vittorio,

Thank you very much. Yes, these two fields in the Infinispan CRD fulfill all of our needs (And even more).

Regarding the question of @dmvolod. Right now I cannot think about a use case where we would need different labels in services and pods, but maybe there could be in the future. For my current use case, with just labels on pods or services is enough.

Thank you very much, have a nice day!
Álvaro.

dmvolod:
Hi, @alvarolop 
Probably, I wrote not so correct sentence, my question was about to separate labels list between Infinispan CR's or setup them globally on the Operator level.

rigazilla:
@dmvolod imo users will like the possibility to setup labels at cluster level. I would like to provide an implementation for the global/operator level on top on that.

[...]

Issue #517, opened 2020-08-06
Apply Infinispan CR changes to generated resources


dmvolod:
If possible we need to invoke controllerutil.CreateOrUpdate for resources to react on CR parameters change
The following objects/controllers need to be implemented

- [x] ConfigMap
- [x] Service
- [x] Route/Ingress
- [x] StatefulSet
- [x] Secret (not sure we need it)

dmvolod:
This option is implemented for ConfigMap only

dmvolod:
Need to add more logic for the #837

PR #810, opened 2021-02-10
Doc for custom labels 

oraNod:
Add docu for #650

### DIFF
--- a/documentation/asciidoc/stories/assembly_creating_services.adoc
+++ b/documentation/asciidoc/stories/assembly_creating_services.adoc
@@ -14,6 +14,10 @@ include::{topics}/ref_cacheservice_resources.adoc[leveloffset=+2]
 include::{topics}/proc_creating_dg_service.adoc[leveloffset=+1]
 include::{topics}/ref_datagridservice_resources.adoc[leveloffset=+2]
 
+//Labelling
+include::{topics}/proc_adding_custom_labels.adoc[leveloffset=+1]
+include::{topics}/ref_label_environment_variables.adoc[leveloffset=+2]
+
 // Restore the parent context.
 ifdef::parent-context[:context: {parent-context}]
 ifndef::parent-context[:!context:]
--- a/documentation/asciidoc/topics/attributes/links.adoc
+++ b/documentation/asciidoc/topics/attributes/links.adoc
@@ -15,6 +15,12 @@
 
 :link_k8s_node_port: https://kubernetes.io/docs/concepts/services-networking/service/#nodeport[Type NodePort]
 
+:link_k8s_labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/[Labels and Selectors]
+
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -45,6 +45,24 @@ $ cekit build --overrides "{version: <version_default_is_latest>}" docker
 
 ### Running the Infinispan Operator
 
+#### Adding Annotations
+
+Annotations are labels that {ispn_operator} propagates to pods and services that it controls.
```
dmvolod, line 50: This is not annotations from deployment point of view. We are using annotations in the Infinispan CR just for identifying list of the labels for each type: Pods and Services.

README.md
```diff
@@ -45,6 +45,24 @@ $ cekit build --overrides "{version: <version_default_is_latest>}" docker
...
+You should define custom annotations before you install the Infinispan Operator with the following environment variables:
```
dmvolod, line 51: Operator Deployment as well as desired variables can be changed at any time. This will occurs operator redeployment and it will start to work with the new parameters

README.md
```diff
@@ -45,6 +45,24 @@ $ cekit build --overrides "{version: <version_default_is_latest>}" docker
...
+
+```
+INFINISPAN_OPERATOR_TARGET_LABELS - Defines annotations for services.
```
dmvolod, line 54: This is just set of the target label and it's value propagated on the operator level globally for the services

README.md
```diff
@@ -45,6 +45,24 @@ $ cekit build --overrides "{version: <version_default_is_latest>}" docker
...
+INFINISPAN_OPERATOR_POD_TARGET_LABELS - Defines annotations for pods.
```
dmvolod, line 55: The sames as above for Pods

README.md
```diff
@@ -45,6 +45,24 @@ $ cekit build --overrides "{version: <version_default_is_latest>}" docker
...
+```
+
+1. Open the [`operator.yaml`](https://github.com/infinispan/infinispan-operator/blob/master/deploy/operator.yaml) for editing.
+2. Use the `env` field in to set annotations in JSON format, for example:
+
+```
+INFINISPAN_OPERATOR_TARGET_LABELS={"svc-label1":"svc-label2"}
```
dmvolod, line 62: They are pairs of label name and value, not a labels list

README.md
```diff
@@ -167,6 +185,23 @@ spec:
...
 
+#### Custom Labels
+Adds values for labels:
+
+```yaml
+apiVersion: infinispan.org/v1
+kind: Infinispan
+metadata:
+  annotations:
+    infinispan.org/podTargetLabels: svc-label1, svc-label2
```
dmvolod, line 196: I'm not sure do we need to say, that changes annotations and labels in the CR can't be reflected now.
We need to implement part of the #517 for Services and Routes

oraNod: Hmmm... This sounds like something the user should know. If you modify labels in Infinispan CR those changes do not get propagated? Does that include deleting labels or only modifying an existing label?

ryanemerson (approved):

rigazilla (approved):

dmvolod (approved):

oraNod:
thanks for the reviews. I'm going to self-merge.

PR #813, opened 2021-02-11
Apply Infinispan CR labels changes to the Service,Route,Ingress and Pod resources

dmvolod:


dmvolod:
@ryanemerson please pay attention that 12.0.1.Final image didn't solve floating `404 Not Found` error in Backup/Restore test

ryanemerson:
> @ryanemerson please pay attention that 12.0.1.Final image didn't solve floating `404 Not Found` error in Backup/Restore test

The combination of `12.0.1.Final` and  changes in https://github.com/infinispan/infinispan-operator/pull/770 should resolve the 404 issues.

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -428,11 +428,11 @@ func (ispn *Infinispan) ApplyOperatorLabels() error {
 	var errStr string
 	err := applyLabels(ispn, OperatorTargetLabelsEnvVarName, OperatorTargetLabels)
 	if err != nil {
-		errStr = fmt.Sprintf("Error unmarshalling "+OperatorTargetLabelsEnvVarName+" envvar: %v\n", err)
+		errStr = fmt.Sprintf("Error unmarshalling %s environment variable: %v\n", OperatorTargetLabelsEnvVarName, err)
 	}
 	err = applyLabels(ispn, OperatorPodTargetLabelsEnvVarName, OperatorPodTargetLabels)
 	if err != nil {
-		errStr = errStr + fmt.Sprintf("Error unmarshalling "+OperatorPodTargetLabelsEnvVarName+" envvar: %v\n", err)
+		errStr = errStr + fmt.Sprintf("Error unmarshalling %s environment variable: %v\n", OperatorPodTargetLabelsEnvVarName, err)
```
ryanemerson, line 435: Is the "\n" character necessary?

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -428,11 +428,11 @@ func (ispn *Infinispan) ApplyOperatorLabels() error {
 	var errStr string
 	err := applyLabels(ispn, OperatorTargetLabelsEnvVarName, OperatorTargetLabels)
 	if err != nil {
-		errStr = fmt.Sprintf("Error unmarshalling "+OperatorTargetLabelsEnvVarName+" envvar: %v\n", err)
+		errStr = fmt.Sprintf("Error unmarshalling %s environment variable: %v\n", OperatorTargetLabelsEnvVarName, err)
```
ryanemerson, line 431: Is the "\n" character necessary?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -643,7 +647,41 @@ func podAffinity(i *infinispanv1.Infinispan, matchLabels map[string]string) *cor
...
+func (r *ReconcileInfinispan) updatePodsLabels(m *infinispanv1.Infinispan, podList *corev1.PodList) error {
+	if len(podList.Items) == 0 {
+		return nil
+	}
+
+	labelsForPod := PodLabels(m.Name)
+	m.AddOperatorLabelsForPods(labelsForPod)
+	m.AddLabelsForPods(labelsForPod)
+
+	for _, pod := range podList.Items {
```
ryanemerson, line 659: Why are we updating pod's individually instead of just updating the StatefulSet spec?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -643,7 +647,41 @@ func podAffinity(i *infinispanv1.Infinispan, matchLabels map[string]string) *cor
...
 
-// deploymentForInfinispan returns an infinispan Deployment object
+func (r *ReconcileInfinispan) updatePodsLabels(m *infinispanv1.Infinispan, podList *corev1.PodList) error {
+	if len(podList.Items) == 0 {
+		return nil
+	}
+
+	labelsForPod := PodLabels(m.Name)
+	m.AddOperatorLabelsForPods(labelsForPod)
+	m.AddLabelsForPods(labelsForPod)
```
ryanemerson, line 657: @rigazilla Is there any reason we can't add these calls inside `PodLabels` so that it doesn't have to be repeated in multiple places?

dmvolod:
> The combination of `12.0.1.Final` and changes in #770 should resolve the 404 issues.

Good news, will wait until #770 be merged

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -643,7 +647,41 @@ func podAffinity(i *infinispanv1.Infinispan, matchLabels map[string]string) *cor
...
+			return err
+		}
+	}
+	return nil
+}
+
+// statefulSetForInfinispan returns an infinispan StatefulSet object
 func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispan, secret *corev1.Secret, configMap *corev1.ConfigMap) (*appsv1.StatefulSet, error) {
 	reqLogger := log.WithValues("Request.Namespace", m.Namespace, "Request.Name", m.Name)
 	lsPod := PodLabels(m.Name)
```
ryanemerson, line 687: @rigazilla  Why do we need `lsPod` to be distinct from `labelsForPod`? Won't both result in a match as all of the labels will be on the pods?

pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -141,13 +141,37 @@ func (s serviceResource) Process() (reconcile.Result, error) {
...
 
-// reconcileResource creates the resource (Service, Route or Ingress) for Infinispan if needed
+func (s serviceResource) reconcileService(service *corev1.Service) error {
+	findService := &corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      service.GetName(),
+			Namespace: service.GetNamespace(),
+		},
+	}
+	result, err := controllerutil.CreateOrUpdate(context.TODO(), s.client, findService, func() error {
```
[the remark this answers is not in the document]
dmvolod: Seems to be none as multiple foreign controllers could update Service and `CreateOrPatch` could conflict on `Patch` as this is non-blocking operation

pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -141,13 +141,37 @@ func (s serviceResource) Process() (reconcile.Result, error) {
 	return reconcile.Result{}, nil
 }
 
-// reconcileResource creates the resource (Service, Route or Ingress) for Infinispan if needed
+func (s serviceResource) reconcileService(service *corev1.Service) error {
```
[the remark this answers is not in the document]
dmvolod: Yes, sure. Missed that Route and Ingress are also labeled

pkg/apis/infinispan/v1/types_util.go
dmvolod: I don't know as this code was originally implemented by @rigazilla . I just add a fix for using `fmt.Sprintf` for all vars

pkg/apis/infinispan/v1/types_util.go
dmvolod: I don't know as this code was originally implemented by @rigazilla . I just add a fix for using `fmt.Sprintf` for all vars

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Updating `Spec.PodTemplateSpec` will cause a rolling update of the whole cluster just for Pods updating that's necessary in our case

ryanemerson: As the update is only to the metadata labels, does this not cause the individual pods to be restarted?

dmvolod: Which update? Any StatefulSet `PodTemplateSpec` update including annotations or labels will cause rolling update of whole cluster. For individual Pods update only annotations or labels will be updated.

ryanemerson: Sorry I meant "As the update is only to the metadata labels of the *individual pods*, does this not cause the individual pods to be restarted?". But I guess not :slightly_smiling_face:

pkg/apis/infinispan/v1/types_util.go
rigazilla: yeah it's not needed

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I wanted to leave the possibility to update only CR or operator labels, I'm not sure if that case will ever happens anyway

pkg/apis/infinispan/v1/types_util.go
dmvolod: Thanks, @rigazilla will remove trailing `\n`

pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -141,26 +140,82 @@ func (s serviceResource) Process() (reconcile.Result, error) {
...
+		return nil
+	})
+	if err != nil {
+		s.log.Error(err, "failed to create or update Ingress", ingress)
+		return err
+	}
+	if result != controllerutil.OperationResultNone {
+		s.log.Info(fmt.Sprintf("%s Ingress %s", strings.Title(string(result)), findIngress.GetName()))
+	}
+	return nil
```
ryanemerson, line 218: AFAICT all three reconcile functions are identical, except the object type being updated. Could we use `ToUnstructured` to create a generic method to handle the following part of the code:

```
		result, err := controllerutil.CreateOrUpdate(context.TODO(), s.client, findIngress, func() error {
		if findIngress.CreationTimestamp.IsZero() {
			controllerutil.SetControllerReference(s.infinispan, findIngress, s.scheme)
			findIngress.Spec = ingress.Spec
			findIngress.Annotations = ingress.Annotations
		}
		findIngress.Labels = ingress.Labels
		return nil
	})
	if err != nil {
		s.log.Error(err, "failed to create or update Ingress", ingress)
		return err
	}
	if result != controllerutil.OperationResultNone {
		s.log.Info(fmt.Sprintf("%s Ingress %s", strings.Title(string(result)), findIngress.GetName()))
	}
```

dmvolod: This could be an option, but I we would like to change some `Spec` structures not sure that will be a possible. For `ObjectMeta` it should be possible

pkg/controller/infinispan/infinispan_controller.go
rigazilla: The match will probably work, but I would not mix up internal labels `lsPod` with user labels `labelsForPod` which come for the user application domain and are not intended to be used as pods selector.

pkg/controller/infinispan/resources/service/service_controller.go
ryanemerson: I haven't tried it, but I think you could do something like:

```go
		found := unstructured.ToUnstructured(findIngress)
		provided := unstructured.ToUnstructured(ingress)
		unstructured.SetNestedField(found, provided["Spec"], "Spec")
		unstructured.SetNestedField(found, provided["Annotations"], "Annotations")
```
But maybe the saved lines aren't worth it :slightly_smiling_face:

dmvolod: Thanks for idea, @ryanemerson 
Bit polished and used for common function. But we should avoid to use `Unstructured` a lot.

dmvolod: Fixed, but Baclup/Restore test failed with another error

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -659,7 +663,41 @@ func podAffinity(i *infinispanv1.Infinispan, matchLabels map[string]string) *cor
...
+		}
+		for index, value := range labelsForPod {
+			podLabels[index] = value
+		}
+
+		_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, &pod, func() error {
+			if pod.CreationTimestamp.IsZero() {
+				return errors.NewNotFound(corev1.Resource(""), pod.Name)
+			}
+			pod.Labels = podLabels
```
rigazilla, line 690: Does this wipe out all the labels from the pod and then restore only our stuff? If so, it's too invasive imo

rigazilla (changes requested):

pkg/controller/infinispan/infinispan_controller.go
dmvolod: No, the system labels controlled by `consts.StatefulSetPodLabels` and never been removed

rigazilla: I'm referring to labels possibly added by the user for his own business, maybe I'm misreading the code but looks like they are not preserved.
also the `cordinator` label for cross site is removed if present.
I would apply a patch policy instead of rebuilding the labels map from scratch

dmvolod: Good catch related coordinator label for Pods, thanks @rigazilla . Will merge it logic with current update. Not sure that customers will need own labels on Pods controlled by statefulset, they will be removed every rolling upgrade. If customers will need own labels, they will need to set them via Infinispan CR.

dmvolod: > I would apply a patch policy instead of rebuilding the labels map from scratch

Also is unable to implement two-way merge without creating labels map from scratch, but it seems to be similar logic

dmvolod: @rigazilla fixed issue with coordinator Pod label

rigazilla: I still prefer my idea of not actively delete unknown labels, but that's just a detail, overall LGTM.

rigazilla (approved):

pkg/controller/infinispan/infinispan_controller.go
dmvolod: @rigazilla yes, sure, but I can't imagine how to remove user defined label which user was removed.

rigazilla: oh sorry @dmvolod I didn't though to that problem. I got your point now, your are right.

Actually I can something like: if the label name is listed in the annotations but not in the infinispan CR labels, then delete the label on the pod/svc.
this complicates things a bit and I'm not sure if it has side effects.

dmvolod: Yes, that's possible solution, but can't handle edit annotations.

rigazilla: yeah fine. I think we can go with this solution, let's keep overcomplicated solutions if users will complain about disappearing labels.

ryanemerson:
Thanks @dmvolod and sorry for the delay.

PR #814, opened 2021-02-12
Documentation improvements for client connections and route exposure

### LINKED ISSUE #812: Doc improvements: Hot Rod client connections and route exposure
Address feedback about Hot Rod client connections and route exposure

oraNod:
Fixes #812


documentation/asciidoc/topics/ref_hotrod_configuration_builder.adoc
```diff
@@ -0,0 +1,59 @@
...
+               .port(ConfigurationProperties.DEFAULT_HOTROD_PORT)
+             .security().authentication()
+               .username("username")
+               .password("password")
+               .realm("default")
+               .saslQop(SaslQop.AUTH)
+               .saslMechanism("SCRAM-SHA-512")
+             .ssl()
+               .sniHostName("$SERVICE_HOSTNAME")
+               .trustStorePath("/path/to/tls.crt");
```
oraNod, line 29: @ryanemerson @Crumby for Hot Rod clients in k8s to use encryption how do you set `trustStorePath`? can you point to the secret that holds tls.crt? 

does it depend on if the client is in the same namespace as the cluster?

documentation/asciidoc/stories/assembly_connecting_clients.adoc
```diff
@@ -0,0 +1,30 @@
...
+//CLI
+include::{topics}/con_cli_connections.adoc[leveloffset=+1]
+include::{topics}/proc_creating_caches_cli.adoc[leveloffset=+2]
+include::{topics}/proc_creating_caches_batch.adoc[leveloffset=+2]
+
+//Hot Rod
+include::{topics}/con_hotrod_clients.adoc[leveloffset=+1]
+include::{topics}/ref_hotrod_configuration_builder.adoc[leveloffset=+2]
+include::{topics}/ref_hotrod_properties.adoc[leveloffset=+2]
+include::{topics}/proc_creating_caches_hotrod.adoc[leveloffset=+2]
```
oraNod, line 22: @Crumby Perhaps this is going in the opposite direction of the "compact" documentation approach we took for clients in the previous versions. For downstream integrations I can also see the need to provide more detail around Hot Rod client configuration. I think it will be a big benefit for other teams if we can give more recommended examples. So while this PR introduces a lot more doc it's probably worth the effort to tell users the one best way.

oraNod: @rigazilla and @ryanemerson From a previous discussion around dogfooding, what do you guys think about taking the approach where we recommend adding cache configuration to XML|JSON files and creating them via CLI or Console? 

There may be instances where users need to create caches via Hot Rod but would you say it's approach to discourage doing so in docs? Mainly because passing long XML as a string is brutal.

documentation/asciidoc/stories/assembly_connecting_clients.adoc
```diff
@@ -0,0 +1,30 @@
+ifdef::context[:parent-context: {context}]
+[id='connecting_clients']
+:context: clients
+= Establishing Remote Client Connections
+Connect to {brandname} clusters from the {brandname} Console, Command Line Interface (CLI), and remote clients.
+Remote client connections let you create caches from valid configuration in XML or JSON format, operate on data, and perform administrative operations such as retrieving statistics and monitoring cluster health.
```
ryanemerson, line 6: How about simplifying to just:

"Remote client connections allow you to connect to {brandname} clusters from the {brandname} Console, Command Line Interface (CLI), and remote clients."

documentation/asciidoc/stories/assembly_connecting_clients.adoc
ryanemerson: I think the recommended approach to creating caches should be:

1. Cache CR (current shortcomings are being addressed)
2. CLI (Batch CR adds direct integration with the operator, PR coming this week).
3. Console
4. HotRod

documentation/asciidoc/topics/con_cli_connections.adoc
```diff
@@ -0,0 +1,19 @@
+= Connecting with the {brandname} CLI
+Use the command line interface (CLI) to connect to your {brandname} cluster and create caches and perform administrative operations.
```
ryanemerson, line 2: I would remove "and create caches and".

documentation/asciidoc/topics/con_cli_connections.adoc
```diff
@@ -0,0 +1,19 @@
...
+
+The CLI is available as part of the server distribution, which you can run on your local host to establish remote connections to {brandname} clusters on {openshiftshort}.
+
```
ryanemerson, line 5: We also have the `infinispan/cli` image upstream.

documentation/asciidoc/topics/con_cli_connections.adoc
```diff
@@ -0,0 +1,19 @@
...
+[source,options="nowrap",subs=attributes+]
+----
+$ {kubectl_exec}
+----
+
+However, this approach has drawbacks such as:
+
+* The CLI consumes memory allocated to the container, which can lead to out of memory exceptions.
+* Caches that you create this way are not replicated to other nodes.
```
ryanemerson, line 18: Not true. If you create a cache(not local-cache)/template/counter, the server will ensure that it's configuration is replicated to all nodes in the cluster.

documentation/asciidoc/topics/proc_adding_cache_auth_secrets.adoc
```diff
@@ -1,12 +1,10 @@
...
-cluster when creating caches.
+= Adding Credentials for {ispn_operator}
+{ispn_operator} must authenticate with {datagridservice} clusters to create caches.
+You add credentials to a secret so that {ispn_operator} can access your cluster when creating caches.
 
-The following procedure explains how to add credentials to a new secret. If you
-have a custom secret that contains credentials, you can use that instead of
-creating a new one.
+The following procedure explains how to add credentials to a new secret.
+If you have a custom secret that contains credentials, you can use that instead of creating a new one.
```
ryanemerson, line 7: "already have"?

documentation/asciidoc/topics/proc_connecting_console.adoc
```diff
@@ -0,0 +1,12 @@
+[id='creating_caches_console-{context}']
+= Accessing {brandname} Console
+Access the console to create caches, perform adminstrative operations, and monitor your {brandname} clusters.
+
+.Prerequisites
+
+* Retrieve the service hostname and, if necessary, port where the console is available.
```
ryanemerson, line 7: For the console isn't this only possible if a route/loadbalancer/ingress is used? A user can't access the console with just a service as it's not externally accessible by the user's browser.

documentation/asciidoc/topics/proc_connecting_console.adoc
```diff
@@ -0,0 +1,12 @@
...
+
+.Procedure
+
+. Access the console from any browser at `$SERVICE_HOSTNAME:11222`
```
ryanemerson, line 11: Doesn't the port number depend on how the server is externally exposed?

documentation/asciidoc/topics/proc_creating_caches_batch.adoc
```diff
@@ -0,0 +1,41 @@
+[id='creating_caches_batches-{context}']
+= Creating Caches in Batches
+Add multiple caches with batch operations with the {brandname} CLI.
```
ryanemerson, line 3: Once the Batch CR PR is up I think we can remove this in favour of it's documentation. Looks good in the interim though (sorry for any duplicated effort here).

documentation/asciidoc/topics/proc_creating_caches_cli.adoc
```diff
@@ -1,71 +1,64 @@
 [id='creating_caches_cli-{context}']
 = Creating Caches with {brandname} CLI
-Use the command line interface (CLI) to create caches on your {brandname}
-cluster.
+Add cache configurations to your {brandname} cluster with the CLI.
```
ryanemerson, line 3: Just "Add caches to ..."?

The procedure detailed below is how to define and create a cache, not just a configuration.

documentation/asciidoc/topics/proc_creating_caches_hotrod.adoc
```diff
@@ -1,90 +1,71 @@
 [id='creating_caches_hotrod-{context}']
-= Creating Caches with Hot Rod Java Clients
-Configure Hot Rod Java clients to connect to {brandname} clusters running on {openshiftshort}.
+= Creating Caches with Hot Rod Clients
+You can remotely create caches on {brandname} clusters running on {k8s} with Hot Rod clients.
+However, {brandname} recommends that you create caches using {brandname} Console, the CLI, or with `Cache` CRs instead of with Hot Rod clients.
```
ryanemerson, line 4: :+1:

documentation/asciidoc/topics/ref_hotrod_configuration_builder.adoc
ryanemerson: I believe you should just be able to set `.trustStorePath"/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt")` if the pod is on the same k8 cluster. It's been a while since I looked at k8 tls though.

@rigazilla @Crumby does that sound right?

documentation/asciidoc/topics/yaml/cache_service.yaml
```diff
@@ -0,0 +1,51 @@
+apiVersion: infinispan.org/v1
```
ryanemerson, line 1: I must have missed it. Where is this file and `datagrid_service.yaml` referenced?

ryanemerson (changes requested):
Great work @oraNod, just a few little things.

documentation/asciidoc/topics/proc_connecting_console.adoc
oraNod: @ryanemerson I remember accessing the console via load balancer from the OCP cluster that @Crumby has for testing. 

maybe we should remove these instructions for the console altogether?

documentation/asciidoc/topics/proc_creating_caches_batch.adoc
oraNod: no, it's OK. I expect this to be replaced by the Batch CR but I would like to get interim docs out there as a means to "prime" users for the Batch CR.

documentation/asciidoc/topics/yaml/cache_service.yaml
oraNod: For both cache service and data grid service there are full yaml examples at: https://infinispan.org/docs/infinispan-operator/2.0.x/operator.html#ref_datagrid_service_crd-nodes 

I thought it would be useful to have a reference like this, especially downstream where https://github.com/infinispan/infinispan-operator/blob/master/doc/design.adoc is not available.

documentation/asciidoc/topics/proc_connecting_console.adoc
ryanemerson: > @ryanemerson I remember accessing the console via load balancer from the OCP cluster that @Crumby has for testing.

That makes sense. My issue was that I thought these instructions were talking about connecting directly via the normal service.

oraNod: I see. Maybe it's somewhat overloaded to say "service hostname" when we mean the load balancer or node port? I suppose Route isn't technically a service either. 

Should we change to maybe "$NETWORK_HOSTNAME" to denote the hostname/url/IP address that provides access to the Infinispan cluster? This could also be the name of the internal service that the operator creates.

ryanemerson: I'm not sure if a number variable placeholder helps or not tbh. Maybe we need to leave this sentence as is and then add a note somewhere in this section explicilty saying that in order to access the console, you need to configure one of the external services, route etc.

documentation/asciidoc/topics/con_client_connection_details.adoc
```diff
@@ -0,0 +1,41 @@
...
+* Authentication credentials
+* TLS certificate, if you use encryption
+
+[NOTE]
+====
+In client configuration examples, `$SERVICE_HOSTNAME` denotes the hostname that allows access to your {brandname} cluster.
+Client configuration examples also use port `11222`.
+
+You should replace `$SERVICE_HOSTNAME` with the actual hostname for your environment.
+Likewise, if you use a node port or route, you should replace `11222` with the correct port.
```
ryanemerson, line 16: Instead of saying replace 11222 should we use $SERVICE_PORT so that this is more explicit?

I can imagine users copy and pasting `$SERVICE_HOSTNAME:11222`, replacing the var and forgetting to replace 11222 or worse not realise it's necessary as they did not read this section. WDYT?

ryanemerson (commented):
@oraNod Requires a rebase.

documentation/asciidoc/topics/con_client_connection_details.adoc
oraNod: Thanks, @ryanemerson I've changed to `$SERVICE_HOSTNAME:PORT` (maybe it should be `$PORT`) and made it more explicit that users will need to replace these with the actual values.

ryanemerson (approved):

documentation/asciidoc/stories/assembly_connecting_clients.adoc
Crumby: What about REST?

oraNod: @Crumby This might be misguided but I feel like REST is kind of a "generic" option. In fairness it would make things more complete if we mentioned the REST API and perhaps gave an example similar to the console. 

If you're using an HTTP client, such as curl, you can access the REST interface at `$SERVICE_HOSTNAME:$PORT/rest/v2` and then we point off to the API docs.

At the same time I think that the REST API should be more of a "backend" thing for k8s/OpenShift. As you know, both the console and CLI are the frontends to the REST interface. Maybe we should just advocate those tools and steer users away from every possible option. I don't know if there are any scenarios for integration that use REST anyway.

Just seems to me that we should be providing more "opinionated" usage examples and including REST opens another possibility that takes us away from that approach.

Crumby: @oraNod I'm ok with not mentioning REST at all. There is whole section on REST in another doc. The only thing important here is to understand how to expose the cluster outside and the rest for the REST is the same

ryanemerson: > If you're using an HTTP client, such as curl, you can access the REST interface at `$SERVICE_HOSTNAME:$PORT/rest/v2` and then we point off to the API docs.

+1 to not going into usage details, but I think adding this comment somewhere would be good so that the user is at least aware it's an option.

documentation/asciidoc/topics/con_client_connection_details.adoc
```diff
@@ -0,0 +1,32 @@
...
+For clients running outside {k8s}, the service hostname is the location URL if you use a load balancer.
+For a node port service, the service hostname is the node host name.
+For a route, the service hostname is either a custom hostname or a system-defined hostname.
+
+.Ports
+
+Client connections on {k8s} and through load balancers use port `11222`.
+
+Node port services use a port in the range of `30000` to `60000`.
+Routes use either port `80` (http) or `443` (https).
```
Crumby, line 25: 443 can be also used with tls for other protocols. In our case HotRod.

documentation/asciidoc/topics/con_hotrod_clients.adoc
```diff
@@ -0,0 +1,12 @@
...
+Hot Rod is a binary TCP protocol that {brandname} provides for high-performance data transfer capabilities with remote clients.
+
+.Client intelligence
+
+Client intelligence refers to mechanisms the Hot Rod protocol provides so that clients can locate and send requests to {brandname} nodes.
+
+Hot Rod clients running on {k8s} can access internal IP addresses for {brandname} nodes so you can use any client intelligence.
+The default intelligence, `HASH_DISTRIBUTION_AWARE`, is recommended because it allows clients to route requests to primary owners, which improves performance.
+
+Hot Rod clients running outside {k8s} can use `BASIC` intelligence only.
```
Crumby, line 12: Unless HotRod client was improved then `BASIC` is mandatory.

documentation/asciidoc/topics/proc_connecting_console.adoc
```diff
@@ -0,0 +1,16 @@
...
+.Prerequisites
+
+* Expose {brandname} on the network so you can access the console through a browser. +
+For example, configure a load balancer service or create a route.
+
+.Procedure
+
+. Access the console from any browser at `$SERVICE_HOSTNAME`.
++
+Replace `$SERVICE_HOSTNAME:PORT` with the hostname and port where {brandname} is available on the network.
```
Crumby, line 14: `$SERVICE_HOSTNAME:$PORT`?

documentation/asciidoc/topics/proc_creating_caches_batch.adoc
```diff
@@ -0,0 +1,43 @@
...
+----
++
+. Create the caches with the CLI.
++
+[source,bash,options="nowrap",subs=attributes+]
+-----
+$ bin/cli.sh -c https://$SERVICE_HOSTNAME:PORT --trustall -f /tmp/caches.batch
+-----
++
+Replace `$SERVICE_HOSTNAME:PORT` with the hostname and port where {brandname} is available on the network.
```
Crumby, line 37: `$SERVICE_HOSTNAME:$PORT`?

documentation/asciidoc/topics/proc_creating_caches_batch.adoc
```diff
@@ -0,0 +1,43 @@
+[id='creating_caches_batches-{context}']
+= Creating Caches in Batches
+Add multiple caches with batch operations with the {brandname} CLI.
+
+.Prerequisites
+
+* Download the server distribution so you can run the CLI.
+* Retrieve the necessary client connection details.
```
Crumby, line 8: Maybe we could list here what the connection details are?

documentation/asciidoc/topics/yaml/datagrid_service.yaml
```diff
@@ -0,0 +1,67 @@
...
+      name: azure
+      expose:
+        type: LoadBalancer
+      # Configures backup locations.
+      locations:
+      - name: azure
+        url: openshift://api.azure.host:6443
+        secretName: azure-token
+      - name: aws
+        clusterName: {example_crd_name}
```
Crumby, line 29: `clusterName` and `namespace` options aren't available in 2.0.x yet

oraNod: yeah I'll have to make sure this doesn't get included when we backport to 2.0.x

documentation/asciidoc/topics/proc_creating_caches_batch.adoc
oraNod: There's a section in the docs "11.2. Client Connection Details" which I think covers it instead of being too repetitious. Perhaps that's not the best approach though. Users might miss that. Let me know what you think when you review the rendered HTML.

rigazilla:
I'm reviewing but I'm confused, I see creating cache info in par: 9 9.2 9.3 10 10.2 10.3.1 10.3.2
it that correct?

oraNod:
> I'm reviewing but I'm confused, I see creating cache info in par: 9 9.2 9.3 10 10.2 10.3.1 10.3.2
> is that correct?

@rigazilla Hey, yeah, I suppose we're covering many different ways to create caches. "Chapter 9" is all about the Cache CR. "Chapter 10" is about connecting to clusters via remote clients, CLI, and console. Within "Chapter 10" there are also some titles focused on creating caches remotely. Does that answer your question?

rigazilla (commented):
> > I'm reviewing but I'm confused, I see creating cache info in par: 9 9.2 9.3 10 10.2 10.3.1 10.3.2
> > is that correct?
> 
> @rigazilla Hey, yeah, I suppose we're covering many different ways to create caches. "Chapter 9" is all about the Cache CR. "Chapter 10" is about connecting to clusters via remote clients, CLI, and console. Within "Chapter 10" there are also some titles focused on creating caches remotely. Does that answer your question?

I would expect to find them in the same place, but yeah maybe we have to many tools for the same job

documentation/asciidoc/stories/assembly_connecting_clients.adoc
```diff
@@ -0,0 +1,33 @@
+ifdef::context[:parent-context: {context}]
+[id='connecting_clients']
+:context: clients
+= Establishing Remote Client Connections
+Connect to {brandname} clusters from the {brandname} Console, Command Line Interface (CLI), and remote clients.
+
+include::{topics}/con_creating_caches.adoc[leveloffset=+1]
```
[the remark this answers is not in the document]
oraNod: sure. I've moved it. thanks!

rigazilla (approved):
lgtm

documentation/asciidoc/topics/ref_hotrod_configuration_builder.adoc
rigazilla: I've forgot to reply.
It should work, but I've never tried it and since we're doing TLS directly I would do some test before and maybe update docs in a separate PR.

rigazilla:
merged, thanks @oraNod!

Issue #816, opened 2021-02-15
e2e testing failing with "Labels haven't been propagated to pods"


redmark-ham:
I'm testing in the 2.0.x branch and noticed the code change for custom labels. What new parameter(s) are need for testing?

Using project "default".
+ cd infinispan-operator
+ export EXPOSE_SERVICE_TYPE=Route
+ EXPOSE_SERVICE_TYPE=Route
+ export KUBECONFIG=/home/fedora/.kube/config
+ KUBECONFIG=/home/fedora/.kube/config
+ export RUN_LOCAL_OPERATOR=false
+ RUN_LOCAL_OPERATOR=false
+ export TESTING_NAMESPACE=datagrid-suite
+ TESTING_NAMESPACE=datagrid-suite
+ oc new-project datagrid-suite
Now using project "datagrid-suite" on server "https://api.test-ocp4-6.qe13.testing:6443".

You can add applications to this project with the 'new-app' command. For example, try:

    oc new-app ruby~https://github.com/sclorg/ruby-ex.git

to build a new example application in Ruby. Or use kubectl to deploy a simple Kubernetes application:

    kubectl create deployment hello-node --image=gcr.io/hello-minikube-zero-install/hello-node

+ go mod vendor
+ go clean -testcache ./test/e2e
+ go test -v ./test/e2e -timeout 60m
=== RUN   TestNodeStartup
--- FAIL: TestNodeStartup (71.86s)
panic: Labels haven't been propagated to pods [recovered]
	panic: Labels haven't been propagated to pods

dmvolod:
This expected behavior if operator image doesn't contains last labels propagation code changes.
Please provide operator deployment version

Crumby:
@dmvolod @rigazilla The issue is `TestNodeStartup` test newly assumes that Operator is running locally as it sets env using `os.Setenv`. If you execute the test with the Operator already running within the OpenShift then setting up envs like this doesn't have any effect.

redmark-ham:
I'm testing the 8.1.5 multi arch operator.

selfLink: >-
    /apis/operators.coreos.com/v1alpha1/namespaces/datagrid-suite/clusterserviceversions/datagrid-operator.v8.1.5
  resourceVersion: '1664286'
  name: datagrid-operator.v8.1.5
  uid: 4e21048f-f737-4c53-896d-ef392f870114

redmark-ham:
[...]

PR #818, opened 2021-02-16
Fixed labels check on deployed operator #816

rigazilla:
When testing a deployed operator labels need to be retrieved from the deployment. Setting them on the test environment doesn't work.

### DIFF
--- a/test/e2e/main/main_test.go
+++ b/test/e2e/main/main_test.go
@@ -139,8 +139,14 @@ func TestNodeStartup(t *testing.T) {
 			panic("Infinispan CR labels haven't been propagated to pods")
 		}
 	} else {
+		// Get the operator namespace from the env if it's different
+		// from the testsuite one
+		operatorNS := tutils.OperatorNamespace
+		if operatorNS == "" {
+			operatorNS = spec.Namespace
+		}
 		// operator deployed on cluster, labels are set by the deployment
-		if !areOperatorLabelsPropagated(spec.Namespace, ispnv1.OperatorPodTargetLabelsEnvVarName, pod.Labels) {
+		if !areOperatorLabelsPropagated(operatorNS, ispnv1.OperatorPodTargetLabelsEnvVarName, pod.Labels) {
 			panic("Operator labels haven't been propagated to pods")
 		}
 	}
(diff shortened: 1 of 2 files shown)


test/e2e/main/main_test.go
```diff
@@ -124,25 +124,83 @@ func TestNodeStartup(t *testing.T) {
...
+	// from operator environment
+	if tutils.RunLocalOperator == "TRUE" {
+		// running locally, labels are hardcoded and set by the testsuite
+		if pod.Labels["operator-pod-label"] != "operator-pod-value" ||
+			ispn.Labels["operator-pod-label"] != "operator-pod-value" {
+			panic("Infinispan CR labels haven't been propagated to pods")
+		}
+	} else {
+		// operator deployed on cluster, labels are set by the deployment
+		if !areOperatorLabelsPropagated(spec.Namespace, ispnv1.OperatorPodTargetLabelsEnvVarName, pod.Labels) {
```
Crumby, line 143: Operator doesn't necessarily have to run in the same namespace as the Infinispan cluster. Can we parameterize expected Operator namespace?

test/e2e/main/main_test.go
```diff
@@ -124,25 +124,83 @@ func TestNodeStartup(t *testing.T) {
...
+	labelsAsString := ""
+	for _, item := range podList.Items[0].Spec.Containers[0].Env {
+		if item.Name == varName {
+			labelsAsString = item.Value
+		}
+	}
+	if labelsAsString == "" {
+		return true
+	}
+	opLabels := make(map[string]string)
```
Crumby, line 192: If we have set of defined labels expected to present on Operator OLM deployment maybe we could just assert that expected labels with expected values are present on the pods?

Crumby (changes requested):

test/e2e/main/main_test.go
rigazilla: @Crumby I've added a commit for this: operator NS can be configured via `TESTING_OPERATOR_NAMESPACE` envvar

test/e2e/main/main_test.go
rigazilla: there's no a known set of labels: i.e. community operator is shipped without any label.

Crumby (approved):

PR #822, opened 2021-02-17
[2.0.x] Able to setup XSite for different namespaces

dmvolod:


dmvolod:
@oraNod please pay attention, that documentation need to be backported for this fixes.

oraNod:
@dmvolod I'd really prefer to keep backports separate for the different issues. I know it's a little more work but it really helps when you need to go through commit history. On the downstream side, for docs at least, these commits get synchronized and it helps to have them cleanly separated so you know for sure which commit corresponds to which issue.

pruivo (approved):

oraNod (commented):
:+1:

ryanemerson:
@dmvolod @oraNod Can we also keep the original commit message in the backports. When viewing the git logs anywhere but GitHub it means that it's necessasry to first lookup the issue # to deterrmine what the commit is related to

dmvolod:
> @dmvolod @oraNod Can we also keep the original commit message in the backports. When viewing the git logs anywhere but GitHub it means that it's necessasry to first lookup the issue # to deterrmine what the commit is related to

Sure, done for original commit

ryanemerson:
Thanks @dmvolod and @oraNod

Issue #746, opened 2021-01-18
Batch CR


ryanemerson:
# Proposal
We should create a "Batch" CR that allows for multiple Infinispan resources to be created using Infinispan's [CLI batch mode](https://infinispan.org/docs/stable/titles/cli/cli.html#batch_operations). The batch script is then executed by as a one-off event with the CR marked as complete once the initial batch has been executed.

The advantage of this approach is that it allows for users to create multiple resource types, e.g. counters & counters, as well as allowing a batch of arbritrary admin actions to be automated. Furthermore, as the operator is leveraging the CLI's existing capabities, there is no need to write additional code to interact with the Infinispan server.

## Configuration
Users can provide a batch CLI in one of two ways (exclusive):

### Inline
Inline definition in the CR spec `spec.config`:
  - Multi-line yaml string containing batch contents
  - Suitable if no additional files are required

```yaml
apiVersion: infinispan.org/v2alpha1
kind: Batch
metadata:
  name: example-inline-batch
spec:
  cluster: example-infinispan
  config: |
    create cache --template=org.infinispan.DIST_SYNC mybatch
    put --cache=mybatch hello world
    put --cache=mybatch hola mundo
    disconnect
```

### Config Map
File(s) provided by a ConfigMap specified via `spec.configMap`.
```yaml
apiVersion: infinispan.org/v2alpha1
kind: Batch
metadata:
  name: example-cm-batch
spec:
  cluster: example-infinispan
  configMap: batch-config-map
```

ConfigMap key `batch` contains the batch file to execute, as well as any additional files required by the script.
 
Local Dir
```
example-batch
├── batch
└── pcache.xml
```

CLI Batch file
```
create cache --file=pcache.xml pcache
disconnect
```

`kubectl create configmap batch-config-map --from-file=example-batch`

## Implementation
A [k8s job](https://kubernetes.io/docs/concepts/workloads/controllers/job/) can be used to launch the CLI with the user provided ConfigMap mounted to the underlying pod so that the batch file can be executed. On completion of the job, the Batch CRs status can be updated to `Completed|Failed`.

Issue #580, opened 2020-09-19
Use client-go dynamic.Interface as DynamicClient for tests


dmvolod:
WaitFor* polling functions are complex and heavy to execute and not always can handle all expected conditions to met
Research is it possible to replace all or part of these functions with such kind of calls
```
DynamicClient.Resource(info.Mapping.Resource).Namespace(info.Namespace).Watch(context.TODO(), watchOptions)
```
We can use `kubectl` source code as example https://github.com/kubernetes/kubectl

PR #827, opened 2021-02-17
Batch CR #746

ryanemerson:
Support for disabling auth needs adding to batch controller once https://github.com/infinispan/infinispan-operator/pull/792 has been merged.


pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,380 @@
...
+		labels := batchLabels(batch.Name)
+		r.client.DeleteAllOf(ctx, &corev1.Pod{}, []client.DeleteAllOfOption{
+			client.InNamespace(batch.Namespace),
+			client.MatchingLabels(labels),
+		}...)
+	}
+	return reconcile.Result{}, allErrors
+}
+
+func (r *batchResource) loadCluster() (*v1.Infinispan, error) {
```
dmvolod, line 323: @ryanemerson do we need to validate that Infinispan CR is well formed for Batch CR execution?

ryanemerson: Definitely! Good catch :slightly_smiling_face:. Updated.

deploy/role.yaml
```diff
@@ -1,9 +1,10 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: infinispan-operator
+  creationTimestamp: null
```
dmvolod, line 4: This field is not needed

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,392 @@
...
+		return controllerutil.SetControllerReference(batch, job, r.scheme)
+	})
+
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("Unable to create batch job '%s': %w", batch.Name, err)
+	}
+	return reconcile.Result{}, r.UpdatePhase(v2.BatchRunning, nil)
+}
+
+func (r *batchResource) waitToComplete() (reconcile.Result, error) {
```
dmvolod, line 246: @ryanemerson could you please a look at this issues #580 described utilizing `dynamic.Interface`
I planned to utilize it for the tests, but probably you can use it in this case

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,392 @@
...
+func (r *batchResource) cleanup() (reconcile.Result, error) {
+	batch := r.instance
+	meta := metav1.ObjectMeta{
+		Namespace: batch.Namespace,
+		Name:      batch.Name,
+	}
+	var allErrors error
+	// Delete the Job used to execute the batch
+	if err := r.client.Delete(ctx, &batchv1.Job{ObjectMeta: meta}); err != nil && !errors.IsNotFound(err) {
+		allErrors = fmt.Errorf("Unable to delete zero-capacity pod: %w", err)
```
dmvolod, line 304:
```suggestion
		allErrors = fmt.Errorf("Unable to delete Job: %w", err)
```

deploy/role.yaml
```diff
@@ -22,6 +23,12 @@ rules:
   - create
   - update
   - delete
+- apiGroups:
```
dmvolod, line 26: In this case we need to remove hack with Role for Travis CI. Now utilizing it for the SA

deploy/role.yaml
```diff
@@ -90,6 +108,7 @@ rules:
   - restores
   - restores/status
   - restores/finalizers
+  - batches
```
dmvolod, line 111: Seems to be batches\status and batches/finalizers are missing

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,392 @@
...
+								MountPath: BatchVolumeRoot,
+							},
+						},
+					}},
+					RestartPolicy: corev1.RestartPolicyNever,
+					Volumes: []corev1.Volume{{
+						Name: BatchVolumeName,
+						VolumeSource: corev1.VolumeSource{
+							ConfigMap: &corev1.ConfigMapVolumeSource{
+								LocalObjectReference: corev1.LocalObjectReference{Name: batch.Name},
```
dmvolod, line 227: Is ConfigMap name always equals Batch CR name?

test/e2e/batch/batch_test.go
```diff
@@ -0,0 +1,194 @@
...
+func createCluster(name string) *v1.Infinispan {
+	infinispan := tutils.DefaultSpec(testKube)
+	infinispan.ObjectMeta.Name = name
+	testKube.Create(infinispan)
+	testKube.WaitForInfinispanPods(1, tutils.SinglePodTimeout, infinispan.Name, tutils.Namespace)
+	return infinispan
+}
+
+func createBatch(name, cluster, config, configMap string) *v2.Batch {
+	stringToPtr := func(s string) *string {
```
dmvolod, line 135: `pointer.StringPtr()` already implemented

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,392 @@
...
+		Name:      batch.Name,
+		Namespace: batch.Namespace,
+	}
+	if err := r.client.Get(ctx, jobKey, job); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, fmt.Errorf("Job '%s' not found", jobKey.Name)
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, fmt.Errorf("Unable to fetch Job '%s': %w", jobKey.Name, err)
+	}
```
dmvolod, line 259: We have a `LookupResource` function in infinispan_controller for this case. NotFound means not ready atm.

deploy/role.yaml
ryanemerson: Doesn't that also require the "list" and "watch" verbs? That's why I didn't update before.

deploy/role.yaml
ryanemerson: Good spot. I think this was added when the sdk added the batch permissions

dmvolod: Yes, looks like very old SDK bug

deploy/role.yaml
dmvolod: Need to be validated, we can remove it and look at the debug-pods.sh processing in the Travis CI

pkg/controller/batch/batch_controller.go
ryanemerson: Doesn't #580 relate to test code? `waitToComplete()` is only ever called on a reconcile event.

I have updated the code to just return `reconcile.Result{}` instead of `reconcile.Result{RequeueAfter: time.Second}` when the job has not completed, as we already watch Jobs so there's no need to explicitly requeue.

test/e2e/batch/batch_test.go
ryanemerson: The `stringToPtr` function is different. If `s == ""`, then `nil` is returned, whereas `pointer.StringPtr()` returns a pointer to `""`.

But regardless I have just updated the function to `createBatch(name, cluster string, config, configMap *string) *v2.Batch` so neither is required.

pkg/controller/batch/batch_controller.go
ryanemerson: Updated and I have moved `LookupResource` to `pkg/kubernertes/controllerutil.go` as this is a generic method applicable to all of our controllers.

pkg/controller/batch/batch_controller.go
ryanemerson: Nope! Good spot. I have fixed this code and updated `TestBatchConfigMap` to specify a ConfigMap with a different name to the batch.

pkg/controller/batch/batch_controller.go
dmvolod: It's related to watch for Job conditions without reconciling each not successful stage

build/prepare-sa-test.sh
```diff
@@ -9,7 +9,6 @@ KUBE_DEPLOY_SECRET_NAME=`kubectl -n $TESTING_NAMESPACE get sa infinispan-operato
 KUBE_API_EP=`kubectl get ep kubernetes -n default -o jsonpath='{.subsets[0].addresses[0].ip}'`
 KUBE_API_TOKEN=`kubectl -n $TESTING_NAMESPACE get secret $KUBE_DEPLOY_SECRET_NAME -o jsonpath='{.data.token}'|base64 --decode`
 
-cat deploy/role.yaml build/debug-role-patch.yaml | kubectl apply -n $TESTING_NAMESPACE -f -
```
dmvolod, line 12: Ops, only `cat` need to be removed. Role should be kept

pkg/controller/batch/batch_controller.go
ryanemerson: I'm confused. The batch controller needs to watch for successful and unsuccessful job stages, as a job failure causes the batch CR to fail.

dmvolod: Yes, correct. You can use current approach validating on every reconcile loop or use `dynamic.Interface` to watch for changes only. Let's keep as is. Probably we can change it in the future.
But I'm worry about logs length and size placed in to the `Reason` field. Is it too big or not?

ryanemerson: A failed batch currently only logs exceptions, so the logs are typically very small.

When I first tested this, I actually retrieved the server logs and it worked well (newline chars respected etc) when using `kubectl describe Batch` to view the status.

build/prepare-sa-test.sh
ryanemerson: Fixed

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,383 @@
...
+		Name:      clusterName,
+	}
+	if err := r.client.Get(ctx, clusterObjKey, infinispan); err != nil {
+		if errors.IsNotFound(err) {
+			return nil, fmt.Errorf("Infinispan CR '%s' not found", clusterName)
+		}
+		// Error reading the object - requeue the request.
+		return nil, fmt.Errorf("Unable to fetch Infinispan CR '%s': %w", clusterName, err)
+	}
+	return infinispan, nil
```
dmvolod, line 341: Seems to be `LookupResource` could be utilized here

ryanemerson: Much better. This allowed me to remove the `loadCluster` function entirely.

.travis.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         - go mod vendor
       script:
         # Run multinamespace tests as admin (skip role settings check)
-        - make test PARALLEL_COUNT=2 && kubectl config use-context $TESTING_CONTEXT && make multinamespace-test
+        - make test PARALLEL_COUNT=2 && kubectl config use-context $TESTING_CONTEXT && make batch-test && make multinamespace-test
```
dmvolod, line 43: @ryanemerson do you need to run `batch-test` without specific role with admin account? If not, this should run before `kubectl config use-context $TESTING_CONTEXT`

ryanemerson: Updated

deploy/operator-install.yaml
```diff
@@ -15,9 +15,10 @@ metadata:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  name: infinispan-operator
+  creationTimestamp: null
```
dmvolod, line 18:
```suggestion
```

.travis.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         - go mod vendor
       script:
         # Run multinamespace tests as admin (skip role settings check)
-        - make test PARALLEL_COUNT=2 && kubectl config use-context $TESTING_CONTEXT && make multinamespace-test
+        - make test PARALLEL_COUNT=2 && make batch-test && kubectl config use-context $TESTING_CONTEXT && make multinamespace-test
```
dmvolod, line 43: @ryanemerson would you like to split `batch-test` with the separate Stage?

ryanemerson: With you saying this requires an additional VM I thought this would be overkill, in the same way that we don't create a separate stage for the multinamespace-tests.

rigazilla (commented):
Single tests launched by the IDE are failing with a 404 for me with server:12.0
I.e:
```
TestBatchInlineConfig

panic: Expected Status Code 200, received 404 [recovered]
	panic: Expected Status Code 200, received 404

goroutine 271 [running]:
testing.tRunner.func1.1(0x1b8c820, 0xc0006f5fe0)
	/usr/lib/golang/src/testing/testing.go:999 +0x51b
testing.tRunner.func1(0xc0008f2120)
	/usr/lib/golang/src/testing/testing.go:1002 +0x31a
panic(0x1b8c820, 0xc0006f5fe0)
	/usr/lib/golang/src/runtime/panic.go:975 +0x41b
github.com/infinispan/infinispan-operator/test/e2e/batch.assertRestOk(0xc00090d3b0, 0x2b, 0x2064260, 0xc0009ca240)
	/home/rigazilla/go/src/github.com/infinispan/infinispan-operator/test/e2e/batch/batch_test.go:187 +0x1fe
github.com/infinispan/infinispan-operator/test/e2e/batch.TestBatchInlineConfig(0xc0008f2120)
	/home/rigazilla/go/src/github.com/infinispan/infinispan-operator/test/e2e/batch/batch_test.go:51 +0x397
```
is this related to the note above about #792 ?

deploy/operator-install.yaml
ryanemerson: Damn that somehow sneaked back in. Fixed.

pkg/apis/infinispan/v2alpha1/batch_types.go
```diff
@@ -0,0 +1,62 @@
...
+	Phase BatchPhase `json:"phase"`
+	// Reason indicates the reason for any batch related failures.
+	Reason string `json:"reason,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Batch is the Schema for the batches API
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=batches,scope=Namespaced
```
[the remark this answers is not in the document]
ryanemerson: Nope. Updating backup_types.go doesn't seem to have any effect on the generated code though.

ryanemerson:
> Single tests launched by the IDE are failing with a 404 for me with server:12.0

That's strange, I can't reproduce locally. What image sha do you have? The latest `quay.io/infinispan/server@sha256:9c15fbaf399e4087d8cf4e0f98b32f0a30ad9171e0a2f97fe3fba64a3ee93473` passes for me.

> is this related to the note above about #792 ?

No that's outdated and just required the batch_controller to be updated.

.travis.yml
dmvolod: Sorry, I was talking about separating multinamespace-test, but not really know how long take Batch CR test. Probably we not needed a separate Stage for it. It depends of test time and case separating.

rigazilla:
> That's strange, I can't reproduce locally. What image sha do you have? The latest `quay.io/infinispan/server@sha256:9c15fbaf399e4087d8cf4e0f98b32f0a30ad9171e0a2f97fe3fba64a3ee93473` passes for me.

yeah my OCP is pulling an old version, need to check why.

rigazilla:
I've updated the image, but I still have error, I'm check my settings

dmvolod:
install-crds.sh should be updated as well

ryanemerson:
> install-crds.sh should be updated as well

Update to use `cat deploy/crds/infinispan.org*crd.yaml | kubectl apply -f -` so that we don't have to manually add new CRDs

dmvolod:
> Updated to use `cat deploy/crds/infinispan.org*crd.yaml | kubectl apply -f -` so that we don't have to manually add new CRDs

Oh, this doesn't work. You need to add a delimiter between yaml's

ryanemerson:
> Oh, this doesn't work. You need to add a delimiter between yaml's to be run as batch. Please see codegen script

Good point! I had missed some other `*.sh` scripts as well. I think I have updated them all now.

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,365 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName  = "batch-controller"
+	DataMountPath   = "/opt/infinispan/batch"
```
dmvolod, line 32: This is unused const.

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -196,6 +196,10 @@ func (ispn *Infinispan) IsUpgradeCondition() bool {
 	return ispn.IsConditionTrue(ConditionUpgrade)
 }
 
+func (ispn *Infinispan) GetServiceFqn() string {
```
dmvolod, line 199: Not sure we need Service FQN for connection between Job Pod and Infinispan Pod's

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,365 @@
...
+		_, err := controllerutil.CreateOrUpdate(ctx, r.client, configMap, func() error {
+			configMap.Data = map[string]string{BatchFilename: *spec.Config}
+			return controllerutil.SetControllerReference(batch, configMap, r.scheme)
+		})
+
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("Unable to create ConfigMap '%s': %w", configMap.Name, err)
+		}
+
+		updated, err := r.update(func() {
```
dmvolod, line 167: As Batch CR one-time operation, we should set Infinispan CR as Batch CR controller OwnerReference

pkg/apis/infinispan/v1/types_util.go
ryanemerson: Good point. The Batch  CR always has to reside in the same namespace as the target cluster, so the svc name alone should suffice.

ryanemerson:
@dmvolod Batch should work as expected with encryption now. I have also updated the controller so that a batch will fail if both a `spec.Config` and `spec.ConfigMap` are provided by a user.

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,373 @@
...
+	infinispan.AddLabelsForPods(labels)
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      batch.Name,
+			Namespace: batch.Namespace,
+		},
+		Spec: batchv1.JobSpec{
+			BackoffLimit: pointer.Int32Ptr(0),
+			Template: corev1.PodTemplateSpec{
```
dmvolod, line 214: @ryanemerson we also discussed that Job Pods should have a Batch CR as owner reference to prevent orphaned Pods after Batch CR deletion.

ryanemerson: As the pods are owned by the Job, I think the only way todo this would be to manually get the created pod and set the owner reference. Do you know of a better way todo this? For example via the Job podspec?

dmvolod: Yeah, seems to be a problem with Job controller implementation on some old OCP version I've experimented before.

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+func (r *batchResource) cleanup() (reconcile.Result, error) {
+	batch := r.instance
+	meta := metav1.ObjectMeta{
+		Namespace: batch.Namespace,
+		Name:      batch.Name,
+	}
+	var allErrors error
+	// Delete the Job used to execute the batch
```
dmvolod, line 303: It would be nice to keep Job and Job Pods undeleted and cleanup them on Batch CR garbage collection (delete)
If some issues will occurs we will need them for debug and trace

pkg/controller/batch/batch_controller.go
dmvolod: Please see comment about cleanup resources. On latest OCP Job Pods are deleted with Job garbage collection

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,373 @@
...
+		if numConditions > 0 {
+			condition := status.Conditions[numConditions-1]
+
+			if condition.Type == batchv1.JobFailed {
+				var reason string
+				podName, err := r.getJobPodName(batch.Name, batch.Namespace)
+				if err != nil {
+					reason = err.Error()
+				} else {
+					reason, err = kubernetes.Logs(podName, batch.Namespace)
```
dmvolod, line 277: I know that's not related to the Operator code, but could we have more reliable reason message. It's not informative, which command is failed and why some extra signs added
`"\e[0;31mCommandLineParserException: Command and/or sub-command is not valid!\e[0;39m\n"`

.travis.yml
dmvolod: I think six tests are enough to have a separate parallel CI stage. wdyt?

pkg/controller/batch/batch_controller.go
ryanemerson: Agreed. I have created [ISPN-12752](https://issues.redhat.com/browse/ISPN-12752) to address this.

pkg/controller/batch/batch_controller.go
ryanemerson: Updated as a new commit. Let me know what you think. I'll rebase if you're happy with the changes.

dmvolod: Thanks, @ryanemerson , works excelent

ryanemerson: Great. I've rebased to squash the commit.

dmvolod (approved):

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,355 @@
...
+const (
+	ControllerName  = "batch-controller"
+	BatchFilename   = "batch"
+	BatchVolumeName = "batch-volume"
+	BatchVolumeRoot = "/etc/batch"
+)
+
+var (
+	kubernetes *kube.Kubernetes
+	log        = logf.Log.WithName("controller_batch")
```
dmvolod, line 39:
```suggestion
	log        = logf.Log.WithName(ControllerName)
```

pkg/controller/batch/batch_controller.go
```diff
@@ -0,0 +1,355 @@
...
+		Spec: batchv1.JobSpec{
+			BackoffLimit: pointer.Int32Ptr(0),
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Name:    batch.Name,
+						Image:   infinispan.ImageName(),
```
dmvolod, line 222: @ryanemerson is it possible to use small CLI image instead of full server image for the Job?

dmvolod: Thanks, make sense.

rigazilla (approved):

PR #832, opened 2021-02-18
Utilize CreateOrUpdate for all Infinispan CR's updates in test

dmvolod:


rigazilla (approved):

test/e2e/utils/kubernetes.go
```diff
@@ -236,10 +226,26 @@ func (k TestKubernetes) GracefulRestartInfinispan(infinispan *ispnv1.Infinispan,
...
 		return true, nil
 	})
-
 	ExpectNoError(err)
 }
 
+func (k TestKubernetes) UpdateInfinispan(ispn *ispnv1.Infinispan, update func()) error {
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), k.Kubernetes.Client, ispn, func() error {
+		if ispn.CreationTimestamp.IsZero() {
+			return errors.NewNotFound(ispnv1.Resource("infinispan.org"), ispn.Name)
```
ryanemerson, line 235: If we thrown a `NotFound` err on a create event, why not just use `client.Update` directly?

dmvolod: CreateOrUpdate do much more than update, it refresh data from CR, compare and Update if needed. We can create or own solution, but I'm prefer to use a library functions.

ryanemerson: Right makes sense. I thought I was probably missing something.

ryanemerson:
Thanks @dmvolod

Issue #609, opened 2020-10-13
User should be able to deploy custom code for Infinispan


rigazilla:
The operator should support the deployment of custom code with the image.

Some possible approaches:
1. pack the custom code into an image and mount it as a sidecar that shares file system.
1. s2i image workflow could be applied on the solution above
1. provide a full s2i implementation that builds custom code on top of the standard Infinispan image.

The final product should work with OLM/Catalog/Hub.

ADDED:
Operator should take care of all the integration details, i.e.: deploying sidecar, mounting fs, configuring Infinispan

ryanemerson:
My vote is for 1. The side-car approach provides a clean separation between the user dependencies and the base server, with k8 mounts linking the two. This also would have zero impact on the OLM manifest as we always use our images for operator deployments and the custom image(s) for the side-car. Users can provide any custom images via the Infinispan CR spec, with updates to the field triggerring a redeployment of the cluster.

Automating this process via 2 is also worth investigating, but we need to ensure we have a solution for both openshift and k8.

dmvolod:
> * pack the custom code into an image and mount it as a sidecar that shares file system.

I'm not sure that this approach should be an easy than build the own image with docker/podman, push to the registry and use custom image option in the Infinispan CR. That should be measured between approaches for customer usability and spending time 
 
> * s2i image workflow could be applied on the solution above

This should be an optimal option
> * provide a full s2i implementation that builds custom code on top of the standard Infinispan image.

This can be too much complex, need to understand how many cases customer wants to use some additional code in image and possible code delivery options in real life (binary, source local or in git-based repository)

rigazilla:
How about just mount a PV with the jars in?

dmvolod:
> How about just mount a PV with the jars in?

it seems that you will need to create a Pod before putting the data to the PV
https://dzone.com/articles/kubernetes-copying-a-dataset-to-a-statefulsets-persistentvolume

Crumby:
>> How about just mount a PV with the jars in?

> it seems that you will need to create a Pod before putting the data to the PV.

Providing extra single shared PV between all the instances by default could be an another option. User would just need to rsync the jars as it's not possible to put the data in PV before it's mount to a pod. Issue in here is that the server doesn't support hot deployments so we would need to trigger the deployment somehow. Maybe upon change in `spec.deployments.xyz`?

Creating `Deployment CR` would be probably taking heavy machinery for a simple task?

[...]

PR #833, opened 2021-02-18
Allow user to run custom libs #609

rigazilla:
This PR allows the user to add custom libraries to Infinispan server.
User must setup a PV with the custom libraries in (i.e. an NFS folder or prepare with platform specific tools see [GKE example](https://stackoverflow.com/questions/57754103/how-to-pre-populate-a-readonlymany-persistent-volume))
User must create a PVC bound to that PV with a `ReadOnlyMany` access mode, i.e.:
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: my-readonly-pvc
spec:
  accessModes:
    - ReadOnlyMany
  volumeName: custom-libs-volume-claim-name
  resources:
    requests:
      storage: 100Mi
```
I this example PVC is bound via `volumeName`, user can adopt other binding strategies.

CR should be configured like:
```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: example-infinispan
spec:
  replicas: 2
  dependencies:
    volumeClaimName: custom-libs-volume-claim-name
```


ryanemerson:
@rigazilla WDYT to adding a `spec.dependencies` element like I described in #609? This would allow us to provide the ability to mount a dependency volume now, but in the future we can expand this to resolve multiple dependency times. For example:

```go
type InfinispanSpec struct {
	Replicas  int32                   `json:"replicas"`
	Image     *string                 `json:"image,optional,omitempty"`
        ...
        Dependencies InfinispanExternalDependencies  `json:"dependencies,optional,omitempty"`
}

type InfinispanExternalDependencies struct {
//     Images []InfinispanImageDependency  `json:"images,optional,omitempty"`
//     Jars []InfinispanMavenDependency  `json:"mavenDependencies,optional,omitempty"`
//     Repositories []InfinispanGitDependency `json:"respositories,optional,omitempty"`
       Volume string `json:"volumes,optional,omitempty"`
}
```

rigazilla:
> @rigazilla WDYT to adding a `spec.dependencies` element like I described in #609?

@ryanemerson at first glance sounds good to me, I'll implement `spec.dependencies`

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -203,6 +203,26 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	// Persistent volume claim for custom variables
+	if infinispan.Spec.Container.CustomLibrariesPVName != "" {
+		customPvc := &corev1.PersistentVolumeClaim{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: infinispan.Namespace, Name: infinispan.Name + "-custom-libraries"}, customPvc)
+		if err != nil {
+			if errors.IsNotFound(err) {
```
dmvolod, line 211: We have `LookupResource` func for this case

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -203,6 +203,26 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	// Persistent volume claim for custom variables
+	if infinispan.Spec.Container.CustomLibrariesPVName != "" {
+		customPvc := &corev1.PersistentVolumeClaim{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: infinispan.Namespace, Name: infinispan.Name + "-custom-libraries"}, customPvc)
```
dmvolod, line 209: It's better to use `fmt.Sprintf` for this

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -203,6 +203,26 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				reqLogger.Info(fmt.Sprintf("Creating '%s'", infinispan.Spec.Container.CustomLibrariesPVName))
+				customPvc.Name = infinispan.Name + "-custom-libraries"
+				customPvc.Namespace = infinispan.Namespace
+				customPvc.Spec.AccessModes = []corev1.PersistentVolumeAccessMode{"ReadOnlyMany"}
+				customPvc.Spec.Resources.Requests = corev1.ResourceList{corev1.ResourceStorage: consts.DefaultPVSize}
+				customPvc.Spec.VolumeName = infinispan.Spec.Container.CustomLibrariesPVName
+				err = r.client.Create(context.TODO(), customPvc)
```
dmvolod, line 218: Not sure that we need to create PVC if it not present

pkg/controller/infinispan/infinispan_controller.go
rigazilla: yeah. Also I'm not sure if we want to keep pvc creation here or join it with the stateful set creation

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I guess we need it, this pvc is for multiple read access and it's one per cluster so it cannot be created by the statefulset with the pvc template mechanism.

dmvolod: Yes, but I'm wonder how customer will upload libraries to this PV and how server will handle this data
If it's already provided PV with uploaded libraries and linked to the SS it will work, but not sure that created PV by operator should work and handle by server without rebooting

rigazilla: Yeah the PV must be provided by the user.
Operator doesn't create PV, this is up to the user: it should be pretty straightforward to do that with NFS, also GKE provides some tools for that, or he can rely on an ad-hoc data importer image.
Once the PV is prepared user just have to configure the PV name into the infinispan RC.

dmvolod: Yes, thanks. It would be nice to place all storage logic inside SS reconciler.

ryanemerson: > Operator doesn't create PV, this is up to the user: it should be pretty straightforward to do that with NFS, also GKE provides some tools for that, or he can rely on an ad-hoc data importer image.

@rigazilla Do you have a link to any docs or blogs on how todo this?

rigazilla: I'm more familiar with NFS, steps are:
- setup an nfs folder reachable by the ispn nodes
- represent your nfs folder in k8s, ie:
```
apiVersion: v1
kind: PersistentVolume
metadata:
  name: example-custom-libraries
spec:
  storageClass: "custom-libraries"
  accessModes:
  - ReadOnlyMany
  capacity:
    storage: 1Gi
  nfs:
    path: /opt/nfs/custom-libraries
    server: mynode.mydomain.com
  persistentVolumeReclaimPolicy: Retain
  volumeMode: Filesystem
```
- specify the PV name in the infinispan CR

For GKE I've found [this](https://stackoverflow.com/questions/57754103/how-to-pre-populate-a-readonlymany-persistent-volume), though I didn't dig into too much

Creating an ad hoc pod should be as easy as run a busybox image with the PV mounted and scp there the custom jars. Seems there's also a [general solution](https://github.com/kubevirt/containerized-data-importer).

pkg/controller/infinispan/infinispan_controller.go
rigazilla: Is this related to performance? I didn't find any clear answer around

rigazilla:
I changed the code to mount custom-libs under `/opt/infinispan/server/lib/custom-libraries`, this way is still possible to mount other things in `/opt/infinispan/server/lib`

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Not sure that's performance related. It's much better for code reading and avoiding typos

dmvolod:
> I changed the code to mount custom-libs under `/opt/infinispan/server/lib/custom-libraries`, this way is still possible to mount other things in `/opt/infinispan/server/lib`

Do we need to add this path to CLASSPATH or runner script will do this automatically?

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: Imo `Sprintf` is advantagous when doing something that requires multiple levels of string concatenation, e.g. `SprintF("%s:%s", str1, str2)` which otherwise would be `str1 + ":" + str2`. I think when it's simply adding a suffix, using `+ str` is fine, maybe even more readable. This comes down to preference though, so I'm happy either way for the prefix or suffix case.

rigazilla: it was just curiosity, I'm also happy with both the syntax. Changed to sprintf

pkg/controller/infinispan/infinispan_controller.go
rigazilla: `LookupResource` doesn't fit very well here, it supposes that the resources is going to be created by another controller (i.e. service or secret). I would keep pvc creation tight to statetulset creation.

dmvolod: Yeah, agreed

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -184,6 +184,13 @@ type Autoscale struct {
 	Disabled           bool  `json:"disabled,optional,omitempty"`
 }
 
+// InfinispanExternalDependencies describes all the external dependencies
+// used by the Infinispan cluster: i.e. lib folder with custom jar, maven artifact, images ...
+type InfinispanExternalDependencies struct {
+	// Name of the persistent volumes with custom libraries
+	CustomLibrariesPVName string `json:"customLibrariesPVName,optional,omitempty"`
```
dmvolod, line 191: Probably simple `volume` name much easy for users that's complex one

rigazilla:
> > I changed the code to mount custom-libs under `/opt/infinispan/server/lib/custom-libraries`, this way is still possible to mount other things in `/opt/infinispan/server/lib`
> 
> Do we need to add this path to CLASSPATH or runner script will pickup jars automatically?

infinispan 12 automatically does the job

dmvolod:
> infinispan 12 automatically does the job

Cool, thanks.
Does it means, that if we will implement multiple operand with operator this will not work for Infinispan 11.x?

ryanemerson:
> Does it means, that if we will implement multiple operand with operator this will not work for Infinispan 11.x?

Multiple operand would always have to be from a defined version going forward. It's not something we have to worry about it here. Furthermore, it seems like multi-operand support might not be necessary after all.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -682,6 +684,43 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
...
+		customPvc := &corev1.PersistentVolumeClaim{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{Namespace: m.Namespace, Name: fmt.Sprintf("%s-custom-libraries", m.Name)}, customPvc)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				reqLogger.Info(fmt.Sprintf("Creating '%s'", m.Spec.Dependencies.CustomLibrariesPVName))
+				customPvc.Name = fmt.Sprintf("%s-custom-libraries", m.Name)
+				customPvc.Namespace = m.Namespace
+				customPvc.Spec.AccessModes = []corev1.PersistentVolumeAccessMode{"ReadOnlyMany"}
+				customPvc.Spec.Resources.Requests = corev1.ResourceList{corev1.ResourceStorage: consts.DefaultPVSize}
+				customPvc.Spec.VolumeName = m.Spec.Dependencies.CustomLibrariesPVName
```
ryanemerson, line 714: Instead of updating individual fields on `customPvc` can you update this to create a new object like we do elsewhere as it makes it much easier to read. e.g:

```go
customPvc = corev1.PersistentVolumeClaim{ObjectMeta: metav1.ObjectMeta{
			Name:      ...,
			Namespace: m.Namespace,
		},
			Spec: corev1.PersistentVolumeClaimSpec{
				AccessModes: []corev1.PersistentVolumeAccessMode{
					corev1.ReadOnlyMany,
				},
				Resources: corev1.ResourceRequirements{
					Requests: corev1.ResourceList{
						corev1.ResourceStorage: consts.DefaultPVSize,
					},
				},
			},
		}
```

rigazilla: I need to pass that instance to the client.Get method, is it safe to do? I mean on not found my instance is returned unchanged?

dmvolod: Seems to a be a problem if we need to update it, not a problem with `Get` and `Create` however `CreateOrUpdate` is a an option or two entries with `ObjectMeta` only and another whole structure to Create

dmvolod:
Also I'm wonder if we could create a test this feature with Travis CI, but not sure that's so easy to create interoperable test for both Travis and standalone

pkg/controller/infinispan/infinispan_controller.go
rigazilla: changed creation to literals. hard to say if go readability has been improved :D

rigazilla:
> Also I'm wonder if we could create a test this feature with Travis CI, but not sure that's so easy to create interoperable test for both Travis and standalone

yeah I can't imagine a simple test for this. I would be great to have a mock for the client an check calls in a unit test.

dmvolod:
> > Also I'm wonder if we could create a test this feature with Travis CI, but not sure that's so easy to create interoperable test for both Travis and standalone
> 
> yeah I can't imagine a simple test for this. I would be great to have a mock for the client an check calls in a unit test.

@rigazilla do you know any simple use-case where customer can use a custom code. Probably we can try to adopt this build and wrap inside test for this feature. I know that's possible to use a custom based store, i.e. database but it's not easy to use for the rest. Any other simple cases? Probably @ryanemerson or @tristantarrant could help with ideas.

rigazilla:
> > > Also I'm wonder if we could create a test this feature with Travis CI, but not sure that's so easy to create interoperable test for both Travis and standalone
> > 
> > 
> > yeah I can't imagine a simple test for this. I would be great to have a mock for the client an check calls in a unit test.
> 
> @rigazilla do you know any simple use-case where customer can use a custom code. Probably we can try to adopt this build and wrap inside test for this feature. I know that's possible to use a custom based store, i.e. database but it's not easy to use for the rest. Any other simple cases? Probably @ryanemerson or @tristantarrant could help with ideas.

I've used the hello-task use case for dev: (8.1.1)[https://infinispan.org/docs/dev/titles/server/server.html#tasks].
Step to reproduce are:
- build the .jar
- copy into the nfs folder
- create the PV
- configure the infinispan CR to use it
- deploy the CR
- execute the remote task via rest or CLI

dmvolod:
> Step to reproduce are:
> 
> * build the .jar
> * copy into the nfs folder
> * create the PV
> * configure the infinispan CR to use it
> * deploy the CR
> * execute the remote task via rest or CLI

Thanks, @rigazilla . We could try to repeat these steps on the CI and for standalone tests as well

ryanemerson:
+1 to creating an automated test that runs on CI, but I think this can come later as we need this functionality ASAP for demo purposes.

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: Should the user be providing a reference to `PersistentVolumClaim` instead of a `PersistentVolume`? The latter seems too low level.

There is a one-to-one mapping between a PVC and a PV, so if we create the PVC for the user this means that the read-only contents can only be used by a single Infinispan CR at once.

Consider a case where a company that has multiple Infinispan clusters, but they want to use the same scripts/marshalling/db-drivers in multiple clusters. If we create a PVC based upon the provided volume, it will be necessary for a volume to be created per cluster. However, if we allow the user to provide a `ReadOnlyMany` PVC we can mount that on all of the clusters.

rigazilla: Good point @ryanemerson. 
I would like to keep the creation of the pvc on the operator side though, how about this policy:
- user provides the `volumeName`
- operator looks for the pvc with name `volumeName-infinispan-custom-libraries` and:
      - create it if doesn't exist 
      - otherwise reuses the existent
could it work?

ryanemerson: > I would like to keep the creation of the pvc on the operator side though, how about this policy:

Any particular reason? Creating a PVC is a higher level abstraction than a PV, so IMO this is more user friendly.

What's the advantage of allowing a user to provide a PV instead of a PVC?

> * operator looks for the pvc with name `volumeName-infinispan-custom-libraries` and:
>   - create it if doesn't exist
>   - otherwise reuses the existent
>   could it work?

Imo this seems harder to explain to a user then simply provide the name of an existing PVC.

rigazilla: > > I would like to keep the creation of the pvc on the operator side though, how about this policy:
> 
> Any particular reason? Creating a PVC is a higher level abstraction than a PV, so IMO this is more user friendly.
> 
> What's the advantage of allowing a user to provide a PV instead of a PVC?

Even using the PVC the user must provide the PV name in the PVC, this because he wants to mount a specific volume and isn't asking for generic storage. I'm just suggesting to let the user to only specify the volume name and let the operator do the rest of the PVC configuration.

> 
> > * operator looks for the pvc with name `volumeName-infinispan-custom-libraries` and:
> >   
> >   * create it if doesn't exist
> >   * otherwise reuses the existent
> >     could it work?
> 
> IMO this is harder to explain to a user then "provide the name of an existing PVC".

he also needs to create the PVC if it doesn't exist and it's something that we could easily do for him since we know how we want to use it.

so basically the instruction for the user should be "provide the name of the PV with your jars"

ryanemerson: > Even using the PVC the user must provide the PV name in the PVC, this because he wants to mount a specific volume and isn't asking for generic storage.

That doesn't have to be the case. With a PVC the user can utilise generic storage as per the k8 cluster defaults.

For example, consider the busybox way of populating the data that you described. The user doesn't care what type of storage is used, so they just create a PVC with `ReadWriteOnce`, the PVC and volume are then mounted to a pod where they copy the required the files, the pod is then shutdown. This PVC is then mounted by the operator as readOnly in one or more Infinispan clusters.

Side note, but this is something that we could simplify with the CLI.

> he also needs to create the PVC if it doesn't exist and it's something that we could easily do for him since we know how we want to use it.

It's possible that the user just creates a PVC and names that, no need to create a volume.

I'm not sure, but in some environments doesn't the creation of a Volume require more permissions than a PVC?

> so basically the instruction for the user should be "provide the name of the PV with your jars"

Fair point. I was thinking that the user would need to be aware of this behaviour, but the operator can just do this behind the scenes.

rigazilla: @ryanemerson I though it over, your approach is ok, let go ahead and leave the pvc creation to the user for now.

rigazilla:
updated the initial description with the new use case with PVC

dmvolod (approved):

ryanemerson:
Thanks @rigazilla

@oraNod Docs for this will have to come in a separate PR as we need these changes ASAP

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: Thanks for changing :+1:

This should greatly simplify creating a test for deployments, as we can do the following using the default volume type:

1. Create PVC
2. Mount PVC to pod
3. Copy files to pod
4. Shutdown pod
5. Start server with pvc and verify that the provided task etc is available via REST.

PR #835, opened 2021-02-19
Split Core and Backup/Restore tests with separate stages

dmvolod:
This changes able to run separate tests (Core/Multinamespace and Backup/Restore) in parallel and add additional separation and visibility to them

image


.travis.yml
```diff
@@ -29,30 +29,37 @@ jobs:
...
-        - make test PARALLEL_COUNT=2 && make backuprestore-test && kubectl config use-context $TESTING_CONTEXT && make multinamespace-test
+        - make test PARALLEL_COUNT=2 && kubectl config use-context $TESTING_CONTEXT && make multinamespace-test
+      after_failure:
+        - cat debug.log
+        - kubectl config use-context $TESTING_CONTEXT
+        - kubectl get events --all-namespaces
+        - kubectl cluster-info
+        - df -h
+        - docker ps -a
+        - for log in $(docker ps -qa | xargs); do docker logs --tail 500 $log; done
```
ryanemerson, line 51: Should we add the `after_failure` steps to a script so that we can just reference this in each stage?

.travis.yml
```diff
@@ -29,30 +29,37 @@ jobs:
...
-        - export TESTING_CONTEXT=`kubectl --insecure-skip-tls-verify config current-context`
-        - ./build/prepare-sa-test.sh
-        - ./build/debug-pods.sh &
-      script:
+        - source ./build/travis/configure-travis-ci.sh
         - go mod vendor
+      script:
         # Run multinamespace tests as admin (skip role settings check)
-        - make test PARALLEL_COUNT=2 && make backuprestore-test && kubectl config use-context $TESTING_CONTEXT && make multinamespace-test
+        - make test PARALLEL_COUNT=2 && kubectl config use-context $TESTING_CONTEXT && make multinamespace-test
```
ryanemerson, line 43: WDYT to adding a stage per test category? Currently this would mean ["main", "multinamespace", "backup/restore"] I can then can update my Batch PR to add an additional stage.

.travis.yml
dmvolod: Yes, it could be better, but using single script will loose expanding data for each category

.travis.yml
dmvolod: It cold be an option, a bit overbooked resources with multi-namespace (short test will require a new VM)

.travis.yml
ryanemerson: Argh I see. I didn't realise that. I had a quick google to see if you can somehow create a reusable version of those commands, but I couldn't find anything.

ryanemerson (approved):

ryanemerson:
Thanks @dmvolod

PR #836, opened 2021-02-19
[2.0.x] doc issue #812 client connection details

oraNod:
Backport for fixes to #812


documentation/asciidoc/topics/proc_creating_caches_hotrod.adoc
```diff
@@ -1,90 +1,71 @@
...
+    manager.administration().getOrCreateCache(cacheName, new XMLStringConfiguration(xml));
+    System.out.println("Cache with configuration exists or is created.");
+}
+----
+
+[discrete]
+== Using Hot Rod client properties
 
-.Reference
+When you invoke `cache.get()` calls for named caches that do not exist, {brandname} creates them from the Hot Rod client properties instead of returning null.
```
rigazilla, line 61: this is not clear to me what is happening here.
is this a `cacheManager.getCache("my-cache")` actually and this call creates a cache from the properties?
Or is really a `cache.get("key")` ? in this case, if the cache is created with that call I'm expecting to have a `null` retuned.

oraNod: Hmm... Good catch. It is `cacheManager.getCache("my-cache");`

For example: https://github.com/infinispan/infinispan-simple-tutorials/blob/00d80cc878ebf98684e0eff2f7249adf5c6fb200/remote-per-cache/src/main/java/org/infinispan/tutorial/simple/remote/percache/InfinispanRemotePerCache.java#L46

rigazilla (approved):

rigazilla:
thanks @oraNod !

PR #839, opened 2021-02-24
Revisit CreateOrUpdate usage (separate Spec and Status.Conditions updates)

dmvolod:


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -630,13 +635,17 @@ func (r *ReconcileInfinispan) scheduleUpgradeIfNeeded(infinispan *infinispanv1.I
...
 	if podDefaultImage != desiredImage {
+		logger.Info("schedule an Infinispan cluster upgrade", "pod default image", podDefaultImage, "desired image", desiredImage)
 		if err := r.update(infinispan, func() {
-			logger.Info("schedule an Infinispan cluster upgrade", "pod default image", podDefaultImage, "desired image", desiredImage)
-			infinispan.SetCondition(infinispanv1.ConditionUpgrade, metav1.ConditionTrue, "")
 			infinispan.Spec.Replicas = 0
 		}); err != nil {
 			return &reconcile.Result{}, err
 		}
+		if err := r.update(infinispan, func() {
```
ryanemerson, line 644: Won't the initial call to `r.update(..infinispan.Spec.Replicas = 0..)` result in a new reconcile event? as well as the subsequent call to `r.update` for the condition?

dmvolod: Any update of the `Spec` should raise a new reconcile event, but `Status` not. `CreateOrUpdate` has the same logic inside. It's sequential `Spec` and `Status` update if needed. There are no more changes. However we could review shutdown/startup logic to avoid more `Spec` updates.

ryanemerson (approved):

rigazilla (approved):
lgtm

ryanemerson:
Thanks @dmvolod

PR #840, opened 2021-02-24
[2.0.x] Revisit CreateOrUpdate usage (separate Spec and Status.Conditions updates)

dmvolod:


dmvolod:
Close, as fix need to be placed for `controllerutil.CreateOrPatch` code

PR #842, opened 2021-02-24
Docs: remove target labels for services

oraNod:
#821 Removes target labels for services.


rigazilla (commented):
I would keep the documentation about the service labels: feature is still available to the user.
and just remove the downstream part for service

documentation/asciidoc/topics/ref_label_environment_variables.adoc
```diff
@@ -1,6 +1,6 @@
 [id='label_environment_variables-{context}']
 = Global Labels for {ispn_operator}
-Global labels are automatically propagated to all {brandname} resources.
+Global labels are automatically propagated to all {brandname} pods.
```
rigazilla, line 3:
```suggestion
Global labels are automatically propagated to all {brandname} pods and services.
```

documentation/asciidoc/topics/yaml/publisher_labels_downstream.yaml
```diff
@@ -1,13 +1,3 @@
-# Defines global labels for services.
```
rigazilla, line 1: Can we leave a comment that `INFINISPAN_OPERATOR_TARGET_LABELS` is also available for services though we don't use it atm?

oraNod: @rigazilla Sure but we should expose `INFINISPAN_OPERATOR_TARGET_LABELS` in community docs only, right? Or is it also possible to have this downstream but it's just ignored at the moment?

documentation/asciidoc/topics/ref_label_environment_variables.adoc
oraNod: @rigazilla Thanks, this relates to my other question about the svc labels. I changed "resources" to "pods" because it felt like "resources" was too vague. 

If we say "pods and services" will that make sense downstream?

documentation/asciidoc/topics/yaml/publisher_labels_downstream.yaml
rigazilla: @oraNod it's unused atm, but it's possible to have it downstream also.

documentation/asciidoc/topics/ref_label_environment_variables.adoc
rigazilla: I think it makes sense if we add the comment below about `INFINISPAN_OPERATOR_TARGET_LABELS`

rigazilla (approved):
lgtm

documentation/asciidoc/topics/yaml/publisher_labels_downstream.yaml
```diff
@@ -1,14 +1,11 @@
 # Defines global labels for services.
 - name: INFINISPAN_OPERATOR_TARGET_LABELS
```
Crumby, line 2: Does it makes sense to provide operator env labels for downstream? User can't reasonable modify them. (User would have to set them everytime he upgrades in OLM deployments)

documentation/asciidoc/topics/yaml/publisher_labels_downstream.yaml
```diff
@@ -1,14 +1,11 @@
...
-     "com.redhat.product-name":"{product-name}",
-     "com.redhat.product-version":"{product-version}",
-     "custom-svc-label1":"svc-value1",
-     "custom-svc-label2":"svc-value2"}
+    {"svc-label1":"svc-value1",
+     "svc-label2":"svc-value2"}
 # Defines global labels for pods.
+# The operator applies com.redhat.* labels automatically.
+# You can remove com.redhat.* labels if necessary.
 - name: INFINISPAN_OPERATOR_POD_TARGET_LABELS
```
Crumby, line 9: Same as above

documentation/asciidoc/topics/yaml/publisher_labels_downstream.yaml
oraNod: @rigazilla WDYT? Sounds like maybe it should be community only?

rigazilla: @oraNod actually we're the only publisher for downstream, so it's ok to not show this info to the customer.
If we want to go this way I would remove the whole chapter "Global labels for " downstream.

oraNod: @rigazilla and @Crumby thanks, guys. I've removed the global labels section from downstream.

Crumby (approved):

oraNod:
self-merge. thanks for the reviews and approvals @rigazilla and @Crumby

Issue #847, opened 2021-02-25
Add automated upgrade test


ryanemerson:
Currently the only way to test the controller behaviour during an upgrade is to run the manual `make upgrade-test`. In addition to this we should add an automated upgrade test that ensures that the GracefulShutdown upgrade policy works as expected. It's not necessary for this to test an upgrade from one channel to another, e.g. 2.0.x -> 2.1.x, but instead we can check that the rolling back of the statefulset occurs by manually setting the DEFAULT_IMAGE env var on the operator, installing a cluster and then restarting the operator with a different DEFAULT_IMAGE value.

This will help with implementing #658 as additional tests will also have to be created for this use-case.

ryanemerson:
https://github.com/infinispan/infinispan-operator/pull/1248

Issue #658, opened 2020-11-20
Upgrade via cluster migration


rigazilla:
A zero downtime zero dataloss upgrade procedure based on remote-store and hotrod data migration should be implemented.
Design here: https://github.com/infinispan/infinispan-designs/pull/21

PR #848, opened 2021-02-25
Add automated upgrade test #847

ryanemerson:


test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,83 @@
...
+	},
+}
+
+func TestGracefulShutdown(t *testing.T) {
+	namespace := tutils.Namespace
+	testKube.NewNamespace(namespace)
+	// Utilise the sha256 of the current Infinispan image to trick the operator into thinking a upgrade is required
+	oldImage := getDockerImageSha()
+	tutils.ExpectNoError(os.Setenv("DEFAULT_IMAGE", oldImage))
+	stopCh := testKube.InstallAndRunOperator(namespace, CrdPath, false)
```
dmvolod, line 45: @ryanemerson I can't understand how this could help with real operator upgrade. The operator upgrade idea is that initial infra (Pods, Services, Routes, etc.) and CR's are forming with one operator version and after upgrade become to continue with process inside another one.

dmvolod: Oh, I see the case in the description, this is not full life-cycle test, but just an upgrade inside one operator version.

ryanemerson: This allows us to verify that, upon detecting that an upgrade is required, the correct actions are taken by the Operator. In the case of GracefulShutdown this requires scaling down/up the StatefulSet. When #658 is implemented, we will need to create a similar test to ensure that zero-downtime migration occurs.

These tests are a mock of an upgrade and allow for us to ensure that a given PR does not break the upgrade workflow. It does not protect against version incompatibilities and is intended to supplement, rather than replace, the existing manual tests.

ryanemerson:
My changes are causing port conflicts on CI, so I must have messed up the cleaning of test resources. I'll take another look.

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func getDockerImageSha() string {
+	sha := tutils.ImageSha
+	if sha != "" {
+		return sha
+	}
+	// If the INFINISPAN_IMAGE_SHA env variable has not been set, attempt to retreive
+	cmd := exec.Command("skopeo inspect docker://quay.io/infinispan/server:12.0 | jq -r '.Digest'")
```
dmvolod, line 79: @ryanemerson I'm wonder why you are trying to inspect local docker storage for containers which is using inside `kind` since
- `kind` not integrated with local docker registry by default, see https://kind.sigs.k8s.io/docs/user/local-registry/ how to implement this
- `kind` uses `containerd` inside for running images, not `docker` runtime
```
00:37 $ docker ps
CONTAINER ID   IMAGE                   COMMAND                  CREATED       STATUS      PORTS                                                 NAMES
7c4dff1916ca   kindest/node:v1.16.15   "/usr/local/bin/entr…"   10 days ago   Up 4 days   127.0.0.1:37351->6443/tcp, 0.0.0.0:11222->30222/tcp   kind-control-plane
00:37 $ docker exec -it 7c4dff1916ca /bin/bash
root@kind-control-plane:/# ctr --namespace=k8s.io images list | grep infinispan
quay.io/infinispan/server:12.0                                                                    application/vnd.docker.distribution.manifest.v2+json sha256:9c15fbaf399e4087d8cf4e0f98b32f0a30ad9171e0a2f97fe3fba64a3ee93473 192.2 MiB linux/amd64 io.cri-containerd.image=managed 
quay.io/infinispan/server:12.1                                                                    application/vnd.docker.distribution.manifest.v2+json sha256:d9c765412f4b4889457b3e75934137c659057e75d4226cbc04ab76fd9a32a3ad 192.5 MiB linux/amd64 io.cri-containerd.image=managed 
quay.io/infinispan/server@sha256:9c15fbaf399e4087d8cf4e0f98b32f0a30ad9171e0a2f97fe3fba64a3ee93473 application/vnd.docker.distribution.manifest.v2+json sha256:9c15fbaf399e4087d8cf4e0f98b32f0a30ad9171e0a2f97fe3fba64a3ee93473 192.2 MiB linux/amd64 io.cri-containerd.image=managed 
quay.io/infinispan/server@sha256:d9c765412f4b4889457b3e75934137c659057e75d4226cbc04ab76fd9a32a3ad application/vnd.docker.distribution.manifest.v2+json sha256:d9c765412f4b4889457b3e75934137c659057e75d4226cbc04ab76fd9a32a3ad 192.5 MiB linux/amd64 io.cri-containerd.image=managed
```

ryanemerson: I'm trying to retrieve the sha256 of the remote image, that's why I am installing `skopeo` to perform the inspection.

dmvolod: Oh, sure. But you can pull it and use a local inspection to reduce complexity. It the image already using by operator, it can be inspected also.

ryanemerson: > Oh, sure. But you can pull it and use a local inspection to reduce complexity.

Won't that result in the image being pulled per CI run in the k8 cluster AND the local machine? Either way it still requires the execution of an external process, so I don't think it increases complexibility particularly.

> It the image already using by operator, it can be inspected also.

The image is pulled into the k8 cluster though, not the local repository where the tests are being executed. Do you know how we could get the digest of the image from k8s? Keep in mind that currently no Infinispan cluster is created before the sha is required, so we can't inspect the pods.

dmvolod: I think a single way to integrate local docker repository with Kind and pull images before checking sha256.
It can be also load images from the local repository into the `Kind` using 'kind load ...` command
Don't see any other ways and seems to be there is no way to use this kind of tests for both CI and personal (remote OCP/OKD/minikube cluster usage) usage. But need to investigate more.

ryanemerson: One hacky solution that doesn't involve any external processes and should work on all envs is to do the following before the test begins:

1. Create a standalone pod with the server image
2. Retrieve the image from the spec as it's in the sha256 format
3. Cleanup pod and execute the test

ryanemerson:
Closing in favour of https://github.com/infinispan/infinispan-operator/pull/1248

PR #849, opened 2021-02-25
2.0.x: [test-integration] Stabilize Autoscaling test

Crumby:
Prevents test failures in case last entry is removed from the cluster but the pods above the minimum weren't terminated yet throwing out `IllegalStateException`.


ryanemerson:
Thanks  @Crumby

PR #853, opened 2021-02-25
Tests should evaluates for cluster is wellFormed

dmvolod:


ryanemerson:
@dmvolod The change makes sense :+1:. Can you add these checks to `backup_restore_test.go`?

dmvolod:
> @dmvolod The change makes sense . Can you add these checks to `backup_restore_test.go`?

@ryanemerson it seems to Backup/Restore controller do this itself and we don't need to add additional validation

ryanemerson:
> @ryanemerson it seems to Backup/Restore controller do this itself and we don't need to add additional validation

Good point! I had forgot about that :smile:

ryanemerson (approved):

rigazilla (approved):

ryanemerson:
Thanks @dmvolod

Issue #844, opened 2021-02-24
Dockerfile still uses a base image hosted on Dockerhub


ryanemerson:
The `Dockerfile` used to build images on quay.io still references the `golang` image that is hosted on DockerHub. We should update this to an alternative.

For example: https://catalog.redhat.com/software/containers/ubi8/go-toolset/5ce8713aac3db925c03774d1?gti-tabs=unauthenticated&container-tabs=gti

PR #854, opened 2021-02-25
Dockerfile still uses a base image hosted on Dockerhub #844

ryanemerson:


dmvolod:
Do we need to update Dockerfile.single file also?

ryanemerson:
> Do we need to update Dockerfile.single file also?

Updated. Previously it wasn't using a tag, so it would have pulled the latest 8.3 image anyway, but I have explicitly set it now so that the two Dockerfiles are the same.

dmvolod (approved):

PR #860, opened 2021-03-02
Status conditions should indicate that Cross-Site view is also formed

dmvolod:
This is example of feature which is strongly bind to the specific Infinispan server version, in this case 12.1.0 need to be released

### DIFF
--- a/documentation/asciidoc/topics/proc_configuring_xsite_within_clusters.adoc
+++ b/documentation/asciidoc/topics/proc_configuring_xsite_within_clusters.adoc
@@ -38,9 +38,12 @@ spec:
 +
 . Configure nodes with any other {datagridservice} resources.
 . Apply the `Infinispan` CRs.
-. Check node logs to verify that {brandname} clusters form a cross-site view, for example:
+. Verify that {brandname} clusters form a cross-site view.
+.. Retrieve the `Infinispan` CR.
 +
 [source,options="nowrap",subs=attributes+]
 ----
-include::cmd_examples/oc_grep_xsite.adoc[]
+include::cmd_examples/get_infinispan.adoc[]
 ----
++
+.. Check for the `type: CrossSiteViewFormed` condition.
--- a/pkg/apis/infinispan/v1/infinispan_types.go
+++ b/pkg/apis/infinispan/v1/infinispan_types.go
@@ -246,11 +246,12 @@ type InfinispanSpec struct {
 type ConditionType string
 
 const (
-	ConditionPrelimChecksPassed ConditionType = "PreliminaryChecksPassed"
-	ConditionGracefulShutdown   ConditionType = "GracefulShutdown"
-	ConditionStopping           ConditionType = "Stopping"
-	ConditionUpgrade            ConditionType = "Upgrade"
-	ConditionWellFormed         ConditionType = "WellFormed"
+	ConditionPrelimChecksPassed  ConditionType = "PreliminaryChecksPassed"
+	ConditionGracefulShutdown    ConditionType = "GracefulShutdown"
+	ConditionStopping            ConditionType = "Stopping"
+	ConditionUpgrade             ConditionType = "Upgrade"
+	ConditionWellFormed          ConditionType = "WellFormed"
+	ConditionCrossSiteViewFormed ConditionType = "CrossSiteViewFormed"
 )
 
 // InfinispanCondition define a condition of the cluster
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -324,17 +324,9 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		reqLogger.Info("Pods IPs are not ready yet")
 		return reconcile.Result{}, r.update(infinispan, func() {
 			infinispan.SetCondition(infinispanv1.ConditionWellFormed, metav1.ConditionUnknown, "Pods are not ready")
+			infinispan.RemoveCondition(infinispanv1.ConditionCrossSiteViewFormed)
 			infinispan.Status.StatefulSetName = statefulSet.Name
 		})
-	}
-
-	// If x-site enable configure the coordinator pods to be selected by the x-site service
-	if infinispan.HasSites() {
-		found := applyLabelsToCoordinatorsPod(podList, cluster, r.client, reqLogger)
-		if !found {
-			// If a coordinator is not found then requeue early
-			return reconcile.Result{Requeue: true}, nil
-		}
 	}
 
 	// All pods ready start autoscaler if needed
@@ -354,7 +346,7 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 	// View didn't form, requeue until view has formed
 	if infinispan.NotClusterFormed(len(podList.Items), int(infinispan.Spec.Replicas)) {
 		reqLogger.Info("notClusterFormed")
-		return reconcile.Result{Requeue: true, RequeueAfter: consts.DefaultWaitClusterNotWellFormed}, nil
+		return reconcile.Result{RequeueAfter: consts.DefaultWaitClusterNotWellFormed}, nil
 	}
 
 	// Below the code for a wellFormed cluster
@@ -440,6 +432,20 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 			infinispan.Status.ConsoleUrl = nil
 		}); err != nil {
 			return reconcile.Result{}, err
+		}
+	}
+
+	// If x-site enable configure the coordinator pods to be selected by the x-site service
+	if infinispan.HasSites() {
+		crossSiteViewCondition, err := r.applyLabelsToCoordinatorsPod(podList, infinispan.GetSitesLocations(""), cluster)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		err = r.update(infinispan, func() {
+			infinispan.SetConditions([]infinispanv1.InfinispanCondition{*crossSiteViewCondition})
+		})
+		if err != nil || crossSiteViewCondition.Status != metav1.ConditionTrue {
+			return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCluster}, err
 		}
 	}
 
@@ -784,10 +790,10 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 							{Name: "CONFIG_HASH", Value: hashString(configMap.Data[consts.ServerConfigFilename])},
 							{Name: "ADMIN_IDENTITIES_HASH", Value: hashString(string(adminSecret.Data[consts.ServerIdentitiesFilename]))},
 						}),
-						LivenessProbe:  PodLivenessProbe(m),
+						LivenessProbe:  PodLivenessProbe(),
 						Ports:          PodPortsWithXsite(m),
-						ReadinessProbe: PodReadinessProbe(m),
-						StartupProbe:   PodStartupProbe(m),
+						ReadinessProbe: PodReadinessProbe(),
+						StartupProbe:   PodStartupProbe(),
 						Resources:      *podResources,
 						VolumeMounts:   volumeMounts,
 					}},
@@ -918,20 +924,20 @@ func PodPortsWithXsite(i *infinispanv1.Infinispan) []corev1.ContainerPort {
 	return ports
 }
 
-func PodLivenessProbe(i *infinispanv1.Infinispan) *corev1.Probe {
-	return probe(i, 5, 10, 10, 1, 80)
+func PodLivenessProbe() *corev1.Probe {
+	return probe(5, 10, 10, 1, 80)
 }
 
-func PodReadinessProbe(i *infinispanv1.Infinispan) *corev1.Probe {
-	return probe(i, 5, 10, 10, 1, 80)
+func PodReadinessProbe() *corev1.Probe {
+	return probe(5, 10, 10, 1, 80)
 }
 
-func PodStartupProbe(i *infinispanv1.Infinispan) *corev1.Probe {
+func PodStartupProbe() *corev1.Probe {
 	// Maximum 10 minutes (60 * 10s) to finish startup
-	return probe(i, 60, 10, 10, 1, 80)
+	return probe(60, 10, 10, 1, 80)
 }
 
-func probe(i *infinispanv1.Infinispan, failureThreshold, initialDelay, period, successThreshold, timeout int32) *corev1.Probe {
+func probe(failureThreshold, initialDelay, period, successThreshold, timeout int32) *corev1.Probe {
 	return &corev1.Probe{
 		Handler: corev1.Handler{
 			HTTPGet: &corev1.HTTPGetAction{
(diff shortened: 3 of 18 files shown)


pruivo:
hmmm. do we really want to include xsite? is it only check during startup?

dmvolod:
> hmmm. do we really want to include xsite? is it only check during startup?

Not only startup, validation performs on every Pod status change, i,e. during rolling upgrade, rescheduling, etc.
Why not using this option, as customer need to inspect logs manually to make sure that cluster is well formed

pruivo:
will it block use access to the site? i.e. not all caches backup to the remote site (internal caches, like the ones used to store server tasks and protobuf schemas, counters, locks). Is the user be able to access them?

Infinispan doesn't have such restriction and sites are independent.They can be connected later and the data can be sync.

dmvolod:
> will it block use access to the site? i.e. not all caches backup to the remote site (internal caches, like the ones used to store server tasks and protobuf schemas, counters, locks). Is the user be able to access them?
> 
> Infinispan doesn't have such restriction and sites are independent.They can be connected later and the data can be sync.

Good catch, thanks @pruivo 
wellFormed status will be block to start Batch/Backup/Restore CR's execution only, not blocking to access to the Router or Service. In this case we could introduce a new status to avoid user to check that xsite view is well formed manually. Do you know cases on OCP/OKD/k8s when users can setup xsite on one site, but not setup on another in suitable time. Is it good approach?

pruivo:
> Do you know cases on OCP/OKD/k8s when users can setup xsite on one site, but not setup on another in suitable time.

not really :rofl:. at least, not in production. I've done some dev/test cases where I only have a site up to test the client. 

I just wanted to make sure the access isn't blocked and users can still reading and writing data to Infinispan.

If it is possible, a separate status would be nicer
* pod joined the local cluster
* x-site set properly

dmvolod:
> If it is possible, a separate status would be nicer

Sure, will introduce a new `Status.Conditions` for this xsite view
@pruivo another case I've found ducting static configuration test: I don't really know how's xsite remote host code looks like, but seems to be host resolving performs once and never resolving later. This means, if second static Infinispan instance (Pods and Services) was not stated during first instance start, xiste view never appears. Is it make sense or I'm doing something wrong

pruivo:
JGroups doesn't have a concept of xsite-view. So, we are creating the view based on the events when site connects or disconnects. The SiteMaster will get the correct view but all the pods started after it, they don't receive any events and don't form the view.

Would it make sense for the REST request to query the SiteMaster for the view? 
An alternative is to check the view and if"coordinator==true"; the pod is always "well formed", in xsite context" if "coordinator=false". 
Does it make sense?
Let me know which one makes your/our life easier :)

dmvolod:
> JGroups doesn't have a concept of xsite-view.

Probably, I not correctly explain situation. Seems to be JGroops not reconnect to the site if IP or  host are not present. In cloud world this is k8s service which is dynamic entity and can be created later.

dmvolod:
> Would it make sense for the REST request to query the SiteMaster for the view?
> An alternative is to check the view and if"coordinator==true"; the pod is always "well formed", in xsite context" if "coordinator=false".
> Does it make sense?
> Let me know which one makes your/our life easier :)

Yeah, it make sense, thanks, but for getting `coordinator` status we need to go around all Pods, but for SitesView, I need to call it once from any Pods or via Service. Also, I got list of sites from one view. You've implemented very desirable and  useful feature exposing sites view to the REST API.

pkg/infinispan/cluster.go
```diff
@@ -117,6 +118,22 @@ func (c Cluster) GetClusterMembers(podName string) ([]string, error) {
...
 }
 
+func (c Cluster) GetSitesView(podName string) (map[string]bool, error) {
+	cacheManagerInfo, err := c.GetCacheManagerInfo(consts.DefaultCacheManagerName, podName)
+	if err != nil {
+		return nil, err
+	}
+	siteViews := cacheManagerInfo["sites_view"]
+	if siteViews == nil {
+		return nil, fmt.Errorf("getting Cross-Site view is not supported")
```
oraNod, line 128: @dmvolod When would you hit this error? Is it more of a case that cross-site is not enabled so it's not possible to get an xsite view?

pkg/controller/infinispan/infinispan_controller_test.go
```diff
@@ -61,6 +61,10 @@ func (m mockCluster) GetClusterMembers(podName string) ([]string, error) {
 	return nil, errors.New("error in getting view")
```
oraNod, line 61: How about here? Could this be when a network condition or some other fault interrupts the operator ability to get the cluster view?

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -214,6 +214,7 @@ const (
 	ConditionStopping           ConditionType = "Stopping"
 	ConditionUpgrade            ConditionType = "Upgrade"
 	ConditionWellFormed         ConditionType = "WellFormed"
+	ConditionCrossSiteFormed    ConditionType = "CrossSiteFormed"
```
oraNod, line 217: @dmvolod Are hyphens allowed or are there character restrictions?

@pruivo What do you think about being consistent with bare metal? While we're at this I'm tempted to propose that we align cluster view messages a little. 

What about:

"WellFormed" -> "ClusterViewFormed" 
"CrossSiteFormed" -> "CrossSiteViewFormed"

If that sounds acceptable I can send a PR to Infinispan and align the messages across platforms.

dmvolod: It would be nice, but I'm not sure that we can change `WellFormed` -> `ConditionWellFormed` now as this cause a migration problems or we should support both `ClusterViewFormed` and `WellFormed` statuses for well formed state.
@rigazilla @ryanemerson what do you think about it?

pkg/infinispan/cluster.go
dmvolod: This problem can be occurred when someone will try to use current operator version with Infinispan server image version lower than 12.1.0 release. It's not supported in downstream but anyone can provide own image version.

pkg/controller/infinispan/infinispan_controller_test.go
dmvolod: @oraNod it's a mock test, please ignore.

pkg/apis/infinispan/v1/infinispan_types.go
oraNod: This might not be the PR to start tackling the conditions but I think we should make sure they are meaningful in isolation and written from a more DevOPs perspective.

@Crumby had a question before about `PreliminaryChecksPassed` because it seemed a little mysterious. IIRC the operator provided an error message and detail on how to resolve it. But the condition itself seemed incomplete or vague. 

For this PR at least I think `CrossSiteViewFormed` is a little more descriptive than just `CrossSiteFormed`. It follows the same language we use in docs "clusters can form a cross-site view" and is clear that we're talking about the view.

Also, maybe nevermind about the hyphens. Let's stick with camel case.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1078,17 +1079,42 @@ func getInfinispanConditions(pods []corev1.Pod, m *infinispanv1.Infinispan, clus
...
+				crossSiteFormed := infinispanv1.InfinispanCondition{Type: infinispanv1.ConditionCrossSiteFormed, Status: metav1.ConditionTrue}
+				sitesView, err := cluster.GetSitesView(pods[0].Name)
+				if err == nil {
+					locations := m.GetSitesLocations("")
+					sites := make([]string, len(locations))
+					for i, location := range locations {
+						sites[i] = location.Name
+						if !sitesView[location.Name] {
+							crossSiteFormed.Status = metav1.ConditionFalse
+							crossSiteFormed.Message = fmt.Sprintf("Site '%s' not ready", location.Name)
```
oraNod, line 1096: @dmvolod When we say the site is not ready, is it that the remote cluster is not fully started and the user just needs to wait? Or could there potentially be an error?

pkg/infinispan/cluster.go
oraNod, line 128:
```suggestion
		return nil, fmt.Errorf("Cross-site replication is not supported with the server image you are using.")
```

oraNod: Thanks. I think the error should just say that. If it only occurs under that condition.

pkg/apis/infinispan/v1/infinispan_types.go
pruivo: > What do you think about being consistent with bare metal? 

@oraNod what do you mean?

dmvolod: > @Crumby had a question before about `PreliminaryChecksPassed` because it seemed a little mysterious. IIRC the operator provided an error message and detail on how to resolve it. But the condition itself seemed incomplete or vague.

This issue should be submitted separately.

dmvolod: I'm not sure we can change `ConditionWellFormed` value in current time and changes, it's better to align types, but need to rewrite a number of code

pkg/controller/infinispan/infinispan_controller.go
dmvolod: I'm not sure we have an additional info for this except situation we discussing with @pruivo privately now

pkg/infinispan/cluster.go
```diff
@@ -125,7 +125,7 @@ func (c Cluster) GetSitesView(podName string) (map[string]bool, error) {
 	}
 	siteViews := cacheManagerInfo["sites_view"]
 	if siteViews == nil {
-		return nil, fmt.Errorf("getting Cross-Site view is not supported")
+		return nil, fmt.Errorf("Cross-site replication is not supported with the server image you are using.")
```
dmvolod, line 128: Oh, @oraNod I missed this, not a Cross-site replication supported, getting Cross-Site view not supported. How to change this phrase?

pkg/apis/infinispan/v1/infinispan_types.go
oraNod: > > What do you think about being consistent with bare metal?
> 
> @oraNod what do you mean?

@pruivo I mean on bare metal you get `ISPN000439: Received new x-site view:` in logs. My thought here was to have the same, or at least similar, message with the Operator.

pkg/infinispan/cluster.go
```diff
@@ -117,6 +118,22 @@ func (c Cluster) GetClusterMembers(podName string) ([]string, error) {
...
 }
 
+func (c Cluster) GetSitesView(podName string) (map[string]bool, error) {
+	cacheManagerInfo, err := c.GetCacheManagerInfo(consts.DefaultCacheManagerName, podName)
+	if err != nil {
+		return nil, err
+	}
+	siteViews := cacheManagerInfo["sites_view"]
+	if siteViews == nil {
+		return nil, fmt.Errorf("Cross-site replication is not supported with the server image you are using.")
```
oraNod, line 128:
```suggestion
		return nil, fmt.Errorf("Retrieving the cross-site view is not supported with the server image you are using.")
```

pkg/infinispan/cluster.go
oraNod: We should probably just say "retrieving the cross-site view is not supported"

pkg/apis/infinispan/v1/infinispan_types.go
rigazilla: > It would be nice, but I'm not sure that we can change `WellFormed` -> `ClusterViewFormed` now as this cause a migration problems or we should support both `ClusterViewFormed` and `WellFormed` statuses for well formed state.
> @rigazilla @ryanemerson what do you think about it?

I would keep both for backward comp.
I also agree with @oraNod that conditions should be designed with a devOp pov, maybe in a separated PR

oraNod: @dmvolod maybe let's hold off on changing `WellFormed` or adding an alternative in this PR. I think it's getting into scope creep. I've created this issue to review status conditions from a DevOPs POV: https://github.com/infinispan/infinispan-operator/issues/887 

For this PR, I think `CrossSiteViewFormed` is the way to go.

dmvolod: Yes, sure @oraNod will fix

pkg/controller/infinispan/infinispan_controller.go
pruivo: (sorry to revive this)
we don't have info about the cause. It just means the remote site isn't connected to this site. It doesn't know if the remote site is started, or there is a connection error or a configuration error (no way t know)

@dmvolod is `pods[0]` in `cluster.GetSitesView(pods[0].Name)` the coordinator?

pruivo (approved):

pkg/controller/infinispan/infinispan_controller.go
dmvolod: @pruivo no, it's just first Pod, but we can choose a Pod with `coordinator` label also. However `GetSitesView` works fine with any Pod, probably I test a single Pod deployment, need to investigate more.

pruivo: please choose the coordinator, it is the one with the real xsite view.
the other pods may have a partial xsite view or no xsite view at all.

dmvolod: Yes, sure. Will fix, thanks.

dmvolod:
Unable to test until Infinispan server image 12.x based on infinispan-image-artifacts v2.1.3 will be released

pruivo (approved):

oraNod (approved):
LGTM. Thanks @dmvolod

documentation/asciidoc/topics/proc_configuring_xsite_within_clusters.adoc
```diff
@@ -38,9 +38,12 @@ spec:
 +
 . Configure nodes with any other {datagridservice} resources.
 . Apply the `Infinispan` CRs.
-. Check node logs to verify that {brandname} clusters form a cross-site view, for example:
+. Verify that {brandname} clusters form a cross-site view.
+.. Retrieve the `Infinispan` CR.
 +
 [source,options="nowrap",subs=attributes+]
 ----
-include::cmd_examples/oc_grep_xsite.adoc[]
```
dmvolod, line 45: @oraNod do we need to remove `oc_grep_xsite.adoc` from the documentation?

documentation/asciidoc/topics/proc_configuring_xsite_within_clusters.adoc
```diff
@@ -38,9 +38,12 @@ spec:
...
+include::cmd_examples/get_infinispan.adoc[]
 ----
++
+.. Check for the `type: CrossSiteViewFormed` condition.
```
dmvolod, line 49: We can also add following command for the `CrossSiteViewFormed` condition validation
`kubectl/oc wait --for condition=CrossSiteViewFormed infinispan example-infinispan`

ryanemerson:
@dmvolod `TestPermanentCache` is failing.

dmvolod:
> `TestPermanentCache` is failing.

Thanks, @ryanemerson I see, but seems to not related to this PR. But I'm trying to run retest.
It can be related to the #913

documentation/asciidoc/topics/proc_configuring_xsite_within_clusters.adoc
oraNod: Yes, please.

dmvolod:
@ryanemerson seems to #956 adds more stability to tests
I'm waiting for #939 to be merged to remove explicit image version setup in the specific test

dmvolod:
@ryanemerson , @rigazilla please review when possible

ryanemerson:
Thanks @dmvolod

Issue #872, opened 2021-03-04
Add status.Phase or Conditions to Cache CR


ryanemerson:
Currently the Cache CR does not utilise a status field to determine the current phase of execution, consequently log files and other transformations will occur on every reconcile event unless properly guarded against.

We can simplify the code by adding a `status.Phase` field as used by the Backup/Batch/Restore CRs or use the `Conditions` based approach used by the Infinispan CR.

ryanemerson:
I think this is probably a blocker for #809, but it has repercussions beyond just cache errors.

dmvolod:
Cache CR has `Status.Conditions` fields an utilize just one `Status=Ready` value that's not enough for full life-cycle support
https://github.com/infinispan/infinispan-operator/blob/4bf99b09b87555f8061f1cc32f6b0c09d1e84483/pkg/controller/cache/cache_controller.go#L208
Can't guess if we need multiple `Conditions` in the future, but probably bi-directional sync will need it. Also reflecting Infinispan CR in the Cache CR could utilize multiple `Conditions` simultaneously

PR #873, opened 2021-03-04
Decouple endpoints

ryanemerson:
This is a preview PR atm as I want to make sure that you're happy with the approach before releasing a new image @dmvolod @rigazilla. I will release the image and update the PR when you're satisfied.

Also, I have mde the Cache CR changes in a separate commit to make it clear how `spec.AdminAuth` is now ignored.

@oraNod There's some doc ramifications with these changes. I think I have addressed most of them, but if you could double check the changes and make sure I haven't missed any sections that would be appreciated.

Image PRs: https://github.com/infinispan/infinispan-images/pull/121 https://github.com/infinispan/infinispan-image-artifacts/pull/36


pkg/controller/cache/cache_controller.go
```diff
@@ -99,6 +97,10 @@ func (r *ReconcileCache) Reconcile(request reconcile.Request) (reconcile.Result,
 		return reconcile.Result{}, err
 	}
 
+	if instance.Spec.AdminAuth != nil {
```
dmvolod, line 100: Would you like to remove this section from CR if it present and update a CR?

pkg/apis/infinispan/v2alpha1/cache_types.go
```diff
@@ -21,7 +21,7 @@ type AdminAuth struct {
 // CacheSpec defines the desired state of Cache
 type CacheSpec struct {
 	// Authentication info
-	AdminAuth AdminAuth `json:"adminAuth,omitempty"`
```
dmvolod, line 24: We can add a comment that this field is deprecated

pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -107,6 +124,56 @@ func (s secretResource) computeSecret(identities []byte) *corev1.Secret {
...
+			Namespace: namespace,
+			Labels:    infinispan.LabelsResource(s.infinispan.Name, "infinispan-secret-admin-identities"),
+		},
+		Type:       corev1.SecretTypeOpaque,
+		StringData: map[string]string{consts.ServerIdentitiesFilename: string(identities)},
+	}
+
+	s.log.Info(fmt.Sprintf("Creating Operator Identities Secret %s", s.infinispan.GetSecretName()))
+	_, err = controllerutil.CreateOrUpdate(ctx, s.client, secret, func() error {
+		return controllerutil.SetControllerReference(s.infinispan, secret, s.scheme)
```
dmvolod, line 161: This will be updated each Infinispan CR reconcile loop and every Infinispan CR will try to owned this resource
It would be better to have personal admin Secret for each cluster

pkg/controller/constants/constants.go
```diff
@@ -34,16 +34,19 @@ var (
...
 	// DefaultOperatorUser users to access the cluster rest API
 	DefaultOperatorUser = "operator"
 	// DefaultDeveloperUser users to access the cluster rest API
 	DefaultDeveloperUser = "developer"
 	// DefaultCacheName default cache name for the CacheService
 	DefaultCacheName         = "default"
-	InfinispanPort           = 11222
-	InfinispanPortName       = "infinispan"
+	InfinispanAdminPort      = 11223
+	InfinispanAdminPortName  = "infinispanadmin"
```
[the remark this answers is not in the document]
ryanemerson: Originally I had a dash, but then we exceed the 15 char limit. I could change it to `ispn-admin`

dmvolod:
It would be nice to merge #813 as it contains common resource reconcilern for services

pkg/controller/cache/cache_controller.go
ryanemerson: Is there any advantage of this approach if the struct is always ignored?

pkg/controller/constants/constants.go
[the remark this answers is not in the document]
dmvolod: Good value, or depends how other systems will display it inside and users will see

pkg/controller/cache/cache_controller.go
dmvolod: Seems to be no value for us, probably b customers will have better and clear view.

pkg/controller/constants/constants.go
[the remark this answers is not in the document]
dmvolod: Probably `infinispan-adm` could be also possible

pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -75,25 +75,42 @@ func Add(mgr manager.Manager) error {
...
 func (s *secretResource) Process() (reconcile.Result, error) {
-	if !s.infinispan.IsGeneratedSecret() {
-		return reconcile.Result{}, nil
+	operatorNs, err := kube.GetOperatorNamespace()
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("Unable to get operator namespace: %w", err)
 	}
-	if err := s.client.Get(context.TODO(), types.NamespacedName{Namespace: s.infinispan.Namespace, Name: s.infinispan.GetSecretName()}, &corev1.Secret{}); err == nil || !errors.IsNotFound(err) {
+
+	secretExists, err := s.secretExists(consts.AdminSecretName, operatorNs)
```
dmvolod, line 83: I'm not sure that this should work in multi-namespace deployment. Secrets can't be shared between namespaces
https://kubernetes.io/docs/concepts/configuration/secret/#details
Each Infinispan CR should has it's own admin secret

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -930,6 +939,7 @@ func PodEnv(i *infinispanv1.Infinispan, systemEnv *[]corev1.EnvVar) []corev1.Env
 		{Name: "JAVA_OPTIONS", Value: i.GetJavaOptions()},
 		{Name: "EXTRA_JAVA_OPTIONS", Value: i.Spec.Container.ExtraJvmOpts},
 		{Name: "DEFAULT_IMAGE", Value: consts.DefaultImageName},
+		{Name: "ADMIN_IDENTITIES_PATH", Value: consts.ServerAdminIdentitiesPath},
```
dmvolod, line 942: Do we need to add admin secret reflection mechanism. I think we need to give clients the ability to change the admin password for some security reason.

pkg/controller/cache/cache_controller.go
ryanemerson: Actually, it's better for us as well as it will stop the repeated log calls on each reconcile event without requiring #872

pkg/controller/infinispan/resources/secret/secret_controller.go
ryanemerson: The secret is contained within the operator namespace and only the operator accesses it, regardless of where the Infinispan CR is created.

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: It is possible for the user to create the admin credentials by creating the admin secret before installing/restarting the operator.

deploy/crds/infinispan.org_caches_crd.yaml
```diff
@@ -34,7 +34,8 @@ spec:
           description: CacheSpec defines the desired state of Cache
           properties:
             adminAuth:
-              description: Authentication info
+              description: Deprecated. This no longer has any effect. The operator's
```
oraNod, line 37:
```suggestion
              description: Deprecated and no longer takes effect.
```

deploy/operator-install.yaml
```diff
@@ -1102,7 +1102,8 @@ spec:
           description: CacheSpec defines the desired state of Cache
           properties:
             adminAuth:
-              description: Authentication info
+              description: Deprecated. This no longer has any effect. The operator's
+                admin credentials are now used to perform cache operations
```
oraNod, line 1106:
```suggestion
                Operator admin credentials are used for cache operations instead.
```

deploy/operator-install.yaml
```diff
@@ -1102,7 +1102,8 @@ spec:
           description: CacheSpec defines the desired state of Cache
           properties:
             adminAuth:
-              description: Authentication info
+              description: Deprecated. This no longer has any effect. The operator's
```
oraNod, line 1105:
```suggestion
              description: Deprecated and no longer takes effect.
```

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -9,14 +9,22 @@ Define a service monitor instances that configures Prometheus to monitor your
...
-<1> Specifies an application user. `developer` is the default.
+. Create an authentication secret that contains the credentials. +
+This allows Prometheus to authenticate with your {brandname} cluster and retrieve metrics.
++
+[source,options="nowrap",subs=attributes+]
+----
+include::yaml/basic_auth.yaml[]
+----
++
+<1> Specifies the **operator** user.
```
oraNod, line 27: @ryanemerson Do you think users might object to re-using the operator credentials for Prometheus authentication? It's convenient but maybe more security-minded users would prefer to keep credentials strictly separate. Just a thought. I'm not sure if it's possible to add user-defined admin credentials.

pkg/controller/infinispan/resources/secret/secret_controller.go
dmvolod: You are mounting this Secret into the Pod Volume. This is not possible for multi-namespace installation
```
  volumes:
    - name: data-volume
      persistentVolumeClaim:
        claimName: data-volume-example-infinispan-0
    - name: config-volume
      configMap:
        name: example-infinispan-configuration
        defaultMode: 420
    - name: admin-identities-volume
      secret:
        secretName: infinispan-operator-credentials
        defaultMode: 420
```

ryanemerson: Great point @dmvolod! I hadn't thought it in that way. So I guess the options are:

1. Have a Secret in the operator namespace with the generated credentials, and then create a copy of this for each Infinispan deployment. The advantages of this is that the same credentials can be used for all clusters and only need to be updated in one place. Updates to the operator secret would have to be reflected across all per cluster secrets.

2. Generate admin secrets per cluster. Simpler, but would require users to create the admin secret per cluster if they wanted to provide custom credentials.

dmvolod: Oh, sure, @ryanemerson I understand your concern.
Please have a look at pros and cons of each solution.

> * Have a Secret in the operator namespace with the generated credentials, and then create a copy of this for each Infinispan deployment. The advantages of this is that the same credentials can be used for all clusters and only need to be updated in one place. Updates to the operator secret would have to be reflected across all per cluster secrets.

pros

- single Secret, easy to manage

cons

- complex code
- need to have separate operator Secret controller
- better to store master Secret in-memory if someone drop it (for better restoring)
- able to create an option to have admin secret per Infinispan CR (some customer want to have this option due to security reasons)

> * Generate admin secrets per cluster. Simpler, but would require users to create the admin secret per cluster if they wanted to provide custom credentials.

pros

- easy to implement

cons

- need to maintain each admin Secret (if needed) individually

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
ryanemerson: > Do you think users might object to re-using the operator credentials for Prometheus authentication?

IMO no, as the k8s users require a certain level of permissions to access the operator secrets in the first place. If a user has that level of access, they can do far worse things (e.g. delete clusters) then configure prometheus :fire: .

pkg/controller/infinispan/resources/secret/secret_controller.go
ryanemerson: I think for the initial implementation lets keep it simple and generate a secret per cluster.

ryanemerson: I've added a commit with these changes.

dmvolod: Good, thanks.

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: It would now be necessary for the user to create the `<infinispan-cr-name>-generated-admin-secret` before creating the Infinispan CR if they require custom credentials. I can add the HASH and restart logic if we think this is required.

dmvolod: I think that ability to change credentials is enough for user

dmvolod:
@ryanemerson please also change Cache CR example with deprecated Auth and its secret

test/e2e/utils/kubernetes.go
```diff
@@ -540,8 +540,9 @@ func RunOperator(m *testing.M, k *TestKubernetes) {
 // Run the operator locally
 func runOperatorLocally(stopCh chan struct{}, namespace string) {
 	kubeConfig := kube.FindKubeConfig()
+	_ = os.Setenv("OPERATOR_NAMESPACE", namespace)
+	_ = os.Setenv("OSDK_FORCE_RUN_MODE", "local")
```
dmvolod, line 544: This env not needed anymore

pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -205,24 +212,32 @@ func setupServiceForEncryption(ispn *ispnv1.Infinispan, service *corev1.Service)
 	}
 }
 
-func computeService(ispn *ispnv1.Infinispan) *corev1.Service {
+func computeAdminService(ispn *ispnv1.Infinispan) *corev1.Service {
+	return computeService(ispn.GetAdminServiceName(), consts.InfinispanAdminPortName, "infinispan-service-admin", consts.InfinispanAdminPort, ispn)
```
dmvolod, line 216: We can expose admin externally with the same address as main Infinispan port, i.e. 11222
For Metrics users port will be the same as for common users

test/e2e/main/main_test.go
```diff
@@ -400,10 +399,10 @@ func TestCacheService(t *testing.T) {
 	testCacheService(t.Name(), nil)
 }
 
-func TestCacheServiceNativeImage(t *testing.T) {
-	t.Parallel()
-	testCacheService(t.Name(), pointer.StringPtr(tutils.NativeImageName))
-}
+// func TestCacheServiceNativeImage(t *testing.T) {
```
dmvolod, line 402: t.SkipNow() can be used for easy maintaining

pkg/controller/infinispan/resources/service/service_controller.go
ryanemerson: We want a different service to the user service, as we never want to expose the admin service externally. It's exclusively for internal k8s usage, i.e. the operator and prometheus monitors.

oraNod (approved):
LGTM thanks @ryanemerson

pkg/controller/infinispan/resources/service/service_controller.go
dmvolod: Exposition doesn't make sense for ocp/k8s internal users
Also have a common posts for admin and users could be nice, as service is mapping between Pod internal IP's and exposition

ryanemerson: I don't fully understand what you're proposing here, can you elaborate?

dmvolod: Sorry, was not correctly explain probably. We can introduce two parameters for compute service, i.e. port (this is external service port) and targetPort (this is internal Pod port) which can be mapped for different services 
admin service: 11222 -> 11223
customer service: 11222 -> 11222
Is this is clear?

ryanemerson: So two distinct services for admin and user, each of which is exposed on port 11222, but the admin service uses `targetPort: 11223`. I think that makes sense, one less port number for the user to worry about :+1:

pkg/controller/infinispan/infinispan_controller.go
dmvolod: @ryanemerson are you plan to add ADMIN_IDENTITIES_HASH for this feature?

rigazilla (commented):
Maybe I've missed it: is the statefulset updated during upgrade or this operator can handle cluster with old statefulset?

pkg/controller/infinispan/resources/service/service_controller.go
ryanemerson: I'm going to remove this after all. It makes the operator internals more complicated as depending on the operation we're trying to perform we need to be careful about which pod we target. Having two distinct ports for admin and user makes things more explicit. The only downside is that the ServiceMonitor has to use port 11223, however I expect most users will copy and paste the `ServiceMonitor` example from our docs anyway.

ryanemerson:
> Maybe I've missed it: is the statefulset updated during upgrade or this operator can handle cluster with old statefulset?

No I missed it :see_no_evil:. I'll have to make it so that on upgrade the developer credentials are extracted from the old secret. 

The question is do we then add the `operator` user to the user context so that we don't break any old applications that may have been using this?

ryanemerson:
Hmm yes. I was thinking that we would have todo something with the secrets, but I guess on upgrade the old secret will be used for user credentials and a new `%v-generated-admin-secret` will be generated for the operator credentials.

rigazilla (approved):
lgtm

ryanemerson:
@dmvolod @rigazilla This should be good to go now. I fixed some issues with zero_controller last night and added a couple of commits to make use of some of our test util functions. This PR now depends on https://github.com/infinispan/infinispan-operator/pull/890, but I can close that and just merge this one if you prefer. I created a separate PR in case there was anything specific that needed to be discussed.

Btw this PR also fixes #878. I would like to get this in today and perform a release if possible.

dmvolod (approved):

rigazilla (approved):
@ryanemerson lgtm, feel free to merge #890 here if you prefer

Issue #877, opened 2021-03-05
`make codegen` should align csv permissions with roles


rigazilla:

rigazilla:
I've two possible solutions:
1. requires mikefarah yq and jq
2. requires python3 with yaml module

any preference?

dmvolod:
@rigazilla further operator SDK version will support it automatic way, if I'm not wrong
We can try to bit hack and use one version for code generating and keep current one for runtime, but need to validate this way

dmvolod:
If not, i'm prefer to use `yq and jq` but install them in automatic way, like operator SDK

PR #879, opened 2021-03-05
Update csv script #877

rigazilla:
adding script and a make rule `make update-csv` to update csv file. Only permissions and clusterpermissions are updated atm.
adding the rule as dependency for `make codegen`

### DIFF
--- a/build/update-csv.sh
+++ b/build/update-csv.sh
@@ -34,11 +34,9 @@ fi
  
 }
 
-
 #Ensure that GOPATH/bin is in the PATH
-printf "Validating yq installation..."
 if ! validateYQ; then
-  printf "Valid yq not found. Trying adding GOPATH/bin\n"
+  printf "Valid yq not found in PATH. Trying adding GOPATH/bin\n"
   PATH=$(go env GOPATH)/bin:$PATH
   if ! validateYQ; then
     installYQ
--- a/build/update-csv.sh
+++ b/build/update-csv.sh
@@ -15,12 +15,11 @@ OPERATOR_ROLES_FILES=(
 
 OPERATOR_CSV_FILE="deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml"
 
-#Ensure that GOPATH/bin is in the PATH
-printf "Validating yq installation..."
+validateYQ() {
+ printf "Validating yq installation..."
 if ! [ -x "$(command -v yq)" ]; then
   printf "Not found\n"
-  PATH=$(go env GOPATH)/bin:$PATH
-  installYQ
+  return 1
 else
   printf "OK\n"
   printf "Validating yq major version..."
@@ -29,7 +28,19 @@ else
     printf "%s\n" "${YQ_MAJ_VERSION}"
   else
     printf "incorrect major version %s\n" "${INSTALLED_YQ_MAJ_VERSION}"
-    PATH=$(go env GOPATH)/bin:$PATH
+    return 1
+  fi
+fi
+ 
+}
+
+
+#Ensure that GOPATH/bin is in the PATH
+printf "Validating yq installation..."
+if ! validateYQ; then
+  printf "Valid yq not found. Trying adding GOPATH/bin\n"
+  PATH=$(go env GOPATH)/bin:$PATH
+  if ! validateYQ; then
     installYQ
   fi
 fi
--- a/build/update-csv.sh
+++ b/build/update-csv.sh
@@ -3,7 +3,8 @@
 YQ_MAJ_VERSION=${1:-4}
 installYQ() {
     printf "Installing yq ..."
-    GO111MODULE=on go get github.com/mikefarah/yq/v4
+    GO111MODULE=off go get -u github.com/myitcv/gobin
+    gobin github.com/mikefarah/yq/v4@v4.6.1
     PATH=$(go env GOPATH)/bin:$PATH
     yq --version
 }
--- a/build/update-csv.sh
+++ b/build/update-csv.sh
@@ -4,6 +4,7 @@ YQ_MAJ_VERSION=${1:-4}
 installYQ() {
     printf "Installing yq ..."
     GO111MODULE=on go get github.com/mikefarah/yq/v4
+    PATH=$GOPATH/bin:$PATH
     yq --version
 }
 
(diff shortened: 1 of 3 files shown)


dmvolod:
changed go.mod and go.sum files for me and then fails
```
./build/update-csv.sh: line 7: yq: command not found
./build/update-csv.sh: line 33: yq: command not found
make: *** [Makefile:36: update-csv] Error 127
```

rigazilla:
let me try with `go install` it should take care of the hw arch.
`go env GOPATH` needs to be in the path

dmvolod:
> `go env GOPATH` needs to be in the path

No, the same result. Seems to be `$GOPATH/bin` needs in the PATH, not whole `$GOPATH`. But I'm never used this config before

I think it's possible to automate such kind of restrictions with bash script
go.* files were updated anymore as well as infinispan-operator.clusterserviceversion.yaml - seems to need to be regenerated in this PR

build/update-csv.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+YQ_MAJ_VERSION=${1:-4}
+installYQ() {
+    printf "Installing yq ..."
+    GO111MODULE=on go get github.com/mikefarah/yq/v4
```
dmvolod, line 6: Can we bind to the specific enough for us `yq` version to prevent updates on some codegen steps

rigazilla: unfortunately not, only major rel can be specified

dmvolod: Not a good news, than we need go away from project folder temporary, i.e. /tmp install yq and than return back

rigazilla:
fixed the PATH setup, csv is modified because indentation is fixed
still looking into the go.mod issue

build/update-csv.sh
```diff
@@ -0,0 +1,35 @@
...
+    PATH=$GOPATH/bin:$PATH
```
dmvolod, line 7: It would be nice to set `GOPATH` if it not defined for example to `$HOME\go`

rigazilla: oh right, my idea was to use `go env` but I forgot to that. fixed

build/update-csv.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+YQ_MAJ_VERSION=${1:-4}
+installYQ() {
+    printf "Installing yq ..."
+    GO111MODULE=on go get github.com/mikefarah/yq/v4
+    PATH=$(go env GOPATH)/bin:$PATH
```
dmvolod, line 7: This need not only for installation, usage needs this also

ryanemerson:
I get the following when trying to execute `make codegen`:

```
./build/update-csv.sh
Validating yq installation...Not found
Installing yq ..../build/update-csv.sh: line 7: gobin: command not found
./build/update-csv.sh: line 9: yq: command not found
./build/update-csv.sh: line 36: yq: command not found
make: *** [Makefile:36: update-csv] Error 127
```

rigazilla:
fixed the PATH setting. thanks @dmvolod

dmvolod:
@rigazilla excellent, works fine now.
Please include regenerated `deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml` file in PR to prevent git change on first usage.

build/update-csv.sh
```diff
@@ -0,0 +1,38 @@
...
+)
+
+OPERATOR_CSV_FILE="deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml"
+
+#Ensure that GOPATH/bin is in the PATH
+printf "Validating yq installation..."
+if ! [ -x "$(command -v yq)" ]; then
+  printf "Not found\n"
+  installYQ
+  PATH=$(go env GOPATH)/bin:$PATH
```
dmvolod, line 24:
```suggestion
```

build/update-csv.sh
```diff
@@ -0,0 +1,38 @@
...
+}
+
+OPERATOR_ROLES_FILES=(
+    "deploy/role.yaml"
+    "deploy/clusterrole.yaml"
+)
+
+OPERATOR_CSV_FILE="deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml"
+
+#Ensure that GOPATH/bin is in the PATH
```
dmvolod, line 19:
```suggestion
PATH=$(go env GOPATH)/bin:$PATH
```

build/update-csv.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+YQ_MAJ_VERSION=${1:-4}
+installYQ() {
+    printf "Installing yq ..."
+    GO111MODULE=off go get -u github.com/myitcv/gobin
+    PATH=$(go env GOPATH)/bin:$PATH
```
dmvolod, line 7:
```suggestion
```

build/update-csv.sh
rigazilla: @dmvolod I would change the user path only if yq is not available, i.e. move this inside the if statem

dmvolod: In this case how to validate `yq` in the PATH if PATH is not set?
For me script always install yq even if it already installed

rigazilla: if the user already have the correct yq in the path nothing is installed

dmvolod: For my laptop GOPATH/bin never installed in the PATH and `yq` always reinstalling on every codegen
It's better to set validate it, set GOPATH/bin into the PATH and validate again

build/update-csv.sh
```diff
@@ -0,0 +1,48 @@
...
+    printf "incorrect major version %s\n" "${INSTALLED_YQ_MAJ_VERSION}"
+    return 1
+  fi
+fi
+ 
+}
+
+
+#Ensure that GOPATH/bin is in the PATH
+printf "Validating yq installation..."
```
dmvolod, line 39:
```suggestion
```

build/update-csv.sh
```diff
@@ -0,0 +1,48 @@
...
+if ! validateYQ; then
+  printf "Valid yq not found. Trying adding GOPATH/bin\n"
```
dmvolod, line 41:
```suggestion
  printf "Valid yq not found in PATH. Trying adding GOPATH/bin\n"
```

build/update-csv.sh
```diff
@@ -0,0 +1,48 @@
...
+    printf "%s\n" "${YQ_MAJ_VERSION}"
+  else
+    printf "incorrect major version %s\n" "${INSTALLED_YQ_MAJ_VERSION}"
+    return 1
+  fi
+fi
+ 
+}
+
+
```
dmvolod, line 37:
```suggestion
```

rigazilla: merged, thanks @dmvolod

dmvolod (approved):

rigazilla:
merged, thanks @dmvolod for the review

Issue #690, opened 2020-12-11
Refactor Logging Levels


ryanemerson:
We should refactor operator logging to better make use of different levels. Currently a large amount of information is output as info, when it should be at the Debug or Trace levels. As an example the following "reconcile" logs are created for a single Infinispan CR:

```
{"level":"info","ts":1607551632.6983733,"logger":"controller_infinispan","msg":"Reconciling Infinispan. Operator Version: RHDG-8.1.0.CD20201127","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551632.8670905,"logger":"controller_infinispan","msg":"Reconciling Infinispan. Operator Version: RHDG-8.1.0.CD20201127","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551632.867159,"logger":"controller_infinispan","msg":"Reconciling Infinispan: update case","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551633.0032003,"logger":"controller_infinispan","msg":"Reconciling Infinispan. Operator Version: RHDG-8.1.0.CD20201127","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551633.0032513,"logger":"controller_infinispan","msg":"Reconciling Infinispan: update case","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551633.0034044,"logger":"controller_infinispan","msg":"Reconciling Infinispan. Operator Version: RHDG-8.1.0.CD20201127","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551633.0034747,"logger":"controller_infinispan","msg":"Reconciling Infinispan: update case","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551698.2401514,"logger":"controller_infinispan","msg":"Reconciling Infinispan. Operator Version: RHDG-8.1.0.CD20201127","Request.Namespace":"datagrid","Request.Name":"datagrid"}
{"level":"info","ts":1607551698.240205,"logger":"controller_infinispan","msg":"Reconciling Infinispan: update case","Request.Namespace":"datagrid","Request.Name":"datagrid"}
[...]

PR #880, opened 2021-03-05
Send user related warnings to the console Events log

### LINKED ISSUE #543: Redundant "Infinispan.Namespace" key and value need to be clear in reqLogger

dmvolod:
In addition to the main functionality (sending user related warnings to the k8s event subsystem and console Events log for the OLM UI or ), this issue applies a new implementation of the idea of `base.ReconcilerBase` which introduces a core reconciler structure for better maintaining Logger, EventRecorder (for sending k8s events) and core functions required for all our controllers. We can move/add some additional common functions as well as additional new structure fields, i.e. `rest.Config`, `rest.Interface`, etc. to this reconciler later if needed.

It can't be possible to separate implementing low level function for sending events and high level refactoring with `ReconcilerBase` as we need to abstract from passing logging and eventing too much via functions parameters.

fixes #543 also


pkg/controller/base/reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	EventReasonResourceNotReady = "ResourceNotReady"
+)
+
+type ReconcilerBase struct {
```
ryanemerson, line 25: `ReconcilerBase` is an interesting idea, however I'm a little worried by two things.

1. It feels like we're trying to use a inheritance based approach by using a "base" struct, wheras we should be favouring composition.
2. `ReconcilerBase` functions cover a lot of unrelated functionality. Logging, resource lookup, pod labels.

IMO a better approach would be to create a interface that's dedicated to Logs & events. This interface can then be passed as the args to our functions used throughout our code base. For example, `LookupResource` is a generic kubernetes operation so it probably makes more sense in `pkg/kubernetes/controllerutil.go` as it's not controller specific. We then update it's parameters to take the new logs/events interface instead of `logr.Logger` so that we can log and send the event.

Having a dedicated logs/events interace would also help #690.

dmvolod: Yes, I agree that combining the `LookupResource` and mixing other functions inside the `ReconcileBase` may have a kind of confusion of concepts. I would keep this structure as the basis for all our controllers and consolidate all the required structures for all potential controllers to avoid redundant declaration and initialization every time we need it, i.e.
```
type ReconcilerBase struct {
	Name string
	client.Client
	kubernetes.Kubernetes
	// TBD some log and event recorder interface or separate interfaces
}
```
Also other `*Label` related common methods should be moved as static go file inside `kubernetes` package like pods.go as they are not API server related
I would like to re-implement common `UpdateResource` functions based on `infinispan_operator.update` inside `kubernetes` package to use `controllerutil.CreateOrPatch` in more shared way for update Infinispan Group CR's (Batch, Cache, Infinispan, etc.)
`LookupResource` and others functions which will require to access log subsystem will need to accept a new log and event recorder interface or separate construction

dmvolod:
@ryanemerson , @rigazilla please have a look at this implementation for Events support

rigazilla:
I like the feature itself.
I don't have strong opinion on the refactoring part, I would prefer to minimize the impact on the existing code as a general though.

ryanemerson:
> I like the feature itself.

+1 to the feature, but I would rather avoid the `reconcilerBase` in favour of implementing a small package/abstraction for events & logging.

rigazilla:
@dmvolod how about moving the common behaviours in an external object that operates on our controllers?
I mean: all the controllers must be initialized and to reuse initialization code define a ControllerInitializer that operates on controllers and initializes them in a common way?
This doesn't introduce hierarchy and focus on behaviours instead of structures.

rigazilla:
ouch I shouldn't have written `objecŧ` in a discussion on golang :smile:

rigazilla:
@dmvolod are you still working on this?
Otherwise I would like to propose an implementation based on my proposal above:
https://github.com/infinispan/infinispan-operator/pull/880#issuecomment-818535642

dmvolod:
@rigazilla yes, sure, you are welcome to implement this option
I'm closing this PR

PR #883, opened 2021-03-08
[test-integration] Wait for status presence

Crumby:
Prevents `NullPointerException` in case the `waitFor()` is syncs up earlier then Operator processes the Infinispan CR for the first time


dmvolod (approved):

ryanemerson:
Thanks @Crumby

PR #885, opened 2021-03-09
Operator upgrade fails without Sites.Locations configuration

dmvolod:


dmvolod:
Thanks, @ryanemerson , true, probably thought about different while edit commit message

ryanemerson:
Thanks @dmvolod

PR #886, opened 2021-03-10
Providing Infinispan console URI in status field

### LINKED ISSUE #687: Avoid back update Route/Ingress host or Service NodePort inside Spec.Expose
We need to discuss potential remove back update of the `Spec.Expose` structure
This can't be helpful when users would like to make this update manually and need to be validated
Probably, this will not take any problem

* Otherwise, additional fields in the `Status` can be used for this kind of back updates like Console URI described in #730
* This cause a set of errors described in #711 also the same problem with NodePort port during upgrade. As alternative solution, can be keep services and other non persisted structures

dmvolod:
Will also be displayed on the OLM UI console
image

Also fixes #687

### DIFF
--- a/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
@@ -214,11 +214,16 @@ spec:
             x-descriptors:
               - 'urn:alm:descriptor:io.kubernetes:PersistentVolumeClaim'
         statusDescriptors:
-          - description: Displays the current status of pods.
+          - description: The current pods
             displayName: Pod Status
             path: podStatus
             x-descriptors:
               - 'urn:alm:descriptor:com.tectonic.ui:podStatuses'
+          - displayName: Infinispan Console URL
+            description: Infinispan Console location
+            path: consoleUrl
+            x-descriptors:
+              - 'urn:alm:descriptor:org.w3:link'
   description: |
     Infinispan is a distributed in-memory data store built on open-source technology.
 
--- a/pkg/kubernetes/kubernetes.go
+++ b/pkg/kubernetes/kubernetes.go
@@ -6,15 +6,15 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"net/url"
 
 	"github.com/go-logr/logr"
 	consts "github.com/infinispan/infinispan-operator/pkg/controller/constants"
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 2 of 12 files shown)


deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -102,6 +102,11 @@ spec:
             path: podStatus
             x-descriptors:
               - "urn:alm:descriptor:com.tectonic.ui:podStatuses"
+          - displayName: Infinispan Console URL
```
oraNod, line 105: @dmvolod Is it possible to change the product name for downstream? s/Infinispan/Data Grid

deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -102,6 +102,11 @@ spec:
...
+            description: Route to access Infinispan Console
```
oraNod, line 106:
```suggestion
            description: Infinispan Console location
```

deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
dmvolod: @oraNod seems to be we aren't changing another entrie names abd descriptions moving from upstream to downstream. I think it's possible but at the downstream site only.
@rigazilla wdyt?

rigazilla: I think it's possible changing the CSV for Data Grid

dmvolod: Yes, sure, thanks @rigazilla 
Do we need to change it now or can be done during product build?

pkg/kubernetes/kubernetes.go
```diff
@@ -188,32 +188,21 @@ func resolveConfig(ctx string) *rest.Config {
...
 }
 
-// PublicIP returns the public IP address of the Kubernetes cluster
-func (k Kubernetes) PublicIP() string {
-	u, _ := url.Parse(k.RestConfig.Host)
-	return u.Hostname()
-}
-
 // GetNodesHost return the addresses of the k8s nodes
 func (k Kubernetes) GetNodesHost() ([]string, error) {
```
rigazilla, line 196: I wondering if we can get rid of this since it's used only here: https://github.com/infinispan/infinispan-operator/blob/7ebaad9a1eca9761709fa6253a96e3b67d4b1a61/test/e2e/utils/kubernetes.go#L319-L320

rigazilla (approved):
looks go to me. I've added a comment that can be address here or in a separate PR

deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
rigazilla: @dmvolod we will do on the product side

pkg/kubernetes/kubernetes.go
dmvolod: Yeah, good point, will try to move to the test package

rigazilla: maybe it can be just replaced with a GetNodeHost, the test just pickup the first result

dmvolod: Switch to the `k.Kubernetes.GetNodeHost`. Seems to be logic a bit similar with unnecessary variations

rigazilla:
@dmvolod, rebase needed

rigazilla:
merged, thanks @dmvolod

Issue #891, opened 2021-03-10
WaitForExternalService should get IP of host containing service pod


ryanemerson:
Currently the `test/e2e/utils/kubernetes.go#WaitForExternalService` retrieves all hosts via `Kubernetes.GetNodesHost()` and just selects the first host in the list.  However, if the tests are executed on a cluster with many nodes it's possible that the used node will not contain an Infinispan pod. Therefore, we should update this method to only retrieve nodes containing Pods that match the required services selector labels.

PR #892, opened 2021-03-10
WaitForExternalService should get IP of host containing service pod #891

ryanemerson:


dmvolod:
@ryanemerson what's the reason of these changes? Please see how `kube-proxy` works for NodePort at diagram below
image

ryanemerson:
@dmvolod You're totally right. I'm inventing problems it seems :upside_down_face:

PR #894, opened 2021-03-11
Grafana ispn 12702

rigazilla:
New PR for grafana dashboard creation.
I've moved pieces from #803  here instead of rebasing and force push on the old PR.

Changes reflect the discussion we had in #803.
 
I've recorded my setup procedure in this gist: https://gist.github.com/rigazilla/015e4a256bd9ef2e9835d296468695d6

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -38,9 +38,16 @@ rules:
   - apps
   resources:
   - replicasets
+  verbs:
+  - get
+- apiGroups:
+  - apps
+  resources:
   - deployments
   verbs:
   - get
+  - list
+  - watch
 - apiGroups:
   - batch
   resources:
@@ -117,3 +124,13 @@ rules:
   - create
   - update
   - patch
+- apiGroups:
+    - integreatly.org
+  resources:
+    - grafanadashboards
+  verbs:
+    - get
+    - list
+    - watch
+    - create
+    - delete
--- a/pkg/controller/infinispan/resources/operatorconfig/grafana.go
+++ b/pkg/controller/infinispan/resources/operatorconfig/grafana.go
@@ -1,0 +1,107 @@
+package operatorconfig
+
+import (
+	consts "github.com/infinispan/infinispan-operator/pkg/controller/constants"
+	kube "github.com/infinispan/infinispan-operator/pkg/kubernetes"
+	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// reconcileGrafana reconciles grafana object status with the operator configuration settings
+func (r *ReconcileOperatorConfig) reconcileGrafana(config, currentConfig map[string]string, operatorNs string) (*reconcile.Result, error) {
+	grafanaNs := config[grafanaDashboardNamespaceKey]
+
+	// Delete current grafana dashboard if namespace is changes
+	if err := r.deleteDashboardOnKeyChanged(config, currentConfig); err != nil {
+		return &reconcile.Result{}, err
+	}
+
+	if grafanaNs == "" {
+		return &reconcile.Result{}, nil
+	}
+	// detect the GrafanaDashboard resource type resourceExists on the cluster
+	resourceExists, err := r.Kube.IsGroupVersionSupported(grafanav1alpha1.SchemeGroupVersion.String(), grafanav1alpha1.GrafanaDashboardKind)
+	if err != nil {
+		r.Log.Error(err, "Error checking Grafana support")
+		return &reconcile.Result{Requeue: true}, nil
+	}
+	if !resourceExists {
+		r.Log.Info("Grafana CRD not present - not installing dashboard CR")
+		return &reconcile.Result{RequeueAfter: consts.DefaultLongWaitOnCreateResource}, nil
+	}
+	infinispanDashboard := emptyDashboard(config)
+	if _, err = controllerutil.CreateOrUpdate(ctx, r.client, infinispanDashboard, func() error {
+		if infinispanDashboard.CreationTimestamp.IsZero() {
+			if grafanaNs == operatorNs {
+				operatorDeployment := &appsv1.Deployment{}
+				if result, err := kube.LookupResource("infinispan-operator", operatorNs, operatorDeployment, r.client, r.Log); result == nil {
+					// If running in cluster set controller config
+					if err = controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, r.scheme); err != nil {
+						return err
+					}
+				}
+			} else {
+				r.Log.Info("Not setting controller reference, cause Infinispan and Grafana are in different namespaces.")
+			}
+		}
+		populateDashboard(infinispanDashboard, config)
+		return nil
+	}); err != nil {
+		return &reconcile.Result{}, err
+	}
+
+	currentConfig[grafanaDashboardNamespaceKey] = grafanaNs
+	currentConfig[grafanaDashboardNameKey] = config[grafanaDashboardNameKey]
+	return &reconcile.Result{}, nil
+}
+
+func emptyDashboard(config map[string]string) *grafanav1alpha1.GrafanaDashboard {
+	return &grafanav1alpha1.GrafanaDashboard{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      config[grafanaDashboardNameKey],
+			Namespace: config[grafanaDashboardNamespaceKey],
+		},
+	}
+}
+
+func populateDashboard(dashboard *grafanav1alpha1.GrafanaDashboard, config map[string]string) {
+	dashboard.ObjectMeta.Labels = map[string]string{
+		"monitoring-key": config[grafanaDashboardMonitoringKey],
+		"app":            "grafana",
+	}
+	dashboard.Spec = grafanav1alpha1.GrafanaDashboardSpec{
+		Json: dashboardJSON,
+		Name: "infinispan.json",
+		Datasources: []grafanav1alpha1.GrafanaDashboardDatasource{
+			{
+				InputName:      "DS_PROMETHEUS",
+				DatasourceName: "Prometheus",
+			},
+		},
+	}
+}
+
+func (r *ReconcileOperatorConfig) deleteDashboardOnKeyChanged(newCfg, curCfg map[string]string) error {
+	// If key is changed and old key is not nil, delete old grafana dashboard
+	if (newCfg[grafanaDashboardNameKey] != curCfg[grafanaDashboardNameKey] ||
+		newCfg[grafanaDashboardNamespaceKey] != curCfg[grafanaDashboardNamespaceKey]) &&
+		curCfg[grafanaDashboardNameKey] != "" &&
+		curCfg[grafanaDashboardNamespaceKey] != "" {
+		currGrafana := &grafanav1alpha1.GrafanaDashboard{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      curCfg[grafanaDashboardNameKey],
+				Namespace: curCfg[grafanaDashboardNamespaceKey],
+			},
+		}
+		if err := r.Kube.Client.Delete(ctx, currGrafana); err != nil && !errors.IsNotFound(err) {
+			return err
+		}
+		currentConfig[grafanaDashboardNamespaceKey] = ""
+		currentConfig[grafanaDashboardNameKey] = ""
+	}
+	return nil
+}
--- a/pkg/controller/infinispan/resources/operatorconfig/monitoring_constants.go
+++ b/pkg/controller/infinispan/resources/operatorconfig/monitoring_constants.go
@@ -1,0 +1,1684 @@
+package operatorconfig
+
+const dashboardJSON = `{
+	"__inputs": [
+	  {
+		"name": "DS_PROMETHEUS",
+		"label": "Prometheus",
+		"description": "",
+		"type": "datasource",
+		"pluginId": "prometheus",
+		"pluginName": "Prometheus"
+	  }
+	],
+	"__requires": [
+      {
+        "type": "grafana",
+        "id": "grafana",
+        "name": "Grafana",
+        "version": "6.2.1"
+      },
+      {
+        "type": "panel",
+        "id": "graph",
+        "name": "Graph",
+        "version": ""
+      },
+      {
+        "type": "datasource",
+        "id": "prometheus",
+        "name": "Prometheus",
+        "version": "1.0.0"
+      },
+      {
+        "type": "panel",
+        "name": "Singlestat",
+        "version": ""
+      }
+    ],
+		"annotations": {
+		  "list": [
+			{
+			  "builtIn": 1,
+			  "datasource": "-- Grafana --",
+			  "enable": true,
+			  "hide": true,
+			  "iconColor": "rgba(0, 211, 255, 1)",
+			  "name": "Annotations & Alerts",
+			  "type": "dashboard"
+			}
+		  ]
+		},
+		"editable": true,
+		"gnetId": null,
+		"graphTooltip": 0,
+		"id": 1,
+		"iteration": 1610657246634,
+		"links": [],
+		"panels": [
+		  {
+			"collapsed": false,
+			"datasource": null,
+			"gridPos": {
+			  "h": 1,
+			  "w": 24,
+			  "x": 0,
+			  "y": 0
+			},
+			"id": 33,
+			"panels": [],
+			"title": "Summary",
+			"type": "row"
+		  },
+		  {
+			"cacheTimeout": null,
+			"colorBackground": false,
+			"colorValue": false,
+			"colors": [
+			  "#299c46",
+			  "rgba(237, 129, 40, 0.89)",
+			  "#d44a3a"
+			],
+			"datasource": null,
+			"description": "The number of pods in the cluster",
+			"fieldConfig": {
+			  "defaults": {
+				"custom": {}
+			  },
+			  "overrides": []
+			},
+			"format": "none",
+			"gauge": {
+			  "maxValue": 100,
+			  "minValue": 0,
+			  "show": false,
+			  "thresholdLabels": false,
+			  "thresholdMarkers": true
+			},
+			"gridPos": {
+			  "h": 6,
+			  "w": 2,
+			  "x": 0,
+			  "y": 1
+			},
+			"id": 52,
+			"interval": null,
+			"links": [],
+			"mappingType": 1,
+			"mappingTypes": [
+			  {
+				"name": "value to text",
+				"value": 1
+			  },
+			  {
+				"name": "range to text",
+				"value": 2
+			  }
+			],
+			"maxDataPoints": 100,
+			"nullPointMode": "connected",
+			"nullText": null,
+			"pluginVersion": "6.2.4",
+			"postfix": "",
+			"postfixFontSize": "50%",
+			"prefix": "",
+			"prefixFontSize": "50%",
+			"rangeMaps": [
+			  {
+				"from": "null",
+				"text": "N/A",
+				"to": "null"
+			  }
+			],
+			"sparkline": {
+			  "fillColor": "rgba(31, 118, 189, 0.18)",
+			  "full": false,
+			  "lineColor": "rgb(31, 120, 193)",
+			  "show": false
+			},
+			"tableColumn": "",
+			"targets": [
+			  {
+				"expr": "sum(kube_pod_status_ready{namespace=~\"$namespace\",pod !~ \"infinispan-operator.*\"}) by (namespace) # kube specific, need OS?",
+				"format": "time_series",
+				"interval": "",
+				"intervalFactor": 1,
+				"legendFormat": "Replicas",
+				"refId": "A"
+			  }
+			],
+			"thresholds": "",
+			"timeFrom": null,
+			"timeShift": null,
+			"title": "Replicas (up)",
+			"type": "singlestat",
+			"valueFontSize": "80%",
+			"valueMaps": [
+			  {
+				"op": "=",
+				"text": "N/A",
+				"value": "null"
+			  }
+			],
+			"valueName": "avg"
+		  },
+		  {
+			"aliasColors": {},
+			"bars": false,
+			"dashLength": 10,
+			"dashes": false,
+			"datasource": null,
+			"fieldConfig": {
+			  "defaults": {
+				"custom": {}
+			  },
+			  "overrides": []
+			},
+			"fill": 1,
+			"fillGradient": 0,
+			"gridPos": {
+			  "h": 6,
+			  "w": 5,
+			  "x": 2,
+			  "y": 1
+			},
+			"hiddenSeries": false,
+			"id": 25,
+			"legend": {
+			  "avg": false,
+			  "current": false,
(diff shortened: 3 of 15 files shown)


pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
 	}
 	return err
 }
+
+// addInfinispanConfig installs a reconciler for a specific object.
+// Only events coming from operatorNs/infinispan-operator-config will be processed
+func addInfinispanConfig(mgr manager.Manager) error {
+	r := newInfinispanOperatorConfigReconciler(mgr)
+	// Create a new controller
+	c, err := controller.New("infinispan-controller", mgr, controller.Options{Reconciler: r})
```
ryanemerson, line 176: "infinispan-config-controller"?

Also it would be good to define this as a constant at the top of this file.

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+	if err != nil {
+		return err
+	}
+	operatorNS, err := getOperatorNamespace()
+	if err != nil && err != k8sutil.ErrRunLocal {
+		return err
+	}
+	operatorConfigPredicate := predicate.Funcs{
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return e.Meta.GetName() == "infinispan-operator-config" && e.Meta.GetNamespace() == operatorNS
```
ryanemerson, line 186: A constant would be good here.

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+	// Watch for changes to global operator config
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, operatorConfigPredicate)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// newInfinispanOperatorConfigReconciler returns a new reconcile.Reconciler
+func newInfinispanOperatorConfigReconciler(mgr manager.Manager) reconcile.Reconciler {
```
ryanemerson, line 207: Is there a need for the function as it's only called once? It's less verbose to just create the struct directly in `addInfinispanConfig`.

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+	return &ReconcileInfinispanOperatorConfig{client: mgr.GetClient(), scheme: mgr.GetScheme(), Log: logf.Log.WithName("operator-config-controller"), Kube: kube.NewKubernetesFromController(mgr)}
```
ryanemerson, line 208: "operator-config-controller" as a constant and maybe we want to use the same value here as the controller name?

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+type ReconcileInfinispanOperatorConfig struct {
+	client client.Client
+	scheme *runtime.Scheme
+	Log    logr.Logger
+	Kube   *kube.Kubernetes
+}
+
+// Reconcile reads the operator configuration in the `infinispan-operator-config` ConfigMap and
+// applies the configuration to the cluster.
+func (r *ReconcileInfinispanOperatorConfig) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
ryanemerson, line 223: As we have a reconcile method that is independent of the existing `Process()` method used by the `Reconciler` inteface, maybe it's best to have the operator configuration controller as a separate file? It seems like their is minimal shared code between the two atm and the separation would make things much easier to read.

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+	isRunningLocal := false
+	if err == k8sutil.ErrRunLocal {
+		isRunningLocal = true
+	}
+	if err != nil && !isRunningLocal {
+		return err
+	}
+	r.client.Get(ctx, types.NamespacedName{Namespace: operatorNs, Name: "infinispan-operator-config"}, config)
+	// Reconcile grafana
+	grafanaNs := config.Data["grafanaDashboardNameSpace"]
```
ryanemerson, line 246: I think we should keep the grafana CRD creation in `pkg/infinispan/grafana/dashboard_controller.go` and call it from this reconcile loop. Adding all of the resource creation logic to the configuration reconcilation logic does not scale as we add additional configuration options.

ryanemerson (changes requested):

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: I would go this way:
-  split the config controller in a separate file
-  and move grafana utils in a separate file but in the same package, I don't think we need to have them global

ryanemerson: > * and move grafana utils in a separate file but in the same package, I don't think we need to have them global

Just to be clear. Are you proposing moving just the CR reconcilation to another file?

I think there's merit in keeping `pkg/infinispan/grafana/` for the implementation parts like the .json etc.

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+}
+
+// getOperatorNamespace returns the operator namespace or the WATCH_NAMESPACE
+// value if the operator is running outside the cluster.
+// "OSDK_FORCE_RUN_MODE"="local" must be set if running locally (not in the cluster)
+func getOperatorNamespace() (string, error) {
+	operatorNs, err := k8sutil.GetOperatorNamespace()
+	// This makes everything works even running outside the cluster
+	if err != nil && err == k8sutil.ErrRunLocal {
+		operatorNs = os.Getenv("WATCH_NAMESPACE")
```
dmvolod, line 293: There is a function, called `k8sutil.GetWatchNamespace()` present. But what happens if this value will have multiple records (comma separated)

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+	}
+	return reconcile.Result{}, nil
+}
+
+// reconcileOperatorConfigMap reconcile the operator ConfigMap
+// Operator configMap is unique for an operator install and placed in k8sutil.GetOperatorNamespace()/infinispan-operator-config
+func (r ReconcileInfinispanOperatorConfig) reconcileOperatorConfigMap() error {
+	config := &corev1.ConfigMap{}
+	operatorNs, err := getOperatorNamespace()
+	isRunningLocal := false
```
dmvolod, line 237: We can identify local running more smarter way, if there is no Infinispan operator Deployment present inside operatorNs means that operator running locally.

pkg/infinispan/grafana/dashboard_controller.go
```diff
@@ -0,0 +1,43 @@
+package grafana
```
dmvolod, line 1: for me `dashboard_controller.go` a bit confusing name as it's not classic controller with Watchers and Reconciler
Probably we need a new name for it, but if we will need to control deletion of this CR, we need to implement controller part for it and keep this name the same

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -152,3 +167,130 @@ func (c configResource) computeAndReconcileConfigMap(xsite *configuration.XSite)
...
+			if err != nil {
+				r.Log.Error(err, "There was a problem looking up the infinispan operator to add the dashboard as a secondary resource")
+				return err
+			}
+			if result != nil {
+				r.Log.Info("Infinispan operator resource not initialized, not installing Grafana Dashboard")
+				return nil
+			}
+			// result and err should be analyzed
+			if err = controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, r.scheme); err != nil {
```
dmvolod, line 270: I think it will be always fail when grafanaNs != operatorNs or will create a broken reference
Cross-namespace owner references are disallowed by design.

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: This is mostly added for dev, I think we can just pickup the first ns in the list

deploy/role.yaml
```diff
@@ -97,3 +97,13 @@ rules:
...
+- apiGroups:
+    - integreatly.org
+  resources:
+    - grafanadashboards
+  verbs:
+    - get
+    - list
+    - watch
+    - create
+    - delete
```
dmvolod, line 109: Do we really need a `delete` rights?

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: I've reused the sdk approach here, but I think we can rely on the deployment presence.

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: mmm yeah, good catch. I'll set the controller only if ns match

deploy/role.yaml
rigazilla: though I don't know how to get the old namespace for deleting

dmvolod: Oh, it could be a global var or prefer struct for storing previous collected value and if it changed, old must be deleted and new one can be created

pkg/controller/infinispan/resources/config/config_controller.go
dmvolod: Not sure that will be actual for dev only, in upstream without OLM and several watching namespaces may take place potentially

pkg/controller/infinispan/resources/operatorconfig/operatorconfig_controller.go
```diff
@@ -0,0 +1,135 @@
...
+		if operatorNs == "" {
+			return "", nil
+		}
+		return strings.Split(operatorNs, ",")[0], nil
+	}
+	return operatorNs, err
+}
+
+func getDefaultOperatorConfig() map[string]string {
+	return map[string]string{}
```
ryanemerson, line 134: Should we just make this a var?

pkg/controller/infinispan/resources/operatorconfig/grafana.go
```diff
@@ -0,0 +1,80 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	grafanaDashboardNamespace = "grafanaDashboardNamespace"
+)
+
+func (r *ReconcileOperatorConfig) reconcileGrafana(config map[string]string, operatorNs string) (*reconcile.Result, error) {
+	grafanaNs := config[grafanaDashboardNamespace]
+	if grafanaNs != "" {
```
ryanemerson, line 22: If we invert the check we can avoid the whole function being  indented.

```go
if grafanaNs == "" {
   return &reconcile.Result{}, nil
}
...
```

pkg/controller/infinispan/resources/operatorconfig/grafana.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		if result == nil {
+			if err = controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, r.scheme); err != nil {
+				r.Log.Error(err, "Unable to set Grafana Dashboard controller reference to Infinispan operator - skipping creation")
+				return &reconcile.Result{Requeue: true}, nil
+			}
+		} else {
+			r.Log.Info("Infinispan operator resource not found, assuming operator is running locally.")
+		}
+		if _, err = controllerutil.CreateOrUpdate(context.Background(), r.client, infinispanDashboard, func() error {
```
ryanemerson, line 48: we can use `ctx` here.

ryanemerson: As we're not using the `mutate` function, can we just call `client.Create` here?

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: do you mean with operator running locally? Anyway I'll go for taking the first one in the list, until we will need any special policy

deploy/role.yaml
rigazilla: it needs some though, we need something that survives restart or upgrade

rigazilla: oh, it can be populated at startup...

pkg/controller/infinispan/resources/operatorconfig/operatorconfig_controller.go
rigazilla: let's design this when we actually need default config

pkg/controller/infinispan/resources/operatorconfig/operatorconfig_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const (
+	ControllerName = "operator-config-controller"
+	configMapName  = "infinispan-operator-config"
```
ryanemerson, line 27: I think we need to append the operator major version or channel to this. In the model where multiple Infinispan operators exist at the global level, we need a way to distinguish between each operator's configuration.

rigazilla: I see your point... how about using Infinispan CR API version?
Different operators must watch different API versions otherwise they clash, so something like:
`infinispan-operator-config-v1' `infinispan-operator-config-v2alpha1`
wdyt?

ryanemerson: > Different operators must watch different API versions otherwise they clash

This isn't the case with the proposed architecture. Multiple operators will watch the same CRD versions, however a field in the CRs determines which operator performs the reconciliation.

I think we need the name to contain the specific operator version.

dmvolod: As version a bit long, we can introduce version in the ConfigMap filed and validates it on reconcile stage.
In this case we should bind watch to the specific ConfigMap template name, not exact value only

rigazilla: Looks like the coexistence of multiple configs needs a proper design and since seems a little OT respect the grafana issue I would propose to move it in a separate issue/pr. wdyt?

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: I've moved grafana under the controller package since implementation is only used there

pkg/controller/infinispan/resources/operatorconfig/grafana.go
```diff
@@ -0,0 +1,95 @@
...
+	grafanav1alpha1 "github.com/integr8ly/grafana-operator/v3/pkg/apis/integreatly/v1alpha1"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	grafanaDashboardNamespace = "grafana.dashboard.namespace"
```
dmvolod, line 16: This should be probably named with `Key` postfix

pkg/controller/infinispan/resources/operatorconfig/grafana.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	currentConfig[grafanaDashboardNamespace] = ""
+	if grafanaNs == "" {
+		return &reconcile.Result{}, nil
+	}
+	// detect the GrafanaDashboard resource type resourceExists on the cluster
+	resourceExists, err := r.Kube.IsGroupVersionSupported(grafanav1alpha1.SchemeGroupVersion.String(), grafanav1alpha1.GrafanaDashboardKind)
+	if err != nil {
+		r.Log.Error(err, "Error checking Grafana support")
+		return &reconcile.Result{Requeue: true}, nil
```
dmvolod, line 43: We should add a delay with this operation, as adding Grafana operator with CRD's is a customer facing action

dmvolod, line 43: Why wen need Requeue without error propagation?

pkg/controller/infinispan/resources/operatorconfig/grafana.go
```diff
@@ -0,0 +1,95 @@
...
+
+// reconcileGrafana reconciles grafana object status with the operator configuration settings
+func (r *ReconcileOperatorConfig) reconcileGrafana(config map[string]string, currentConfig map[string]string, operatorNs string) (*reconcile.Result, error) {
+	grafanaNs := config[grafanaDashboardNamespace]
+	currGrafanaNs := currentConfig[grafanaDashboardNamespace]
+	// Delete current grafana dashboard if namespace is changes
+	if grafanaNs != currGrafanaNs && currGrafanaNs != "" {
+		currGrafana := &grafanav1alpha1.GrafanaDashboard{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      constants.GetEnvWithDefault(ApplicationNameVariable, DefaultApplicationName),
```
dmvolod, line 27: Is it possible to store all required fields inside config. Why we need to mix evv variables and ConfigMap values?

pkg/controller/infinispan/resources/operatorconfig/grafana.go
```diff
@@ -0,0 +1,95 @@
...
+	if result == nil && grafanaNs == operatorNs {
+		if err = controllerutil.SetControllerReference(operatorDeployment, infinispanDashboard, r.scheme); err != nil {
+			r.Log.Error(err, "Unable to set Grafana Dashboard controller reference to Infinispan operator - skipping creation")
+			return &reconcile.Result{Requeue: true}, nil
+		}
+	} else {
+		r.Log.Info("Infinispan operator resource not found, assuming operator is running locally.")
+	}
+
+	if err = r.Kube.Client.Create(ctx, infinispanDashboard); err != nil && !errors.IsAlreadyExists(err) {
```
dmvolod, line 65: Can we use CreateOrUpdate in case of Grafana CR changing while parameters updates (env variable change)?

pkg/controller/infinispan/resources/operatorconfig/grafana.go
rigazilla: @dmvolod what do you mean? Please example.

pkg/controller/infinispan/resources/operatorconfig/grafana.go
rigazilla: This sound really good, thanks!

pkg/controller/infinispan/resources/operatorconfig/grafana.go
dmvolod, line 16:
```suggestion
	grafanaDashboardNamespaceKey = "grafana.dashboard.namespace"
```

pkg/controller/infinispan/resources/operatorconfig/grafana.go
rigazilla: changed to the common LookupResource handling

pkg/controller/infinispan/resources/operatorconfig/operatorconfig_controller.go
rigazilla: populated with valued and moved to var

pkg/controller/infinispan/resources/operatorconfig/grafana.go
rigazilla: I've added a key suffix to all the property keys in the last commit

pkg/controller/infinispan/resources/operatorconfig/grafana.go
```diff
@@ -0,0 +1,104 @@
...
+	if err != nil {
+		r.Log.Error(err, "Error checking Grafana support")
+		return &reconcile.Result{Requeue: true}, nil
+	}
+	if !resourceExists {
+		r.Log.Info("Grafana CRD not present - not installing dashboard CR")
+		return &reconcile.Result{RequeueAfter: consts.DefaultWaitOnCreateResource}, nil
+	}
+	infinispanDashboard := emptyDashboard(config)
+	if _, err = controllerutil.CreateOrUpdate(ctx, r.client, infinispanDashboard, func() error {
```
dmvolod, line 38: CreateOrUpdate should validate for CreationTimestamp and if it IsZero() that's creation and need to invoke `controllerutil.SetControllerReference` otherwise need to update some changed fields if necessary

rigazilla: uh right, `SetControllerReference` is only on create. thanks

pkg/controller/infinispan/resources/operatorconfig/operatorconfig_controller.go
```diff
@@ -0,0 +1,137 @@
...
+	// Watch for changes to global operator config
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, &handler.EnqueueRequestForObject{}, operatorConfigPredicate)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+var _ reconcile.Reconciler = &ReconcileOperatorConfig{}
+var currentConfig map[string]string = make(map[string]string)
```
dmvolod, line 92: It would be nice to have this config available to other controllers, but due to concurrent access we need to use a blocking mechanism for it

rigazilla: yeah but I would do that in a separate PR when concurrent access will be needed

dmvolod:
> @dmvolod, I would go with this if it seems acceptable

@rigazilla let me run this PR locally, I will return with results in 5-10 minutes

dmvolod:
@rigazilla overall LGTM, just minor changes are required
- OSDK_FORCE_RUN_MODE=local setup for `make run-local` script
- Example ConfigMap `infinispan-operator-config` (preferable for upstream)
- Too often `Grafana CRD not present - not installing dashboard CR` on reconciling. Probably we can add additional delay for this kind of resource wait
- Failed to list *v1.Deployment - need to add special rights to Role

dmvolod (approved):

pkg/controller/infinispan/resources/operatorconfig/monitoring_constants.go
```diff
@@ -0,0 +1,1684 @@
+package operatorconfig
+
+const dashboardJSON = `{
```
dmvolod, line 3: Just idea for next PRs and issues: we could provide the ability to customize standard Grafana Dashboard json configuration by adding key to the ConfigMap and create it based on this value instead of hard-coded constant if needed

rigazilla: +1
yeah, that could be a good enhancement

rigazilla:
merged, thanks @dmvolod your review was invaluable!

Issue #852, opened 2021-02-25
Decouple user and operator endpoints


ryanemerson:
The operator uses the same Infinispan endpoint that is exposed to users in order to perform it's required admin/lifecycle operations. This means that any authentication/authorization settings on this endpoint have to be reflected in the operator code. So far this has been acceptable, as the operator has only allowed credential based authentication, however additional authentication mechanisms such as #851 means that this approach is no longer viable. Therefore, we need to decouple the user and operator admin endpoints.

The two endpoints, "operator" and "user", should each have their own security-realm, so that user configuration does not impact the operator. The operator endpoint would only be exposed via an internal service and it's realm would contain the operator user.

# Endpoints

## Operator
A REST endpoint is exposed on a port other than 11222.

### Credentials
The operator credential is always generated and stored in a dedicated secret per operator instance.

### Metrics
Prometheus only supports BASIC authentication. Either:

1. We allow BASIC auth on the endpoint. In which case we also create a "prometheus" user for this purpose with restricted authorization permissions.
2. We set `metrics-auth=false` so that /metrics can be consumed without authentication

### Encryption
No encryption?
Communication between endpoint and operator/prometheus is always internal within k8s cluster.

## User
The user endpoint works as before, with the exception of credential generation.

### Credentials
If no user credentials are specified, then a "developer" user is generated as before. The "operator" user is no longer contained within this secret. If a user specifies a "operator" user in a secret, then this is treated as a normal user and applied to the user security-realm.

# Image Changes
This requires an additional endpoint to be exposed by the underlying server image along with a dedicated security-realm. Furthermore, an additional identities file will need to be provided for the operator credentials e.g. ADMIN_IDENTITIES_PATH="" & IDENTITIES_PATH="".

rigazilla:
I don't think we need to expose the operator endpoint via internal service since we usually access pods directly

dmvolod:
> I don't think we need to expose the operator endpoint via internal service since we usually access pods directly

Seems to be beginning Batch CR will access Infinispan cluster via service

rigazilla:
Oh, maybe we need it then. I was thinking that a balancer between the operator and pods could split commands on different nodes and cause troubles. We should probably check that we're using the sticky mode.

ryanemerson:
> I don't think we need to expose the operator endpoint via internal service since we usually access pods directly

We need to expose an internal service for the Prometheus ServiceMonitor(s).

dmvolod:
[...]

PR #895, opened 2021-03-11
Use 11222 for the admin service port #852

ryanemerson:
@dmvolod I realised that my initial attempt at this was convoluted, which was what was adding complexity. You were right that this is the better approach.


dmvolod (approved):

dmvolod:
Thanks, @ryanemerson 
Looks good

PR #897, opened 2021-03-11
Incorrect GroupVersion combination for Route and Ingress in service controller

dmvolod:


dmvolod:
@ryanemerson it would be nice to fit this PR into release if possible

ryanemerson:
Thanks @dmvolod. The release won't be until tomorrow at the earliest now as we're waiting on the server release.

PR #910, opened 2021-03-16
revert docs for #702

oraNod:
@rigazilla Hey, as discussed on the call today we need to revert docs for xsite with different cluster names and namespaces for 2.0.x.


oraNod:
@dmvolod Thanks, I think the decision was to just keep current docs "as-is" so I reverted the part for different cluster names as well. But, yeah, you're correct about the namespaces distinction. I should've been clear in the initial PR description.

Crumby (approved):

oraNod:
@rigazilla and @dmvolod going to self-merge this one as it's needed downstream. thanks.

PR #912, opened 2021-03-16
Disabling endpoint encryption is not working after CR is created

dmvolod:


rigazilla:
overall looks good, with this it's possible to disable encryption.
However I'm not able to do the opposite: create cluster with encryption.type None and switch it to Service.
Tested with:
```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: example-infinispan
spec:
  replicas: 2
  expose:
    type: Route
    host: www.example.org
  security:
    endpointEncryption:
      type: None
```

```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: example-infinispan
spec:
  replicas: 2
  expose:
    type: Route
    host: www.example.org
  security:
    endpointEncryption:
      type: Service
```

dmvolod:
> overall looks good, with this it's possible to disable encryption.
> However I'm not able to do the opposite: create cluster with encryption.type None and switch it to Service.
> Tested with:

Oh, @rigazilla a bit strange. But what happens in this case? It seems to be forward changes was working before.

rigazilla:
> 
> Oh, @rigazilla a bit strange. But what happens in this case? It seems to be forward changes was working before.

The change is applied to the infinispan cr and then nothing happens: no certificates setup, no tls on route applied

dmvolod:
> The change is applied to the infinispan cr and then nothing happens: no certificates setup, no tls on route applied

Yeah, I see and can reproduce, but this is bit related to `ApplyEndpointEncryptionSettings` change implementation

dmvolod:
@rigazilla please have a look again, when possible. Seems to both cases have fixed.
Thanks for pointing with case.

rigazilla:
@dmvolod I'm experiencing a different problem now: tls is not enabled by default on OCP when deploying `cr_route.yaml`

rigazilla:
Not related to this, but I've noticed a mismatch between `status.podStatus` and `oc get pods`:
```
...
    podStatus:
      ready:
      - example-infinispan-1
      - example-infinispan-2
    statefulSetName: example-infinispan
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
✔ ~/go/src/github.com/infinispan/infinispan-operator [issue-837|…166⚑ 104] 
10:24 $ oc get pods
NAME                                 READY   STATUS    RESTARTS   AGE
example-infinispan-0                 1/1     Running   0          3m56s
example-infinispan-1                 1/1     Running   0          3m23s
```

dmvolod:
> Not related to this, but I've noticed a mismatch between `status.podStatus` and `oc get pods`

Oh, these statuses a but specific for OLM view and provided by library. This deployment state view not direct pod status mapping.

dmvolod:
> I'm experiencing a different problem now: tls is not enabled by default on OCP when deploying `cr_route.yaml`
Yeah, thanks, @rigazilla again, this is regression of dynamic client usage for this controller. Probably much easy to create three reconcilers rather than play with Unstructured fields, but will try to fix this before. The second problem, that we don't have Route/Ingress specific tests in CI.

rigazilla:
> The second problem, that we don't have Route/Ingress specific tests in CI.

the problem seems not strictly related to Route, encryption is not enabled also deploying `cr_minimal.yaml`.

dmvolod:
> the problem seems not strictly related to Route, encryption is not enabled also deploying `cr_minimal.yaml`.

Yes, you are totally right, it was some regression in `ApplyEndpointEncryptionSettings` related to #930 as I'm bit confusing with three functions implementations

dmvolod:
@rigazilla seems to be all issues has been fixed
Thanks

rigazilla:
> @rigazilla seems to be all issues has been fixed
> Thanks

@dmvolod yeah, reconciliation seems to work both for encryption and routes now. great!

rigazilla (approved):

ryanemerson:
Thanks @dmvolod

PR #915, opened 2021-03-18
Upgrade Policy

### LINKED ISSUE #859: When the operator upgrades user should have control of operands upgrade
Allow the user to decide whether or not upgrade the Infinispan cluster to the new operator default image.

An initial proposal (@ryanemerson) is to add the following fields to the CR:
```
type UpgradePolicy
const (
    UpgradeStrategyManual UpgradeStrategy = "Manual"
    // Example future policy. Only update base image changes, stay on same Infinispan version
    // UpgradeStrategyLatest UpgradeStrategy = "CVE"
    UpgradeStrategyLatest UpgradeStrategy = "Latest"
)

type UpgradeRollout string
const (
    UpgradeGracefulShutdown               UpgradeRollout = "Shutdown"
    UpgradeMigrateUpdateStrategy      UpgradeRollout = "Migrate"
)

type InfinispanUpgrade struct {
    Rollout *UpgradeRollout `json:"rollout,optional,omitempty"`
    Policy   *UpgradePolicy `json:"strategy,optional,omitempty"`
}
```

these new fields should allow the user to:
- decide the upgrade strategy
- stay with the current image, upgrade to latest
      - eventually more control over installed version could be provided (update only on CVE, update to specific version)

rigazilla:
Adding upgrade policy configuration via infinispan cr. Fixes #859

Current implementation is:
- Policy= "Latest" usual behaviour 
- Policy= "Manual" triggers an upgrade only if `infinispan.spec.image != pod.container[0].image`. If `infinispan.spec.image=""` the operator default image is used for the check


pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -49,6 +49,24 @@ type InfinispanServiceContainerSpec struct {
...
+	// UpgradePolicyCVE cluster upgrade is triggered automatically only if a CVE upgrade is available
+	UpgradePolicyCVE UpgradePolicy = "CVE"
+	// UpgradePolicyLatest cluster upgrade is triggered automatically if a new image is available
+	UpgradePolicyLatest UpgradePolicy = "Latest"
+)
+
+// InfinispanUpgrade represents the upgrade strategy
+type InfinispanUpgrade struct {
+	// Policy represent the current upgrade policy for this cluster
+	Policy *UpgradePolicy `json:"strategy,optional,omitempty"`
```
dmvolod, line 67: We don't need to have a pointer on simple types of we don't want to remove this field

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -49,6 +49,24 @@ type InfinispanServiceContainerSpec struct {
 	StorageClassName string  `json:"storageClassName,optional,omitempty"`
 }
 
+// UpgradePolicy specifies all the possible policies that rule the ispn cluster upgrade
```
dmvolod, line 52: It would bi nice to add `+kubebuilder:validation:Enum` annotation for all possible values

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -628,6 +628,19 @@ func (r *ReconcileInfinispan) scheduleUpgradeIfNeeded(infinispan *infinispanv1.I
...
 	desiredImage := consts.DefaultImageName
 
+	switch infinispan.Spec.Upgrade.Policy {
+	case infinispanv1.UpgradePolicyManual:
+		// In manual mode upgrade is triggered by the user changing infinispan.spec.Image field
+		// if that field is empty the operator default image is used (same as policy latest in this case)
+		if *infinispan.Spec.Image != "" {
+			desiredImage = *infinispan.Spec.Image
+		}
+	case infinispanv1.UpgradePolicyCVE: // All the rest upgrade mode works as usual atm
```
ryanemerson, line 638: I think we should only provide the "Manual" and "latest" strategies until we're able to implement this properly, otherwise it will cause confusion with users.

dmvolod (approved):

rigazilla:
obsolete

PR #927, opened 2021-03-22
Rootless Operator SDK installation

dmvolod:


build/crds-gen.sh
```diff
@@ -3,74 +3,70 @@
...
 # shellcheck disable=SC2155
 export GOROOT=$(go env GOROOT)
+OPERATOR_SDK_INSTALL_PATH=$(go env GOPATH)/bin
 
 installOperatorSDK() {
-  validateROOT
   printf "Installing Operator SDK version %s ..." "${OPERATOR_SDK_VERSION}"
   curl -LOs "${OPERATOR_SDK_URL}"
-  if grep -Fxq "Not Found" "${OPERATOR_SDK_BUNDLE}" &> /dev/null; then
+  if grep -Fxq "Not Found" "${OPERATOR_SDK_BUNDLE}" &>/dev/null; then
```
rigazilla, line 21: Just curious, how does this work?

dmvolod: This protects from incorrect Operator SDK installation. but we could use a `--fail` flag and remove this code.
Good catch, thanks, will fix it. Still working, but can be implement easily.

rigazilla (approved):
lgtm!

PR #933, opened 2021-03-24
[2.0.x] Rootless Operator SDK installation

dmvolod:


deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -106,87 +104,95 @@ spec:
...
+                - get
+                - list
+                - watch
+            - apiGroups:
+                - storage.k8s.io
+              resources:
+                - storageclasses
+              verbs:
+                - get
+                - watch
```
rigazilla, line 187: I guess this could fix something in the current released version

rigazilla (approved):

deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
dmvolod: yeah, but it's a just small case with custom storageClassName which is not so popular

dmvolod:
Thanks for review, @rigazilla 
Merged into 2.0.x

PR #941, opened 2021-03-24
Remove redundant doc dir

ryanemerson:
I think `backup.md` was added to help @oraNod originally, but no longer has any value.

`doc/design.adoc` is very out of date at this point and I don't think it adds anything that isn't already in the CRDs and actual documentation.

@oraNod @rigazilla @dmvolod WDYT to removing the dir?


dmvolod:
@ryanemerson I agree with you from time spend aspect, but I think design doc can be interesting for upstream customer who want to know deeply about CRD structure but can't read source code
@oraNod pointed to this doc several days ago
Of course, this document needs to be kept up to date, and from an architectural point of view, the structures need to be updated before coding every time we need to change / create a CRD's.

oraNod:
@dmvolod @ryanemerson What about porting `doc/design.adoc` to the documentation directory? I definitely think we need to keep a CRD reference for downstream and it's easier to maintain as part of the documentation. I say nuke it and I can add the full CRD when I tackle cache service deprecation. Pretty sure everything is in this file anyway: https://github.com/infinispan/infinispan-operator/blob/master/documentation/asciidoc/topics/yaml/datagrid_service.yaml

oraNod:
@ryanemerson Fine with me to remove the `doc` directory altogether. The intention is to have a place where contributors can add doc in md/txt/adoc files that we can then incorporate into the main documentation given that it has a specific structure and asciidoc conventions. If we need it again we can just add it.

rigazilla:
@ryanemerson fine by me to remove the `doc` dir. I'm also in favour of deleting `design.adoc`, it would be great to have a similar doc generate from the source, I think it's the only realistic way to have it aligned with the code.

ryanemerson:
> The intention is to have a place where contributors can add doc in md/txt/adoc files that we can then incorporate into the main documentation given that it has a specific structure and asciidoc conventions. I

I think we can also use GH issues, wiki or gists for this purpose without having to add anything to the repo itself.

ryanemerson:
I think we're in agreement on removal, with a view to potentially generating a similar doc in the future. Merging.

oraNod:
Created https://github.com/infinispan/infinispan-operator/issues/953 to generate a CRD reference.

PR #942, opened 2021-03-24
Reconcile loop can hangs if setup LoadBalancer expose type not supported for the hosting platform

dmvolod:


ryanemerson:
Thanks @dmvolod

PR #948, opened 2021-03-25
zero_controller should inherit PodSecurityContext used by Infinispan …

ryanemerson:


pkg/controller/zerocapacity/zero_controller.go
```diff
@@ -216,6 +215,20 @@ func (z *Controller) initializeResources(request reconcile.Request, instance Res
 		return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCluster}, nil
 	}
 
+	podList := &corev1.PodList{}
+	podLabels := ispnCtrl.PodLabels(infinispan.Name)
+	if err := z.Kube.ResourcesList(infinispan.Namespace, podLabels, podList); err != nil {
+		z.Log.Error(err, "Failed to list pods", "Infinispan.Namespace")
```
dmvolod, line 221:
```suggestion
		z.Log.Error(err, "Failed to list pods")
```

pkg/controller/zerocapacity/zero_controller.go
```diff
@@ -216,6 +215,20 @@ func (z *Controller) initializeResources(request reconcile.Request, instance Res
...
+		return reconcile.Result{}, err
+	}
+
+	if len(podList.Items) < 1 {
```
dmvolod, line 225: We can keep this code, but probably it can be a bit redundant, as WellFormed cluster has at least one Pod

ryanemerson: I actually missed that we returned false in `kube.ArePodIPsReady(podList)` if the number of pods is 0. I'll remove this code as it's redundant.

dmvolod:
Looks good, thanks, @ryanemerson 
But need to be validate on OCP cluster without admin permissions

dmvolod (approved):

rigazilla:
merged, thanks @ryanemerson !

Issue #949, opened 2021-03-25
Batch and Cache CRs should not be deleted with Infinispan CR


ryanemerson:
Currently the Cache and Batch CRs are deleted when the Infinispan CR that they target are removed. This should not happen as the CRs are user created and it could cause confusion.

In the event of a Infinispan CR being deleted we should do the following:

Cache CRs shouldn't be deleted and it should be reconciled again once a Infinispan CR with the same name has been recreated.

Batch CRs shouldn't be deleted, but are a one-time event, so if the Batch is beyond the "initializing" phase then it's not possible to replay this batch.

PR #951, opened 2021-03-25
Batch and Cache CRs should not be deleted with Infinispan CR #949

ryanemerson:
For the Cache CR I have only removed the controllerRef. If a Infinispan CR is deleted and recreated, then the cache controller will create the cache if it doesn't exist in the new cluster, however this only occurs if a reconcilliation event occurs. For this to be more robust, I think we need to watch Infinispan CR resources.


pkg/controller/batch/batch_controller.go
```diff
@@ -187,7 +187,12 @@ func (r *batchResource) initializeResources() (reconcile.Result, error) {
 	}
 
 	// We update the phase separately to the spec as the status update is ignored when in the update mutate function
-	return reconcile.Result{}, r.UpdatePhase(v2.BatchInitialized, nil)
+	_, err := r.update(func() error {
```
dmvolod, line 190: .SetControllerReference for Batch CR still present, see L177-L179

pkg/apis/infinispan/v2alpha1/batch_types.go
```diff
@@ -33,6 +34,8 @@ type BatchStatus struct {
 	Phase BatchPhase `json:"phase"`
 	// Reason indicates the reason for any batch related failures.
 	Reason string `json:"reason,omitempty"`
+	// The UUID of the Infinispan instance that the Batch is associated with
+	InfinispanUUID types.UID `json:"infinispanUUID"`
```
dmvolod, line 38: We can call it for example `clusterUUID` the same as `cluser` for Infinispan CR name
Also we can use a pointer if we will need to remove it, but probably nothing happens for Batch CR

deploy/cr/batch/batch-inline.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: infinispan.org/v2alpha1
+kind: Batch
+metadata:
+  name: example-batch-inline
+spec:
+  cluster: example-infinispan
+  config: |
+    create cache --template=org.infinispan.DIST_SYNC mycache
+    put --cache=mycache hello world
+    put --cache=mycache hola mundo
```
dmvolod, line 10:
```suggestion
    put --cache=mycache hola mundo

```

pkg/apis/infinispan/v2alpha1/batch_types.go
dmvolod: And would be nice to call postfix `UID`, not `UUID`

pkg/controller/batch/batch_controller.go
ryanemerson: Whoops, that was a result of ctrl+z when chaning my initial approach :see_no_evil:

pkg/apis/infinispan/v2alpha1/batch_types.go
ryanemerson: I've renamed it to `ClusterUID`. A pointer seems overkill to me, I think this will always be needed to prevent batch replay.

dmvolod (approved):

dmvolod:
lgtm, thanks, @ryanemerson

pkg/apis/infinispan/v2alpha1/batch_types.go
ryanemerson: I went for the pointer in the end :smile:

rigazilla (approved):

rigazilla:
merged thanks @ryanemerson !

PR #972, opened 2021-04-05
Operator config controller should utilize POD_NAME env var for Deployment lookup

dmvolod:


ryanemerson (approved):

pkg/kubernetes/pods.go
```diff
@@ -53,3 +62,45 @@ func GetEnvVarIndex(envVarName string, env *[]corev1.EnvVar) int {
...
+func GetOperatorPodOwnerRef(ns string, client crclient.Client) (*metav1.OwnerReference, error) {
+	// Get current Pod the operator is running in
+	pod, err := k8sutil.GetPod(context.TODO(), client, ns)
+	if err != nil {
+		return nil, err
+	}
+	podOwnerRefs := metav1.NewControllerRef(pod, pod.GroupVersionKind())
+	// Get Owner that the Pod belongs to
+	ownerRef := metav1.GetControllerOf(pod)
+	finalOwnerRef, err := findFinalOwnerRef(ns, ownerRef, client)
```
rigazilla, line 75: When installed via OLM the CSV owns the Deployment. Does this produce any drawback? At first sight seems not...

dmvolod: This code ported from the Operator SDK which is using for Service owner reference. Not sure that's will not work for Grafana Dashboard

dmvolod: And we don't another alternatives way if Deployment name has been changed without labels which will be supported at least in the OCP 4.10
https://github.com/operator-framework/operator-lifecycle-manager/pull/2054

rigazilla: It seems to me that Csv and Deployment have the same lifespan, so I think it's almost the same to have one or the other as owner for the dashboard object.

rigazilla (approved):

dmvolod:
I think we can merge this PR
Thanks @ryanemerson and @rigazilla for review

PR #975, opened 2021-04-06
Automatically Create Cluster ServiceMonitor

dmvolod:
Please let me know, if we wold like to remove `ServiceMonitor` and `Secret` in case of switching off (set "infinispan.org/monitoring" label to false from true or default state)

### DIFF
--- a/documentation/asciidoc/topics/proc_changing_operator_password.adoc
+++ b/documentation/asciidoc/topics/proc_changing_operator_password.adoc
@@ -1,0 +1,18 @@
+[id='changing-operator-password_{context}']
+= Changing the Operator Password
+You can change the password for the `operator` user if you do not want to use the automatically generated password.
+
+.Procedure
+
+* Update the `password` key in the `{example_crd_name}-generated-operator-secret` secret as follows:
++
+[source,bash,options="nowrap",subs=attributes+]
+----
+{oc} patch secret {example_crd_name}-generated-operator-secret -p='{"stringData":{"password": "supersecretoperatorpassword"}}'
+----
++
+[NOTE]
+====
+You should update only the `password` key in the `generated-operator-secret` secret.
+When you update the password, {ispn_operator} automatically refreshes other keys in that secret.
+====
--- a/documentation/asciidoc/topics/proc_creating_service_monitor.adoc
+++ b/documentation/asciidoc/topics/proc_creating_service_monitor.adoc
@@ -1,52 +1,25 @@
 [id='creating-service-monitor_{context}']
 = Creating a Prometheus Service Monitor
-Create a Prometheus `ServiceMonitor` that scrapes metrics from a {brandname} cluster.
-
-.Prerequisites
-
-* Have an `oc` client.
-* Enable monitoring for user-defined projects on {ocp}.
+{ispn_operator} automatically creates a Prometheus `ServiceMonitor` that scrapes metrics from your {brandname} cluster.
 
 .Procedure
 
-. Create an authentication secret for the `operator` user.
+Enable monitoring for user-defined projects on {ocp}.
+
+When the `ServiceMonitor` resource is detected, {ispn_operator} does the following:
+
+* Creates a `ServiceMonitor` named `<cluster_name>-monitor`.
+* Adds the `infinispan.org/monitoring: 'true'` annotation to your `Infinispan` CR:
 +
-Prometheus needs credentials to authenticate with {brandname} and must use the `operator` user and corresponding password.
-+
-.. Retrieve the `operator` user credentials.
-+
-[source,options="nowrap",subs=attributes+]
+[source,yaml,options="nowrap",subs=attributes+]
 ----
-include::cmd_examples/oc_get_secret_operator_creds.adoc[]
+include::yaml/infinispan-monitoring-true.yaml[]
 ----
-+
-.. Create an authentication secret, for example:
-+
-[source,options="nowrap",subs=attributes+]
-----
-include::yaml/basic_auth.yaml[]
-----
-+
-.. Add the authentication secret to your {brandname} cluster namespace.
-+
-[source,options="nowrap",subs=attributes+]
-----
-$ oc apply -f basic-auth.yaml
-----
-+
-. Create a Prometheus `ServiceMonitor` that scrapes {brandname} metrics.
-+
-[source,options="nowrap",subs=attributes+]
-----
-include::yaml/prometheus_service_monitor.yaml[]
-----
-+
-. Add the `ServiceMonitor`.
-+
-[source,options="nowrap",subs=attributes+]
-----
-$ oc apply -f service-monitor.yaml
-----
+
+[NOTE]
+====
+To authenticate with {brandname}, Prometheus uses the `operator` credentials.
+====
 
 .Verification
 
--- a/documentation/asciidoc/topics/yaml/cache_service.yaml
+++ b/documentation/asciidoc/topics/yaml/cache_service.yaml
@@ -3,6 +3,9 @@ kind: Infinispan
 metadata:
   # Names the cluster.
   name: {example_crd_name}
+  annotations:
+    # Automatically creates a ServiceMonitor for the cluster.
+    infinispan.org/monitoring: 'true'
 spec:
   # Specifies the number of nodes in the cluster.
   replicas: 4
--- a/documentation/asciidoc/topics/yaml/datagrid_service.yaml
+++ b/documentation/asciidoc/topics/yaml/datagrid_service.yaml
@@ -3,6 +3,9 @@ kind: Infinispan
 metadata:
   # Names the cluster.
   name: {example_crd_name}
+  annotations:
+    # Automatically creates a ServiceMonitor for the cluster.
+    infinispan.org/monitoring: 'true'
 spec:
   # Specifies the number of nodes in the cluster.
   replicas: 6
--- a/pkg/apis/infinispan/v1/types_util.go
+++ b/pkg/apis/infinispan/v1/types_util.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/go-logr/logr"
@@ -32,6 +33,8 @@ const (
 	PodTargetLabels string = "infinispan.org/podTargetLabels"
 	// TargetLabels labels propagated to services/ingresses/routes
 	TargetLabels string = "infinispan.org/targetLabels"
+	// ServiceMonitoringLabel defines if we need to create ServiceMonitor or not
+	ServiceMonitoringLabel string = "infinispan.org/monitoring"
 	// OperatorPodTargetLabels labels propagated by the operator to pods
 	OperatorPodTargetLabels string = "infinispan.org/operatorPodTargetLabels"
 	// OperatorTargetLabels labels propagated by the operator to services/ingresses/routes
@@ -144,6 +147,16 @@ func (ispn *Infinispan) ApplyDefaults() {
 	}
 	if ispn.Spec.Security.EndpointAuthentication == nil {
 		ispn.Spec.Security.EndpointAuthentication = pointer.BoolPtr(true)
+	}
+}
+
+func (ispn *Infinispan) ApplyMonitoringAnnotation() {
+	if ispn.Annotations == nil {
+		ispn.Annotations = make(map[string]string)
+	}
+	_, ok := ispn.GetAnnotations()[ServiceMonitoringLabel]
+	if !ok {
+		ispn.Annotations[ServiceMonitoringLabel] = strconv.FormatBool(true)
 	}
 }
 
@@ -293,6 +306,11 @@ func (ispn *Infinispan) IsGeneratedSecret() bool {
 // GetConfigName returns the ConfigMap name for the cluster
 func (ispn *Infinispan) GetConfigName() string {
 	return fmt.Sprintf("%v-configuration", ispn.Name)
+}
+
+// GetServiceMonitorName returns the ServiceMonitor name for the cluster
+func (ispn *Infinispan) GetServiceMonitorName() string {
+	return fmt.Sprintf("%v-monitor", ispn.Name)
 }
 
 // GetEncryptionSecretName ...
@@ -514,4 +532,14 @@ func applyLabels(ispn *Infinispan, envvar, annotationName string) error {
 // HasCustomLibraries true if custom libraries are defined
 func (ispn *Infinispan) HasCustomLibraries() bool {
 	return ispn.Spec.Dependencies != nil && ispn.Spec.Dependencies.VolumeClaimName != ""
+}
+
+// IsServiceMonitorEnabled validates that "infinispan.org/monitoring":true annotation defines or not
+func (ispn *Infinispan) IsServiceMonitorEnabled() bool {
+	monitor, ok := ispn.GetAnnotations()[ServiceMonitoringLabel]
+	if ok {
+		isMonitor, err := strconv.ParseBool(monitor)
+		return err == nil && isMonitor
+	}
+	return false
 }
--- a/pkg/controller/constants/constants.go
+++ b/pkg/controller/constants/constants.go
@@ -47,6 +47,8 @@ const (
 	DefaultDeveloperUser = "developer"
 	// DefaultCacheName default cache name for the CacheService
 	DefaultCacheName         = "default"
+	AdminUsernameKey         = "username"
+	AdminPasswordKey         = "password"
 	InfinispanAdminPort      = 11223
 	InfinispanAdminPortName  = "infinispan-adm"
 	InfinispanUserPortName   = "infinispan"
@@ -115,6 +117,13 @@ const (
 	DefaultLongWaitOnCreateResource = 60 * time.Second
 	//DefaultWaitClusterNotWellFormed wait delay until cluster is not well formed
 	DefaultWaitClusterNotWellFormed = 15 * time.Second
+)
+
+const (
+	ExternalTypeService = "Service"
+	ExternalTypeRoute   = "Route"
+	ExternalTypeIngress = "Ingress"
+	ServiceMonitorType  = "ServiceMonitor"
 )
 
 const DefaultKubeConfig = "~/.kube/config"
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -13,9 +13,11 @@ import (
 	"time"
 
 	"github.com/RHsyseng/operator-utils/pkg/olm"
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/go-logr/logr"
 	infinispanv1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
 	consts "github.com/infinispan/infinispan-operator/pkg/controller/constants"
+	"github.com/infinispan/infinispan-operator/pkg/controller/infinispan/resources"
 	ispn "github.com/infinispan/infinispan-operator/pkg/infinispan"
 	"github.com/infinispan/infinispan-operator/pkg/infinispan/caches"
 	config "github.com/infinispan/infinispan-operator/pkg/infinispan/configuration"
@@ -101,6 +103,16 @@ func secondaryResourceTypes() []SecondaryResourceType {
 	}
 }
 
+var supportedTypes = map[string]*resources.ReconcileType{
+	consts.ExternalTypeRoute:   {ObjectType: &routev1.Route{}, GroupVersion: routev1.GroupVersion, GroupVersionSupported: false},
+	consts.ExternalTypeIngress: {ObjectType: &networkingv1beta1.Ingress{}, GroupVersion: networkingv1beta1.SchemeGroupVersion, GroupVersionSupported: false},
+	consts.ServiceMonitorType:  {ObjectType: &monitoringv1.ServiceMonitor{}, GroupVersion: monitoringv1.SchemeGroupVersion, GroupVersionSupported: false},
+}
+
+func isTypeSupported(kind string) bool {
+	return supportedTypes[kind].GroupVersionSupported
+}
+
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
@@ -113,6 +125,16 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	err = c.Watch(&source.Kind{Type: &infinispanv1.Infinispan{}}, &handler.EnqueueRequestForObject{})
 	if err != nil {
 		return err
+	}
+
+	for index, obj := range supportedTypes {
+		// Validate that GroupVersion is supported on runtime platform
+		ok, err := kubernetes.IsGroupVersionSupported(obj.GroupVersion.String(), obj.Kind())
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Failed to check if GVK '%s' is supported", obj.GroupVersionKind()))
+			continue
+		}
+		supportedTypes[index].GroupVersionSupported = ok
 	}
 
 	// TODO(user): Modify this to be the types you create that are owned by the primary resource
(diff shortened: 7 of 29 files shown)


Crumby:
@dmvolod `infinispan.org/monitoring: true` should be automatically added between the annotations if it's not already there. If set to `false` the the related resources must be removed otherwise the monitoring will continue.

dmvolod:
> @dmvolod `infinispan.org/monitoring: true` should be automatically added between the annotations if it's not already there. If set to `false` the the related resources must be removed otherwise the monitoring will continue.

Yes, thanks @Crumby fixed. Please review.

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -135,6 +138,18 @@ func (ispn *Infinispan) ApplyDefaults() {
 	}
 }
 
+func (ispn *Infinispan) ApplyMonitoringAnnotation(supported bool) {
+	if supported {
```
ryanemerson, line 142: This seems backwards to me. If you're providing the `boolean` flag as a parameter, and no other action is taken if this flag is false, then it makes more sense for the call the `ApplyMonitoringAnnotation` to be surrounded by an `if` statement where it's called.

For example in `infinispan_controller.go`

```go
if isTypeSupported(consts.ServiceMonitorType) {
    infinispan.ApplyMonitoringAnnotation()
}
```

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -115,6 +127,19 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
 
+	for index, obj := range supportedTypes {
+		// Validate that GroupVersion is supported on runtime platform
+		ok, err := kubernetes.IsGroupVersionSupported(obj.GroupVersion.String(), obj.Kind())
+		if err != nil {
+			log.Error(err, fmt.Sprintf("Failed to check if GVK '%s' is supported", obj.GroupVersionKind()))
+			continue
+		}
+		supportedTypes[index].GroupVersionSupported = ok
+		if !ok {
```
ryanemerson, line 138: Isn't this `if` & `continue` redundant as nothing proceeds it in the loop?

pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -209,6 +208,61 @@ func (s serviceResource) cleanupExternalExpose(excludeKind string) error {
 	return nil
 }
 
+func (s serviceResource) reconcileServiceMonitor(service *corev1.Service) (reconcile.Result, error) {
+	if s.infinispan.HasServiceMonitor() && isTypeSupported(consts.ServiceMonitorType) {
```
ryanemerson, line 212: We can simplify this by starting the function with:

```go
if !isTypeSupported(consts.ServiceMonitorType) {
}

if s.infinispan.HasServiceMonitor() {

} else {

}
```

pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -209,6 +208,61 @@ func (s serviceResource) cleanupExternalExpose(excludeKind string) error {
...
+		operatorSecret := &corev1.Secret{}
+		if result, err := kube.LookupResource(s.infinispan.GetAdminSecretName(), s.infinispan.Namespace, operatorSecret, s.client, s.log); result != nil {
+			return *result, err
+		}
+		if operatorPassword, err := users.AdminPassword(operatorSecret.Name, operatorSecret.Namespace, s.kube); err != nil {
+			return reconcile.Result{}, err
+		} else {
```
ryanemerson, line 219: No need for `else` nesting.

pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -209,6 +208,61 @@ func (s serviceResource) cleanupExternalExpose(excludeKind string) error {
...
+			monitorSecret := computeServiceMonitorSecret(s.infinispan.GetServiceMonitorSecretName(), s.infinispan.Namespace, operatorPassword)
+			if _, err := controllerutil.CreateOrUpdate(ctx, s.client, monitorSecret, func() error {
+				creationTimestamp := monitorSecret.GetCreationTimestamp()
+				if creationTimestamp.IsZero() {
+					return controllerutil.SetControllerReference(service, monitorSecret, s.scheme)
+				}
+				return nil
+			}); err != nil {
+				return reconcile.Result{}, err
+			} else {
```
ryanemerson, line 229: No need for `else` nesting.

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -492,3 +517,13 @@ func applyLabels(ispn *Infinispan, envvar, annotationName string) error {
 func (ispn *Infinispan) HasCustomLibraries() bool {
 	return ispn.Spec.Dependencies != nil && ispn.Spec.Dependencies.VolumeClaimName != ""
 }
+
+// HasServiceMonitor returns true by default for ServiceMonitor
+func (ispn *Infinispan) HasServiceMonitor() bool {
```
ryanemerson, line 522: `IsServiceMonitorEnabled`?

ryanemerson (changes requested):

pkg/apis/infinispan/v1/types_util.go
dmvolod: Yes, make sense, will fix

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yeah, it was copy/paste from resource controller where logic a bit complex. Will fix

pkg/controller/infinispan/resources/service/service_controller.go
dmvolod: Yes, could be easy to define. Will fix

pkg/apis/infinispan/v1/types_util.go
dmvolod: Yes, thanks, didn't find better name, but this could be fine. Will change

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -32,6 +33,8 @@ const (
 	PodTargetLabels string = "infinispan.org/podTargetLabels"
 	// TargetLabels labels propagated to services/ingresses/routes
 	TargetLabels string = "infinispan.org/targetLabels"
+	// ServiceMonitoringLabel defines if we need to create ServiceMonitor or not
+	ServiceMonitoringLabel string = "infinispan.org/monitoring"
```
oraNod, line 37: @dmvolod For end users would this work by setting this in the Infinispan CR like:

```
metadata:
  annotations:
    infinispan.org/monitoring: true
```

dmvolod: Yes, @oraNod , will look like, `true` is string
```
metadata:
  annotations:
    infinispan.org/monitoring: 'true'
```

dmvolod:
Decided to move ServiceMonitor Secret to the secret_controller to support Infinispan CR operator Secret update

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -101,6 +103,16 @@ func secondaryResourceTypes() []SecondaryResourceType {
 	}
 }
 
+var supportedTypes = map[string]*resources.ReconcileType{
```
Crumby, line 106: Why is this PR touching Routes? How is it related to automatic ServiceMonitor creation?

dmvolod: Oh, @Crumby I changed a bit logic for optimized performance and reduce API server access, also we need an additional option to enable/disable Watch objects and don't do this for all dynamic discovery API's

Crumby (approved):

ryanemerson (approved):
@rigazilla Anything to add before I merge?

oraNod:
> @rigazilla Anything to add before I merge?

@ryanemerson Can I push a docs update before merging?

documentation/asciidoc/topics/yaml/cache_service.yaml
```diff
@@ -3,6 +3,9 @@ kind: Infinispan
 metadata:
   # Names the cluster.
   name: {example_crd_name}
+  annotations:
+    # Creates a ServiceMonitor for the cluster.
+    infinispan.org/monitoring: 'false'
```
dmvolod, line 8: `false` means disable ServiceMonitor
```suggestion
    infinispan.org/monitoring: 'true'
```

documentation/asciidoc/topics/yaml/datagrid_service.yaml
```diff
@@ -3,6 +3,9 @@ kind: Infinispan
 metadata:
   # Names the cluster.
   name: {example_crd_name}
+  annotations:
+    # Creates a ServiceMonitor for the cluster.
+    infinispan.org/monitoring: 'false'
```
dmvolod, line 8: false means disable ServiceMonitor
```suggestion
    infinispan.org/monitoring: 'true'
```

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,52 +1,26 @@
 [id='creating-service-monitor_{context}']
 = Creating a Prometheus Service Monitor
-Create a Prometheus `ServiceMonitor` that scrapes metrics from a {brandname} cluster.
+{ispn_operator} can create a Prometheus `ServiceMonitor` that scrapes metrics from a {brandname} cluster.
```
Crumby, line 3: @dmvolod @oraNod Creates automatically if ServiceMonitor resource is available?

dmvolod: Yes, @Crumby correct.

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,52 +1,26 @@
...
 
 .Prerequisites
 
-* Have an `oc` client.
 * Enable monitoring for user-defined projects on {ocp}.
 
 .Procedure
 
-. Create an authentication secret for the `operator` user.
+. Set `'true'` as the value for the `infinispan.org/monitoring` annotation in your `Infinispan` CR.
```
Crumby: Set `false` to disable? This is tricky as it's going to be platform dependent.

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
oraNod: @Crumby yes, my understanding wasn't 100% but @dmvolod has set me straight. the operator adds the annotation when it detects that the prometheus operator is installed.

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,53 +1,21 @@
...
+{ispn_operator} automatically creates a Prometheus `ServiceMonitor` that scrapes metrics from your {brandname} cluster.
 
-.Prerequisites
+.Procedure
 
-* Have an `oc` client.
 * Enable monitoring for user-defined projects on {ocp}.
 
-.Procedure
+When it detects that the Prometheus Operator is installed, {ispn_operator} adds the `infinispan.org/monitoring` annotation to your `Infinispan` CR.
```
dmvolod, line 9:
```suggestion
When it detects that the Prometheus Operator is installed, {ispn_operator} adds the `infinispan.org/monitoring: 'true'` annotation to your `Infinispan` CR.
```

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
oraNod: we call out the fact that the instructions are specific to OCP. however this is an interesting point. @dmvolod can users explicitly set `false` to disable monitoring for Infinispan clusters?

dmvolod: Correct, @oraNod explicitly setting `false` means disable even if ServiceMonitor CRD is present

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,53 +1,21 @@
...
+{ispn_operator} automatically creates a Prometheus `ServiceMonitor` that scrapes metrics from your {brandname} cluster.
 
-.Prerequisites
+.Procedure
 
-* Have an `oc` client.
 * Enable monitoring for user-defined projects on {ocp}.
 
-.Procedure
+When it detects that the Prometheus Operator is installed, {ispn_operator} adds the `infinispan.org/monitoring: 'true'` annotation to your `Infinispan` CR.
```
Crumby, line 9: Can we change Prometheus Operator to `ServiceMonitor`? Eg: `When ServiceMonitor resource presence is detected on the cluster, {ispn_operator}...`. With user workload monitoring you don't exactly deal with the Prometheus Operator and it may be confusing

rigazilla (approved):

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
oraNod: Good point, Pavel. This might also be confusing with the note that we make about the community Prometheus Operator. I've changed the wording.

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,53 +1,21 @@
...
-+
-. Add the `ServiceMonitor`.
-+
-[source,options="nowrap",subs=attributes+]
-----
-$ oc apply -f service-monitor.yaml
+include::yaml/infinispan-monitoring-true.yaml[]
 ----
 
+{ispn_operator} also creates a `ServiceMonitor` named `<cluster_name>-monitor` and an authentication secret named `<cluster_name>-monitor-secret`.
```
dmvolod, line 16: @oraNod please pay attention, that operator no longer creates a separate Secret and just utilizing single Secret for all functions

dmvolod:
It seems to be we need to update ServiceMonitor CR in case of the Secret password change to restart query process with the new password. Need to investigate more

dmvolod:
@ryanemerson , @rigazilla I changed password management logic inside operator Secret a bit. Now `password` key is single source of truth if present and possible migration steps have been added (if `password` key is not present)
Please have a look again.

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
oraNod: fixed

pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -76,25 +73,47 @@ func Add(mgr manager.Manager) error {
...
-		updated, err := s.update(secret, func() error {
-			return s.addCliProperties(secret)
-		})
-
-		if updated || err != nil {
-			return reconcile.Result{}, err
+		pass, ok := adminSecret.Data[consts.ServiceMonitorPasswordKey]
+		password := string(pass)
+		if !ok || password == "" {
+			var usrErr error
```
ryanemerson, line 101: Do we need the explicit declaration here?

ryanemerson (commented):
> @ryanemerson , @rigazilla I changed password management logic inside operator Secret a bit. Now `password` key is single source of truth if present and possible migration steps have been added (if `password` key is not present)
> Please have a look again.

Awesome @dmvolod! I tried to create a PR for this logic on Friday, but my branch wasn't working. Here's a testcase for updating the password if you want to incorporate it as part of this PR:

```go
func TestUpdateOperatorPassword(t *testing.T) {
	t.Parallel()
	// Create a resource without passing any config
	spec := tutils.DefaultSpec(testKube)
	name := strcase.ToKebab(t.Name())
	spec.Name = name
	// Register it
	testKube.CreateInfinispan(spec, tutils.Namespace)
	defer testKube.DeleteInfinispan(spec, tutils.SinglePodTimeout)
	testKube.WaitForInfinispanPods(1, tutils.SinglePodTimeout, spec.Name, tutils.Namespace)
	testKube.WaitForInfinispanCondition(spec.Name, spec.Namespace, ispnv1.ConditionWellFormed)

	newPassword := "supersecretoperatorpassword"
	secret := testKube.GetSecret(spec.GetAdminSecretName(), spec.Namespace)
	kube.CreateOrPatch(context.TODO(), testKube.Kubernetes.Client, secret, func() error {
		secret.Data["password"] = []byte(newPassword)
		return nil
	})

	err := wait.Poll(tutils.DefaultPollPeriod, tutils.SinglePodTimeout, func() (bool, error) {
		secret = testKube.GetSecret(spec.GetAdminSecretName(), spec.Namespace)
		identities := secret.Data[cconsts.ServerIdentitiesFilename]
		pwd, err := users.FindPassword(cconsts.DefaultOperatorUser, identities)
		tutils.ExpectNoError(err)
		fmt.Printf("Pwd=%s, Identities=%s", string(pwd), string(identities))
		return pwd == newPassword, nil
	})
	tutils.ExpectNoError(err)
}
```

pkg/controller/constants/constants.go
```diff
@@ -117,6 +117,16 @@ const (
...
 )
 
+const (
+	ExternalTypeService = "Service"
+	ExternalTypeRoute   = "Route"
+	ExternalTypeIngress = "Ingress"
+	ServiceMonitorType  = "ServiceMonitor"
+
+	ServiceMonitorUserKey     = "username"
+	ServiceMonitorPasswordKey = "password"
```
ryanemerson, line 127: `AdminUsernameKey` & `AdminPasswordKey`?

pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -76,25 +73,47 @@ func Add(mgr manager.Manager) error {
 }
 
 func (s *secretResource) Process() (reconcile.Result, error) {
-	secret, err := s.getSecret(s.infinispan.GetAdminSecretName())
-	if err != nil {
-		return reconcile.Result{}, err
+	adminSecret := &corev1.Secret{
```
ryanemerson, line 76: It would be cleaner if we add the admin secret logic to a function that's called within process so that it's self contained.

pkg/controller/constants/constants.go
dmvolod: Yeah, right, now entity has been changed

pkg/controller/infinispan/resources/secret/secret_controller.go
dmvolod: Thanks, @ryanemerson will embed secret update into test.
We need to declare var here, to play with the password var

pkg/controller/infinispan/resources/secret/secret_controller.go
dmvolod: Yes, correct, will create a separate function. It will be cleaner solution

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,53 +1,26 @@
...
+
+[source,yaml,options="nowrap",subs=attributes+]
 ----
-$ oc apply -f service-monitor.yaml
+include::yaml/infinispan-monitoring-true.yaml[]
 ----
 
+[NOTE]
+====
+To authenticate with {brandname}, Prometheus uses the `operator` credentials.
```
dmvolod, line 21: @ryanemerson do we need to document how to change operator admin password and that `password` filed is primary truth entry?

ryanemerson: I tried pushing the following commit to this PR, but it was rejected. Here's the docs I wrote on Friday:

https://github.com/infinispan/infinispan-operator/compare/master...ryanemerson:password_doc_updates?expand=1

rigazilla (approved):
code looks good to me.
maybe it's just my concern, but it would be great to have a non admin password to share just for scraping metrics. maybe in a separate PR

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
oraNod: @dmvolod and @ryanemerson I can add those docs to this PR. weird that that commit was rejected.

dmvolod:
> maybe it's just my concern, but it would be great to have a non admin password to share just for scraping metrics. maybe in a separate PR

@rigazilla can we fit this option now, I'm worry about migration path. It becomes more complex and complex

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
dmvolod: Thanks @oraNod and @ryanemerson

rigazilla:
> > maybe it's just my concern, but it would be great to have a non admin password to share just for scraping metrics. maybe in a separate PR
> 
> @rigazilla can we fit this option now, I'm worry about migration path. It becomes more complex and complex

yeah @dmvolod let's go with this. I think we can evaluate improvement later

documentation/asciidoc/topics/proc_changing_operator_password.adoc
```diff
@@ -0,0 +1,18 @@
...
++
+[source,bash,options="nowrap",subs=attributes+]
+----
+{oc} patch secret {example_crd_name}-generated-operator-secret -p='{"stringData":{"password": "supersecretoperatorpassword"}}'
+----
++
+[NOTE]
+====
+You should update only the `password` key in the `generated-operator-secret` secret.
+When you update the password, {ispn_operator} automatically refreshes other keys in that secret.
```
Crumby, line 17: @dmvolod @ryanemerson What happens at this point, is the whole cluster restarted?

Prometheus has side container that's watching for the changes in the resources provided by secrets/configmaps and when the change happens it triggers config reload on server so it doesn't restart the nodes.

dmvolod: @Crumby this is related not only for Prometheus, Infinispan cluster need to be restarted when admin password has been changed

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,53 +1,26 @@
...
 = Creating a Prometheus Service Monitor
-Create a Prometheus `ServiceMonitor` that scrapes metrics from a {brandname} cluster.
+{ispn_operator} automatically creates a Prometheus `ServiceMonitor` that scrapes metrics from your {brandname} cluster.
 
-.Prerequisites
+.Procedure
 
-* Have an `oc` client.
-* Enable monitoring for user-defined projects on {ocp}.
+Enable monitoring for user-defined projects on {ocp} so that the `ServiceMonitor` resource is available on the {openshiftshort} cluster.
```
Crumby, line 7: I believe that `ServiceMonitor` resource is always available on OCP clusters. But if user workload monitoring isn't enabled it's instance presence won't have any effect.

documentation/asciidoc/topics/proc_changing_operator_password.adoc
dmvolod: Also I can't get Prometheus side container works without ServiceMonitor update. I tried to change Infinispan cluster password, but after that Prometheus metrics never been collected

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
```diff
@@ -1,53 +1,26 @@
...
 
-.Procedure
+When it detects the `ServiceMonitor` resource, {ispn_operator} does the following:
```
Crumby, line 9: Would it make sense to use rather _When the `ServiceMonitor` resrouce is detected on the cluster, ..._. Calling the Operator _it_ doesn't fits in here as the context is missing

documentation/asciidoc/topics/proc_creating_service_monitor.adoc
oraNod: makes sense. I'll change.

documentation/asciidoc/topics/proc_changing_operator_password.adoc
Crumby: > I can't get Prometheus side container works without ServiceMonitor update

Right, that's not good, is the ServiceMonitor recreated then so the Prometheus catches up or is there anything needed on user side?

My initial idea was about the need to restart the server though and I used Prometheus CR as an example of implementation which might caused misunderstanding.

Infinispan server itself doesn't need to be restarted to catch on updated credentials. If we had side container catching up on the secret changes we could trigger config-generator reload which would update all user.properties file on fly while server would be still up and running. But this is way out of scope of this PR

dmvolod: > Right, that's not good, is the ServiceMonitor recreated then so the Prometheus catches up or is there anything needed on user side?

Nothing need, I updated ServiceMonitor CR with the password hash annotation

dmvolod: > > I can't get Prometheus side container works without ServiceMonitor update
> 
> Right, that's not good, is the ServiceMonitor recreated then so the Prometheus catches up or is there anything needed on user side?
> 
> My initial idea was about the need to restart the server though and I used Prometheus CR as an example of implementation which might caused misunderstanding.
> 
> Infinispan server itself doesn't need to be restarted to catch on updated credentials. If we had side container catching up on the secret changes we could trigger config-generator reload which would update all user.properties file on fly while server would be still up and running. But this is way out of scope of this PR

Yes, true, We would have side effect in the future, but better to implement as separate PR

oraNod (approved):
LGTM

ryanemerson:
@dmvolod LGTM as well. Can you just rebase so that "pdrobek comments" and "Docs" commit are combined and then I will merge?

dmvolod:
> Can you just rebase so that "pdrobek comments" and "Docs" commit are combined and then I will merge?

Yes, sure, done. Thanks, @ryanemerson

ryanemerson:
Great work @dmvolod, automatically creating the `ServiceMonitor` and the changes to the admin secret will greatly simplify things for the user :slightly_smiling_face:

PR #977, opened 2021-04-06
ISPN-12916 [Operator] Grafana is Querying on stat that exists for def…

wburns:
…ault cache only

https://issues.redhat.com/browse/ISPN-12916


Crumby (approved):

ryanemerson:
Thanks @wburns

PR #980, opened 2021-04-07
Backup/Restore investigation

ryanemerson:


pkg/controller/zerocapacity/zero_controller.go
```diff
@@ -293,12 +293,10 @@ func (z *Controller) cleanupResources(httpClient http.HttpClient, request reconc
 	var allErrors error
 
 	// Stop the zero-capacity server so that it leaves the Infinispan cluster
-	rsp, err, reason := httpClient.Post(request.Name, consts.ServerHTTPServerStop, "", nil)
+	rsp, err, reason := httpClient.Post(meta.Name, consts.ServerHTTPServerStop, "", nil)
```
dmvolod, line 296: @ryanemerson it looks like we have duplicate methods for shutdown server or it's temporary solution just for tests as we have `cluster.GracefulShutdown` method

ryanemerson: Graceful shuts down the cluster not the individual server.

dmvolod: Oh, true, missed this.
In this case we should review some cluster vs individual server interaction and try to optimize it. In some cases Pod[0] can be very heated, but we can interact with cluster via service, as it's much more faster than call curl inside Pod.

ryanemerson: In this case we need to speak to a specific pod, the zero-capacity one, so I'm not sure we can go via the service. I could make a Server interface similar to `ClusterInterface` to abstract the details of the REST call for future reuse

dmvolod: Yes, sure, this not for current case, just to discuss and remember it.

test/e2e/utils/httpClient.go
```diff
@@ -85,16 +82,13 @@ func (c *httpClientConfig) Post(path, payload string, headers map[string]string)
...
 func (c *httpClientConfig) exec(req *http.Request, headers map[string]string) (*http.Response, error) {
 	if c.isAuthRequired() {
-		if c.authRealm == nil {
-			rsp, err := c.Do(req)
-			ExpectNoError(err)
-			if rsp.StatusCode != http.StatusUnauthorized {
-				return rsp, fmt.Errorf("Expected 401 DIGEST response before content: %v", rsp)
-			}
-			c.authRealm = getAuthorization(*c.username, *c.password, rsp)
+		rsp, err := c.Do(req)
```
dmvolod, line 85: @ryanemerson probably this situation could happens when later `rsp.Body.Close()` is not calling

.travis.yml
```diff
@@ -31,27 +31,6 @@ jobs:
         - make unit-test
         - make lint
```
dmvolod, line 32:
```suggestion
```

test/e2e/utils/httpClient.go
dmvolod, line 85: @ryanemerson probably we make `rsp, err := c.Do(req)` calls twice and didn't close first `rsp` and this is the reason why after CR deletion connection still open and tries to invoke on the finish

ryanemerson: Good point, I had missunderstood what you meant about the closing of requests previously. Updated

ryanemerson: I think I finally have it :sweat_smile: 

I have added a DEBUG constant to `httpClient.go` that dumps all http req/rsp headers.

ryanemerson:
The backup/restore tests will pass once the `12.1.2.Final` image has been released. As these changes only impact backup/restore tests and these are already failing, this PR should be ok to merge now.

dmvolod (approved):

rigazilla:
merged thanks @ryanemerson !

Issue #981, opened 2021-04-07
path is out of place in the OLM catalog yaml


oraNod:
For both backup and restore CR, `"path": "asdasd"` does not belong and needs to be removed.

oraNod:
https://github.com/infinispan/infinispan-operator/pull/982

PR #982, opened 2021-04-07
Remove path field from OLM entry for Backup and Restore

oraNod:
issue #981

### DIFF
--- a/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
@@ -26,8 +26,7 @@ metadata:
               "cpu": "1000m",
               "extraJvmOpts": "-Djava.property=me",
               "memory": "1Gi"
-            },
-            "path": "asdasd"
+            }
           }
         },
         {
@@ -56,8 +55,7 @@ metadata:
               "cpu": "1000m",
               "extraJvmOpts": "-Djava.property=me",
               "memory": "1Gi"
-            },
-            "path": "asdasd"
+            }
           }
         },
         {


deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -27,7 +27,6 @@ metadata:
               "extraJvmOpts": "-Djava.property=me",
               "memory": "1Gi"
             },
-            "path": "asdasd"
```
Crumby, line 30: Isn't the field mandatory and we should change the `asdasd` to some more meaningful example default? @ryanemerson?

ryanemerson: @Crumby `path` is not part of the `Backup.Spec`.

ryanemerson (approved):

deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -27,7 +27,6 @@ metadata:
               "extraJvmOpts": "-Djava.property=me",
               "memory": "1Gi"
             },
```
Crumby, line 29: `,` needs to be removed as the block is now last element, otherwise it's going to result in invalid JSON

deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -57,7 +56,6 @@ metadata:
...
```
Crumby, line 58: `,` needs to be removed as the block is now last element, otherwise it's going to result in invalid JSON

Crumby (changes requested):

Issue #986, opened 2021-04-08
Server overlay configuration support


rigazilla:
Operator should support an overlay mechanism to run Infinispan with a user provided set of configuration files (infinispan.xml, jgroups.xml...).
This is a dependency for #958

Issue #958, opened 2021-03-30
Operator Future and beyond!!!!


ryanemerson:
# Goal
Allow users the freedom to customise Infinispan deployments without having to expose all configuration options via the Infinispan CR.

# Proposal
Allow users to provide Infinispan configuration files explicitly via a ConfigMap.

1. A ConfigMap containing the required configuration is created by the user; any updates to these files results in the cluster being rolled out again.

2. The operator overlays any additional configuration required to interact with the operand. So far this includes:
    - Admin endpoint
    - Admin security-realm
    - JGroups transport/ping protocol
    - JGroups xsite container & jgroups protocols

3. Launch the server using the layered configuration

## User Configuration Snippets
 So that it's not required for the user ot provide an entire Infinispan configuration, we can also make it so that it's only necessary for users to provide "snippets" corresponding to a particular area of configuration. For example, provide a snippet for the container, transport, authentication and user endpoint.

Configuration Snippets would be provided via the ConfigMap using specific keys, e.g. container.conf -> container configuration file.

## Static Cache Configuration
Users are able to provide static cache configurations via the Infinispan configuration that will be available on server startup. Such caches need to be marked as immutable by the server so that they cannot be deleted at runtime. Without this, it's possible for a cache to be deleted at runtime and then recreated when a cluster is rolledout again during a Gracefulshutdown upgrade.

## Operator Configuration
Currently the Infinispan server is configured by providing a yaml file to the Infinispan image, which transforms the yaml to the required xml via the config-generator artifact. This approach simplifies the operator code, however it requires the maintainence and release of an additional artifact outside of the code base. As the overlay configuration required by the operator has no application outside of the operator, we should generate the required configuration (XML or Yaml) in the operator directly and simply pass these to the server artifacts in the Infinispan image.

## Infinispan CR Development
This proposal provides a mechanim for users to configure the infinite Infinispan knobs directly. Therefore, we should limit future Infinispan CR spec additions to options that directly affect the k8s deployment and/or where we provide some kind of automation/simplification. A good example of a Infinispan configuration that is appropriate in this CR is `spec.Security`, as this provides a high-level spec that is utilised by the operator alongside with the capabilities of the local k8s cluster, to automate Encryption configuration. Cluster content configuration should then be limited to appropriate CRs, such as the `Batch` or `Cache` CR.

# Blockers
- Server overlay configuration support
- Immutable static cache configuration

PR #987, opened 2021-04-08
Overlay configuration #986

rigazilla:
Support for users that want to run infinispan with a full custom image provided via configmap.
Current solution replaces the whole `/opt/infinispan/server/conf` folder with the content of the provided config map.

Should be enable via Infinispan CR this way:
```
spec:
  serverConfiguration:
    type: free
    configmap: my-config-map
```

I working example of configmap that mimic the minimal Infinispan configuration deployed with the `minimal_cr.yaml` file is here: https://gist.github.com/rigazilla/531fcedd74fda800caef88e131579281


ryanemerson:
Isn't this premature without support for overlaying multiple configuration files on the server? Allowing the user to supply the entire `server/conf` directory without this, means that in order for the operator to continue functioning the user has to provide some very specific configuration elements in order for the operator to continue functioning. For example, to ensure that the operator is able to communicate with the server endpoint the user needs to include:


```xml
         <security>
                <security-realm name="admin">
                    <properties-realm groups-attribute="Roles">
                        <user-properties path="admin/users.properties" relative-to="infinispan.server.config.path"
                                         plain-text="true"/>
                        <group-properties path="admin/groups.properties" relative-to="infinispan.server.config.path"/>
                    </properties-realm>
                </security-realm>
       </security>
        <endpoints socket-binding="admin" security-realm="admin">
            <rest-connector>
                <authentication mechanisms="BASIC DIGEST"/>
            </rest-connector>
        </endpoints>
```

It's also necessary for the user to ensure that JGroups is configured appropriately. I don't even want to consider what's required for XSite :smile:.

IMO for customer configuration to be a viable option we need a solution that allows the operator to ensure that the bare-minimum configuration is present, which means waiting for overlay support on the server.

rigazilla:
@ryanemerson the provided configuration has everything inside to run a cr_minimal.yaml cluster (even the admin setup), this just because I've copied it from a Config Generated pod instance. This is just an experimental pr to see how things can go far down this way atm.
but you're right atm the user is leaved alone in the wild :)

I wondering if the operator can add more intelligence inspecting the provided xml

ryanemerson:
> the provided configuration

I'm not sure which configuration you're referring to? I don't see anything in this PR.

> I wondering if the operator can add more intelligence inspecting the provided xml

I think the idea of xml/config overriding was that we generate the minimum required config to apply on top of the user provided config. So that way there is no need for the operator to worry about config semantics, we just generate what's required and let the server do the rest.

ryanemerson:
Overlay capabilities will be available sooner rather than later https://github.com/infinispan/infinispan/pull/9226 :slightly_smiling_face:

rigazilla:
> > the provided configuration
> 
> I'm not sure which configuration you're referring to? I don't see anything in this PR.
> 

it's the gist linked in the first comment, it's a configmap with a full user configuration in (https://gist.github.com/rigazilla/531fcedd74fda800caef88e131579281)


> I think the idea of xml/config overriding was that we generate the minimum required config to apply on top of the user provided config. So that way there is no need for the operator to worry about config semantics, we just generate what's required and let the server do the rest.

I think I'm following more or less that direction, atm the user must provide everything, but we can add more steps to automatically setup admin and jgroups (as per [proposal](https://github.com/infinispan/infinispan-operator/issues/958)). overlay on server can be of help in this. Though I think we need to verifiy that the overlay approach can fit all the cases.

> > I wondering if the operator can add more intelligence inspecting the provided xml
> 
this is probably not something for now... i.e. the operator should be able to deal with configuration that requires persistence on DB (if a setup is required on k8s)

ryanemerson:
> it's the gist linked in the first comment, it's a configmap with a full user configuration in (https://gist.github.com/rigazilla/531fcedd74fda800caef88e131579281)

Sorry I looked at everything but the description :see_no_evil:

ryanemerson:
> I think I'm following more or less that direction, atm the user must provide everything, but we can add more steps to automatically setup admin and jgroups (as per [proposal](https://github.com/infinispan/infinispan-operator/issues/958)). overlay on server can be of help in this. Though I think we need to verifiy that the overlay approach can fit all the cases.


[ISPN-12931](https://issues.redhat.com/browse/ISPN-12931) adds overlay support to the server. This overlay support is format independent, so xml or yaml files can be provided. The latter would be the most convenient for the operator, as we can just create structs to represent the hierarchy in a similiar way to how currently create the image config in `configuration.go`.

The following is the yaml overlay required to add the admin endpoint, security realms and the DNS ping protocol to the tcp stack. My thoughts were that we would simply overlay this over whatever config the user provides. For example, the following provides all of the bits required by the operator for non-xsite use:

`./bin/server.sh -c user-infinispan.[xml | yaml] -c operator-override.yaml`
```
infinispan:
  jgroups:
    stacks:
      - stack:
          name: image-tcp
          extends: tcp
          TCP:
            bind_addr: ${jgroups.bind.address,jgroups.tcp.address:127.0.0,1}
            bind_port: ${jgroups.bind.port,jgroups.tcp.port:7800}
            enable_diagnostics: true
          dns.DNS_PING:
            dns_query: example-infinispan-ping.default.svc.cluster.local
            dns_record_type: A
            stack.combine: REPLACE
            stack.position: MPING
  cacheContainer:
    transport:
      cluster: ${infinispan.cluster.name:example-infinispan}
      nodeName: ${infinispan.node.name:}
      stack: image-tcp # Make this configurable via the Infinispan CR so that a complete custom stack can be used if specified by the user (dangerzone)
  server:
    socketBindings:
      defaultInterface: public
      portOffset: 0
      bindings:
        - socketBinding:
            name: admin
            port: 11223
    security:
      securityRealms:
        - securityRealm:
            name: admin
            propertiesRealm:
              userProperties:
                path: users.properties
                plainText: true
              groupProperties:
                path: groups.properties
    endpoints:
      socketBinding: admin
      securityRealm: admin
      connectors:
        restConnector:
          authentication:
            mechanisms: BASIC DIGEST
```


The only gotcha, is  configuring the JGroups stack, which I think we may need a way for the user to configure explicitly OR we have multiple overlays so that some operator defaults are overridable and some not (e.g. the admin endpoint). So the server args would look something like:

`./bin/server/sh  -c infinispan.xml -c operator-defaults.yaml -c user-infinispan.[xml | yaml] -c operator-essentials.yaml`

rigazilla:
I would like to see both the options:
- user-base+infinispan-overlay
- infinispan-base+user-overlay (supposing for now that user is educated and doesn't mess up mandatory setup)
il setup a prototype as soon as the server will support overlay

imo stay with XML is a better solution: XML seems to me more close to the Infinispan configuration language. since we're doing something new I would consider to skip the yaml step if feasible

ryanemerson:
> I would like to see both the options in action:
> 
> * user-base+infinispan-overlay
> * infinispan-base+user-overlay (supposing for now that user is educated and doesn't mess up mandatory setup)

I was thinking that the mandatory overlay would contain the bare minimum required for the operator to work, i.e. the admin endpoint and security realm. IMO there's no advantage to not applying this last, as without it we're blindly relying on the assumption the user always knows what they're doing.

> imo stay with XML is a better solution: XML seems to me more close to the Infinispan configuration language. since we're doing something new I would consider to skip the yaml step if feasible

To be clear, it's possible to mix `xml` and `yaml` overlays. So it's possible for us to provide our overlays in `yaml` or `xml` and the same for the user.

An advantage of us using yaml for the operator overlays is that we can just create structs to represent the underlying server configuration and then use `.Marshall()` similar to how we use `configuration.go` atm. I'm not familiar with go's xml capabilities though, so maybe generating xml is just as trivial.

tristantarrant:
> To be clear, it's possible to mix `xml` and `yaml` overlays. So it's possible for us to provide our overlays in `yaml` or `xml` and the same for the user.

And soon: JSON.

rigazilla:
> And soon: JSON.

that's good for sure on the user side.
I'm just wondering if staying with XML internally could simplify things and reduce risk of bugs, but yeah I know we already have yaml inside
I'll check the XML support for go anyway

ryanemerson:
FYI @rigazilla I already converted the config-generator to use `yaml` as part of my review of Tristan's yaml parser and with an eye on future features (Authorization in particular). My branch maybe of interest so you can see what it looks like in action: https://github.com/infinispan/infinispan-image-artifacts/compare/master...ryanemerson:3.0.x

I'm not sure refactoring the config-generator ^ is the correct thing todo anymore. The flexibility that overlays provide (most of the config can be static), combined with the ability to keep all the code in one place (much more extensible, one less thing to maintain), makes me think handling all configuration  in the operator could be a good goal longer term.

rigazilla:
> FYI @rigazilla I already converted the config-generator to use `yaml` as part of my review of Tristan's yaml parser and with an eye on future features (Authorization in particular). My branch maybe of interest so you can see what it looks like in action: [infinispan/infinispan-image-artifacts@master...ryanemerson:3.0.x](https://github.com/infinispan/infinispan-image-artifacts/compare/master...ryanemerson:3.0.x)

that's fine, I was not suggesting to introduce XML in the config-generator

> 
> I'm not sure refactoring the config-generator ^ is the correct thing todo anymore. The flexibility that overlays provide (most of the config can be static), combined with the ability to keep all the code in one place (much more extensible, one less thing to maintain), makes me think handling all configuration in the operator could be a good goal longer term.

+100 
this is by far my preferred option

PR #988, opened 2021-04-08
ISPN-12926 [Operator] Grafana dashboard does not differentiate betwee…

wburns:
…n the clusters with same names

https://issues.redhat.com/browse/ISPN-12926


ryanemerson:
Thanks @wburns

Issue #989, opened 2021-04-09
Batch credentials are too exposed


ryanemerson:
Currently the Batch CR cli command exposes the operator credentials in plain text when outputting the created job/pod. Instead we should utilise a `~/.config/infinispan/cli.properties` file with the `autoconnect-url` property so that only the following will be exposed:

```
/opt/infinispan/bin/cli.sh --file '/etc/batch/batch'
```

PR #990, opened 2021-04-09
Batch credentials are too exposed #989

ryanemerson:


Crumby:
@ryanemerson Is the secret with `cli.properties` created for each Batch? Could we create a single one altogether with the initial cluster that'd be used with all the batches targeting that particular cluster?

pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -75,16 +76,25 @@ func Add(mgr manager.Manager) error {
...
 	// If the operator secret does not already exist create it with generated identities
-	if !secretExists {
+	if secret == nil {
 		if err = s.createAdminIdentitiesSecret(); err != nil {
 			return reconcile.Result{}, err
 		}
+	} else {
+		// Patch secret to include any generated fields
+		updated, err := s.update(secret, func() error {
+			return s.addCliProperties(secret)
```
ryanemerson, line 92: @dmvolod I was envisaging that you could add an additional call to this update function and in `createSecret()` to a method that generates the monitoring `username` & `password` fields

dmvolod: Yes, cool, @ryanemerson 
We need to merge this PR before and I think this section will be permanent and never been deleted

Crumby (approved):

dmvolod:
@ryanemerson you missed `patch` rights in Role for Secret

dmvolod (approved):

Issue #962, opened 2021-03-31
Infinispan Cloud Event integration must be configurable by the user


rigazilla:
The user must be allowed to configure Cloud Event integration in the infinispan configuration.

A suitable solution could be:
- add in the infinispan CR a cloudEvent field as follow:
```
cloudevent:
  bootstrap-servers: 
  acks:
  cache-entries-topic:
```
- propagate it to the config-generator that will add the following xml fragment in the infinispan.xml:
```
 <cloudevents bootstrap-servers="127.0.0.1:9092" acks="0" cache-entries-topic="cache-events"/>
 ```

dmvolod:
@rigazilla can we close this issue as seems to be implemented

PR #996, opened 2021-04-13
cloud event docs #962

oraNod:
Docs for https://github.com/infinispan/infinispan-operator/pull/964


rigazilla (approved):
lgtm

documentation/asciidoc/topics/yaml/cloud_events.yaml
```diff
@@ -0,0 +1,8 @@
+spec:
+  cloudEvents:
+    # Configures the number of acks.
+    acks: "1"
```
dmvolod, line 4: just for information, acks could be number or "all" value, thanks the reason why it's a string
https://betterprogramming.pub/kafka-acks-explained-c0515b3b707e

oraNod: nice one @dmvolod even though that might be kafka specific I've updated the comment with some additional detail

dmvolod (approved):

documentation/asciidoc/topics/yaml/cloud_events.yaml
dmvolod: @rigazilla could we add a validation for the acks values in CRD?

PR #1006, opened 2021-04-20
Able to download custom code as external artifacts from http/ftp source

dmvolod:
This very simple method to embed third-party libraries to be running together with the Infinispan server runtime utilizing init containers and ephemeral storage. Persistence volume embedding already implemented will have higher priority
Pros

- No additional permanent storage needed
- Simplicity
- Fastest start for users

Cons

- Embedding will be running each time Pod's are creating. In complex solutions Build could be utilized, but for builds two path are requires, Kaniko builds for k8s, and special separate Builds for OCP and OKD 
- No permanent image will be prepared, see build approach above
- Additional http/ftp server endpoint needed if not already present (not needed for maven based libraries)


pkg/apis/infinispan/v1/types_util.go
```diff
@@ -525,9 +534,31 @@ func applyLabels(ispn *Infinispan, envvar, annotationName string) error {
 	return err
 }
 
-// HasCustomLibraries true if custom libraries are defined
-func (ispn *Infinispan) HasCustomLibraries() bool {
-	return ispn.Spec.Dependencies != nil && ispn.Spec.Dependencies.VolumeClaimName != ""
+// GetExternalDependencyType return external dependency type if defined
+func (ispn *Infinispan) GetExternalDependencyType() ExternalDependencyType {
```
ryanemerson, line 538: I think only allowing one dependency type is too limiting. Infinispan Server recursively searches directories in `/opt/infinispan/server/lib` for files, so it's possible for us to support both `Dependencies.VolumeClaimName` & `Dependencies.ExternalLibraries` at the same time by using different mount paths within `server/lib`.

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -218,6 +218,12 @@ type InfinispanExternalDependencies struct {
 	// Name of the persistent volume claim with custom libraries
 	// +optional
 	VolumeClaimName string `json:"volumeClaimName,omitempty"`
+	// +optional
+	ExternalLibraries []InfinispanExternalDependencyLibrary `json:"externalLibraries,omitempty"`
```
ryanemerson, line 222: Files placed in `server/libs` don't have to be libraries. WDYT about keeping this simple and just having:

```
Remote []string `json:"remote, omitempty"`
```

This means the user would just do:

```yaml
spec:
  dependencies:
    remote:
      - http://example.com/lib1.jar
      - http://example.com/lib2.jar
```

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -525,9 +534,31 @@ func applyLabels(ispn *Infinispan, envvar, annotationName string) error {
...
+		if ispn.Spec.Dependencies.VolumeClaimName != "" {
+			return ExternalDependencyTypeVolume
+		} else if len(ispn.Spec.Dependencies.ExternalLibraries) > 0 {
+			return ExternalDependencyTypeLibrary
+		}
+	}
+	return ExternalDependencyTypeNone
+}
+
+func (ispn *Infinispan) GetExternalDependencyLibrariesExtractCommand(mountPath string) string {
```
ryanemerson, line 549: I think this logic should belong in the controller as it defines the `initController` behaviour.

ryanemerson (changes requested):
Great work @dmvolod, just a few points.

pkg/apis/infinispan/v1/types_util.go
tristantarrant: Additionally, aside from jars, the server supports plain files (properties, etc). Is this allowed in the PR ?

pkg/apis/infinispan/v1/infinispan_types.go
dmvolod: Agree with naming like `remote` or `artifacts` (Strimzi using this type name)
Yeah, original implementation was simple like you provided, but if we will plan to have additional entries, like explicitly defined type (i.e. `zip`, `tgz`, etc.), sha512sum, etc it will be not possible to extend without migration.

Please see how AMQ Streams (Strimzi) defines plugin downloads
https://access.redhat.com/documentation/en-us/red_hat_amq/2021.q2/html-single/using_amq_streams_on_openshift/index#type-Build-reference
```
artifacts:
  - type: tgz
  url: https://repo1.maven.org/maven2/io/debezium/debezium-connector-postgres/1.3.1.Final/debezium-connector-postgres-1.3.1.Final-plugin.tar.gz
  sha512sum: 962a12151bdf9a5a30627eebac739955a4fd95a08d373b86bdcea2b4d0c27dd6e1edd5cb548045e115e33a9e69b1b2a352bee24df035a0447cb820077af00c03
```

pkg/apis/infinispan/v1/types_util.go
dmvolod: Yes, agreed. Originally it was here, but moved to reduce main controller code size

pkg/apis/infinispan/v1/types_util.go
dmvolod: > I think only allowing one dependency type is too limiting. Infinispan Server recursively searches directories in `/opt/infinispan/server/lib` for files, so it's possible for us to support both `Dependencies.VolumeClaimName` & `Dependencies.ExternalLibraries` at the same time by using different mount paths within `server/lib`.

Good point, @ryanemerson will add this implementation to support both option working in parallel.

dmvolod: > Additionally, aside from jars, the server supports plain files (properties, etc). Is this allowed in the PR ?

Yes, any type of the file can be download in two options:
- inside archive (preferred for multiple files)
- if extension not archive or skipped (URL rewriting) will download as original file

pkg/apis/infinispan/v1/infinispan_types.go
ryanemerson: Thanks for the Strimzi example. I didn't consider optional sha's before ... that would be a great addition.

Do we really need sha-512 for verification? I think sha-256 is what's used in most places, e.g. container images.

How about we make it so the `type` and `sha` fields are optional. If no type is specified, then we can use the extension to determine how to handle the file, however if the type is provided we always process the artifact in that manner. Initially we could provide support for `type = file | zip | tgz` as you have already implemented this.

For example:

```yaml
spec:
  dependencies:
    artifacts:
      # Copy file as is
      - url: http://example.com/lib1.jar
      # Extract jar contents
      - url: http://example.com/lib1.jar
        type: zip
      # Don't extract as type explicitly configured
      - url: http://example.com/lib1.zip
        type: file
      # Extract as no type, but .zip extension
      - url: http://example.com/lib1.zip
```

dmvolod: > Do we really need sha-512 for verification? I think sha-256 is what's used in most places, e.g. container images.

I'm not sure, but probably `sha-1` is enough for validation in most cases, I see it uses for distribution download validation on infinispan.org as well as for maven also in addition to the `md5` @ryanemerson wdyt about it?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1026,6 +1033,64 @@ func PodEnv(i *infinispanv1.Infinispan, systemEnv *[]corev1.EnvVar) []corev1.Env
 	return envVars
 }
 
+func applyExternalArtifactsDownload(ispn *infinispanv1.Infinispan, spec *corev1.PodSpec) (updated bool) {
```
ryanemerson, line 1036: WDYT to splitting this logic into a separate `dependencies.go` file in the same package? `infinispan_controller.go` already has a lot of content and with https://github.com/infinispan/infinispan-operator/pull/1007 we will do the same for encryption.

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -525,11 +527,17 @@ func applyLabels(ispn *Infinispan, envvar, annotationName string) error {
...
 }
 
+
 // HasCustomLibraries true if custom libraries are defined
 func (ispn *Infinispan) HasCustomLibraries() bool {
 	return ispn.Spec.Dependencies != nil && ispn.Spec.Dependencies.VolumeClaimName != ""
 }
 
+// HasExternalArtifacts true if external artifacts are defined
+func (ispn *Infinispan) HasExternalArtifacts() bool {
```
ryanemerson, line 537: We should probably change `HasCustomLibraries` to `HasPvcArtifacts` or something similar.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1026,6 +1033,64 @@ func PodEnv(i *infinispanv1.Infinispan, systemEnv *[]corev1.EnvVar) []corev1.Env
...
+			*c = append(spec.InitContainers[:position], spec.InitContainers[position+1:]...)
+			*volumes = append(spec.Volumes[:volumePosition], spec.Volumes[volumePosition+1:]...)
+			*volumeMounts = append(spec.Containers[0].VolumeMounts[:volumeMountPosition], spec.Containers[0].VolumeMounts[volumeMountPosition+1:]...)
+			updated = true
+		}
+	}
+	return
+}
+
+func externalArtifactsExtractCommand(ispn *infinispanv1.Infinispan, mountPath string) string {
```
ryanemerson, line 1075: No need for `mountPath` to be a parameter as it's always `ExternalArtifactsMountPath`.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1026,6 +1033,64 @@ func PodEnv(i *infinispanv1.Infinispan, systemEnv *[]corev1.EnvVar) []corev1.Env
...
+	for i, artifact := range ispn.Spec.Dependencies.Artifacts {
+		fileName := strings.ToLower(path.Base(artifact.Url))
+		if artifact.Type == infinispanv1.ExternalArtifactTypeZip || (artifact.Type == "" && strings.HasSuffix(fileName, ExternalArtifactsZipExtension)) {
+			downloadFileName := fmt.Sprintf("file%d%s", i, ExternalArtifactsZipExtension)
+			extractCommands = extractCommands + fmt.Sprintf("curl --insecure -L %s -o %s && unzip -oq %s && rm %s;", artifact.Url, downloadFileName, downloadFileName, downloadFileName)
+			continue
+		}
+		if artifact.Type == infinispanv1.ExternalArtifactTypeTarGz || (artifact.Type == "" && strings.HasSuffix(fileName, ExternalArtifactsTarGzExtension)) {
+			downloadFileName := fmt.Sprintf("file%d%s", i, ExternalArtifactsTarGzExtension)
+			extractCommands = extractCommands + fmt.Sprintf("curl --insecure -L %s -o %s && tar xf %s && rm %s;", artifact.Url, downloadFileName, downloadFileName, downloadFileName)
```
ryanemerson, line 1086: We can just make this `extractCommands = fmt.Sprintf("%s curl ...` and pass `extractCommands` to `Sprintf`. There's also no need to specify downloadFileName three times, instead you can provide it once and use positional args in the string.

As the `.zip` & `.tar.gz` logic is the same except the extract command, I think we create a function to perform the common operations that simply takes the `artifact.Type`, extension and command as args.

dmvolod: Oh, yes, very good point. Thanks

pkg/apis/infinispan/v1/types_util.go
dmvolod: Yes, will rename it to the `HasDependenciesVolume`

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, good point. Moving it into the `dependencies.go` file

ryanemerson (approved):
@dmvolod Great work! A rebase is required, but the code LGTM.

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -419,6 +419,23 @@ spec:
             dependencies:
               description: External dependencies needed by the Infinispan cluster
               properties:
+                artifacts:
+                  items:
+                    properties:
+                      type:
+                        description: ExternalArtifactType external artifact file type
```
Crumby, line 426: Can we improve this description? This will appear in OCP console

deploy/operator-install.yaml
```diff
@@ -645,6 +645,23 @@ spec:
             dependencies:
               description: External dependencies needed by the Infinispan cluster
               properties:
+                artifacts:
+                  items:
+                    properties:
+                      type:
+                        description: ExternalArtifactType external artifact file type
```
Crumby, line 652: Same here

deploy/crds/infinispan.org_infinispans_crd.yaml
dmvolod: Yes, @Crumby 
Probably we need to brake golang comment convention and add more human readable notes to all our types in the CRD

pkg/apis/infinispan/v1/infinispan_types.go
ryanemerson: I don't have strong views on what sha mechanism should be used for verification (any thoughts @tristantarrant @Crumby), but I think we should have this available before merging as it could save a lot of trouble when debugging user cases.

tristantarrant: I think `sha1` is sufficient.
A couple of notes:
* Should we only allow `https` by default ? Maybe with a `allowInsecureUrls: true` option to override
* Do we support credentials ?
* It would be awesome if we also supported Maven GAVs:
` - maven: groupId:artifactId:version[:packaging[:classifier]]]`
This should default to Maven central https://repo1.maven.org/maven2/ with the possibility of overriding the repo

ryanemerson: > * This should default to Maven central https://repo1.maven.org/maven2/ with the possibility of overriding the repo

IMO per Infinispan CR overriding would be overkill, but we could allow a maven "settings.xml" to be provided as part of the global operator configuration.

dmvolod: We can probably add real maven support to the OCP Build (k8s Kaniko build) option and build a real image just one time
Also it's possible to download maven artifacts with the direct URL link. Credential can be also provided via URL (not secure way), but probably will be always insecure with Init containers utilization. The single secure way us using builds

dmvolod: I've added any hash algorithms to be available for hashing (sha1|sha224|sha256|sha384|sha512|md5)

deploy/crds/infinispan.org_infinispans_crd.yaml
dmvolod: Done, thanks

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -218,6 +218,32 @@ type InfinispanExternalDependencies struct {
...
+	// Url from which the file will be downloaded
+	Url string `json:"url"`
+	// +optional
+	// Type of file (archive) that will be downloaded.
+	// If not specified, the file type will be automatically determined by its extension or downloaded without unpacking
+	Type ExternalArtifactType `json:"type,omitempty"`
+	// +optional
+	// +kubebuilder:validation:Pattern=`^(sha1|sha224|sha256|sha384|sha512|md5):[a-z0-9]+`
+	// Hash sum of the downloaded file to validate
+	Hash string `json:"hash,omitempty"`
```
ryanemerson, line 246: I like the flexibility of this a lot :slightly_smiling_face:

ryanemerson (approved):
LGTM. I think we might have to hold off merging until we branch for the next major release though. I'm trying to clarify this at the moment.

rigazilla:
@dmvolod I've added a wrong sha256 for the `.jar` artifact in the `TestExternalDependenciesHttp` and even with a wrong sha the jar is loaded by Infinispan.

I've investigated a little, the whole command line is:
```
cd /opt/infinispan/server/lib/external-artifacts;
mkdir -p ./tmp && rm -rf ./tmp/* && cd ./tmp && curl --insecure -LO http://external-libs-web-server:8080/task01-1.0.0.jar && FILENAME=$(ls -1 . | head -n1) && cd .. && echo 1234567890d ./tmp/$FILENAME | sha256sum -c && mv ./tmp/$FILENAME .;
curl --insecure -L http://external-libs-web-server:8080/task02-1.0.0.zip -o file1.zip  && unzip -oq file1.zip && rm file1.zip;
```
I suppose what happens is that the whole command doesn't fail because the last single command returns 0 and the .jar is leaved in the tmp dir and then loaded by Infinispan

dmvolod:
> I suppose what happens is that the whole command doesn't fail because the last single command returns 0 and the .jar is leaved in the tmp dir and then loaded by Infinispan

Thanks, @rigazilla 
Will look into this issue. I tested wrong way manually as it's was working, but probably we need to add `set -e` at the start of the script and this relates to the set of the command not individual one.

dmvolod:
> Will look into this issue. I tested wrong way manually as it's was working, but probably we need to add `set -e` at the start of the script and this relates to the set of the command not individual one.

I remember why we need to don't brake init container in case of incorrect hash, otherwise we never start cluster and never fix this issue. In case of the incorrect hash we just blocking embedding downloaded file into the server code.

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -419,6 +419,28 @@ spec:
             dependencies:
               description: External dependencies needed by the Infinispan cluster
               properties:
+                artifacts:
+                  items:
+                    properties:
+                      hash:
+                        description: Hash sum of the downloaded file to validate
```
oraNod, line 426:
```suggestion
                        description: Checksum that you can use to verify downloaded files.
```

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -419,6 +419,28 @@ spec:
...
+                        pattern: ^(sha1|sha224|sha256|sha384|sha512|md5):[a-z0-9]+
+                        type: string
+                      type:
+                        description: Type of file (archive) that will be downloaded. If not specified, the file type will be automatically determined by its extension or downloaded without unpacking
```
oraNod, line 430:
```suggestion
                        description: File type for the archive that you want to download. If not specified, the file type is automatically determined from the extension.
```

oraNod: @dmvolod Sorry but what does "or downloaded without unpacking" mean here? If you specify the file type then the archive is unpacked after download / if you don't specify the file type then it is not unpacked

deploy/crds/infinispan.org_infinispans_crd.yaml
```diff
@@ -419,6 +419,28 @@ spec:
...
+                        enum:
+                          - file
+                          - zip
+                          - tgz
+                        type: string
+                      url:
+                        description: Url from which the file will be downloaded
```
oraNod, line 437:
```suggestion
                        description: URL where the archive is available for download.
```

deploy/operator-install.yaml
```diff
@@ -653,6 +653,28 @@ spec:
...
+                        description: Type of file (archive) that will be downloaded. If not specified, the file type will be automatically determined by its extension or downloaded without unpacking
+                        enum:
+                          - file
+                          - zip
+                          - tgz
+                        type: string
+                      url:
+                        description: Url from which the file will be downloaded
+                        pattern: ^(https?|ftp)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]
+                        type: string
```
oraNod, line 673: Same comments for `hash` `type` and `url` descriptions above. :wink:

oraNod (commented):
Thanks @dmvolod I guess I'm a little behind with reviews and the operator PR queue but I've made some comments. I've created a separate Jira and can update the docs for this async so we don't hold up code merge.

deploy/crds/infinispan.org_infinispans_crd.yaml
dmvolod: @oraNod it's possible to download files in thee ways: download any file type and place it into the folder, unpack archive with explicitly defined type (tar.gz o zip) and unpack archive based on the extension detection. In this case file is not archive only, it can be any file type customer wants to embedded to the server.

deploy/crds/infinispan.org_infinispans_crd.yaml
dmvolod: @oraNod please see my comment above. It's not only archive, can be any file to download. What is the correct message?

deploy/crds/infinispan.org_infinispans_crd.yaml
oraNod: @dmvolod Thanks! I updated the suggestion. That seems like a lot of info to fit into a single description without making things confusing. How about if we tackle the different ways to download files in docs?

dmvolod: Thanks, @oraNod . Agree, it's better to explain all actions in doc. This description just for common information.

documentation/asciidoc/topics/proc_copying_code.adoc
```diff
@@ -76,6 +76,18 @@ $ {oc_copy} --no-preserve=true libs datagrid-libs-pod:/tmp/
 ----
 $ {oc_delete_pod} datagrid-libs-pod
 ----
++
+Specify the persistent volume with `spec.dependencies.VolumeClaimName` in your `Infinispan` CR and then apply the changes.
```
dmvolod, line 80:
```suggestion
Specify the persistent volume with `spec.dependencies.volumeClaimName` in your `Infinispan` CR and then apply the changes.
```

documentation/asciidoc/topics/yaml/dependencies_download.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: infinispan.org/v1
+kind: Infinispan
+metadata:
+  name: {example_crd_name}
+spec:
+  replicas: 2
+  dependencies:
+    artifacts:
+      - url: ftp://<user>:<password>@example.com:21/path
```
dmvolod, line 9: @oraNod I think it's better to use http example

ryanemerson:
> I remember why we need to don't brake init container in case of incorrect hash, otherwise we never start cluster and never fix this issue. In case of the incorrect hash we just blocking embedding downloaded file into the server code.

IMO if an expected dependency is not available (download fails of sha validation fails) it's better that the cluster fails-fast and does not start without the required user dependencies. This will be much easier to debug than the cluster starting and the user having to investigate why `x` feature does not work as expected.

dmvolod:
> IMO if an expected dependency is not available (download fails of sha validation fails) it's better that the cluster fails-fast and does not start without the required user dependencies.

@ryanemerson it's not good approach for current (only after successful well formed state) StatefulSet update implementation. Cluster never starts in this case and never been updated. For debug purposes output download trace is placed inside init container log.

ryanemerson:
> @ryanemerson it's not good approach for current (after successful well formed state only) StatefulSet update implementation. Cluster never starts in this case and never been updated.

Maybe I'm miss-understanding, but if an update is triggered by adding a dependency and that dependency fails to download then I wouldn't expect the new cluster to start.

If a StatefulSet is updated due to a different spec change, then I still wouldn't expect the new cluster to start if all dependencies aren't available. That being said, we can make the code more robust against corrupt downloads, by retrying the download `x` number of times if the sha does not match.

dmvolod:
> Maybe I'm miss-understanding, but if an update is triggered by adding a dependency and that dependency fails to download then I wouldn't expect the new cluster to start.

Yes, correct. But in our current implementation cluster will never updated in case of you will fix `sha` or artifact download `url`. Infinispan CR need to be created from scratch every fix.

ryanemerson:
> Yes, correct. But in our current implementation cluster will never updated in case of you will fix `sha` or artifact download `url`. Infinispan CR need to be created from scratch every fix.

Is this an oversight? Shouldn't we use a HASH env like we do for other .spec elements to ensure the reconcilliation and redeploy happens on changes to the dependencies?

dmvolod:
> Is this an oversight? Shouldn't we use a HASH env like we do for other .spec elements to ensure the reconcilliation and redeploy happens on changes to the dependencies?

No, it's updating but too late after WellFormed validation. We should utilize CreateOrPatch for most of the StatefulSet fields (probably except `replicas` required for shutdown control) to be able update not well formed cluster.

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -249,6 +249,31 @@ type InfinispanExternalDependencies struct {
 	// Name of the persistent volume claim with custom libraries
 	// +optional
 	VolumeClaimName string `json:"volumeClaimName,omitempty"`
+	// +optional
+	Artifacts []InfinispanExternalArtifacts `json:"artifacts,omitempty"`
```
dmvolod, line 253: @ryanemerson , @rigazilla what do you think about to add a downloading type strategy (stop | ignore) and give a user to decide what to do in case of fails download in any case (hash validation or incorrect or unavailable location)

rigazilla: > @ryanemerson , @rigazilla what do you think about to add a downloading type strategy (stop | ignore) and give a user to decide what to do in case of fails download in any case (hash validation or incorrect or unavailable location)

IMO starting a cluster with a missing piece is not a common use case. I would keep things as they are and maybe document the fact that CR needs to be deleted and recreated, if usability can't be improved.

dmvolod: Good point, agree. In this case I would prefer to keep this PR on hold and implement #1076 before. I will to take about it.
Than, I will plan to thinking about OCP/Kaniko build integration.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -336,6 +334,18 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
+	// Recover Pods with updated init containers in case of fails
```
dmvolod, line 337: @rigazilla , @ryanemerson I've added specific recovery logic from the failed pods in case of error happens during downloading or incorrect hash sum.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -336,6 +334,18 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+	for _, pod := range podList.Items {
+		if !kube.IsInitContainersEqual(statefulSet.Spec.Template.Spec.InitContainers, pod.Spec.InitContainers) {
```
ryanemerson, line 339: Why do we check for container equality here? If this is to guard against upgrades, shouldn't we just check the conditions?

pkg/kubernetes/pods.go
```diff
@@ -63,6 +64,41 @@ func GetEnvVarIndex(envVarName string, env *[]corev1.EnvVar) int {
...
+			if !reflect.DeepEqual(srcInitContainer.Command, destContainer[dstInitContainerIdx].Command) ||
+				!reflect.DeepEqual(srcInitContainer.Args, destContainer[dstInitContainerIdx].Args) {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func IsInitContainerTerminated(initContainerStatuses []corev1.ContainerStatus) bool {
```
ryanemerson, line 84: I think `InitContainerFailed` is more descriptive as we're checking that a container has terminated and with a non-zero exit code.

pkg/kubernetes/pods.go
```diff
@@ -63,6 +64,41 @@ func GetEnvVarIndex(envVarName string, env *[]corev1.EnvVar) int {
...
+	for _, initContainerStatus := range initContainerStatuses {
+		if initContainerStatus.LastTerminationState.Terminated != nil && initContainerStatus.LastTerminationState.Terminated.ExitCode != 0 {
+			return true
+		}
+	}
+	return false
+}
+
+func FindInitContainer(initContainers []corev1.Container, containerName string) int {
```
ryanemerson, line 93: How about: `func ContainerIndex(containers []corev1.Container, name string) int {}`

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: Also, couldn't we just do `!reflect.DeepEqual(statefulSet.Spec.Template.Spec.InitContainers, pod.Spec.InitContainers)` instead of using `IsInitContainersEqual`?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -336,6 +334,18 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
...
+			if kube.IsInitContainerTerminated(pod.Status.InitContainerStatuses) {
+				if err = r.client.Delete(context.TODO(), &pod); err != nil {
+					return reconcile.Result{}, err
+				}
+				continue
+			}
+		}
+	}
```
ryanemerson, line 347: Having the init container retry logic in the operator with no upper bound will result in the pods being created/deleted repeatedly if it's not possible for a download to ever succeed (e.g. typo in the sha).

If we have the retry logic in the init container itself we can easily set an upper bound on the number of retries and we don't need any additional logic in the controller. If the download fails `x` times, then the initcontainer fails and the cluster can't be created.

pkg/kubernetes/pods.go
dmvolod: Yes, sure, will fix.

pkg/controller/infinispan/infinispan_controller.go
dmvolod: > Why do we check for container equality here? If this is to guard against upgrades, shouldn't we just check the conditions?

It's a special hack for recovery Pod in case of StatefulSet controller logic. If customer set incorrect `sha` or `url` and Pod becomes fails, StatefulSet controller never been updated this Pod and we should remove it and than controller will create a correct one from the StatefulSet configuration.

dmvolod: > Also, couldn't we just do `!reflect.DeepEqual(statefulSet.Spec.Template.Spec.InitContainers, pod.Spec.InitContainers)` instead of using `IsInitContainersEqual`?

It's a good point, thanks. In this case this probably should work

pkg/kubernetes/pods.go
dmvolod: Yes, good point. Will change.

pkg/controller/infinispan/infinispan_controller.go
dmvolod: > Having the init container retry logic in the operator with no upper bound will result in the pods being created/deleted repeatedly if it's not possible for a download to ever succeed (e.g. typo in the sha).

Pods will never recreated if Init container is fail. Not sure we could embed retry logic into the download artifact script. I think that retry logic managed by k8s controller much better that own one.

pkg/controller/infinispan/infinispan_controller.go
dmvolod: > It's a good point, thanks. In this case this probably should work

Doesn't work as expected as Pod controller embedding additional resources and structures becomes not match. Separate function need to be implemented for validation.

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: > Not sure we could embed retry logic into the download artifact script.

It's simple really, we just utilise a bash for loop e.g. `for i in 1 2 3 4 5; do command && break || sleep 1; done`

If `command` succeeds, then when don't retry as the exit code will be 0.

dmvolod: Thanks. Will try to embed. Script logic become complex and complex :)

dmvolod: Yes, good idea, but in this case we need to implement additional containers comparison logic. Lets try to use old approach.

dmvolod: Also, I would prefer to make retry count and interval configurable, but we need to have Configuration CR first. Many project who start to keep configuration in the ConfigMap are moving away from this method to CRD as this is much flexible way.

ryanemerson: It would be good to have the retry count+interval configurable, but I don't think we need to worry too much about that unless there is demand from users.

+1 to a InfinispanConfiguration CR. That was my preference when the ConfigMap was initially introduced. I think this should probably wait until we have settled on a plan for CRD versioning.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -336,6 +334,17 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		return reconcile.Result{}, err
 	}
 
+	// Recover Pods with updated init containers in case of fails
```
ryanemerson, line 337: Can't remove this now that the retry mechanism is part of the initContainer?

dmvolod: No, this is another type of the recovery. It protects from the incorrect `sha` sum entering, not from `http` host unavailability.

ryanemerson:
@dmvolod Please see [my branch](https://github.com/ryanemerson/infinispan-operator/tree/external_dep_templates) which reworks `dependencies.go` so that a go template is used to generate the init-container command. I have tested against `TestExternalDependenciesHttp` and it passes fine locally.

Let me know what you think about this approach. When working on this I found it very difficult to determine what commands should be executed, so hopefully this way it's a bit clearer (it does to me at least).

ryanemerson:
> I will take it and polish a bit, to store template as separate file and embedding it via go-rice for better maintaining.
> wdyt?

Yeah even better :slightly_smiling_face:

pkg/controller/infinispan/infinispan_controller.go
dmvolod: Yes, sure, @ryanemerson . The scenario is:
- create cluster with or without dependencies and correct forming
- add or change (add hash or url) dependencies
- starting cluster rolling upgrade
- init container fails and first Pod become invalid
- change somewhere in dependencies to fix previous errors
- this code kill the Pod, otherwise StatefulSet controller never rolling update filed Pods

ryanemerson: Shouldn't this code only be executed if a update is needed then? Currently this code will execute causing the following scenario:

- Create cluster with incorrect hash/url
- init-container fails
- Pods are deleted
- init-container fails again

dmvolod: > Shouldn't this code only be executed if a update is needed then? Currently this code will execute causing the following scenario

No, comparing init container in Pod and propagated to StatefulSet are protects from this scenario.

pkg/controller/infinispan/templates/dependencies.gotmpl
```diff
@@ -0,0 +1,36 @@
...
+    fi
+}
+cd {{ .MountPath }}
+{{- range $i, $artifact := .Artifacts }}
+    mkdir -p ./tmp && rm -rf ./tmp/* && cd ./tmp
+    retry "curl --insecure -LO {{ $artifact.Url }}"
+    FILENAME=$(ls -1 . | head -n1)
+    {{ hashCmd $artifact "$FILENAME" }}
+    cd .. && mv ./tmp/$FILENAME .
+    unpack {{ $artifact.Type }} $FILENAME
```
ryanemerson, line 35: I think we should do `rm -rf ./tmp` as the final step of the loop. That way the `./tmp` dir is deleted on the final loop (no currently the case.

So:

```
mkdir -p ./tmp
cd ./tmp
retry "curl --insecure -LO {{ $artifact.Url }}"
FILENAME=$(ls -1 . | head -n1)
{{ hashCmd $artifact "$FILENAME" }}
cd ..
mv ./tmp/$FILENAME .
unpack {{ $artifact.Type }} $FILENAME
rm -rf ./tmp
```

For readability I also think it's better if we don't chain the commands and have them as a invidivual lines now we're using a template.

pkg/apis/infinispan/v1/infinispan_types.go
```diff
@@ -249,6 +249,31 @@ type InfinispanExternalDependencies struct {
...
+	// +kubebuilder:validation:Pattern=`^(https?|ftp)://[-a-zA-Z0-9+&@#/%?=~_|!:,.;]*[-a-zA-Z0-9+&@#/%=~_|]`
+	// URL of the file you want to download.
+	Url string `json:"url"`
+	// +optional
+	// Specifies the type of file you want to download. If not specified, the file type is automatically determined from the extension.
+	Type ExternalArtifactType `json:"type,omitempty"`
+	// +optional
+	// +kubebuilder:validation:Pattern=`^(sha1|sha224|sha256|sha384|sha512|md5):[a-z0-9]+`
+	// Checksum that you can use to verify downloaded files.
+	Hash string `json:"hash,omitempty"`
```
ryanemerson, line 276: In the documentation we describe this as a "Checksum". Should we also name this field "Checksum" so it's intent is more explicit?  \\cc @oraNod

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: Thanks for explaining. I understand now.

pkg/controller/infinispan/templates/dependencies.gotmpl
dmvolod: Yes, thanks, will fix.

pkg/apis/infinispan/v1/infinispan_types.go
dmvolod: We could probably add a fix to the docs. `hash` is a filed, and checksum is an entity

ryanemerson:
Great work @dmvolod, thanks

PR #1008, opened 2021-04-21
sent log to the events system #346

rigazilla:
I've:
1. ported here the code for events from PR #880
2. tried to group code for events and log in one package

I'm not sure about the result of point 2

### DIFF
--- a/pkg/controller/batch/batch_controller.go
+++ b/pkg/controller/batch/batch_controller.go
@@ -178,7 +178,7 @@ func (r *batchResource) initializeResources() (reconcile.Result, error) {
 	spec := batch.Spec
 	// Ensure the Infinispan cluster exists
 	infinispan := &v1.Infinispan{}
-	if result, err := kube.LookupResource(spec.Cluster, batch.Namespace, infinispan, r); result != nil {
+	if result, err := kube.LookupResource(spec.Cluster, batch.Namespace, infinispan, r, &r.client); result != nil {
 		return *result, err
 	}
 
@@ -228,7 +228,7 @@ func (r *batchResource) initializeResources() (reconcile.Result, error) {
 func (r *batchResource) execute() (reconcile.Result, error) {
 	batch := r.instance
 	infinispan := &v1.Infinispan{}
-	if result, err := kube.LookupResource(batch.Spec.Cluster, batch.Namespace, infinispan, r); result != nil {
+	if result, err := kube.LookupResource(batch.Spec.Cluster, batch.Namespace, infinispan, r, &r.client); result != nil {
 		return *result, err
 	}
 
@@ -309,7 +309,7 @@ func (r *batchResource) execute() (reconcile.Result, error) {
 func (r *batchResource) waitToComplete() (reconcile.Result, error) {
 	batch := r.instance
 	job := &batchv1.Job{}
-	if result, err := kube.LookupResource(batch.Name, batch.Namespace, job, r); result != nil {
+	if result, err := kube.LookupResource(batch.Name, batch.Namespace, job, r, &r.client); result != nil {
 		return *result, err
 	}
 
--- a/pkg/controller/batch/batch_controller.go
+++ b/pkg/controller/batch/batch_controller.go
@@ -49,7 +49,7 @@ func Add(mgr manager.Manager) error {
 
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
 	kubernetes = kube.NewKubernetesFromController(mgr)
-	return &reconcileBatch{Client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &reconcileBatch{Client: mgr.GetClient(), scheme: mgr.GetScheme(), eventRecorder: mgr.GetEventRecorderFor(ControllerName)}
 }
 
 func add(mgr manager.Manager, r reconcile.Reconciler) error {
@@ -79,8 +79,9 @@ var _ reconcile.Reconciler = &reconcileBatch{}
 // reconcileBatch reconciles a Batch object
 type reconcileBatch struct {
 	client.Client
-	scheme *runtime.Scheme
-	logger logr.Logger
+	scheme        *runtime.Scheme
+	logger        logr.Logger
+	eventRecorder record.EventRecorder
 }
 
 type batchResource struct {
@@ -98,6 +99,10 @@ func (rec *reconcileBatch) Logger() *logr.Logger {
 
 func (rec *reconcileBatch) Name() string {
 	return ControllerName
+}
+
+func (rec *reconcileBatch) EventRecorder() *record.EventRecorder {
+	return &rec.eventRecorder
 }
 
 func (bat *batchResource) Logger() *logr.Logger {
@@ -135,11 +140,12 @@ func (r *reconcileBatch) Reconcile(request reconcile.Request) (reconcile.Result,
 	}
 
 	batch := &batchResource{
-		instance:    instance,
-		client:      r.Client,
-		scheme:      r.scheme,
-		requestName: request.NamespacedName,
-		logger:      r.logger,
+		instance:      instance,
+		client:        r.Client,
+		scheme:        r.scheme,
+		requestName:   request.NamespacedName,
+		logger:        r.logger,
+		eventRecorder: r.eventRecorder,
 	}
 
 	phase := instance.Status.Phase
--- a/pkg/controller/batch/batch_controller.go
+++ b/pkg/controller/batch/batch_controller.go
@@ -97,10 +97,6 @@ func (rec *reconcileBatch) Logger() *logr.Logger {
 	return &rec.logger
 }
 
-func (rec *reconcileBatch) Name() string {
-	return ControllerName
-}
-
 func (rec *reconcileBatch) EventRecorder() *record.EventRecorder {
 	return &rec.eventRecorder
 }
@@ -117,12 +113,8 @@ func (bat *batchResource) Client() *client.Client {
 	return &bat.client
 }
 
-func (bat *batchResource) Name() string {
-	return ControllerName
-}
-
 func (r *reconcileBatch) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := eventlog.ValuedLogger(r, "Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger := eventlog.ValuedLogger(r, ControllerName, "Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Batch")
 
 	// Fetch the Batch instance
--- a/pkg/controller/batch/batch_controller.go
+++ b/pkg/controller/batch/batch_controller.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/go-logr/logr"
 	v1 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v1"
 	v2 "github.com/infinispan/infinispan-operator/pkg/apis/infinispan/v2alpha1"
 	consts "github.com/infinispan/infinispan-operator/pkg/controller/constants"
+	"github.com/infinispan/infinispan-operator/pkg/controller/eventlog"
 	ispnCtrl "github.com/infinispan/infinispan-operator/pkg/controller/infinispan"
 	kube "github.com/infinispan/infinispan-operator/pkg/kubernetes"
 	batchv1 "k8s.io/api/batch/v1"
@@ -16,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -77,17 +80,44 @@ var _ reconcile.Reconciler = &reconcileBatch{}
 type reconcileBatch struct {
 	client.Client
 	scheme *runtime.Scheme
+	logger logr.Logger
 }
 
 type batchResource struct {
-	instance    *v2.Batch
-	client      client.Client
-	scheme      *runtime.Scheme
-	requestName types.NamespacedName
+	instance      *v2.Batch
+	client        client.Client
+	scheme        *runtime.Scheme
+	requestName   types.NamespacedName
+	logger        logr.Logger
+	eventRecorder record.EventRecorder
+}
+
+func (rec *reconcileBatch) Logger() *logr.Logger {
+	return &rec.logger
+}
+
+func (rec *reconcileBatch) Name() string {
+	return ControllerName
+}
+
+func (bat *batchResource) Logger() *logr.Logger {
+	return &bat.logger
+}
+
+func (bat *batchResource) EventRecorder() *record.EventRecorder {
+	return &bat.eventRecorder
+}
+
+func (bat *batchResource) Client() *client.Client {
+	return &bat.client
+}
+
+func (bat *batchResource) Name() string {
+	return ControllerName
 }
 
 func (r *reconcileBatch) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger := eventlog.ValuedLogger(r, "Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling Batch")
 
 	// Fetch the Batch instance
@@ -109,6 +139,7 @@ func (r *reconcileBatch) Reconcile(request reconcile.Request) (reconcile.Result,
 		client:      r.Client,
 		scheme:      r.scheme,
 		requestName: request.NamespacedName,
+		logger:      r.logger,
 	}
 
 	phase := instance.Status.Phase
@@ -147,7 +178,7 @@ func (r *batchResource) initializeResources() (reconcile.Result, error) {
 	spec := batch.Spec
 	// Ensure the Infinispan cluster exists
 	infinispan := &v1.Infinispan{}
-	if result, err := kube.LookupResource(spec.Cluster, batch.Namespace, infinispan, r.client, log); result != nil {
+	if result, err := kube.LookupResource(spec.Cluster, batch.Namespace, infinispan, r); result != nil {
 		return *result, err
 	}
 
@@ -197,7 +228,7 @@ func (r *batchResource) initializeResources() (reconcile.Result, error) {
 func (r *batchResource) execute() (reconcile.Result, error) {
 	batch := r.instance
 	infinispan := &v1.Infinispan{}
-	if result, err := kube.LookupResource(batch.Spec.Cluster, batch.Namespace, infinispan, r.client, log); result != nil {
+	if result, err := kube.LookupResource(batch.Spec.Cluster, batch.Namespace, infinispan, r); result != nil {
 		return *result, err
 	}
 
@@ -278,7 +309,7 @@ func (r *batchResource) execute() (reconcile.Result, error) {
 func (r *batchResource) waitToComplete() (reconcile.Result, error) {
 	batch := r.instance
 	job := &batchv1.Job{}
-	if result, err := kube.LookupResource(batch.Name, batch.Namespace, job, r.client, log); result != nil {
+	if result, err := kube.LookupResource(batch.Name, batch.Namespace, job, r); result != nil {
 		return *result, err
 	}
 
--- a/pkg/controller/eventlog/eventlog.go
+++ b/pkg/controller/eventlog/eventlog.go
@@ -44,7 +44,7 @@ func Logger(cwl ControllerWithLogger) logr.Logger {
 	return *cwl.Logger()
 }
 
-func LogAndSendEvent(cwe ControllerWithEvents, owner runtime.Object, message, reason string) {
+func LogAndSendEvent(cwe ControllerWithEvents, owner runtime.Object, reason, message string) {
 	(*cwe.Logger()).Info(message)
 	if *cwe.EventRecorder() != nil && owner != nil {
 		(*cwe.EventRecorder()).Event(owner, corev1.EventTypeWarning, reason, message)
--- a/pkg/controller/eventlog/eventlog.go
+++ b/pkg/controller/eventlog/eventlog.go
@@ -10,13 +10,9 @@ import (
 
 type ReasonType string
 
-type ControllerWithEvents interface {
-	EventRecorder() *record.EventRecorder
-	Logger() *logr.Logger
-}
-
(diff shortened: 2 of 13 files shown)


pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -152,12 +157,32 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 
 var _ reconcile.Reconciler = &ReconcileInfinispan{}
 
+var log = logf.Log.WithName("controller_infinispan")
```
dmvolod, line 160: Seems to be utilize ControllerName for logger also

pkg/kubernetes/controllerutil.go
```diff
@@ -152,7 +153,7 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj runtime.Object)
 }
 
 // LookupResource lookup for resource to be created by separate resource controller
-func LookupResource(name, namespace string, resource runtime.Object, client client.Client, logger logr.Logger) (*reconcile.Result, error) {
+func LookupResource1(name, namespace string, resource runtime.Object, client client.Client, logger logr.Logger) (*reconcile.Result, error) {
```
dmvolod, line 156: Do we need function with number in name, i.e. `LookupResource1`

pkg/controller/logandevent/logandevent.go
```diff
@@ -0,0 +1,42 @@
+package logandevent
```
dmvolod, line 1: Probably we can skip `and` in the middle for better read and call package as `eventlog` or anything else

dmvolod:
@rigazilla overall LGTM, probably less structured, but should work
The rights is also missing

pkg/controller/logandevent/logandevent.go
rigazilla: fine with `eventlog`

pkg/kubernetes/controllerutil.go
rigazilla: no, this is non more needed

rigazilla:
added rights for events to role. thanks @dmvolod 

I'm wondering if we can go down this road and group some common behaviours for controller initialization the same way

pkg/kubernetes/controllerutil.go
```diff
@@ -151,12 +151,12 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj runtime.Object)
 	return nil
 }
 
-// LookupResource lookup for resource to be created by separate resource controller
-func LookupResource(name, namespace string, resource runtime.Object, client client.Client, logger logr.Logger) (*reconcile.Result, error) {
-	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: name}, resource)
+// LookupResourceForController lookup for resource to be created by separate resource controller
+func LookupResource(name, namespace string, resource runtime.Object, cwe eventlog.ControllerWithEvents) (*reconcile.Result, error) {
```
dmvolod, line 155: I think we should have an option, to skip sending events for some cases

rigazilla: I've added an optional bool arg so user can disable event in func call

pkg/kubernetes/controllerutil.go
```diff
@@ -151,12 +151,21 @@ func mutate(f controllerutil.MutateFn, key client.ObjectKey, obj runtime.Object)
 	return nil
 }
 
-// LookupResource lookup for resource to be created by separate resource controller
-func LookupResource(name, namespace string, resource runtime.Object, client client.Client, logger logr.Logger) (*reconcile.Result, error) {
-	err := client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: name}, resource)
+// LookupResourceForController lookup for resource to be created by separate resource controller
+func LookupResource(name, namespace string, resource runtime.Object, cwe eventlog.ControllerWithEvents, sendNotFoundEvent ...bool) (*reconcile.Result, error) {
```
dmvolod, line 155: @rigazilla `sendNotFoundEvent` a bit confusing me, WDYT about `skipSendEvent`
Probably we can define  `cwe eventlog.ControllerWithEvents` as pointer and analyse it for the `nil` or not

rigazilla: > @rigazilla `sendNotFoundEvent` a bit confusing me, WDYT about `skipSendEvent`

yeah maybe better  

> Probably we can define `cwe eventlog.ControllerWithEvents` as pointer and analyse it for the `nil` or not

this doesn't allow a 'per call' choice, only at controller level

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -47,19 +49,22 @@ import (
...
+	DataMountPath                    = "/opt/infinispan/server/data"
+	DataMountVolume                  = "data-volume"
+	CustomLibrariesMountPath         = "/opt/infinispan/server/lib/custom-libraries"
+	CustomLibrariesVolumeName        = "custom-libraries"
+	EncryptMountPath                 = "/etc/encrypt"
+	ConfigVolumeName                 = "config-volume"
+	EncryptVolumeName                = "encrypt-volume"
+	IdentitiesVolumeName             = "identities-volume"
+	AdminIdentitiesVolumeName        = "admin-identities-volume"
+	EventReasonPrelimChecksFailed    = "PrelimChecksFailed"
```
ryanemerson, line 61: Never used.

ryanemerson (commented):
@rigazilla The "github.com/infinispan/infinispan-operator/pkg/controller/eventlog" package is missing from the commits.

rigazilla:
rebased and addressed issues

Crumby:
`"system:serviceaccount:namespace-for-testing:infinispan-operator" cannot create resource "events" in API group "" in the namespace "default"` Look like Operator is trying to do something illegal

dmvolod:
@rigazilla could you please to add sending events and logs for the following messages in the code
https://github.com/infinispan/infinispan-operator/blob/0f8825d8f8b51b49bb9967ece54cbd01333029d0/pkg/controller/infinispan/infinispan_controller.go#L417
 https://github.com/infinispan/infinispan-operator/blob/0f8825d8f8b51b49bb9967ece54cbd01333029d0/pkg/controller/infinispan/resources/config/xsite.go#L177

dmvolod:
@rigazilla I saw following error in the cross-site log
```
E0521 13:10:11.440361       1 event.go:237] Server rejected event '&v1.Event{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, ObjectMeta:v1.ObjectMeta{Name:"test-cross-site-view-kubernetes-node-port-xsite1-configuration.168116fbc4b3c553", GenerateName:"", Namespace:"default", SelfLink:"", UID:"", ResourceVersion:"", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ClusterName:"", ManagedFields:[]v1.ManagedFieldsEntry(nil)}, InvolvedObject:v1.ObjectReference{Kind:"ConfigMap", Namespace:"", Name:"test-cross-site-view-kubernetes-node-port-xsite1-configuration", UID:"", APIVersion:"v1", ResourceVersion:"", FieldPath:""}, Reason:"ResourceNotReady", Message:"ConfigMap resource 'test-cross-site-view-kubernetes-node-port-xsite1-configuration' not ready", Source:v1.EventSource{Component:"infinispan-controller", Host:""}, FirstTimestamp:v1.Time{Time:time.Time{wall:0xc0220a2cd9338753, ext:150097839017, loc:(*time.Location)(0x24aa8a0)}}, LastTimestamp:v1.Time{Time:time.Time{wall:0xc0220a2cda17f573, ext:150112809409, loc:(*time.Location)(0x24aa8a0)}}, Count:3, Type:"Warning", EventTime:v1.MicroTime{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, Series:(*v1.EventSeries)(nil), Action:"", Related:(*v1.ObjectReference)(nil), ReportingController:"", ReportingInstance:""}': 'events "test-cross-site-view-kubernetes-node-port-xsite1-configuration.168116fbc4b3c553" is forbidden: User "system:serviceaccount:namespace-for-testing-xsite1:infinispan-operator" cannot patch resource "events" in API group "" in the namespace "default"' (will not retry!)
```

dmvolod (approved):

pkg/controller/eventlog/eventlog.go
```diff
@@ -0,0 +1,52 @@
...
+
+func Logger(cwl ControllerWithLogger) logr.Logger {
+	if *cwl.Logger() != nil {
+		return *cwl.Logger()
+	}
+	*cwl.Logger() = logf.Log.WithName(cwl.Name())
+	return *cwl.Logger()
+}
+
+func LogAndSendEvent(cwe ControllerWithEvents, owner runtime.Object, reason, message string) {
```
dmvolod, line 47: @rigazilla what do you think to introduce ReasonType to avoid misprints between reason and message positions?

rigazilla: @dmvolod sounds good

rigazilla: @dmvolod,  mmm my idea was to add something like
```
type ReasonType string
```
but it doesn't enforce that much cause string is implicitly converted to ReasonType.
Any better easy idea? Or I would leave things like this

dmvolod: @rigazilla you should define const as `ReasonType` and seems to you don't need to convert except last call to the event API.
Probably I'm wrong, but should work

rigazilla: actually it works, but it doesn't enforce anything on the content of the reason parameter, since any  free string is implicitly converted to ReasonType without any warning.
I've just realized that also the k8s.Event uses string for both reason and message args, I would do same

pkg/controller/eventlog/eventlog.go
```diff
@@ -0,0 +1,52 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type ControllerWithEvents interface {
+	EventRecorder() *record.EventRecorder
+	Logger() *logr.Logger
+	Client() *client.Client
```
ryanemerson, line 15: I don't think this belongs as part of this interface. It's only use is in `controllerutil.LookupResource`, so I think it would be better to just pass `client client.Client` and `ControllerWithEvents` separately.

pkg/controller/eventlog/eventlog.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+type ControllerWithLogger interface {
```
ryanemerson, line 18: Is ther any advantage in having `ControllerWithLogger` and `ControllerWithEvents` as separate interfaces? It seems like all controller satisfy both interfaces. How about having a single interface `EventLogger`:

```go
type EventLogger interface {
    EventRecorder() *record.EventRecorder
    Logger() *logr.Logger
}
```

pkg/controller/eventlog/eventlog.go
```diff
@@ -0,0 +1,52 @@
...
+	Name() string
+	Logger() *logr.Logger
+}
+
+type ErrorEvent struct {
+	E      error
+	Reason string
+	Owner  *runtime.Object
```
ryanemerson, line 26: `Owner` is never used.

ryanemerson: I'm also not sure what the advantage of having this struct is?

pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -79,14 +98,18 @@ func (c *configResource) Process() (reconcile.Result, error) {
...
-		if result, err := kube.LookupResource(c.infinispan.GetSiteServiceName(), c.infinispan.Namespace, siteService, c.client, c.log); result != nil {
+		if result, err := kube.LookupResource(c.infinispan.GetSiteServiceName(), c.infinispan.Namespace, siteService, c); result != nil {
 			return *result, err
 		}
 
 		var err error
 		xsite, err = ComputeXSite(c.infinispan, c.kube, siteService, c.log)
 		if err != nil {
-			c.log.Error(err, "Error in computeXSite configuration")
+			if errEv, ok := err.(*eventlog.ErrorEvent); ok {
```
ryanemerson, line 108: This doesn't seem like the right way to do things to me. If `xsite.go#appendKubernetesRemoteLocation` is supposed to return just an `error`, then we should just use `fmt.Errorf("%s: %w", reason, err)` in order to return a `error` directly from `appendKubernetesRemoteLocation`.

pkg/controller/eventlog/eventlog.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+type ControllerWithEvents interface {
+	EventRecorder() *record.EventRecorder
+	Logger() *logr.Logger
+	Client() *client.Client
+}
+
+type ControllerWithLogger interface {
+	Name() string
```
ryanemerson, line 19: Requiring all controllers implement a `Name()` function seems very verbose just to avoid passing a `name` parameter in to `ValuedLogger` function.

How about just having:

```go
func ValuedLogger(name string, cwl ControllerWithLogger, keysAndValues ...interface{}) logr.Logger {
...
}
```

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -167,7 +193,7 @@ type ReconcileInfinispan struct {
 // The Controller will requeue the Request to be processed again if the returned error is non-nil or
 // Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
 func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger := eventlog.ValuedLogger(r, "Request.Namespace", request.Namespace, "Request.Name", request.Name)
```
ryanemerson, line 196: I think we can simplify things by having the `eventlog.go` simply be:

```go
type EventLogger interface {
	Logger() *logr.Logger
	EventRecorder() *record.EventRecorder
}

func LogAndSendEvent(el EventLogger, owner runtime.Object, reason, message string) {
	(*el.Logger()).Info(message)
	if *el.EventRecorder() != nil && owner != nil {
		(*el.EventRecorder()).Event(owner, corev1.EventTypeWarning, reason, message)
	}
}
```

In the controllers, we then always create the log instance explicitly (less indirection), i.e.:

```go
r.log = logf.Log.WithName(ControllerName).WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
```

ryanemerson (changes requested):

pkg/controller/eventlog/eventlog.go
rigazilla: sounds good. done

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: this is to allow events when errors occurs [here](https://github.com/infinispan/infinispan-operator/pull/1008#issuecomment-845870043)
`ComputeXSite` doesn't accept a controller as argument so event can't be generated inside, also is used by the testsuite where there's no controller at all. So the idea was to minimize changes and generate events basing on the type of the returned error.

`error` is actually an interface and imo it's ok to return an object that implements it: I would say that `xsite.go#appendKubernetesRemoteLocation` returns "at least" an `error`

pkg/controller/eventlog/eventlog.go
rigazilla: mmm, I don't remember why I did that... need to review

pkg/controller/infinispan/resources/config/config_controller.go
ryanemerson: > `error` is actually an interface and imo it's ok to return an object that implements it: I would say that `xsite.go#appendKubernetesRemoteLocation` returns "at least" an `error`

That's a fair point. I didn't realise that was what was happening.



> `ComputeXSite` doesn't accept a controller as argument so event can't be generated inside, also is used by the testsuite where there's no controller at all. So the idea was to minimize changes and generate events basing on the type of the returned error.

As we're already modifying the behaviour of `ComputeXsite` and `appendKubernetesRemoteLocation`, can't we just make it so that `ComputeXsite` takes the `EventLogger` interface and logs the event there? We can avoid using a specific `error` implementation and  just do the following in `config_controller.go`:

```go
if err != nil {
   reconcile.Result{RequeueAfter: consts.DefaultWaitOnCreateResource}, nil
}
```

pkg/controller/eventlog/eventlog.go
rigazilla: seems that `batchController` doesn't implement the event part

ryanemerson: IMO it's better to add the EventRecorder to the batch_controller now, as I'm sure that utilising events for batch could be useful, instead of splitting into two interfaces for this one case.

rigazilla: fixed

pkg/controller/eventlog/eventlog.go
rigazilla: done

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: @ryanemerson yep, it can be done, though the changes spread a little because there are some functions call between `ComputeXsite` and `appendKube...`. Also we need a `fake` EventLogger for the testsuite.
I don't have preference anyway I can change the implementation like that

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -48,19 +50,23 @@ import (
...
+	DataMountPath                    = "/opt/infinispan/server/data"
+	DataMountVolume                  = "data-volume"
+	CustomLibrariesMountPath         = "/opt/infinispan/server/lib/custom-libraries"
+	CustomLibrariesVolumeName        = "custom-libraries"
+	EncryptMountPath                 = "/etc/encrypt"
+	ConfigVolumeName                 = "config-volume"
+	EncryptVolumeName                = "encrypt-volume"
+	IdentitiesVolumeName             = "identities-volume"
+	AdminIdentitiesVolumeName        = "admin-identities-volume"
+	EventReasonPrelimChecksFailed    = eventlog.ReasonType("PrelimChecksFailed")
```
dmvolod, line 62: probably following syntax a bit clear
`EventReasonPrelimChecksFailed eventlog.ReasonType = "PrelimChecksFailed"`

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -866,12 +891,13 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 		}
 
 		if m.IsDataGrid() && m.StorageSize() != "" {
-			var pvErr error
-			pvSize, pvErr = resource.ParseQuantity(m.StorageSize())
+			pvSize, pvErr := resource.ParseQuantity(m.StorageSize())
 			if pvErr != nil {
+				eventlog.LogAndSendEvent(r, m, EventReasonParseValueProblem, err.Error())
```
dmvolod, line 896:
```suggestion
				eventlog.LogAndSendEvent(r, m, EventReasonParseValueProblem, pvErr.Error())
```

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -866,12 +891,13 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
...
 				return nil, pvErr
 			}
 			if pvSize.Cmp(memory) < 0 {
+				eventlog.LogAndSendEvent(r, m, EventReasonLowPersistenceStorage, err.Error())
```
dmvolod, line 900: Seems to message should be different than `err.Error()`

pkg/controller/infinispan/infinispan_controller.go
rigazilla: I'll remove the `ReasonType` for now

pkg/controller/eventlog/eventlog.go
rigazilla: removed

pkg/controller/infinispan/resources/config/config_controller.go
rigazilla: changed implementation

pkg/controller/infinispan/infinispan_controller.go
rigazilla: the initial idea was to keep in a common place all the initialization stuff for logs and events, honestly I'm this PR is getting old and confused: better to stay simple and limit changes to only the send event feature

ryanemerson (commented):
Seems I forgot to submit my review yesterday.

pkg/controller/eventlog/eventlog.go
```diff
@@ -0,0 +1,29 @@
...
+	"k8s.io/client-go/tools/record"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+type EventLogger interface {
+	EventRecorder() *record.EventRecorder
+	Logger() *logr.Logger
+}
+
+func Logger(el EventLogger, name string) logr.Logger {
```
ryanemerson, line 16: No longer required as it's never called.

pkg/controller/batch/batch_controller.go
```diff
@@ -76,18 +79,42 @@ var _ reconcile.Reconciler = &reconcileBatch{}
...
+	eventRecorder record.EventRecorder
+}
+
+func (rec *reconcileBatch) Logger() *logr.Logger {
+	return &rec.logger
+}
+
+func (rec *reconcileBatch) EventRecorder() *record.EventRecorder {
+	return &rec.eventRecorder
+}
```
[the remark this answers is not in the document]
rigazilla: LookupResource now needs to send events and it's used in batch:
https://github.com/infinispan/infinispan-operator/blob/50f56e50ef3e2c19c0197cf6113776f7bfc4b97a/pkg/controller/batch/batch_controller.go#L150

https://github.com/infinispan/infinispan-operator/blob/50f56e50ef3e2c19c0197cf6113776f7bfc4b97a/pkg/controller/batch/batch_controller.go#L200

rigazilla:
@ryanemerson I would propose to close this PR and open a new one that just sends events when required

pkg/controller/batch/batch_controller.go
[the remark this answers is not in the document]
ryanemerson: In both cases they are using `batchResource` not `reconcileBatch`.

ryanemerson:
> @ryanemerson I would propose to close this PR and open a new one that just sends events when required

Sure. It seems to me that the only time having the interface is really useful is when calling `lookupresource`, but we can easily just pass a log and eventrecorder as two params here.

PR #1009, opened 2021-04-21
[test-integration] ServiceMonitor is now created ootb

Crumby:
Removes ServiceMonitor and related Secret creation as those resources are now created out of the box by the Operator

### DIFF
--- a/test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
+++ b/test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
@@ -3,16 +3,11 @@ package org.infinispan;
 import java.io.File;
 import java.io.IOException;
 import java.util.Base64;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BooleanSupplier;
 
-import io.fabric8.kubernetes.api.model.LabelSelectorBuilder;
-import io.fabric8.kubernetes.api.model.Secret;
-import io.fabric8.kubernetes.api.model.SecretBuilder;
-import io.fabric8.openshift.api.model.monitoring.v1.*;
 import org.infinispan.cr.InfinispanObject;
 import org.infinispan.cr.status.Condition;
 import org.infinispan.crd.InfinispanContextProvider;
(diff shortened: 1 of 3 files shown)


test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
```diff
@@ -102,59 +97,4 @@ private Credentials getCredentials(String secretName, String username) throws IO
       Identities identities = new ObjectMapper(new YAMLFactory()).readValue(identitiesYaml, Identities.class);
       return identities.getCredentials(username);
    }
-
-   public void createServiceMonitor() throws IOException {
```
dmvolod, line 106: @Crumby can we add `ServiceMonitor` validation to this code?

Crumby: @dmvolod Can you elaborate more on `ServiceMonitor` validation? `MonitoringStackIT` already validates that endpoints are scraped by the Prometheus so there is that validation. I should probably add `ServiceMonitor` to cleanup validator though.

On the identiteis. If you mean the code above, then that code is related user secrets

dmvolod: > Can you elaborate more on `ServiceMonitor` validation? `MonitoringStackIT` already validates that endpoints are scraped by the Prometheus so there is that validation. I should probably add `ServiceMonitor` to cleanup validator though.
> 
Oh, I missed that `MonitoringStackIT` validates `ServiceMonitor`.

> On the identities. If you mean the code above, then that code is related user secrets

Yes, I see that you got username and password from `identities.yaml` but now Secret has `username` and `password` separate fields and not need to parse yaml file anymore

dmvolod (approved):

test-integration/operator-tests/src/main/java/org/infinispan/Infinispan.java
Crumby: @dmvolod But that applies only for the Operator credentials, right? The tests themselves are working with user credentials which are still only provided via `identities.yaml`

dmvolod: Yeah, @Crumby right, you are testing from the customer point of view.

dmvolod:
Thanks, @Crumby 
Merged into master

PR #1011, opened 2021-04-22
[test-integration] Custom libs test

Crumby:
Tests `Custom libs` feature of the Infinispan CR

### DIFF
--- a/test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
+++ b/test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
@@ -1,0 +1,110 @@
+package org.infinispan.operator;
+
+import cz.xtf.core.config.XTFConfig;
+import cz.xtf.core.openshift.OpenShift;
+import cz.xtf.core.openshift.OpenShiftWaiters;
+import cz.xtf.core.openshift.OpenShifts;
+import cz.xtf.junit5.annotations.CleanBeforeAll;
+import io.fabric8.kubernetes.api.model.*;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.maven.it.VerificationException;
+import org.apache.maven.it.Verifier;
+import org.assertj.core.api.Assertions;
+import org.infinispan.Infinispan;
+import org.infinispan.Infinispans;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Paths;
+import java.util.Collections;
+
+@Slf4j
+@CleanBeforeAll
+public class CustomLibsIT {
+    private static final OpenShift openShift = OpenShifts.master();
+    private static final OpenShiftWaiters openShiftWaiters = OpenShiftWaiters.get(openShift, () -> false);
+    private static final Infinispan infinispan = Infinispans.customLibs();
+
+    @BeforeAll
+    static void prepare() throws IOException {
+        preparePVC();
+        preparePod();
+
+        openShiftWaiters.areExactlyNPodsRunning(1, "app", "infinispan-libs").waitFor();
+
+        buildLibs();
+        uploadLibs();
+
+        infinispan.deploy();
+        infinispan.waitFor();
+    }
+
+    @Test
+    void test() {
+        Pod pod = openShift.getPod("custom-libs-0");
+        String logs = openShift.getPodLog(pod);
+
+        Assertions.assertThat(logs).contains("Loaded extension 'basic-filter-factory'");
+    }
+
+    private static void preparePVC() {
+        PersistentVolumeClaimBuilder pvc = new PersistentVolumeClaimBuilder();
+        pvc.withNewMetadata().withName("infinispan-libs").endMetadata();
+        pvc.withNewSpec()
+                .withAccessModes("ReadWriteOnce")
+                .withNewResources().withRequests(Collections.singletonMap("storage", new Quantity("100Mi")))
+                .endResources().endSpec();
+
+        openShift.createPersistentVolumeClaim(pvc.build());
+    }
+
+    private static void preparePod() {
+        String serverImage = XTFConfig.get("org.infinispan.image.server", "quay.io/infinispan/server:latest");
+
+        VolumeBuilder vb = new VolumeBuilder();
+        vb.withName("lib-pv-storage").withNewPersistentVolumeClaim("infinispan-libs", false);
+
+        VolumeMountBuilder vmb = new VolumeMountBuilder();
+        vmb.withMountPath("/tmp/libs");
+        vmb.withName("lib-pv-storage");
+
+        ContainerBuilder cb = new ContainerBuilder();
+        cb.withName("lib-pv-container")
+                .withImage(serverImage)
+                .withVolumeMounts(vmb.build());
+
+        PodBuilder pod = new PodBuilder();
+        pod.withNewMetadata()
+                .withName("infinispan-libs")
+                .withLabels(Collections.singletonMap("app", "infinispan-libs"))
+                .endMetadata();
+        pod.withNewSpec()
+                .withNewSecurityContext().withFsGroup(2000L).endSecurityContext()
+                .withVolumes(vb.build())
+                .withContainers(cb.build())
+                .endSpec();
+
+        openShift.createPod(pod.build());
+    }
+
+    private static void buildLibs() {
+        try {
+            String projectPath = Paths.get("src/test/resources/libs/custom-filter").toAbsolutePath().toString();
+            Verifier maven = new Verifier(projectPath);
+
+            maven.setAutoclean(true);
+            maven.executeGoals(Collections.singletonList("package"));
+            maven.resetStreams();
+        } catch (VerificationException e) {
+            throw new IllegalStateException("Failed to build test project locally", e);
+        }
+    }
+
+    private static void uploadLibs() {
+        File file = new File("src/test/resources/libs/custom-filter/target/custom-filter-1.0.jar");
+
+        openShift.pods().withName("infinispan-libs").file("/tmp/libs/custom-filter-1.0.jar").upload(file.toPath());
+    }
+}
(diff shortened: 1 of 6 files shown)


test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
```diff
@@ -0,0 +1,107 @@
...
+        pvc.withNewSpec()
+                .withAccessModes("ReadWriteOnce")
+                .withNewResources().withRequests(Collections.singletonMap("storage", new Quantity("100Mi")))
+                .endResources().endSpec();
+
+        openShift.createPersistentVolumeClaim(pvc.build());
+    }
+
+    private static void preparePod() {
+        String serverImage = XTFConfig.get("org.infinispan.image.server", "quay.io/infinispan/server:latest");
```
dmvolod, line 64: @Crumby do you really need to boot Pod with Infinispan sever for file exchange? I think `busybox` will be enough.

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
```diff
@@ -0,0 +1,107 @@
...
+        VolumeMountBuilder vmb = new VolumeMountBuilder();
+        vmb.withMountPath("/tmp/libs");
+        vmb.withName("lib-pv-storage");
+
+        ContainerBuilder cb = new ContainerBuilder();
+        cb.withName("lib-pv-container")
+                .withImage(serverImage)
+                .withVolumeMounts(vmb.build());
+
+        PodBuilder pod = new PodBuilder();
```
dmvolod, line 78: I tried to run this code on my OCP but it hangs with
```
[INFO] Running org.infinispan.operator.CustomLibsIT
[19:17:04.958] INFO - Waiting up to 20 seconds. Reason: Cleaning project.
[19:17:07.096] INFO - Waiting up to 3 minutes. Reason: Waiting for exactly 1 pods to be running.
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 215.183 s <<< FAILURE! - in org.infinispan.operator.CustomLibsIT
[ERROR] org.infinispan.operator.CustomLibsIT  Time elapsed: 215.181 s  <<< ERROR!
cz.xtf.core.waiting.WaiterException
```
however Pod was created successfully
```
✔ /u01/git/infinispan/infinispan-operator/test-integration [pr-1011 L|✚ 3] 
19:20 $ oc get pods
NAME              READY   STATUS    RESTARTS   AGE
infinispan-libs   1/1     Running   0          8m36s
✔ /u01/git/infinispan/infinispan-operator/test-integration [pr-1011 L|✚ 3]
```

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
Crumby: Probably not, I just folled up what is documented. I was trying `ubi8` image but I had issues with that one and even if I suceded then ubi8 does not contain tar so `oc cp ...` would work anyway. `busybox` may be better.

test-integration/operator-tests/src/test/resources/libs/custom-filter/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.redhat.xpaas.test.jdg</groupId>
+    <artifactId>custom-filter</artifactId>
+    <version>1.0</version>
+    <packaging>jar</packaging>
+    <name>Client Authentication Mapper</name>
+
+    <properties>
+        <version.org.infinispan>12.1.0.Final</version.org.infinispan>
```
dmvolod, line 13: We should update Infinispan version in all others IT pom.xml files

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
```diff
@@ -0,0 +1,107 @@
...
+        pod.withNewMetadata().withName("infinispan-libs").endMetadata();
+        pod.withNewSpec()
+                .withNewSecurityContext().withFsGroup(2000L).endSecurityContext()
+                .withVolumes(vb.build())
+                .withContainers(cb.build())
+                .endSpec();
+
+        openShift.createPod(pod.build());
```
dmvolod, line 86: Also test (probably others) didn't cleanup resources in case of fails. It would be nice to have this option.

dmvolod:
Also would be nice to add `-Dxtf.openshift.admin.kubeconfig` to README. `xtf-cz/xtf` excellent project and supports kubeconfig location (would be better to analyze KUBECONFIG env var also) , as I logging-in once and newer care about API URL and token values

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
Crumby: Each test contains `@CleanBeforeAll` which is responsible for cleaning up namespace resources before anything in the test is called. But it's true that tests do no cleanup after themselves beside removing Inifnispan CR to validate that everything was removed properly.

Shouldn't be an issue to add similar optional functionality for after the execution

dmvolod: `@CleanBeforeAll` seem to very danger method which can be cleanup all resources in namespace

Crumby: That's intentional design to keep namespaces clean for running the tests and the tests only. It should also prevent any unintentional leftovers by any other tests in the namespace

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
Crumby: I haven't tried it yet. The design is that Operator runs in different one as it's technically a static resource. You can try adding labels `xtf.cz/keep` to all the Operator resources so they are not wiped out. But if you are going to run the Operator in the same namespace the waiter expecting only one pod running in the namespace will fail and would have to be updated to be more specif

[1] https://github.com/xtf-cz/xtf/blob/master/core/src/main/java/cz/xtf/core/openshift/OpenShift.java#L95

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
dmvolod: Yes, true, but limits to have a separate namespace for operator deployment. Not able to test in simple way. This design could  have but a bit limiting with short tests.

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
dmvolod: Oh, good, thanks @Crumby , will try to execute and probably add this options to local operator deployment script

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
Crumby: I agree. I would actually like to make the Operator installation part of the test execution altogether with different install modes testing but that's out of scope for now as there are higher priority tasks

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final OpenShift openShift = OpenShifts.master();
+    private static final OpenShiftWaiters openShiftWaiters = OpenShiftWaiters.get(openShift, () -> false);
+    private static final Infinispan infinispan = Infinispans.customLibs();
+
+    @BeforeAll
+    static void prepare() throws IOException {
+        preparePVC();
+        preparePod();
+
+        openShiftWaiters.areExactlyNPodsRunning(1).waitFor();
```
dmvolod, line 35: @Crumby I can't to get it works in the same namespace with operator deployment without your help.
This code just wait for a single Pod (in this case this is operator deployment Pod) without any name or label identification. Could you please to add labels while Pod creation and validate it there by this value.

Also `xtf.cz/keep: "true"` labels validation a bit aggressive, as I expected that nested resources which `ownerReference` has this label defined should not be deleted, but it's not match. Probably you can fix this issue in the `xtf-cz/xtf` project.

Crumby: Waiter should check only the libs loader pod now.

> I expected that nested resources which ownerReference has this label defined should not be deleted, but it's not match

Right, this would be non-trivial though. Since most of the resources are now managed by Operator and tests by themselves are creating only few of them then let the tests manage the resources might be simpler

dmvolod: Thanks, @Crumby . Test working now after your fix and #1012

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
Crumby: @dmvolod I tried `busybox` but it suffers from the same issue as `ubi8`: endless restart. It could be fixed by adding some manual sleep cmd to the pod but we need to consider that this tests are going to be executed on different architectures. And busybox is not available for s390x or ppc.

dmvolod: Oh, yeah, @Crumby good point, let's keep as is. Probably we can play with entry-point to avoid whole Infinispan server boot on startup

dmvolod (approved):

test-integration/operator-tests/src/test/java/org/infinispan/operator/CustomLibsIT.java
dmvolod: One more question, @Crumby do you know, why `busybox` is not available on `s390x`. Docker hub says it's present for that archs.

Crumby: @dmvolod You are correct, it seems I've been looking at wrong page.

dmvolod:
Merged into master
Thanks, @Crumby

PR #1016, opened 2021-04-28
Remove TestCacheServiceNativeImage test

dmvolod:


Crumby (approved):

ryanemerson:
@dmvolod I think if we're going to keep this test, then by default it should be enabled on the upstream CI, otherwise it will always be ignored and we might as well remove the test. So I think we need to set the env in the travis file.

Crumby:
@ryanemerson What about removing the test and running the whole testsuite with the native server image as discussed on Zulip? Or is that out of scope for now?

ryanemerson:
@Crumby It would be cool to have, however I think a lot of tests would be broken at this stage. I think given the low priority of the native server at this time it would be better to focus our efforts elsewhere.

Crumby:
@ryanemerson Does it makes sense to have the test present at all at this moment? I mean from your statement it seems it's not usable by users and there is a lot of work to be done before it'll be working. It's using Cache service which is intended to be abandoned. Running the test takes CI time with each PR

ryanemerson:
Good point @Crumby. I would be happy to remove the test altogether.

dmvolod:
@Crumby , @ryanemerson are you decide to remove all TestCacheService* from the test suite?
In this case I will close this PR and create a new to remove them at all.

Crumby:
@dmvolod I meant specifically `TestCacheServiceNativeImage` as that's the only tests using native server image.

dmvolod:
> I meant specifically `TestCacheServiceNativeImage` as that's the only tests using native server image.

Yes, true, @Crumby . Current changes disables `TestCacheServiceNativeImage` both for CI and default test in separate cluster

ryanemerson:
> @dmvolod I meant specifically `TestCacheServiceNativeImage` as that's the only tests using native server image.

+1 to only removing the Native test.

Crumby (approved):

ryanemerson:
Thanks @dmvolod

PR #1017, opened 2021-04-29
[test-integration] Minor testsuite improvements

Crumby:
Updates integration testsuite:
* Repurposes MinimalSetup deployment to DevSetup with disable security and monitoring
  * Adds HotRod test with disabled authentication
  * Verifies that ServiceMonitor is not created when the annotation is set to false
* Renames AdvancedSetupITs tests to corresponding DataGridServiceIT and CacheServiceIT as those are more meaningful in the context
* Extend README file to describe way of running the tests without having to setup properties
* Adds ServiceMonitor cleanup validator
* Updates Infinispan dependency to 12.1.1


test-integration/README.md
```diff
@@ -5,13 +5,16 @@ Brief description on how to run integration tests.
...
   * Operator has to watch different namespace then one it's installed in
+* Either valid context in `kubeconfig` file or set properties below.
 
 ## Running the tests
 * `mvn clean verify` to run all the tests
 * `mvn clean verify -Dit.test=${testName}` to run particular test
 * `mvn clean verify -Dit.test=org.infinispan.operator.*IT` to run functional tests that do not require cluster-admin access. User needs to have rights to manipulate with Infinispan CR objects however.
 * `mvn clean verify -Dit.test=org.infinispan.operator.installation.*IT` to run Operator installation tests that require cluster-admin access
 
+Tests retreive OpenShift context configuration from the current context in the `~/kube/config` file. You can override current context and fix the target environment using properties below.
```
dmvolod, line 16: @Crumby we can provide `-Dxtf.openshift.admin.kubeconfig` property explicitly

Crumby: @dmvolod Yeah, I didn't wanted to include all the possibilities as there is really lot of them. Best would be to point to xtf framework documentation but it's not greatly documented there either

dmvolod: Yes, @Crumby, would be nice and choose a section with connection options directly

ryanemerson:
Thanks @Crumby

PR #1029, opened 2021-05-05
Fix authentication request uri when executed with query

Crumby:
When authentication is requested client should return also query part as part of the uri otherwise request won't pass OpenShift HA proxy.


dmvolod:
@ryanemerson please pay attention that after 12.1.2.Final image release and test with the current PR Infinispan server still fails with ISPN000196 and ISPN000217 errors in CI.

ryanemerson:
Thanks @Crumby.

@dmvolod This fix is definitely needed, so I have merged this and will investigate the failures on a different PR.

PR #1030, opened 2021-05-05
rice embed-go revises git stage for unchanged files

dmvolod:


dmvolod:
@rigazilla , @ryanemerson please review this PR and merge as due to this problem git stage always changes on every local build/test

ryanemerson:
@dmvolod Is there any way that we can just exclude the problem fields, `FileModTime` etc, from the generated file? If not, then I'm happy with this workaround.

dmvolod:
> Is there any way that we can just exclude the problem fields, `FileModTime` etc, from the generated file? If not, then I'm happy with this workaround.

@ryanemerson not sure, that's easy to do this. We can create our own generator, but it's a bit complex solution for now.

ryanemerson (approved):

PR #1031, opened 2021-05-06
Backup/Restore Improvements

ryanemerson:


test/e2e/backup-restore/backup_restore_test.go
```diff
@@ -126,7 +124,7 @@ func testBackupRestore(t *testing.T, clusterSpec clusterSpec, clusterSize int) {
 	sourceCluster := name + "-source"
 	infinispan := clusterSpec(sourceCluster, namespace, clusterSize)
 	testKube.Create(infinispan)
-	defer testKube.DeleteInfinispan(infinispan, tutils.SinglePodTimeout)
+	defer testKube.CleanNamespaceAndLogOnPanic(namespace)
```
Crumby, line 127: Is there a way to disable cleanup via flag/property  on failure for debugging purposes?

ryanemerson: I have updated the code now, so it would just be a case of commenting the `DeleteAllOf` commands in `CleanNamespaceAndLogOnPanic` .. if you think it would be useful, we can add a env var to toggle this. Any thoughts on what it should be called?

I'm just investigation why backup/restore is not  playing nice with travis atm, but my inteion is to make use of this method in all tests.

build/travis/debug-pods.sh
```diff
@@ -9,10 +9,8 @@ do
 	for SS in "${SSS[@]}"; do
 	  if [ -n "${SS}" ]; then
 	    echo "${SS}" "is ready for logs"
-	    kubectl get pods -o yaml >> debug.log
```
dmvolod, line 12: @ryanemerson I think we can remove this script as you've implemented more intelligent solution for Pod logs, and this script was a quick-fix workaround

ryanemerson: I didn't know if the Infinispan and Statefulset yaml might be useful? I suppose I can add this to the test method as well.

dmvolod: Sometimes we need SS status to identify why Pods can't be created and Infinispan CR status in other cases, but we can trace them inside code also as your solution for Pods

test/e2e/backup-restore/backup_restore_test.go
Crumby: Maybe `KEEP_DEPLOYMENT_RUNNING`?

ryanemerson:
@dmvolod The most recent [test run](https://travis-ci.org/github/infinispan/infinispan-operator/jobs/769916005) failed because the backup pod was stuck in the "pending" state with the following tolerations:

```yaml
tolerations:
- key: node.kubernetes.io/not-ready
  operator: Exists
  value: ""
  effect: NoExecute
  tolerationseconds: 300
- key: node.kubernetes.io/unreachable
  operator: Exists
  value: ""
  effect: NoExecute
  tolerationseconds: 300
```

The interesting thing is that the `TestBackupRestore` test passed with no problems, but executing `TestBackupRestoreNoAuth` is when  the issue occurred.

test/e2e/backup-restore/backup_restore_test.go
ryanemerson: `CLEANUP_INFINISPAN_ON_FINISH` already exists, so we can just reuse that.

dmvolod (approved):

dmvolod:
@ryanemerson very good, I'm eager to see CI green for all our PR's
Is it ready to be merge now?

ryanemerson:
> I'm eager to see CI green for all our PR's

:crossed_fingers: 

It's ready, thanks

dmvolod:
Thanks, @ryanemerson 
Merged into master

PR #1032, opened 2021-05-06
[test] Shorten Backup/Restore test clusters names

Crumby:
Hostname generated by TestBackupRestore/DataGridNoAuth is too long. (54 characters by itself leaving only 9 characters for namespace and OpenShift domain)


ryanemerson:
Thanks @Crumby

PR #1036, opened 2021-05-10
Documentation change

oraNod:
Rewording for creating caches via Cache CR. Comes from downstream. We should not suggest using tech preview feature as the "primary mechanism".

### DIFF
--- a/documentation/asciidoc/topics/con_creating_caches.adoc
+++ b/documentation/asciidoc/topics/con_creating_caches.adoc
@@ -1,8 +1,8 @@
 [id='creating_caches-{context}']
 = Creating {brandname} Caches
-To create caches when running {brandname} on {k8s}, you should:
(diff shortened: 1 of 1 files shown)


documentation/asciidoc/topics/con_creating_caches.adoc
```diff
@@ -2,7 +2,7 @@
 = Creating {brandname} Caches
 To create caches when running {brandname} on {k8s}, you should:
 
-* Use `Cache` CR as the primary mechanism for creating caches.
+* Use `Cache` CR as the mechanism for creating caches through the {openshiftshort} front end.
```
oraNod, line 5: @Crumby This is the same wording as on the "master" branch. Changing from saying "primary" mechanism to something more generic. Sound OK to you?

ryanemerson:
Thanks @oraNod

PR #1043, opened 2021-05-13
Support for cross-site expose type (LoadBalancer and NodePort) reflection

dmvolod:


dmvolod:
NodePort number should be reflected also

ryanemerson (approved):

ryanemerson:
Thanks @dmvolod

Issue #1053, opened 2021-05-17
Credentials generated deterministically


ryanemerson:
User credentials are generated deterministically as they use the `math/rand` packages, instead the `crypto/rand` package should be used to ensure that credentials are randomly generated.

PR #1054, opened 2021-05-17
Ensure genereated credentials are random #1053

ryanemerson:


dmvolod:
@ryanemerson just for my information: is it means, that before this fix password was the same for the same cluster or I missed something?

dmvolod (approved):

ryanemerson:
> @ryanemerson just for my information: is it means, that before this fix password was the same for the same cluster or I missed something?

Previously every call to `rand.Int` with the same size arg would cycle through the same numbers, which meant that for each operator deployment the generated passwords would be predictable for a given deployment. For example, the first Infinispan CR cluster to be created by an operator would always have the same credentials, the second cluster created by all operators would have the same credentials.

ryanemerson:
Just repushing to fix a type in the commit message.

rigazilla:
merged thanks @ryanemerson !

PR #1056, opened 2021-05-17
CleanNamespaceAndLogOnPanic in test should utilize Unstructured type for print yaml structure camelCase

dmvolod:


ryanemerson:
Thanks @dmvolod

PR #1078, opened 2021-05-25
Authorization

ryanemerson:
[Image PR](https://github.com/infinispan/infinispan-images/pull/135)
[Config Generator PR](https://github.com/infinispan/infinispan-image-artifacts/pull/41)


dmvolod:
@ryanemerson please pay attention that you catch  #1079 in this test

dmvolod (approved):

test/e2e/utils/httpClient.go
```diff
@@ -75,6 +75,7 @@ func (c *httpClientConfig) exec(method, path, payload string, headers map[string
 	ExpectNoError(err)
 	fmt.Printf("%s: %s\n", method, httpURL)
 	rsp, err := c.request(httpURL, method, payload, headers)
+	ExpectNoError(err)
```
dmvolod, line 78: @ryanemerson I've identified that sometimes request return `net.IsConnectionRefused` error. Do you have ideas why it happens and how we can handle this? Otherwise we will get error on test.
`CheckExternalAddress` has this error handle, probably we need to embedded this handler to the test httpClient

ryanemerson: > I've identified that sometimes request return `net.IsConnectionRefused` error. Do you have ideas why it happens and how we can handle this?

No idea the cause tbh. Adding some kind of retry logic to the client makes sense to me :+1:

ryanemerson:
Self merging as required for `master` to pass now that the Infinispan images have Authorization enabled by default.

PR #1090, opened 2021-05-31
Automatic cross-site static service discovery inside single cluster

dmvolod:


pkg/controller/constants/constants.go
```diff
@@ -73,7 +73,8 @@ const (
 	NativeImageMarker         = "native"
 	GeneratedSecretSuffix     = "generated-secret"
 	InfinispanFinalizer       = "finalizer.infinispan.org"
-	SiteServiceTemplate       = "%v-site"
+	SiteServiceNameTemplate   = "%v-site"
+	SiteServiceFQNTemplate    = "%s.%s.svc.cluster.local"
```
ryanemerson, line 77: Does this need to be a constant? It's only used once.

dmvolod: Yes, exactly. I will move both `SiteServiceNameTemplate` and `SiteServiceFQNTemplate` to the `types_util.go`

ryanemerson (approved):

oraNod:
@dmvolod Does this eliminate the need for users to specify any `spec.service.sites` configuration for cross-site in a single cluster?

dmvolod:
> Does this eliminate the need for users to specify any `spec.service.sites` configuration for cross-site in a single cluster?

Yes, @oraNod correct question. I will update an examples,for static-internal in this case

pruivo (approved):

ryanemerson:
Thanks @dmvolod

Issue #851, opened 2021-02-25
Allow client certificate authentication


ryanemerson:
It should be possible for a user to provide a truststore via the `tls-secret` and configure whether client cert validation or authentication should be enabled.

```yaml
spec:
  security:
    endpointEncryption: 
            type: Secret 
            certSecretName: tls-secret 
            clientCert: None | Validate | Authenticate
```

A user secret would then be as follows:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: tls-secret
type: Opaque
stringData:
    alias: server 
    password: password 
data:
    keystore.p12:  "MIIKDgIBAzCCCdQGCSqGSIb3DQEHA..." 
    truststore.p12: "MIIKDgIBAzCCCdQGCSqGSIb3DQEHA..." 
```

N.B. With this proposal truststores can't be configured with `spec.security.endpointEncryption.type = Service`

### Validate
Only a truststore is created:

```xml
         <server-identities>
            <ssl>
               <keystore path="server.pfx" relative-to="infinispan.server.config.path" keystore-password="secret"
                         alias="server"/>
               <truststore path="ca.pfx"  relative-to="infinispan.server.config.path" password="secret"/>
            </ssl>
         </server-identities>
```

### Authenticate
Both a truststore and a trustore-realm is created.

```xml
   <security-realms>
      <security-realm name="default">
         <server-identities>
            <ssl>
               <keystore path="server.pfx" relative-to="infinispan.server.config.path" keystore-password="secret"
                         alias="server"/>
               <truststore path="trust.pfx"  relative-to="infinispan.server.config.path" password="secret"/>
            </ssl>
         </server-identities>
         <truststore-realm/>
      </security-realm>
   </security-realms>
```

# Operator 2.0.x
Only `clientCert: None | Authenticate` can be supported as truststore-realms are mandatory so Validate is not possible.

dmvolod:
> N.B. With this proposal truststores can't be configured with `spec.security.endpointEncryption.type = Service`

we could use separate names for trust and keystore in case of client and service certificate but seems to be should be supported by server config and generator.

rigazilla:
I think it would be good to support this feature also for `spec.security.endpointEncryption.type = Service`

ryanemerson:
> I think it would be good to support this feature also for `spec.security.endpointEncryption.type = Service`

I originally avoided this as there was no way to provide the truststore, but I think the following could be possible:

[...]

PR #1099, opened 2021-06-04
Docs for client certificate authentication

oraNod:
Docs for issue #851


documentation/asciidoc/topics/proc_enabling_client_cert_authn.adoc
```diff
@@ -0,0 +1,36 @@
+[id='enabling-client-certificate-authentication_{context}']
+= Enabling client certificate authentication
+
+[role="_abstract"]
+{brandname} clusters do not use client certificate authentication by default.
+Complete the following procedure so that clients must present valid certificates before they can connect to {brandname} endpoints.
```
ryanemerson, line 6: "must present valid certificates before they can connect to {brandname} endpoints." etc has already been said, maybe just:

"{brandname} clusters do not use client certificate authentication by default. Complete the following procedure to enable it."

documentation/asciidoc/topics/proc_enabling_client_cert_authn.adoc
```diff
@@ -0,0 +1,36 @@
...
+
+You can configure {brandname} to verify client identities from certificates in a trust store in two ways:
+
+`Validate`:: Requires a trust store that contains only the signing certificate, which is typically a Certificate Authority (CA). Any client that presents a signed certificate can connect to {brandname}.
```
ryanemerson, line 10: "Any client that presents a certificate signed by the CA can connect to {brandname}".

documentation/asciidoc/topics/proc_enabling_client_cert_authn.adoc
```diff
@@ -0,0 +1,36 @@
...
+.Procedure
+
+. Set either `Validate` or `Authenticate` as the value for the `spec.security.endpointEncryption.clientCert` field in your `Infinispan` CR.
++
+[NOTE]
+====
+The default value is `None`.
+====
++
+. Optionally specify a name for the secret that contains the client trust store with the `spec.security.endpointEncryption.clientCertSecretName` field.
```
ryanemerson, line 22: This isn't optional. The user must always provide `clientCertSecretName`, otherwise there are 1) no certificates to include in a truststore created by the operator or 2) No existing truststore to use.

ryanemerson (commented):
Overall looks great :slightly_smiling_face:

documentation/asciidoc/topics/proc_enabling_client_cert_authn.adoc
oraNod: @ryanemerson Right after this we state: "By default Infinispan Operator adds the truststore to a secret named `example-infinispan-client-cert-secret`.

So users should specify the default secret name if not using a custom secret?

ryanemerson: > "By default Infinispan Operator adds the truststore to a secret named `example-infinispan-client-cert-secret`.

This is wrong. A trustsore secret is never generated by the operator, it has to be provided by the user as this is the only way to get the truststore or certs to create the truststore. 

I'm guessing I failed to update my docs primer with this detail. Originally the idea was that a truststore could be generated using the Openshift service CA, but this doesn't make sense and was scrapped.

oraNod: I'm suffering from brain fog myself (and the internet is acting weird today). I've updated with some changes, hopefully they make sense.

ryanemerson (approved):

dmvolod:
I don't know exact place, but we need to add notes, that operator will set owner reference for the provided secrets and will garbage collected them when Infinispan CR will be deleted, Also is not possible to share Secrets between instances.
@ryanemerson wdyt about this?

documentation/asciidoc/topics/proc_enabling_client_cert_authn.adoc
```diff
@@ -0,0 +1,38 @@
...
+====
+The default value is `None`.
+====
++
+. Optionally specify the secret that contains the client trust store with the `spec.security.endpointEncryption.clientCertSecretName` field.
++
+By default {ispn_operator} uses a trust store secret named `{example_crd_name}-client-cert-secret`.
+If you want to use this secret you do not need to specify the `clientCertSecretName` field.
+However, you must create the secret if it does not already exist.
++
```
ryanemerson, line 27: > I don't know exact place, but we need to add notes, that operator will set owner reference for the provided secrets and will garbage collected them when Infinispan CR will be deleted, Also is not possible to share Secrets between instances.
> @ryanemerson wdyt about this?

Agreed @dmvolod.

We should maybe add this here as a `[NOTE]` @oraNod.

Something along the lines of:

"In order to support the updating of truststore secret values at runtime, it's necessary for the operator to add a owner reference with `controller: true` to the truststore Secret. Consequently, it's not possible for truststore Secrets to be utilised by multiple Infinispan CRs. Furhermore, upon Infinispan CR deletion the configured Secret will also be deleted."

[GC](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/) docs if it helps to understand what's going on.

This also applies to the TLS keystore secrets, so we need to update the docs in that section as well.

oraNod: @ryanemerson when you say TLS keystore secrets are you referring to `spec.security.endpointEncryption.certSecretName`?

ryanemerson: Bingo!

oraNod: thanks @ryanemerson and @dmvolod I added a note to two spots in the doc. hopefully the changes look good.

dmvolod (approved):

ryanemerson:
Thanks @oraNod

Issue #1102, opened 2021-06-07
Cluster configuration ConfigMap updates on start 


dmvolod:
Seems to be this is regression of the #1007
Startup configuration ConfigMap
```
infinispan:
  authorization:
    enabled: false
    roleMapper: cluster
  clusterName: example-infinispan
  zeroCapacityNode: false
  locks: {}
jgroups:
  transport: tcp
  dnsPing:
    query: example-infinispan-ping.infinispan-operator.svc.cluster.local
  diagnostics: false
endpoints:
  auth: true
  dedicatedAdmin: true

```

`keystore` embedding later, that's wrong and forcing cluster rolling upgrade
```
infinispan:
  authorization:
    enabled: false
    roleMapper: cluster
  clusterName: example-infinispan
  zeroCapacityNode: false
  locks: {}
jgroups:
  transport: tcp
  dnsPing:
    query: example-infinispan-ping.infinispan-operator.svc.cluster.local
  diagnostics: false
keystore:
  path: /opt/infinispan/server/conf/keystore
  password: password
  alias: server
  crtPath: /etc/encrypt/keystore
endpoints:
  auth: true
  dedicatedAdmin: true
```

Probably this cause regression in cross-site forming also

dmvolod:
@ryanemerson could you please to have a look.

ryanemerson:
@dmvolod sure

ryanemerson:
@dmvolod How do you reproduce this behaviour? Is it just in the testsuite or have you observed it by just installing a CR? I looked at #1106 but I am unable to reproduce locally.

dmvolod:
@ryanemerson I can see configmap update info in the logs on OCP, than set breakpoint and look at the configmap structure after create and on the next reconcile step. They were different. Seems to not actual on k8s, but need to be validated.

dmvolod:
Seems to refactored code regression a bit related to the #1103 that also need to be fixed. I added second node to the kind cluster in #1094 and issue #1103 can be reproduced in stable way

PR #1109, opened 2021-06-08
Cluster configuration ConfigMap updates on start #1102

ryanemerson:
@dmvolod It turns out I also wasn't configuring the `c.truststore.Password` correctly, which resulted in the ConfigMap values changing on second reconcilliation.


dmvolod (approved):

ryanemerson:
Thanks for the review @dmvolod

PR #1111, opened 2021-06-10
Tests should setup OSDK_FORCE_RUN_MODE=local if operator runs locally

### LINKED ISSUE #929: Tests should setup OSDK_FORCE_RUN_MODE=local if operator runs locally
Currently, OSDK_FORCE_RUN_MODE=local set only for run-local and inside travis.yaml
It should be set for all tests with setup RUN_LOCAL_OPERATOR=true (default behavior)

dmvolod:
Can't wait for someone could fix this issue, as need to be remember this env var option every local test run :smile:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,6 @@ jobs:
         - docker
       env:
         - KIND_VERSION=v0.11.0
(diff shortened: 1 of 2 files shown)


ryanemerson:
Thanks @dmvolod

PR #1115, opened 2021-06-10
Print error in case of error only in config controller

dmvolod:
Error should be printed in case of the real error (when resource not ready empty error can be printed)


ryanemerson:
Good spot @dmvolod. Can you create a PR for `master` as well.

dmvolod:
> Good spot @dmvolod. Can you create a PR for `master` as well.

Fix already in PR #1113

Issue #1118, opened 2021-06-11
x509: certificate relies on legacy Common Name field, use SANs


ryanemerson:
The `TestEndpointEncryptionUpdate` test will fail with the following error when using go >= 1.15.x

```
Get "https://172.18.0.2:32189/rest/v2/cache-managers/default/health": x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0 [recovered]
```

PR #1119, opened 2021-06-11
x509: certificate relies on legacy Common Name field, use SANs #1118

ryanemerson:


dmvolod (approved):

Crumby (approved):

test/e2e/utils/kubernetes.go
```diff
@@ -466,6 +467,16 @@ func CheckExternalAddress(c HTTPClient, hostAndPort string) bool {
 	return resp.StatusCode == http.StatusOK
 }
 
+func isTlsHandskakeTimeoutError(err error) bool {
+	if err != nil {
+		if urlError, ok := err.(*url.Error); ok {
```
dmvolod, line 472: new linter will fails with 
`test/e2e/utils/kubernetes.go:472:22: type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors (errorlint)`

test/e2e/utils/kubernetes.go
```diff
@@ -457,7 +458,7 @@ func getNodePort(service *corev1.Service) int32 {
 func CheckExternalAddress(c HTTPClient, hostAndPort string) bool {
 	httpURL := fmt.Sprintf("%s/%s", hostAndPort, consts.ServerHTTPHealthPath)
 	resp, err := c.Get(httpURL, nil)
-	if net.IsConnectionRefused(err) {
+	if net.IsConnectionRefused(err) || isTlsHandskakeTimeoutError(err) {
```
dmvolod, line 461: @ryanemerson I got following error on my OCP cluster
```
--- FAIL: TestCacheService (59.79s)
panic: Get "https://test-cache-service-external-namespace-for-testing.apps.cluster-msk-fd24.msk-fd24.example.opentlc.com/rest/v2/cache-managers/default/health": EOF [recovered]
	panic: Get "https://test-cache-service-external-namespace-for-testing.apps.cluster-msk-fd24.msk-fd24.example.opentlc.com/rest/v2/cache-managers/default/health": EOF
```
Seems to we need to add `errors.Is(err, io.EOF)` for retry

ryanemerson: I originally had this logic, but on my cluster EOF was only happening when there was a config error. I can re-add it though if you're seeing this intermittingly.

dmvolod: Also, this errors should be handled inside `func (c *httpClientConfig) exec(method, path, payload string, headers map[string]string) (*http.Response, error) {` not there, as this is too late and errors occurs much early

dmvolod: @ryanemerson please also replace `ExpectNoError(err)` inside `func (c *httpClientConfig) exec(method, path, payload string, headers map[string]string) (*http.Response, error)` with return.

ryanemerson: I have updated to make use of `errors.As` and I have made the handling more robust by checking for `io.EoF` as well as any errors that implement the `Temperorary` or `Timeout` interfaces.

I agree that the rety/error handling should be moved to `httpClient.go` but this needs to be properly thought through and I think it should come in a subsequent PR.

ryanemerson: > @ryanemerson please also replace `ExpectNoError(err)` inside `func (c *httpClientConfig) exec(method, path, payload string, headers map[string]string) (*http.Response, error)` with return.

Sure, I'll update todo that :+1:

dmvolod: > I agree that the rety/error handling should be moved to `httpClient.go` but this needs to be properly thought through and I think it should come in a subsequent PR.

Yes, agree. But without return instead of `ExpectNoError(err)` your error handlers code never reach

ryanemerson: Updated. I hadn't read your second comment when I replied initially :slightly_smiling_face:

dmvolod:
Thanks, @ryanemerson 
Merged into main

Issue #1125, opened 2021-06-15
Encryption secret VolumeMount not updated on upgrade


ryanemerson:
The changes required for client-cert authentication means that tls secrets are now mounted at `/etc/encrypt/keystore` instead of `/etc/encrypt`. Consequently, it's now necessary for the VolumeMount to be updated to the new value when performing an upgrade.

dmvolod:
Fixed

PR #1126, opened 2021-06-15
Encryption secret VolumeMount not updated on upgrade #1125

ryanemerson:


Crumby:
I wonder if it fully fixes the issue with upgrade. If I understand it correctly then the issue is that ConfigMap is updated and then the server is gracefully restarted via rest endpoint before shutting down and ugprading. Once it restarts, it loads new configuration that is not valid due to incorrect mount point. If we update volumemount aren't we going to lose graceful restart? So shouldn't be ConfigMap update postponed until the cluster is down?

ryanemerson:
> If we update volumemount aren't we going to lose graceful restart?

I'm not sure what you mean here.

> So shouldn't be ConfigMap update postponed until the cluster is down?

This isn't something we handle for any other config change when it comes to upgrades. The operator checks whether an upgrade is required and sets `ConditionUpgrade: true` before comparing `CONFIG_HASH` values. So the pods are gracefully shutdown before the StatefulSet spec is updated.

ryanemerson:
> * Configuration ConfigMap is updated to point to a new expected keystore path

The ConfigMap change has no impact on the running pods as the file is only read on the server startup.

> The issue happens when pods are gracefully shut down because that triggers restart and reload of new configuration. 

This is guarded against with `ConditionUpgrade`:

1. Operator determines that an upgrade should occur and sets `ConditionUpgrade: true` and returns. Resources still exist at this point
2. Operator sees `ConditionUpgrade: true` via `infinispan.IsUpgradeNeeded()`, so deletes resources and sets `ConditionUpgrade: false`. Return and requeue
3. Operator creates new StatefulSet with correct VolumeMounts and upto date `CONFIG_HASH` val.

dmvolod (approved):

Crumby:
> The ConfigMap change has no impact on the running pods as the file is only read on the server startup

If this would be true we wouldn't be hitting upgrade failures with non-upgraded sever version

Crumby (approved):

Issue #1123, opened 2021-06-15
endpointSecretName field is not initialized when generated


rigazilla:
`endpointSecretName` must be initialized by the operator with the name of the generated secret

dmvolod:
Fixed

PR #1132, opened 2021-06-16
Initialize endpointSecretName in case of generate secret #1123

rigazilla:
ported on main fix for #1123


dmvolod (approved):

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -155,6 +155,9 @@ func (ispn *Infinispan) ApplyDefaults() {
 	if ispn.Spec.Security.EndpointAuthentication == nil {
 		ispn.Spec.Security.EndpointAuthentication = pointer.BoolPtr(true)
 	}
+	if *ispn.Spec.Security.EndpointAuthentication {
```
dmvolod, line 158: @rigazilla please pay attention, that `endpointSecretName` never removed when changing `endpointAuthentication` from `true` to `false`

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -155,6 +155,11 @@ func (ispn *Infinispan) ApplyDefaults() {
...
+		ispn.Spec.Security.EndpointSecretName = ispn.GetSecretName()
+	} else if ispn.IsGeneratedSecret() {
```
dmvolod, line 160: `IsGeneratedSecret` should be changed, it worked on assumption, that for generated secret `endpointSecretName` is empty

rigazilla: It also assume that a generated secret has a well-known suffix:
https://github.com/infinispan/infinispan-operator/blob/a4b51250caa91795ae35a50cce95628eaf039a10/pkg/apis/infinispan/v1/types_util.go#L373-L374

so it works on that side, but it also check that authentication is enable that needs a check :)

rigazilla: I would remove the IsAuthenticationEnabled check from the IsGeneratedSecret method and explicitly check for it in the caller (1 reference), @dmvolod wdyt?

dmvolod: Yes, @rigazilla agree.
In this case we can identify that secret is auto-generated by the name only.

rigazilla: @dmvolod you mean assuming that it will never be equal to ""? Actually it should be, but I feel a little thrill

dmvolod: Yes, it will never be equal to "" if auth is enabled. Previous logic defines that empty value means auto-generated.

pkg/apis/infinispan/v1/types_util.go
```diff
@@ -365,7 +370,7 @@ func (ispn *Infinispan) IsClientCertEnabled() bool {
 
 // IsGeneratedSecret verifies that the Secret should be generated by the controller
 func (ispn *Infinispan) IsGeneratedSecret() bool {
-	return ispn.IsAuthenticationEnabled() && (ispn.Spec.Security.EndpointSecretName == "" || strings.HasSuffix(ispn.Spec.Security.EndpointSecretName, consts.GeneratedSecretSuffix))
+	return strings.HasSuffix(ispn.Spec.Security.EndpointSecretName, consts.GeneratedSecretSuffix)
```
dmvolod, line 373: in this case we could validate whole secret name, not just suffix to avoid confusion

dmvolod:
@rigazilla looks very good. Can we merge it into main now?

rigazilla:
merged @dmvolod  and thanks for the review!

Issue #1140, opened 2021-06-17
Grafana replicas should be an Instant metric


wburns:
The Replicas (Up) can be shown as an average which makes no sense over time. Instead it should be instant only so the user knows how many are up right now.

PR #1136, opened 2021-06-16
Change Grafana CPU Load to CPU Usage

wburns:
Show the CPU usage of the JVM process for the pod instead of the system load average.

PR #1141, opened 2021-06-17
Grafana replica instant

wburns:
https://github.com/infinispan/infinispan-operator/issues/1140

Change Replicas to be an instant value so it doesn't show an average.

I cherrypicked https://github.com/infinispan/infinispan-operator/pull/1136 before since otherwise they will conflict.


rigazilla (approved):

wburns:
Rebased.

Issue #1150, opened 2021-06-18
ConfigMap changes on update cause startup failures


ryanemerson:
Currently the `config_controller#Process` function always reconciles the ConfigMap state regardless of whether an upgrade will be scheduled. If a future operator version introduces a Infinispan config change that is not compatible with the old image version, then this will cause the Infinispan pods to fail when they restart during the gracefulshutdown process, which ultimately causes the cluster upgrade to fail.

dmvolod:
> If a future operator version introduces a Infinispan config change that is not compatible with the old image version, then this will cause the Infinispan pods to fail when they restart during the gracefulshutdown process, which ultimately causes the cluster upgrade to fail.

This means that Pods could be restarting during upgrade before new image is placed inside StatefulSet. Mostly it's not a problem with the configuration controller but a bit lack of logic in the upgrade procedure.

ryanemerson:
> This means that Pods could be restarting during upgrade before new image is placed inside StatefulSet. Mostly it's not a problem with the configuration controller but a bit lack of logic in the upgrade procedure.

Exactly. This is a problem with the current GracefulShutdown approach and was the primary motivation for creating #1154. If this behaviour didn't exist, then #1126 would have been sufficient on it's own to fix the upgrade failures.

PR #1151, opened 2021-06-18
[2.1.x] ConfigMap changes on update cause startup failures #1150

ryanemerson:


pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -74,6 +75,16 @@ func Add(mgr manager.Manager) error {
 }
 
 func (c *configResource) Process() (reconcile.Result, error) {
+	// Don't update the ConfigMap if an update will be required
+	if req, err := ispnctrl.IsUpgradeRequired(c.infinispan); req || err != nil {
+		return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCreateResource}, nil
+	}
+
+	// If an upgrade is in progress wait for it to complete before updating the ConfigMap
+	if c.infinispan.IsAnyCondititionTrue(ispnv1.ConditionUpgrade, ispnv1.ConditionStopping) {
```
Crumby, line 84: When is ConditionUpgrade removed or set to false? Is that when podCount == 0 or the when the cluster was upgraded with new image?

ryanemerson: It's set to false once `!ConditionStopping` (podCount == 0) and the old StatefulSet has been deleted.

Crumby (approved):

dmvolod (approved):

Issue #763, opened 2021-01-26
Infinispan cluster shutdown/restart problem


mirostary:
Hello, 
I’m trying to restart the Infinispan cluster on OCP4.6. Basically, I apply this YAML:

```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: a-tests-infinispan-cluster
  namespace: datagrid-cluster
spec:
  replicas: 2
  service:
    type: DataGrid
  container:
    storage: 2Gi
    extraJvmOpts: "-XX:NativeMemoryTracking=summary"
    cpu: "2000m"
    memory: 1Gi
  security:
    endpointSecretName: connect-secret
    endpointEncryption:
      type: Secret
      certSecretName: tls-secret
  expose:
    type: LoadBalancer
```

Then I change the number of replicas to 0 and apply again. And here is the problem, sometimes the cluster got the “stopping” condition (just stop the pods), and sometimes it got the “gracefulShutdown” condition (terminate the pods)… I wasn’t able to find out why I got a different condition when I didn't change anything.

I tried changing the `.spec.container` settings but without success. I wasn’t able to find out the reason in the guide... Do you know where is the problem?

I have installed the Datagrid operator 8.1.3 in another namespace and it is watching the namespace "datagrid-cluster".
The Openshift version is 4.6.9

dmvolod:
Test sometimes hangs with following status, i.e. `Stopping` condition has not been cleared in some cases, however `WellFormed` condition is set to `True`
```
  status:
    conditions:
    - status: "True"
      type: PreliminaryChecksPassed
    - message: 'View: test-check-data-survive-to-shutdown-0-2251'
      status: "True"
      type: WellFormed
    - status: "True"
      type: Stopping
    consoleUrl: http://172.18.0.2:32041/console
    podStatus:
      ready:
      - test-check-data-survive-to-shutdown-1
    statefulSetName: test-check-data-survive-to-shutdown
```

ryanemerson:
Closing as it should be resolved by https://github.com/infinispan/infinispan-operator/issues/1154 in 2.2.4.

PR #1152, opened 2021-06-18
ConfigMap changes on update cause startup failures #1150

ryanemerson:


pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -74,6 +75,16 @@ func Add(mgr manager.Manager) error {
 }
 
 func (c *configResource) Process() (reconcile.Result, error) {
+	// Don't update the ConfigMap if an update will be required
+	if req, err := ispnctrl.IsUpgradeRequired(c.infinispan); req || err != nil {
+		return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCreateResource}, nil
+	}
+
+	// If an upgrade is in progress wait for it to complete before updating the ConfigMap
+	if c.infinispan.IsAnyCondititionTrue(ispnv1.ConditionUpgrade, ispnv1.ConditionStopping) {
```
dmvolod, line 84: If #763 wouldn't be fixed, this code cause continuous loop

ryanemerson: True. I wasn't aware of that issue. That issue needs to be fixed eventually, but I don't think it should stop this fix. Worst case scenario the ConfigMap retains the old data whilst `ConditionStopping` and the config-controller will loop until this is resolved.

dmvolod: Probably, we could validate upgrade condition again here, i.e. no Pods or Pods with DEFAULT_IMAGE different from operator configuration

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -706,16 +719,23 @@ func (r *ReconcileInfinispan) scheduleUpgradeIfNeeded(infinispan *infinispanv1.I
...
+	if err != nil {
+		return false, err
 	}
-	return nil, nil
+	return upgradeRequired(infinispan, podList)
+}
+
+func PodList(infinispan *infinispanv1.Infinispan) (*corev1.PodList, error) {
+	podList := &corev1.PodList{}
+	if err := kubernetes.ResourcesList(infinispan.Namespace, PodLabels(infinispan.Name), podList); err != nil {
```
dmvolod, line 735: can we just return `podList, err` without validation?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -683,19 +683,32 @@ func (r *ReconcileInfinispan) upgradeInfinispan(infinispan *infinispanv1.Infinis
 }
 
 func (r *ReconcileInfinispan) scheduleUpgradeIfNeeded(infinispan *infinispanv1.Infinispan, podList *corev1.PodList, logger logr.Logger) (*reconcile.Result, error) {
+	if upgrade, err := upgradeRequired(infinispan, podList); upgrade || err != nil {
+		if err := r.update(infinispan, func() {
+			logger.Info("schedule an Infinispan cluster upgrade", "pod default image", podList.Items[0].Spec.Containers[0], "desired image", consts.DefaultImageName)
```
dmvolod, line 688: Probably image is missing, whole container will be printing

pkg/controller/infinispan/resources/config/config_controller.go
ryanemerson: That's handled in the call to `if req, err := ispnctrl.IsUpgradeRequired(c.infinispan); req || err != nil {`

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: I don't understand what you mean? In order for `upgradeRequired` to return true, it's necessary for `len(podList) > 0`.

pkg/controller/infinispan/resources/config/config_controller.go
dmvolod: Why we need additional validation ` c.infinispan.IsAnyCondititionTrue` in this case?

pkg/controller/infinispan/infinispan_controller.go
dmvolod: `podList.Items[0].Spec.Containers[0]` returns whole container object, not an image

pkg/controller/infinispan/resources/config/config_controller.go
ryanemerson: They handle two different scenarios. The first if statement is to guard against the config-controller processing Infinispan before infinispan_controller can initiate the upgrade and sets the appropriate conditions. The second if statement ensures that if a upgrade has been initiated, we don't update the Config until after the original cluster pods have been shutdown.

dmvolod: I'm not sure, but seems to new Pods never will have new DEFAULT_IMAGE value till upgrade has finished. Seems to second validation is bit redundant, but maybe I'm wrong

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: Gotcha! My copy and paste skills failed me there. Updated.

pkg/controller/infinispan/resources/config/config_controller.go
ryanemerson: > Seems to second validation is bit redundant, but maybe I'm wrong

Hmm actually I think you're right, as `IsUpgradeRequired` already checks if `ConditionUpgrade` is true. I'll try removing and verify locally.

dmvolod: Yes, `ispnv1.ConditionStopping` a bit danger validation as it's related to graceful shutdown also. But in this case we can free to change any config, as it's not upgrade

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -706,16 +720,20 @@ func (r *ReconcileInfinispan) scheduleUpgradeIfNeeded(infinispan *infinispanv1.I
...
+	if err != nil {
+		return false, err
 	}
-	return nil, nil
+	return upgradeRequired(infinispan, podList)
+}
+
+func PodList(infinispan *infinispanv1.Infinispan) (*corev1.PodList, error) {
+	podList := &corev1.PodList{}
+	return podList, kubernetes.ResourcesList(infinispan.Namespace, PodLabels(infinispan.Name), podList)
```
dmvolod, line 736: seems to we need execute `err := kubernetes.ResourcesList(infinispan.Namespace, PodLabels(infinispan.Name), podList)` and then `return podList, err`
May be I'm wrong

pkg/controller/infinispan/resources/config/config_controller.go
ryanemerson: Do we use graceful shutdown for anything other than upgrades?

dmvolod: Yes, sure. If anyone defines replicas=0 operator will execute graceful shutdown

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: I don't think it matters, as `podList` is just a pointer. The call to `ResourcesList` updates the value associated with the pointer.

pkg/controller/infinispan/resources/config/config_controller.go
ryanemerson: Ok, in that case it should be fine to remove the `ConditionUpgrade` check as scaling the cluster back up will contain the same Infinispan version and so we can't run into this issue.

dmvolod (approved):

PR #1145, opened 2021-06-17
Grafana replica pod filter

wburns:
Fixes https://github.com/infinispan/infinispan-operator/issues/1144

This is based off of https://github.com/infinispan/infinispan-operator/pull/1141 to prevent conflict.

PR #1158, opened 2021-06-21
Grafana memory filter by cluster

### LINKED ISSUE #1157: Grafana Memory requests and limits show other pods in same namespace
We should change the query so it only counts pods that are for our Infinispan cluster.

This should be doable by changing the query to ﻿pod﻿ =~﻿"$cluster-[0-9]+". This relies upon our pod names being derived from the cluster name with - and numbers being appended, which it does currently. This should also prevent clusters with the same name collapsing on each other as well.

Also should add the pod name to the label.

wburns:
Fixes #1157 

This is based on #1145 to prevent conflict.


rigazilla (approved):

wburns:
Rebased

rigazilla:
merged, thanks @wburns

PR #1159, opened 2021-06-22
[2.1.x] ConfigMap changes on update cause startup failures #1150

ryanemerson:
- Add missing check for upgrade in progress

@dmvolod In our discussion [here](https://github.com/infinispan/infinispan-operator/pull/1152#pullrequestreview-687390671) I concluded that we could remove the `IsConditionTrue(ispnv1.ConditionUpgrade)` check. This isn't the case. The first check `IsUpgradeRequired` guards against an upgrade being scheduled in the near future and `IsConditionTrue(ispnv1.ConditionUpgrade)` guards against an upgrade currently being in progress.

### DIFF
--- a/pkg/controller/infinispan/resources/config/config_controller.go
+++ b/pkg/controller/infinispan/resources/config/config_controller.go
@@ -75,8 +75,13 @@ func Add(mgr manager.Manager) error {
 }
 
 func (c *configResource) Process() (reconcile.Result, error) {
-	// Don't update the ConfigMap if an update is required
+	// Don't update the ConfigMap if an update is about to be scheduled
 	if req, err := ispnctrl.IsUpgradeRequired(c.infinispan); req || err != nil {
+		return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCreateResource}, nil
+	}
+
+	// Don't update the configMap if an update is in progress
+	if c.infinispan.IsConditionTrue(ispnv1.ConditionUpgrade) {
 		return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCreateResource}, nil
 	}
 


pkg/controller/infinispan/resources/config/config_controller.go
```diff
@@ -75,11 +75,16 @@ func Add(mgr manager.Manager) error {
...
 
 func (c *configResource) Process() (reconcile.Result, error) {
-	// Don't update the ConfigMap if an update is required
+	// Don't update the ConfigMap if an update is about to be scheduled
 	if req, err := ispnctrl.IsUpgradeRequired(c.infinispan); req || err != nil {
 		return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCreateResource}, nil
 	}
 
+	// Don't update the configMap if an update is in progress
+	if c.infinispan.IsConditionTrue(ispnv1.ConditionUpgrade) {
```
Crumby, line 84: Naming could be confusing. I'd expect that ConditionUpgrade(ing) is true till the cluster is back up and everything was successfully updated. ClusterStopped condition would be better in this place.

dmvolod:
> @dmvolod In our discussion [here](https://github.com/infinispan/infinispan-operator/pull/1152#pullrequestreview-687390671) I concluded that we could remove the `IsConditionTrue(ispnv1.ConditionUpgrade)` check. This isn't the case. The first check `IsUpgradeRequired` guards against an upgrade being scheduled in the near future and `IsConditionTrue(ispnv1.ConditionUpgrade)` guards against an upgrade currently being in progress.

Yes, @ryanemerson I remember this discussion and thought about this issue. But probably upgrade identification based on Pods data which are not permanent not very good practice. We should identify DEFAULT_IMAGE value in the StatefulSet and than just shutdown and validate Pods presence. Which will guard more stable rather than validate Pods which can be rescheduled by controller at any time.

pkg/controller/infinispan/resources/config/config_controller.go
ryanemerson: > Naming could be confusing. I'd expect that ConditionUpgrade(ing) is true till the cluster is back up and everything was successfully updated.

`ConditionUpgrade` is true until the original StatefulSet has been deleted, then it is set to `false` and the reconcile request requeued in order to recreate the new StatefulSet. I don't think we should change the name or semantics of this condition at this stage, as users could be depending on this logic via `kubectl  wait --for condition=`.

ryanemerson:
+1 To moving the validation checks to the StatefulSet in the future.

pkg/controller/infinispan/resources/config/config_controller.go
Crumby: >  I don't think we should change the name or semantics of this condition at this stage

I agree. Just wanted to point out that instinctively I'd expect the condition to behave differently. Upgrade isn't really finished by deleting resources.

PR #1163, opened 2021-06-24
[2.0.x] Support for static XSite configuration #1156

### LINKED ISSUE #1156: Backport support for static XSite configuration to 2.0.x
I'm currently running Keycloak 14.0.0 in cross-datacenter replication mode with Infinispan 11.0.9, which is the latest [recommended](https://www.keycloak.org/docs/latest/server_installation/#setting-up-cross-dc-with-infinispan-11-0-9) version, configured as a remote cache with cross-site replication enabled. For this I'm using Infinispan Operator 2.0.x, which is the latest version that supports Infinispan 11 according to the [documentation](https://infinispan.org/docs/infinispan-operator/2.0.x/operator.html).

Instead of using Kubernetes/OpenShift tokens I want to use static URIs, which is not supported until later releases of the operator. The reason for this is that static URIs are easier to configure (may be subjective) and that it will allow me to use a separate ingress controller.

ericot:
This makes it possible to use the `infinispan+xsite` scheme when configuring XSite locations. It's based primarily on the changes done in commit [33e718c](https://github.com/infinispan/infinispan-operator/commit/33e718caa244d7d98e888edcbee7ed6f68aa3f9f).

Fixes #1156

### DIFF
--- a/deploy/cr/xsite-static-loadbalancer/xsite_a.yaml
+++ b/deploy/cr/xsite-static-loadbalancer/xsite_a.yaml
@@ -1,0 +1,20 @@
+apiVersion: infinispan.org/v1
+kind: Infinispan
+metadata:
+  name: example-infinispan
+spec:
+  replicas: 1
+  expose:
+    type: LoadBalancer
+  service:
+    type: DataGrid
+    sites:
+      local:
+        name: SiteA
+        expose:
+          type: LoadBalancer
+      locations:
+      - name: SiteA
+        url: infinispan+xsite://infinispan-sitea.myhost.com:7900
+      - name: SiteB
+        url: infinispan+xsite://infinispan-siteb.myhost.com:7900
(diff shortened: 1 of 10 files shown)


rigazilla:
I can't see any drawback in backporting this.
@dmvolod , @ryanemerson wdyt?

dmvolod:
> I can't see any drawback in backporting this.

@rigazilla, yes, agreed. Also this PR doesn't pass core tests and we don't have cross-site tests for the 2.0.x and need to test manually and carefully.
Original logic was introduced in this commit https://github.com/infinispan/infinispan-operator/commit/69c8f16f9f2720527ec4c8c6216a9f5f2656361a and than modified in https://github.com/infinispan/infinispan-operator/commit/33e718caa244d7d98e888edcbee7ed6f68aa3f9f

ryanemerson:
Closing as the user has upgraded operator version instead.

PR #1167, opened 2021-06-28
Print info in operator logs when cluster is well formed

varsharain-a11y:
Tried to add " well formed" info in logs by making changes to pkg/controller/infinispan/infinispan_controller.go.


ryanemerson:
@varsharain-a11y Thanks for creating a PR.

In this case I don't think we want to print a "WellFormed" log to the operator at the INFO level. This is a cluster specific log that will not scale well when the operator is managing many clusters.

I think having a cluster specific log like this is desirable at the trace or debug level, however in order for that to happen we first need to refactor the Operator logging https://github.com/infinispan/infinispan-operator/issues/690.

varsharain-a11y:
@ryanemerson thanks for pointing it out. Agree with you, "WellFormed" cluster specific logging at Trace/debug level sounds better. .

ryanemerson:
Closing for now. The WellFormed log will be added as part of #690

PR #1174, opened 2021-07-02
[2.1.x] Operator upgrades from 2.0.x -> 2.1.x fail on restart

ryanemerson:
To test scenario do:

1. Fetch the latest from this branch `sfs_uprade_issues`
2. Fetch the following tags `2.0.6` `2.1.3.Final`
3. Create `infinispan.yaml` and `identities.yaml` in the operator root
4. Execute [olm-simulator.sh](https://gist.github.com/ryanemerson/a17b208c12a5c91a938f499b24d2691b)

infinispan.yaml:
```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: infinispan
spec:
  container:
    memory: 1Gi
  replicas: 2
  security:
    endpointSecretName: auth-secret
  service:
    type: DataGrid
```

identities.yaml
```
credentials:
- username: testuser
  password: testpass
- username: operator
  password: supersecretoperatorpass
```

I still need to add some logic to initiate state-transfer on xsite caches once the upgrade has completed.

For some reason OOM are happening when migrating the server event log .dat file. I'm still investigating this, but strangely when I extract the corrupted .dat file from the Openshift pod, it can be recovered on my local machine using much more restrictive memory settings :shrug: 

Output of OOM:
```
13:06:30,330 INFO  (blocking-thread--p3-t1) [org.infinispan.PERSISTENCE] ISPN000642: Attempting to recover '___event_log_cache' corrupt data ...
java.lang.OutOfMemoryError: Java heap space
Dumping heap to java_pid91.hprof ...
Heap dump file created [46847017 bytes in 0.397 secs]
Terminating due to java.lang.OutOfMemoryError: Java heap space
```

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -491,6 +491,19 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+		// ISPN-13116 If xsite view has been formed, then we must perform state-transfer to all sites if a SFS recovery has occurred
+		if crossSiteViewCondition.Status == metav1.ConditionTrue {
+			podName := podList.Items[0].Name
+			logs, err := kubernetes.Logs(podName, infinispan.Namespace)
+			if err != nil {
+				log.Error(err, fmt.Sprintf("Unable to retrive logs for infinispan pod %s", podName))
+			}
+			if strings.Contains(logs, "ISPN000643") {
+				if err := cluster.XsitePushAllState(podName); err != nil {
+					log.Error(err, "Unable to push xsite state after SFS data recovery")
+				}
+			}
+		}
 		err = r.update(infinispan, func() {
 			infinispan.SetConditions([]infinispanv1.InfinispanCondition{*crossSiteViewCondition})
 		})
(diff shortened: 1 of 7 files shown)


ryanemerson:
@rigazilla @dmvolod @Crumby Any thoughts on what could be happening in k8s for the memory to be even more restrictive than `-Xmx` suggests?

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1307,15 +1307,26 @@ func (r *ReconcileInfinispan) gracefulShutdownReq(ispn *infinispanv1.Infinispan,
...
 			if err != nil {
-				logger.Error(err, "failed to exec shutdown command on pod")
-				continue
+				logger.Error(err, "Error encountered on cluster shutdown")
+
+				// ISPN-13141 causes GracefulShutdown to fail if there are issues with the server
+				logger.Info("Cluster Shutdown failed. Attempting to execute GracefulShutdownTask")
+				if err := cluster.GracefulShutdownTask(pod.GetName()); err != nil {
+					logger.Error(err, "Error encountered on GracefulShutdownTask")
+				}
```
rigazilla, line 1321: if `GracefulShutdownTask` fails and then all the pods are deleted, the js is not successfully executed on the cluster. Is this ok?

ryanemerson: Hmm good point. I will update this so that it tries executing the task on the next ready pod. If the task fails on all pods, then I think we have to proceed with deleting the pods and updating the condition so that the upgrade can continue.

Previously if `GracefulShutdown` failed, we return `nil` which resulted in a strange state where the wrong conditions aren't in place and the upgrade can't proceed or recover.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -1305,28 +1318,42 @@ func (r *ReconcileInfinispan) reconcileGracefulShutdown(ispn *infinispanv1.Infin
...
 		if kube.IsPodReady(pod) {
-			err := cluster.GracefulShutdown(pod.GetName())
-			if err != nil {
-				logger.Error(err, "failed to exec shutdown command on pod")
-				continue
+			if err := cluster.GracefulShutdown(pod.GetName()); err != nil {
+				logger.Error(err, "Error encountered on cluster shutdown")
+
+				// ISPN-13141 causes GracefulShutdown to fail if there are issues with the server
+				logger.Info("Cluster Shutdown failed. Attempting to execute GracefulShutdownTask")
```
dmvolod, line 1329: @ryanemerson I'm not sure, but probably we need to execute shutdown via task in specific case only, not for every error. As error could say about communication problem, not server.

pkg/infinispan/cluster.go
```diff
@@ -114,6 +116,55 @@ func (c Cluster) GracefulShutdown(podName string) error {
...
+			 print("Shutting down cache " + name);
+			 try {
+				cacheManager.getCache(name).shutdown();
+			 } catch (err) {
+				stdErr.println("Encountered error trying to shutdown cache " + name + ": " + err);
+			 }
+	   }
+	}
+	`
+	// Remove all new lines to prevent a "100 continue" response
```
dmvolod, line 156: Probably utilizing go-rice could help, but make sense if this solution will be forever with us

ryanemerson: [ISPN-13141](https://issues.redhat.com/browse/ISPN-13141) will be fixed on the server-side, so this shouldn't be around for too long. Either way, we have to remove the new lines as we're sending the script via a curl string.

pkg/infinispan/cluster.go
```diff
@@ -314,6 +365,40 @@ func (c Cluster) SetLogger(podName, loggerName, loggerLevel string) error {
 	return nil
 }
```
dmvolod, line 367: It's not clear for me, probably i missed something. Is the second site should be updated already or not?

pkg/controller/infinispan/infinispan_controller.go
ryanemerson: Ideally yes, but the problem is identifying when we should/shouldn't execute the task. If the error is caused by a communication problem, then `GracefulShutdownTask` will also fail, so we're only executing a single http request unnecessarily.

IMO I think it's better to have the additional HTTP request instead of risking getting the conditional check wrong and not executing  `GracefulShutdownTask` when it's required.

pkg/infinispan/cluster.go
ryanemerson: The second site will push it's state if on restart it detects that it's SFS state has been corrupted and recovered. The operator is only concerned with pushing the state of the site it's reconciling.

dmvolod:
Looks good, but didn't get a chance to test on OCP instance
Will wonder if we can add e2e test to upstream infra for this issue

ryanemerson:
> Will wonder if we can add e2e test to upstream infra for this issue

I'm going to spend some time this week moving the upstream operator use OLM bundles. It should the be possible for us to add some of downstream tests to this repo.

> Looks good, but didn't get a chance to test on OCP instance

Unfortunately we've ran out of Travis credits as well. I'm just running the full testsuite against Openshift atm.

dmvolod (approved):

PR #1183, opened 2021-07-09
[2.1.x] Sfs upgrade issues part 3

ryanemerson:


pkg/controller/infinispan/client.go
```diff
@@ -12,12 +12,23 @@ import (
...
 )
 
+// Create a cluster for connecting to pods created on the 2.0.x channel
+func NewLegacyCluster(i *v1.Infinispan, kubernetes *kube.Kubernetes) (*ispn.Cluster, error) {
+	// Use the user secret in order to obtain the old credentials used by the legacy server
+	pass, err := users.AdminPassword(i.GetSecretName(), i.Namespace, kubernetes)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to retrieve identities when creating legacy Cluster instance: %w", err)
+	}
+	return ispn.NewCluster(consts.DefaultOperatorUser, pass, i.Namespace, i.GetEndpointScheme(), 11222, kubernetes), nil
```
dmvolod, line 22: We have constant for the 11222 port

dmvolod (approved):

pkg/controller/infinispan/client.go
ryanemerson: I deliberately avoided using the constant because any future updates to this value will result in the legacy behaviour being broken.

dmvolod: Probably we can introduce another constant for better visibility

ryanemerson: The code is only used in one place, so unless we need it elsewhere I think a constant is overkill.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -999,6 +999,20 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 		reqLogger.Info(errMsg)
 	}
 
+	// ISPN-12667 ISPN-13121 If upgrading the cluster ensure that cache .state required properties are added
+	if m.IsConditionTrue(infinispanv1.ConditionGracefulShutdown) {
+		c := &spec.InitContainers
+		*c = append(*c, corev1.Container{
+			Image:   consts.InitContainerImageName,
+			Name:    DataMountVolume,
```
dmvolod, line 1007: Probably we need a specific init container name for this case

dmvolod, line 1007: This should be container name, not the volume name

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -999,6 +999,20 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
...
+			Command: []string{"sh", "-c", "for f in *.state; do (grep -q 'ReplicatedConsistentHash' $f && grep -qv 'capacityFactors' $f) && echo 'capacityFactors=1' >> $f; done"},
+			VolumeMounts: []corev1.VolumeMount{{
+				Name:      "cache-state-repair",
```
dmvolod, line 1010: This is volume name to be mount

PR #1186, opened 2021-07-12
Sfs upgrade issues

ryanemerson:
Forward port of the various upgrade issues already merged on the 2.1.x branch.

### DIFF
--- a/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
+++ b/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
@@ -339,6 +339,7 @@ spec:
                 - update
                 - delete
                 - patch
+                - deletecollection
             - apiGroups:
                 - ""
               resources:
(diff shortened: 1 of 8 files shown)


deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -78,6 +78,7 @@ metadata:
     createdAt: 2021-01-26T09:00:00Z
     support: Infinispan
     certified: "false"
+    olm.skipRange: ">=2.1.0 <2.1.5"
```
Crumby, line 81: I'm not sure this makes sense for master as master should be `2.2.0` potentially replacing `2.1.x` (5)? and `2.1.5` should skip `2.1.0 - 2.1.4`. Also looking at the version name of the Operator in line bellow it should skip itself.

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -980,8 +1006,43 @@ func (r *ReconcileInfinispan) statefulSetForInfinispan(m *infinispanv1.Infinispa
 	if _, err := applyExternalArtifactsDownload(m, &dep.Spec.Template.Spec); err != nil {
 		return nil, err
 	}
-
 	applyExternalDependenciesVolume(m, &dep.Spec.Template.Spec)
+
+	// ISPN-12667 ISPN-13121 If upgrading the cluster ensure that cache .state required properties are added
+	if m.IsConditionTrue(infinispanv1.ConditionGracefulShutdown) {
```
dmvolod, line 1012: @ryanemerson do we need update init container in the main branch? if we will not skip previous operator version it will update to the correct state. Or I'm wrong?

ryanemerson: That's true. It also shouldn't be necessary to have the check for the 11223 port if we make this assumption. I'll update to remove these two commits.

dmvolod (approved):

PR #1187, opened 2021-07-13
Exposed port for LoadBalancer services (application and xsite) should be configurable

dmvolod:


pkg/controller/infinispan/resources/service/service_controller.go
```diff
@@ -416,6 +427,9 @@ func computeSiteService(ispn *ispnv1.Infinispan) *corev1.Service {
 				TargetPort: intstr.IntOrString{IntVal: consts.CrossSitePort},
 			},
 		}
+		if exposeConf.Port > 0 {
```
pruivo, line 430: shouldn't we do this for all `expose types`? `NodePort` and `ClusterIP` has the field `Port` too

dmvolod: For `NodePort` it doesn't make sense as external communication is performing via `NodePort` on all hosts. For `ClusterIP` it make sense but can guess situation why customer need to change port for internal k8s cluster communication.

pruivo: how about the same cluster cross-site (i.e. different namespaces/cluster names)?
 isn't this `port` be used independently of the expose type?

dmvolod: Yes, but for the same cluster service name is unique and port doesn't matter in this case as we are operating with service:port combination

test/e2e/xsite/xsite_test.go
```diff
@@ -99,28 +100,32 @@ func secretSiteName(siteName string) string {
...
 }
 
 func TestCrossSiteViewOpenshiftLoadBalancer(t *testing.T) {
-	testCrossSiteView(t, true, config.SchemeTypeOpenShift, ispnv1.CrossSiteExposeTypeLoadBalancer)
+	testCrossSiteView(t, true, config.SchemeTypeOpenShift, ispnv1.CrossSiteExposeTypeLoadBalancer, 0)
 }
 
-func testCrossSiteView(t *testing.T, isMultiCluster bool, schemeType string, exposeType ispnv1.CrossSiteExposeType) {
+func TestCrossSiteViewLoadBalancerWithPort(t *testing.T) {
+	testCrossSiteView(t, true, config.SchemeTypeOpenShift, ispnv1.CrossSiteExposeTypeLoadBalancer, 1443)
```
dmvolod, line 125: @pruivo , @ryanemerson please pay attention that CI and my local machine unable to test services on privileged ports (less than 1000). Probably it's an MetalLB limitations. but need to be tested on Azure or AWS LoadBalancer

pruivo: I did a test yesterday in my laptop with port 443 and it worked fine (kind 0.11.1 + MetalLB v0.10.2)
for testing purposes, we just need the port to be different from 7900

pruivo (approved):

ryanemerson (approved):

test/e2e/xsite/xsite_test.go
dmvolod: Thanks for confirmation. Will try with newer MetalLB version

PR #1194, opened 2021-07-14
[2.1.x] Exposed port for LoadBalancer services (application and xsite) should be configurable

dmvolod:


pruivo (approved):

rigazilla:
merged, thanks @dmvolod !

PR #1201, opened 2021-07-21
Cross Site  - Gossip Router tunnel support

pruivo:
Adds support to use JGroups' TUNNEL protocol with GossipRouter pods.
It uses the JGroups JAR inside the server image to avoid depending on another image.

PR is open to get feedback sooner than later. No rush to integrate yet.

TODO:

- [ ] Create an entry point in the server image (avoid doing "java -jar ...")
- [ ] Handle server image upgrades (?)

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -81,6 +81,7 @@ rules:
   - apps
   resources:
   - statefulsets
+  - deployments
   - deployments/finalizers
   verbs:
   - get
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -261,6 +261,32 @@ func (r *ReconcileInfinispan) Reconcile(request reconcile.Request) (reconcile.Re
 		}
 	}
 
+	if infinispan.HasSites() {
+        tunnel := &appsv1.Deployment{}
+		err = r.client.Get(context.TODO(),types.NamespacedName{Namespace: infinispan.Namespace, Name: infinispan.Name}, tunnel)
+		if err != nil && errors.IsNotFound(err) {
+			reqLogger.Info("Configuring the Cross-Site Deployment")
+			deployment, err := r.createGossipRouterTunnel(infinispan)
+			if err != nil {
+				reqLogger.Error(err, "failed to configure new Deployment")
+				return reconcile.Result{}, err
+			}
+			reqLogger.Info("Creating a new Deployment", "Deployment.Name", deployment.Name)
+			err = r.client.Create(context.TODO(), deployment)
+			if err != nil {
+				reqLogger.Error(err, "failed to create new Deployment", "Deployment.Name", deployment.Name)
+				return reconcile.Result{}, err
+			}
+
+			// StatefulSet created successfully
+			reqLogger.Info("End of the Cross-Site tunneling Deployment creation")
+		}
+		if err != nil {
+			reqLogger.Error(err, "failed to get Cross-Site Deployment")
+			return reconcile.Result{}, err
+		}
+	}
+
 	// Reconcile the StatefulSet
 	// Check if the StatefulSet already exists, if not create a new one
 	statefulSet := &appsv1.StatefulSet{}
@@ -1547,6 +1573,10 @@ func PodLabels(name string) map[string]string {
 	return LabelsResource(name, "infinispan-pod")
 }
 
+func TunnelPodLabels(name string) map[string]string {
+	return LabelsResource(name, "infinispan-tunnel-pod")
+}
+
 func ServiceLabels(name string) map[string]string {
 	return map[string]string{
 		"clusterName": name,
@@ -1605,18 +1635,21 @@ func remove(list []string, s string) []string {
 }
 
 // returns the tunnel service
-func (r *ReconcileInfinispan) createGossipRouterTunnel(siteServiceName string, m *infinispanv1.Infinispan, logger logr.Logger) (*corev1.Service, error) {
-	lsService := ispncom.LabelsResource(m.ObjectMeta.Name, "infinispan-tunnel-service")
-	lsTunnel := ispncom.LabelsResource(m.ObjectMeta.Name, "infinispan-tunnel")
+func (r *ReconcileInfinispan) createGossipRouterTunnel(m *infinispanv1.Infinispan) (*appsv1.Deployment, error) {
+	name := m.Name + "-tunnel"
+	reqLogger := log.WithValues("Request.Namespace", m.Namespace, "Request.Name", name)
+	//TODO labels
+	lsTunnel := TunnelPodLabels(m.Name)
 
-	logger.Info("Creating GossipRouter (xsite-tunnel) Deployment")
+	// TODO where to create it?
+	reqLogger.Info("Creating GossipRouter (xsite-tunnel) Deployment")
 	deployment := &appsv1.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: "v1",
 			Kind:       "Deployment",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      m.ObjectMeta.Name + "-tunnel",
+			Name:      name,
 			Namespace: m.ObjectMeta.Namespace,
 			Labels:    lsTunnel,
 		},
@@ -1637,7 +1670,7 @@ func (r *ReconcileInfinispan) createGossipRouterTunnel(siteServiceName string, m
 						Ports: []corev1.ContainerPort{
 							{ContainerPort: 8787, Name: "debug", Protocol: corev1.ProtocolTCP},
 							{ContainerPort: 9000, Name: "netcat", Protocol: corev1.ProtocolTCP},
-							{ContainerPort: 12001, Name: "tunnel", Protocol: corev1.ProtocolTCP},
+							{ContainerPort: consts.CrossSitePort, Name: "tunnel", Protocol: corev1.ProtocolTCP},
 						},
 					}},
 				},
@@ -1645,41 +1678,8 @@ func (r *ReconcileInfinispan) createGossipRouterTunnel(siteServiceName string, m
 		},
 	}
 	// Set Infinispan instance as the owner and controller
-	controllerutil.SetControllerReference(m, deployment, r.scheme)
-
-	err := r.client.Create(context.TODO(), deployment)
-	if err != nil && !errors.IsAlreadyExists(err) {
+	if err := controllerutil.SetControllerReference(m, deployment, r.scheme); err != nil {
 		return nil, err
 	}
-
-	logger.Info("Creating GossipRouter (xsite-tunnel) Service")
-	service := &corev1.Service{
-		TypeMeta: metav1.TypeMeta{
-			APIVersion: "v1",
-			Kind:       "Service",
-		},
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      siteServiceName,
-			Namespace: m.ObjectMeta.Namespace,
-			Labels:    lsService,
-		},
-		// TODO! customize spec.type
-		Spec: corev1.ServiceSpec{
-			Type:     m.Spec.Service.Sites.Local.Expose.Type,
-			Selector: lsTunnel,
-			Ports: []corev1.ServicePort{
-				{Port: 12001, Name: "tunnel", Protocol: corev1.ProtocolTCP},
-			},
-		},
-	}
-
-	// Set Infinispan instance as the owner and controller
-	controllerutil.SetControllerReference(m, service, r.scheme)
-
-	err = r.client.Create(context.TODO(), service)
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return nil, err
-	}
-
-	return service, nil
+	return deployment, nil
 }
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -1603,3 +1603,83 @@ func remove(list []string, s string) []string {
 	}
 	return slice
 }
+
+// returns the tunnel service
+func (r *ReconcileInfinispan) createGossipRouterTunnel(siteServiceName string, m *infinispanv1.Infinispan, logger logr.Logger) (*corev1.Service, error) {
+	lsService := ispncom.LabelsResource(m.ObjectMeta.Name, "infinispan-tunnel-service")
+	lsTunnel := ispncom.LabelsResource(m.ObjectMeta.Name, "infinispan-tunnel")
+
+	logger.Info("Creating GossipRouter (xsite-tunnel) Deployment")
+	deployment := &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Deployment",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      m.ObjectMeta.Name + "-tunnel",
+			Namespace: m.ObjectMeta.Namespace,
+			Labels:    lsTunnel,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: lsTunnel,
+			},
+			Template: corev1.PodTemplateSpec{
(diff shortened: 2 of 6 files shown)


pruivo:
depends on 
* https://github.com/infinispan/infinispan-operator/pull/1238
* https://github.com/infinispan/infinispan-operator/pull/1252

controllers/infinispan_controller.go
```diff
@@ -180,7 +180,7 @@ func (r *InfinispanReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // +kubebuilder:rbac:groups=core,resources=pods/exec,verbs=create
 // +kubebuilder:rbac:groups=core;events.k8s.io,resources=events,verbs=create;patch
 
-// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=create;get;list;watch
```
dmvolod, line 183: Seems to delete rights is missing

pruivo: right! added `delete` and `update` permissions.

controllers/infinispan_controller.go
```diff
@@ -290,6 +290,32 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 		}
 	}
 
+	if infinispan.HasSites() {
+		tunnel := &appsv1.Deployment{}
```
dmvolod, line 294: Whole get/create/update cycle could be replaced with controllerutil.CreateOrUpdate function

controllers/infinispan_service_controller.go
```diff
@@ -466,8 +465,7 @@ func computeServiceExternal(ispn *ispnv1.Infinispan) *corev1.Service {
 
 // computeSiteService compute the XSite service
 func computeSiteService(ispn *ispnv1.Infinispan) *corev1.Service {
-	lsPodSelector := PodLabels(ispn.Name)
-	lsPodSelector[consts.CoordinatorPodLabel] = strconv.FormatBool(true)
```
dmvolod, line 470: @pruivo coordinator pod labeling seems to be removed also

dmvolod:
Also we need to watch Deployment resources to catch unpredictable resource remove/delete/update

controllers/infinispan_service_controller.go
pruivo: it is not required for cross-site.

pruivo:
> Also we need to watch Deployment resources to catch unpredictable resource remove/delete/update

I wasn't aware of that. Is there any example in the code?

dmvolod:
> I wasn't aware of that. Is there any example in the code?

You should add `appsv1.Deployment` to the `secondaryResourceTypes` list
https://github.com/infinispan/infinispan-operator/blob/6381d8b0fd4aca538fab2c19c97a40d6374a19b9/controllers/infinispan_controller.go#L126

controllers/infinispan_service_controller.go
dmvolod: I talked about to remove this logic. Do we really need it after GossipRouter Implementation?
https://github.com/infinispan/infinispan-operator/blob/6381d8b0fd4aca538fab2c19c97a40d6374a19b9/controllers/xsite.go#L17-L26

controllers/infinispan_controller.go
```diff
@@ -290,6 +290,32 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
+		err = r.Client.Get(context.TODO(), types.NamespacedName{Namespace: infinispan.Namespace, Name: infinispan.GetGossipRouterDeploymentName()}, tunnel)
+		if err != nil && errors.IsNotFound(err) {
+			reqLogger.Info("Configuring the Cross-Site Deployment")
+			tunnel, err = r.createGossipRouterTunnel(infinispan)
+			if err != nil {
+				reqLogger.Error(err, "failed to configure new Deployment")
+				return reconcile.Result{}, err
+			}
+			reqLogger.Info("Creating a new Deployment", "Deployment.Name", tunnel.Name)
+			err = r.Client.Create(context.TODO(), tunnel)
```
dmvolod, line 304: @pruivo another potential issue, if something happens in the cluster, i.e. no resources or anything else, customers never know that Deployment not created correctly. For cluster Pods controller eventually request data from the Pod status and log it into the Infinispan CR status. Seems to need to implement similar logic for the GossipRouter Deployment

controllers/infinispan_service_controller.go
pruivo: right! I've removed it.

controllers/xsite.go
```diff
@@ -5,25 +5,21 @@ import (
...
 	ispnv1 "github.com/infinispan/infinispan-operator/api/v1"
 	consts "github.com/infinispan/infinispan-operator/controllers/constants"
 	ispn "github.com/infinispan/infinispan-operator/pkg/infinispan"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 func (r *infinispanRequest) applyLabelsToCoordinatorsPod(podList *corev1.PodList, siteLocations []string, cluster ispn.ClusterInterface) (*ispnv1.InfinispanCondition, error) {
```
dmvolod, line 18: This function should be named related to the task it does now, like 'getCrossSiteViewCondition'

pruivo:
updated the code to use `controllerutil.CreateOrUpdate()`.

Also, just crossed my mind, I should add some readiness and liveness  probes to GossipRouter pod.

pruivo:
code updated. 
Added probes
Added condition "GossipRouterReady" example
```yaml
Status:
  Conditions:
    Status:   True
    Type:     PreliminaryChecksPassed
    Status:   True
    Type:     GossipRouterReady
    Message:  View: test-cross-site-view-internal-xsite2-0-63738
    Status:   True
    Type:     WellFormed
    Message:  Cross-Site view: xsite1,xsite2
    Status:   True
    Type:     CrossSiteViewFormed
  ```

File: controllers/infinispan_controller.go [hunk 6]
```diff
@@ -525,7 +556,7 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 
 	// If x-site enable configure the coordinator pods to be selected by the x-site service
```
dmvolod, line 557: comment also need to changed

pruivo:
updated.
Added: shutting down the GossipRouter pod (set Deployment.Spec.Replicas=0) if Infinispan.Spec.Replicas == 0

controllers/xsite.go
```diff
@@ -43,18 +39,64 @@ func (r *infinispanRequest) applyLabelsToCoordinatorsPod(podList *corev1.PodList
...
+			Namespace: m.ObjectMeta.Namespace,
+			Labels:    lsTunnel,
+		},
+		Spec: appsv1.DeploymentSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: lsTunnel,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      m.ObjectMeta.Name,
```
dmvolod, line 75: `ObjectMeta` can be omitted

controllers/infinispan_controller.go
```diff
@@ -290,6 +290,37 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
+		reqLogger.Info("Creating the Cross-Site Deployment (Gossip Router)")
+		tunnel := r.GetGossipRouterDeployment(infinispan)
+		_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, tunnel, func() error {
+			return controllerutil.SetControllerReference(r.infinispan, tunnel, r.scheme)
+		})
+		if err != nil {
+			reqLogger.Error(err, "Failed to configure Cross-Site Deployment")
+			return reconcile.Result{}, err
+		}
+		reqLogger.Info("Cross-Site Deployment created")
```
dmvolod, line 303: @pruivo you will always print  "Cross-Site Deployment created" on each reconcile loop whenever it created or updated
second parameter of the `controllerutil.CreateOrUpdate` returns performed operation

controllers/infinispan_controller.go
```diff
@@ -290,6 +290,37 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 		}
 	}
 
+	if infinispan.HasSites() {
+		reqLogger.Info("Creating the Cross-Site Deployment (Gossip Router)")
+		tunnel := r.GetGossipRouterDeployment(infinispan)
+		_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, tunnel, func() error {
+			return controllerutil.SetControllerReference(r.infinispan, tunnel, r.scheme)
```
dmvolod, line 297: You just update controller reference in the mutate function, not required Deployment parts.
Please have a look how others controllers (service/config, etc.) performing updates for mutating fields

controllers/xsite.go
```diff
@@ -43,18 +39,64 @@ func (r *infinispanRequest) applyLabelsToCoordinatorsPod(podList *corev1.PodList
...
+	replicas := int32(1)
+
+	// if the user configures 0 replicas, shutdown the gossip router pod too.
+	if m.Spec.Replicas <= 0 {
+		replicas = 0
+	}
+
+	deployment := &appsv1.Deployment{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
```
dmvolod, line 61:
```suggestion
			APIVersion: "apps/v1",
```

controllers/xsite.go
```diff
@@ -43,18 +39,64 @@ func (r *infinispanRequest) applyLabelsToCoordinatorsPod(podList *corev1.PodList
...
+			Kind:       "Deployment",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: m.ObjectMeta.Namespace,
```
dmvolod, line 66:
```suggestion
			Namespace: m.Namespace,
```

controllers/xsite.go
```diff
@@ -43,18 +39,64 @@ func (r *infinispanRequest) applyLabelsToCoordinatorsPod(podList *corev1.PodList
...
+	name := m.GetGossipRouterDeploymentName()
+	lsTunnel := GossipRouterPodLabels(m.Name)
+	replicas := int32(1)
+
+	// if the user configures 0 replicas, shutdown the gossip router pod too.
+	if m.Spec.Replicas <= 0 {
+		replicas = 0
+	}
+
+	deployment := &appsv1.Deployment{
```
dmvolod, line 59: @pruivo what do you think about `gossiprouter` sidecar, not separate deployment? Sidecar shares network/storage with the Infinispan container, but number of the sidecars will be the same as cluster size.

controllers/infinispan_controller.go
```diff
@@ -290,6 +290,37 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 		}
 	}
 
+	if infinispan.HasSites() {
```
dmvolod, line 293: If someone remove `sites` section from the CR, Deployment will keep forever. Seems we need validate it and cleanup if needed or set `sites` section as immutable in the webhook controller (not implemented now)

controllers/infinispan_controller.go
pruivo: I have no idea what to do TBH. 
The service seems to copy every field manually (which seems error-prone to me). Configs, secrets, batch just invoke `SetControllerReference()`.

controllers/xsite.go
pruivo: There is no advantage of having more than one GossipRouter. I would have to tag a single pod to be used with the exposed service and that pod will be hit with all the traffic from both sites (and it will negate the multi site master advantage).
If that's ok for you, I would rather a the gossip router in its own pod.

controllers/infinispan_controller.go
pruivo: where can it be set immutable? 

Removing the deployment wouldn't help since JGroups and Infinispan are still configured to use cross-site replication. I guess it would be easier to set the field immutable than restarting and reconfiguring the cluster. 
Do you agree?

controllers/xsite.go
dmvolod: Yes, sure. The sidecar additional value is a tightly coupling within the server Pods. If it's not so important, we could live it as is.

controllers/infinispan_controller.go
dmvolod: Seem to if someone remove `sites` structure, configuration ConfgiMap will be changed and StatefulSet will be rollout to the new config without cross-site setup. Is'n it?

controllers/infinispan_controller.go
dmvolod: The code should like this
```
		tunnelDeployment := &appsv1.Deployment{
			ObjectMeta: metav1.ObjectMeta{
				Name:      infinispan.Name,
				Namespace: infinispan.Namespace,
			},
		}
		result, err := controllerutil.CreateOrUpdate(r.ctx, r.Client, tunnelDeployment, func() error {
			tunnel := r.GetGossipRouterDeployment(infinispan)
			tunnelDeployment.Spec = tunnel.Spec
			tunnelDeployment.Labels = tunnel.Labels
			if tunnelDeployment.CreationTimestamp.IsZero() {
				reqLogger.Info("Creating the Cross-Site Deployment (Gossip Router)")
				return controllerutil.SetControllerReference(r.infinispan, tunnelDeployment, r.scheme)
			}
			return nil
		})
		if err != nil {
			if errors.IsConflict(err) {
				return reconcile.Result{Requeue: true}, nil
			} else {
				reqLogger.Error(err, "Failed to configure Cross-Site Deployment")
			}
		}
		if result == controllerutil.OperationResultCreated {
			reqLogger.Info("Cross-Site Deployment created")
		}
```

controllers/xsite.go
```diff
@@ -43,18 +39,64 @@ func (r *infinispanRequest) applyLabelsToCoordinatorsPod(podList *corev1.PodList
...
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "v1",
+			Kind:       "Deployment",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
```
dmvolod, line 65: Probably we need  a specific name for GossipRouter deployment, not the same as cluster name, i.e. with `tunnel` prefix or postfix

pruivo: name is `name := m.GetGossipRouterDeploymentName()`
`-tunnel` is added to the cluster name.

pruivo: actually, I don't need the variable `name`. I'll remove it to avoid confusion

controllers/infinispan_controller.go
pruivo: I'll test this locally to see what the operator does and acting accordingly. thanks!

controllers/infinispan_controller.go
```diff
@@ -290,6 +290,55 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
+		reqLogger.Info("Creating the Cross-Site Deployment (Gossip Router)")
+		tunnelDeployment := &appsv1.Deployment{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      infinispan.Name,
```
dmvolod, line 297:
```suggestion
				Name:      infinispan.GetGossipRouterDeploymentName(),
```

controllers/infinispan_controller.go
pruivo: @dmvolod you are right! 
I've updated the code to remove the `-site` `service` and the `deployment`.

ryanemerson:
If you're happy with the changes @dmvolod I will merge

dmvolod (approved):

ryanemerson:
Thanks  @pruivo

ryanemerson:
> Also Rolling Upgrade PR need to be adopted for GossipRouter deployment if it will be supported

Good point @dmvolod. FYI @gustavonalle

PR #1207, opened 2021-07-29
Watching for the Secrets resources are not owned by the Infinispan CR

### LINKED ISSUE #1206: Watching for the Secrets resources are not owned by the Infinispan CR
There are possible to use custom created resources, i.e. Secrets which are not owned by the Infinispan CR.
Currently, the custom defined Secrets are updated with Infinispan CR as a Controller OwnerReference. This solution works fine for the triggering changes but will cause an unexpected object deletion when Infinispan CR has deleted.

We need to implement `EnqueueRequestsFromMapFunc` for the `Secret` which are not autogenerated and don't have Infinispan CR as owner reference.
https://github.com/kubernetes-sigs/controller-runtime/blob/release-0.4/pkg/handler/enqueue_mapped.go

Probably, we need to migrate owner references removing it from the Secret. Need to investigate more.
Secret names should be stored in the labels also for the indexing get list of the Infinispan CR's or we need to add required fields to the index for caching.

dmvolod:
fixes #1206 

- [x]  TODO: Migration need to be tested

### DIFF
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -126,6 +126,23 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	// Add Secret name fields to the index for caching
+	if err = mgr.GetFieldIndexer().IndexField(&infinispanv1.Infinispan{}, "spec.security.endpointSecretName", func(obj runtime.Object) []string {
+		return []string{obj.(*infinispanv1.Infinispan).GetSecretName()}
+	}); err != nil {
+		return err
+	}
+	if err = mgr.GetFieldIndexer().IndexField(&infinispanv1.Infinispan{}, "spec.security.endpointEncryption.certSecretName", func(obj runtime.Object) []string {
+		return []string{obj.(*infinispanv1.Infinispan).GetKeystoreSecretName()}
+	}); err != nil {
+		return err
+	}
+	if err = mgr.GetFieldIndexer().IndexField(&infinispanv1.Infinispan{}, "spec.security.endpointEncryption.clientCertSecretName", func(obj runtime.Object) []string {
+		return []string{obj.(*infinispanv1.Infinispan).GetTruststoreSecretName()}
+	}); err != nil {
+		return err
+	}
+
 	// Watch for changes to primary resource Infinispan
 	err = c.Watch(&source.Kind{Type: &infinispanv1.Infinispan{}}, &handler.EnqueueRequestForObject{})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -81,17 +81,13 @@ func Add(mgr manager.Manager) error {
 
 func (s *secretResource) Process() (reconcile.Result, error) {
 	// Reconcile Encryption Secrets
-	if result, err := s.reconcileKeystoreSecret(); result != nil {
-		return *result, err
-	}
-
```
rigazilla, line 87: it's not clear to me why we don't need this any more

rigazilla (commented):
added some questions

pkg/controller/infinispan/resources/secret/secret_controller.go
dmvolod: This function just set owner reference in the code, nothing more

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -704,6 +755,21 @@ func (r *ReconcileInfinispan) upgradeInfinispan(infinispan *infinispanv1.Infinis
...
 }
 
+func (r *ReconcileInfinispan) dropSecretOwnerReference(secretName string, infinispan *infinispanv1.Infinispan, logger logr.Logger) error {
+	secret := &corev1.Secret{}
+	if result, err := kube.LookupResource(secretName, infinispan.Namespace, secret, r.client, logger, eventRec); result != nil {
+		return err
+	}
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, secret, func() error {
+		creationTimestamp := secret.GetCreationTimestamp()
+		if !creationTimestamp.IsZero() {
```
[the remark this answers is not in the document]
dmvolod: The identification passed before in the upper function. This only update

pkg/controller/infinispan/infinispan_controller.go
```diff
@@ -704,6 +755,21 @@ func (r *ReconcileInfinispan) upgradeInfinispan(infinispan *infinispanv1.Infinis
 	})
 }
 
+func (r *ReconcileInfinispan) dropSecretOwnerReference(secretName string, infinispan *infinispanv1.Infinispan, logger logr.Logger) error {
+	secret := &corev1.Secret{}
+	if result, err := kube.LookupResource(secretName, infinispan.Namespace, secret, r.client, logger, eventRec); result != nil {
+		return err
+	}
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, secret, func() error {
```
ryanemerson, line 763: Do we need todo `LookupResource` and `CreateOrUpdate`? Can't we just use `CreateOrPatch`?

The value of `result` is never used, as eventually `destroyResources` always returns `reconcile.Result{}, err` if an error exists.

dmvolod: Yes, good point for optimization. Thanks, @ryanemerson

pkg/controller/infinispan/resources/secret/secret_controller.go
rigazilla: oh keystore is never generated, right?

dmvolod: Seems to yes

rigazilla (approved):

ryanemerson:
Thanks @dmvolod

PR #1208, opened 2021-07-29
[test/e2e] Do not wait on Ready batch pods

Crumby:
If the testing environment is fast enough then the pod is completed sooner then marked as Ready. Causing the wait function to cycle until the end, waiting for the condition that already passed or was never met even though everything goes as expected.


ryanemerson:
Thanks @Crumby

PR #1221, opened 2021-08-16
Operator sdk 1.3.2

ryanemerson:
Builds upon https://github.com/infinispan/infinispan-operator/pull/1220.

~~I have based upon the latest `main` and have made it so that the tests can be executed. Opening this PR as a WIP to see how many failures we get on CI.~~

Work done:
- All Controller implementations are now in the `controllers` package. Package level vars refactored so that no conflicts occur. Controller methods refactored to use `<control-type>Request` which wraps `<control-type>Reconciler>` and adds request specific fields.
- All `*Reconciler` initialisation moved to `SetupWithManager` methods.
- Controllers refactored to always use  the `context.Context` passed to `Reconcile`.
- RBAC annotations added to relevant controllers.
- Makefile updated to use latest operator-sdk structure.


go.mod
```diff
@@ -1,58 +1,35 @@
 module github.com/infinispan/infinispan-operator
 
-go 1.13
+go 1.15
```
dmvolod, line 3: @ryanemerson if go version could be updated to the 1.16 `embed` can be used instead of the third-party `rice` tool
https://pkg.go.dev/embed@go1.16

ryanemerson: Good idea! Once I have worked out the perf issues with this PR, I'll look into adopting that. Possibly as a follow up PR.

ryanemerson: @rigazilla Do you know any reason why we can't use `1.16`?

ryanemerson:
@rigazilla I hope the number of commits don't put you off too much :sweat_smile: 

So far I have concentrated my efforts on refactoring the code, making the tests pass and making it so that the docker image can be built and the operator deployed to a remote k8s cluster.

I have updated the README to reflect how the latest changes to the `Makefile`. As part of this, I have removed a lot of content that was outdated or not directly related to operator development. IMO things like setting up minikube etc are beyond the scope of this document, WDYT? If we're doing things right, then the operator should be usable on any k8s cluster assuming all requirements are met.

Still todo:
[ ] Migrate to OLM bundle format
[ ] Update xsite tests to install manifests via Make and not us build.sh
[ ] Remove all use of `deploy/` in scripts
[ ] Remove use of Dockerfile.single
[ ] Remove redundant scripts/config
[ ] Remove checks for missing deps in .sh scripts, can use `go-get-tool` in Make instead.

A consequence of this PRs changes is that many of the old scripts are now out of date, or in several cases even more out of date than they previously were :smiley:

What are your thoughts on removing any/all of the following:
- `cekit/` - Is this required downstream?
- 'deploy/` - How useful are the various examples in `deploy/cr`? I never use them personally, but we can move these to `config/samples` if you think there is a use.
- `build/minikube` - I have no idea if these are used or functional
- `olm` - Is this ever used? I think it will be redundant once I have updated the CSV to use bundles

go.mod
rigazilla: @ryanemerson I'm not aware of any, `1.16` should be fine

rigazilla:
>     * `cekit/` - Is this required downstream?

@ryanemerson , downstream has his own build. I think we can remove this

rigazilla:
> I have updated the README to reflect how the latest changes to the `Makefile`. As part of this, I have removed a lot of content that was outdated or not directly related to operator development. IMO things like setting up minikube etc are beyond the scope of this document, WDYT? If we're doing things right, then the operator should be usable on any k8s cluster assuming all requirements are met.

@ryanemerson I like this approach, +100!

rigazilla:
>     * 'deploy/`- How useful are the various examples in`deploy/cr`? I never use them personally, but we can move these to `config/samples` if you think there is a use.

I'm used to use them to create ispn cluster with basic features, I would move them under `config/samples`
btw @ryanemerson , how do you create ispn clusters commonly?

rigazilla:
>     * `olm` - Is this ever used? I think it will be redundant once I have updated the CSV to use bundles

also this can be removed imo

ryanemerson:
> btw @ryanemerson , how do you create ispn clusters commonly?

For feature work I create the CR as part of the testsuite and that forms the basis of the tests that accompany my PR.

For reproducers etc I usually just create a CR file locally. I find that in most case the `deploy` examples would need significant changes for my purposes, so it's easier to just create a CR locally.

ryanemerson:
Restructure complete. The only thing that is missing is a replacement for the old `build/upgrade-test.sh`. This will come in a subsequent PR that will add OLM tests that would have made the manual tests obsolete anyway.

There's still an issue with the refactored `configure-xsite.sh` script. Once that's been resolved, I will rebase to a single commit and this PR should be ready.

go.mod
rigazilla: justa note: fedora 34 is needed to run go 1.16

ryanemerson: I won't make this change as part of this PR, but for a follow up PR would fedora 34 being the minimum be an issue for anyone @rigazilla @Crumby @dmvolod ?

rigazilla: I'm on 33, let me check if 34 is ok with all the downstream tools

dmvolod: It can be possible to install `go` tools manually without repo and could use any version
P.S. I'm on WSL2 now and have a very old go tools repo inside embedded Ubuntu image, but easily install go tools from binaries 😁

rigazilla: yeah, but I'd like to keep the distro clean if possible, anyway f34 looks good! I'm upgrading

ryanemerson:
I'll create a follow up PR to:

- Create automated OLM upgrade tests. This will replace the manual upgrade-tests.sh removed in this PR
- Upgrade to `go` 1.16 and remove rice

rigazilla:
I've added two commit:
- operator seems was running clusterwide
- added a check on userSecret, cause it could happen that authentication could be enabled in the middle of a reconc. loop, generating a NPE. `TestEndpointAuthenticationUpdate` sometime triggers this condition.

rigazilla:
Thanks @ryanemerson , great job!

Issue #1171, opened 2021-07-01
Explicitly set `imagePullPolicy: Always` on cluster StatefulSet


ryanemerson:
Without this it's possible for pods distributed across k8s nodes to utilise different versions of the image, as we use the floating tags for images in our development branches. For example with the tag `:12.1` 12.1.2.Final and 12.1.4.Final server could be deployed on different nodes. This will work for most cases, but is not officially supported and we should avoid this where possible.

dmvolod:
We could set exact image version, i.e. `12.1.7.Final` by default in upstream and tests to avoid unexpected behavior when new image has been released.

ryanemerson:
> We could set exact image version, i.e. `12.1.7.Final` by default in upstream and tests to avoid unexpected behavior when new image has been released.

The issue with this is that `12.1.7.Final` is also a floating tag. So if we have to respin `12.1.7` images, we can have `12.1.7.Final-1` and `12.1.7-Final-2` on different k8s nodes.

PR #1226, opened 2021-08-25
Separate admin port into a different service

gustavonalle:
Motivation: so that redirection of services during rolling upgrades can be done only for the "user facing" service and not for admin services, that needs to be associated with the cluster for monitoring purposes

### DIFF
--- a/pkg/controller/infinispan/resources/secret/secret_controller.go
+++ b/pkg/controller/infinispan/resources/secret/secret_controller.go
@@ -250,7 +250,7 @@ func (s secretResource) reconcileAdminSecret() error {
 }
 
 func (s secretResource) addCliProperties(secret *corev1.Secret, password string) {
-	service := s.infinispan.GetServiceName()
+	service := s.infinispan.GetAdminServiceName()
 	url := fmt.Sprintf("http://%s:%s@%s:%d", consts.DefaultOperatorUser, url.QueryEscape(password), service, consts.InfinispanAdminPort)
 	properties := fmt.Sprintf("autoconnect-url=%s", url)
 	secret.Data[consts.CliPropertiesFilename] = []byte(properties)
--- a/build/upgrade-test-local.sh
+++ b/build/upgrade-test-local.sh
@@ -56,6 +56,7 @@ if git checkout "${FROM_UPGRADE_VERSION}"; then
     export DEFAULT_IMAGE=$(sed -n '/name: DEFAULT_IMAGE/ {n;p}' "${TO_OPERATOR_DEPLOY_FILE}" | awk -F '"' '{print $2}')
   else
     export DEFAULT_IMAGE="${TO_OPERATOR_IMAGE}"
+    export EXPECTED_IMAGE="${TO_OPERATOR_IMAGE}"
   fi
 
   if ! "${TO_TEST_SCRIPT_FILE}" "${KUBECONFIG}"; then
--- a/pkg/apis/infinispan/v1/types_util.go
+++ b/pkg/apis/infinispan/v1/types_util.go
@@ -262,6 +262,10 @@ func (ispn *Infinispan) GetServiceName() string {
 	return ispn.Name
 }
 
+func (ispn *Infinispan) GetAdminServiceName() string {
+	return fmt.Sprintf("%s-admin", ispn.Name)
+}
+
 func (ispn *Infinispan) GetPingServiceName() string {
 	return fmt.Sprintf("%s-ping", ispn.Name)
 }
--- a/pkg/controller/infinispan/infinispan_controller.go
+++ b/pkg/controller/infinispan/infinispan_controller.go
@@ -650,6 +650,17 @@ func (r *ReconcileInfinispan) destroyResources(infinispan *infinispanv1.Infinisp
 	err = r.client.Delete(context.TODO(),
 		&corev1.Service{
 			ObjectMeta: metav1.ObjectMeta{
+				Name:      infinispan.GetAdminServiceName(),
+				Namespace: infinispan.Namespace,
+			},
+		})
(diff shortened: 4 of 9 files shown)


pkg/controller/infinispan/resources/secret/secret_controller.go
```diff
@@ -250,7 +250,7 @@ func (s secretResource) reconcileAdminSecret() error {
 }
 
 func (s secretResource) addCliProperties(secret *corev1.Secret, password string) {
-	service := s.infinispan.GetServiceName()
+	service := s.infinispan.GetAdminServiceName()
```
dmvolod, line 253: @gustavonalle we should ensure that Secret will update this entry during update, otherwise Batch will nor work after that

gustavonalle: I suppose the batch after upgrade scenario is not tested currently?

dmvolod: No upgrade tests in CI now. But seems to new CLI will be patched carefully in the `reconcileAdminSecret`

dmvolod: You can run upgrade test with `make`
https://github.com/infinispan/infinispan-operator/blob/fba24c178eab6ed9bbcc3761530f08bd465fa67a/Makefile#L147-L166

https://github.com/infinispan/infinispan-operator/blob/fba24c178eab6ed9bbcc3761530f08bd465fa67a/Makefile#L168-L178

and keep cluster running with CLEANUP_INFINISPAN_ON_FINISH=false option. The one of the success condition for upgrade is Infinispan Server image tags should be different between branches or redefined with env variable

gustavonalle: > You can run upgrade test with make

Thanks, yeah, I was looking at those tests, but can I reproduce the secret issue you mentioned with it?

I tried running from my ```service``` branch, with:

```
make KUBECONFIG=... upgrade-test-local FROM_UPGRADE_VERSION=main
```

But it never does the upgrade, timing out in the second step, I guess 

> The one of the success condition for upgrade is Infinispan Server image tags should be different between branches or redefined with env variable

Is this a manual step?

dmvolod: You can redefine target image with TO_OPERATOR_IMAGE env variables with the different server image tag
But seems to `build/upgrade-test-local.sh` need to be fixed a bit for the `main`

gustavonalle: ok, makes sense, in any case, the batch is not tested after the upgrade. I will add it to the test

gustavonalle: btw, tried with other branches and it failed too:

```
make KUBECONFIG=... upgrade-test-local FROM_UPGRADE_VERSION=2.1.4.Final TO_UPGRADE_VERSION=2.1.5.Final CLEANUP_INFINISPAN_ON_FINISH=false
```

dmvolod: Server images for the `2.1.4.Final` and `2.1.5.Final` are the same, i.e. `quay.io/infinispan/server:12.1`. Upgrade will not trigger and required conditions flow will not pass.

gustavonalle: I thought the CSV file in each tag contained the image version for each operator:

https://github.com/infinispan/infinispan-operator/blob/2.1.4.Final/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml#L312

https://github.com/infinispan/infinispan-operator/blob/2.1.5.Final/deploy/olm-catalog/infinispan-operator.clusterserviceversion.yaml#L313

gustavonalle: So the code uses 12.1 that is a "movable" tag, always points to the latest 12.1.x?

dmvolod: Yes. correct. It get from the operator deployment yaml file. We discussed setting exact image version on update in the deployment here #1171 but images the same for the rest of the versions

gustavonalle: ouch. IMHO, having a single "movable" tag like ```latest``` is fine, but making every tag a movable tag kinds of defeat the purpose of a tag :smile:

dmvolod: Yes, agree. But for full upgrade it's necessary to keep tags up to date in many branches. Seems to be too late now and can be customized with env variable only.

gustavonalle: I added a test and wasn't able to reproduce the secret issue @dmvolod, could you clarify the scenario?

dmvolod: @gustavonalle looks good. can you try to see that upgraded cluster is wellFormed?

gustavonalle: This is already checked here I reckon:

https://github.com/infinispan/infinispan-operator/blob/fba24c178eab6ed9bbcc3761530f08bd465fa67a/test/e2e/main/main_test.go#L64-L66

gustavonalle: https://github.com/infinispan/infinispan-operator/blob/fba24c178eab6ed9bbcc3761530f08bd465fa67a/test/e2e/utils/constants.go#L44

ryanemerson:
Sorry for the delay. Thanks @gustavonalle

Issue #1232, opened 2021-09-01
Docs revision: Clarify auto scale capabilities


oraNod:
Make it clear in documentation that the operator does not autoscale DG service clusters.

PR #1235, opened 2021-09-01
issue #1232 clarify autoscale

oraNod:
Docs for issue #1232 

Needs backport to 2.1.x.


ryanemerson (commented):
`s/nodes/pods`

Otherwise LGTM. I would will wait for @rigazilla to review though as he's the Autoscale Jedi

rigazilla (commented):
yeah, I'm not sure we already addressed this, but as mentioned by @ryanemerson `node` is a problematic word with OCP and Infinispan.

`pod` would be better (implying `1 pod` == `1 infinispan node`) or something like `Infinispan node` or `data grid node`.

otherwise PR lgtm

oraNod:
thanks @rigazilla and @ryanemerson roger the problematic nature of `node` vs `pod`. @Crumby and I discussed this before and I think the intent was that `node` was more familiar term when talking about Infinispan clusters. using the two terms interchangeably doesn't seem quite right and it's kind of a tricky one. I'll update but might also have another look through related sections while I'm at it. cheers.

ryanemerson:
> the intent was that `node` was more familiar term when talking about Infinispan clusters. using the two terms interchangeably doesn't seem quite right and it's kind of a tricky one.

I agree it's tricky. IMO when it comes to the operator docs though we have to follow the k8s conventions as the operator is only deployed in that environment. For the wider Infinispan docs ecosystem, I think the use of node(s) is fine as pods only really make sense in the context of k8s and clearly Infinispan is deployed in many more contexts.

ryanemerson (approved):

oraNod:
@ryanemerson Thanks for picking up on `s/nodes/pods`. I really appreciate this level of detail.

rigazilla (approved):

oraNod:
Merging after offline discussion with @Crumby . Will backport to 2.1.x. Thanks again @rigazilla and @ryanemerson

PR #1244, opened 2021-09-14
Fix xsite test failures

ryanemerson:


ryanemerson:
Self merging as fixes required for `main` CI.

PR #1248, opened 2021-09-20
OLM Integration tests

ryanemerson:


Jenkinsfile
```diff
@@ -13,6 +13,8 @@ pipeline {
         PATH="/opt/go/bin:$PATH"
         RUN_SA_OPERATOR = 'true'
         MAKE_DATADIR_WRITABLE = 'true'
+        OLM_NEXT_RELEASE = '2.1.6'
```
Crumby, line 16: Shouldn't it be 2.2.0 and 2.2.x channel?

ryanemerson:
To test locally:

1. Create the kind cluster: `./scripts/ci/kind-with-olm.sh`
2. Create the operator, bundle and catalog image and install the `CatalogSource` in `namespace-for-testing`
3. Run the test `make upgrade-test`

@Crumby Step 3 can be replaced with `kubectl apply -f scripts/ci/Subscription.yaml ` to manually test the Subscription.

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+	testKube = tutils.NewTestKubernetes(os.Getenv("TESTING_CONTEXT"))
+	// startChannel          = os.Getenv("SUBSCRIPTION_CHANNEL_START")
+	subName               = constants.GetEnvWithDefault("SUBSCRIPTION_NAME", "infinispan-operator")
+	subNamespace          = constants.GetEnvWithDefault("SUBSCRIPTION_NAMESPACE", tutils.Namespace)
+	sourceChannel         = constants.GetEnvWithDefault("SUBSCRIPTION_CHANNEL_SOURCE", "2.1.x")
+	latestChannel         = constants.GetEnvWithDefault("SUBSCRIPTION_CHANNEL_LATEST", "2.2.x")
+	catalogSource         = constants.GetEnvWithDefault("SUBSCRIPTION_CATALOG_SOURCE", "test-catalog")
+	catalogSourcNamespace = constants.GetEnvWithDefault("SUBSCRIPTION_CATALOG_SOURCE_NAMESPACE", tutils.Namespace)
+	subPackage            = constants.GetEnvWithDefault("SUBSCRIPTION_PACKAGE", "infinispan")
+	subStartingCsv        = constants.GetEnvWithDefault("SUBSCRIPTION_STARTING_CSV", "infinispan-operator.v2.1.5")
```
Crumby, line 33: Since the test runs upgrade from `current - 1` channel this value can be omitted

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	testKube = tutils.NewTestKubernetes(os.Getenv("TESTING_CONTEXT"))
+	// startChannel          = os.Getenv("SUBSCRIPTION_CHANNEL_START")
+	subName               = constants.GetEnvWithDefault("SUBSCRIPTION_NAME", "infinispan-operator")
+	subNamespace          = constants.GetEnvWithDefault("SUBSCRIPTION_NAMESPACE", tutils.Namespace)
+	sourceChannel         = constants.GetEnvWithDefault("SUBSCRIPTION_CHANNEL_SOURCE", "2.1.x")
+	latestChannel         = constants.GetEnvWithDefault("SUBSCRIPTION_CHANNEL_LATEST", "2.2.x")
```
Crumby, line 29: This value can be parsed so it doesn't need to be updated with each new release

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	testKube = tutils.NewTestKubernetes(os.Getenv("TESTING_CONTEXT"))
+	// startChannel          = os.Getenv("SUBSCRIPTION_CHANNEL_START")
+	subName               = constants.GetEnvWithDefault("SUBSCRIPTION_NAME", "infinispan-operator")
+	subNamespace          = constants.GetEnvWithDefault("SUBSCRIPTION_NAMESPACE", tutils.Namespace)
+	sourceChannel         = constants.GetEnvWithDefault("SUBSCRIPTION_CHANNEL_SOURCE", "2.1.x")
```
Crumby, line 28: This value can be parsed so it doesn't need to be updated with each release

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, state := range []ispnv1.ConditionType{ispnv1.ConditionUpgrade, ispnv1.ConditionStopping, ispnv1.ConditionWellFormed} {
+		testKube.WaitForInfinispanCondition(strcase.ToKebab(t.Name()), tutils.Namespace, state)
+	}
+
+	// Validates that all pods are running with desired image
+	pods := &corev1.PodList{}
+	err := testKube.Kubernetes.ResourcesList(tutils.Namespace, controllers.PodLabels(spec.Name), pods, context.TODO())
+	tutils.ExpectNoError(err)
+	for _, pod := range pods.Items {
+		if pod.Spec.Containers[0].Image != latestImage {
```
Crumby, line 99: The test performs only single version upgrade and it's from the previously released channel. This will fail with 2.2.1 release as it will expect 2.2.1 related server image but the subscription will be at this point only on 2.2.0

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+	latestChannel         = constants.GetEnvWithDefault("SUBSCRIPTION_CHANNEL_LATEST", "2.2.x")
+	catalogSource         = constants.GetEnvWithDefault("SUBSCRIPTION_CATALOG_SOURCE", "test-catalog")
+	catalogSourcNamespace = constants.GetEnvWithDefault("SUBSCRIPTION_CATALOG_SOURCE_NAMESPACE", tutils.Namespace)
+	subPackage            = constants.GetEnvWithDefault("SUBSCRIPTION_PACKAGE", "infinispan")
+	subStartingCsv        = constants.GetEnvWithDefault("SUBSCRIPTION_STARTING_CSV", "infinispan-operator.v2.1.5")
+
+	latestImage = constants.GetEnvWithDefault("LATEST_OPERAND_IMAGE", constants.DefaultOperandImageOpenJDK)
+)
+
+func TestUpgrade(t *testing.T) {
```
Crumby, line 38: We should probably add 0 pod kill at the end of the test to assert that cluster will recover from failover after upgrade

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// Approve the initial startingCSV InstallPlan
+	testKube.ApproveInstallPlan(subName, subNamespace, 1)
+
+	testKube.WaitForCrd(&apiextv1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "infinispans.infinispan.org",
+		},
+	})
+
+	spec := tutils.DefaultSpec(testKube)
```
Crumby, line 79: Better would be to use at least 2 replicas sized cluster. Lot of issues could be hidden without using actual cluster

test/e2e/upgrade/upgrade_test.go
ryanemerson: Originally I was thinking of just having this as a ENV var without a default and then setting this centrally for CI. However, that makes it more painful to run locally.

Any thoughts on where we could parse the value from for it to work with the centralised Makefile and go testsuite?

Crumby: https://github.com/infinispan/infinispan-operator/pull/1211/files#diff-b2218cd9f8b91e8747cc3b42a41c88f4e770132ba2a117f616c44b61fceac3d4R12-R49 is piece of code in java that covers what you are looking for. It shouldn't be an issue to get the same in go on test init.

test/e2e/upgrade/upgrade_test.go
ryanemerson: As the server image has no relation to the operator version, I'm not sure we can fully automate this.

The test in it's current form will only work with a single upgrade anyway, as it approves the `InstallPlan` instances outside of a loop.

If we update the test to perform multiple upgrades in a loop, then I think we have two options:

1. Only check `pod.Spec.Containers[0].Image` for the final upgrade
2. Require a ENV var that contains a csv of all expected images. CSV will have to take into account operator versions that don't change server version.

My preference would be for 1.

Crumby: The relation between Operator and Server is given by CSV which specifies a particular server image it's supposed to run so it can parsed for each currently installed CSV for verification. And if we want the tests to go through several upgrades we'll need to wait for cluster being upgraded properly before proceeding. Updating CSV before the new version of cluster is installed could result in unpredictable behavior.

Basically approve new CSV version, wait till it's installed, parse it for DEFAULT_IMAGE, wait for cluster being upgraded. (Or simply proceed if the server wasn't updated in particular release)

Crumby: Actually I just realized that using `constants.DefaultOperandImageOpenJDK` won't work nicely with downstream as we'd have to set additional evn that's already available in the CSV

ryanemerson: Thanks for explaining in such detail. I was definitely under thinking it with my initial approach.

I have updated so that the server image is extracted from the currently active CSV.

test/e2e/upgrade/upgrade_test.go
ryanemerson: I have updated it to automatically populate the default value, but it's also possible to override the source/target channels via the `SUBSCRIPTION_CHANNEL_SOURCE` and `SUBSCRIPTION_CHANNEL_TARGET` env variables respectively.

test/e2e/upgrade/upgrade_test.go
ryanemerson: The default value is now the `CurrentCSVName` of the source channel (equivalent to not explicitly configuring it). It's also possible to use an older CSV in the channel by specifying the `SUBSCRIPTION_STARTING_CSV` env var.

Crumby:
@ryanemerson Would be nice if the test had custom cache with persistence created, filled with several entries, prior upgrade, to verify that data can be read correctly after the upgrade

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,185 @@
...
+		Spec: &coreos.SubscriptionSpec{
+			Channel:                sourceChannel.Name,
+			CatalogSource:          catalogSource,
+			CatalogSourceNamespace: catalogSourcNamespace,
+			InstallPlanApproval:    coreos.ApprovalManual,
+			Package:                subPackage,
+			StartingCSV:            subStartingCsv,
+		},
+	}
+	testKube.CreateSubscription(sub)
```
Crumby, line 65: `OperatorGroup` should be also created unless it's global namespace. If the OperatorGroup doesn't exist the installation won't proceed on OpenShift. It's interesting that it passes on Kind in CI though

test/e2e/utils/olm.go
```diff
@@ -0,0 +1,121 @@
...
+	envName := "RELATED_IMAGE_OPENJDK"
+	envVars := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.Containers[0].Env
+	index := kubernetes.GetEnvVarIndex(envName, &envVars)
+	if index < 0 {
+		panic(fmt.Sprintf("Unable to find env var '%s' in CSV %s", envName, csv.Name))
+	}
+	return envVars[index].Value
+}
+
+func (k TestKubernetes) PackageManifest(name, namespace string) *manifests.PackageManifest {
```
Crumby, line 79: This should also filter out the result based on CatalogSourceName. If both `test-catalog` and `released` catalog exist in same namespace then picking up one for parsing is random

test/e2e/upgrade/upgrade_test.go
ryanemerson: The `OperatorGroup` is actually created as part of `scripts/ci/install-catalog-source.sh`. I did this originally as it helped with manual testing as well, but I can move the creation to the `upgrade-test.go` if you prefer?

Crumby: Maybe it would be better so the test is more self-contained? I mean the test creates and removes the namespace where it installs the Operator by default. If the user will only rerun the test it will fail on second attempt as the OperatorGroup will be removed.

test/e2e/utils/olm.go
ryanemerson: I'm not sure this is possible via the [REST API](https://docs.openshift.com/container-platform/4.8/rest_api/operatorhub_apis/packagemanifest-packages-operators-coreos-com-v1.html#apispackages-operators-coreos-comv1namespacesnamespacepackagemanifestsname). 

If two `CatalogSources` exist in the same namespace with entry "infinispan" then the most recently created is always returned. This is the case when executing a GET to `/apis/packages.operators.coreos.com/v1/namespaces/{namespace}/packagemanifests/{name}` and executing `kubectl get PackageManifest {name} -n {namespace}`

The entire behaviour of `PackageManifest` is strange with it not being a traditional k8s resource.

Crumby: It should be possible to retrieve all PackageManifests and filter out the one we want. But it's not a must have. Can be workaroundable by disabling default CatalogSources or creating the test CatalogSource in different namespace

test/e2e/upgrade/upgrade_test.go
```diff
@@ -0,0 +1,185 @@
...
+	defer testKube.DeleteNamespace(tutils.Namespace)
```
Crumby, line 66: We should probably also add defer on removing subscription and csv in case `SUBSCRIPTION_NAMESPACE` != tutils.Namespace

ryanemerson: I've added a deferred call to `cleanup(..)` that will remove the created OLM CRs as well as any Infinispan resources.

test/e2e/utils/olm.go
ryanemerson: I have been unable to find a way todo this so far. Ideally I don't want to spend any more time on this atm, but it's something that we can come back to in the future.

ryanemerson: I've changed it so that we get all manifests via `/apis/packages.operators.coreos.com/v1/packagemanifests` and filter on the catalog name.

PR #1251, opened 2021-09-21
New test for testing pod degradation after OutOfMemory

fabriziosta:


test/e2e/main/main_test.go
```diff
@@ -1450,3 +1450,61 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+// Test if the cluster made up of 2 nodes is working correctly after it was restarted by OutOfMemory exception
+func TestPodDegradationAfterOOM(t *testing.T) {
+	t.Parallel()
+
+	//Creating Infinispan cluster
+	ispn := tutils.DefaultSpec(testKube)
+	ispn.Name = strcase.ToKebab(t.Name())
+	ispn.Labels = map[string]string{"test-name": t.Name()}
+	ispn.Spec.Replicas = 2
+	ispn.Spec.Container.Memory = "512Mi"
```
Crumby, line 1463: This is default value, I would try going a bit lower to speed up test. Eg.: 384Mi or maybe even 256Mi depending on how empty node needs

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1450,61 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+	hostAddr, client := tutils.HTTPClientAndHost(ispn, testKube)
+	createCacheWithXMLTemplate(cacheName, hostAddr, template, client)
+
+	//Generate tons of random entries
+	for key := 1; key < 60000; key++ {
```
Crumby, line 1480: Would be nice to stop the cycle when OOM hits so it doesn't load bunch of other entries

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1450,61 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+		defer tutils.CloseHttpResponse(resp)
+
+		if error != nil {
+			fmt.Printf("ERROR for key=%d, Description=%s", key, error)
+		}
+	}
+
+	//Check if all pods are running and they are not degraded
+	podList := &corev1.PodList{}
+	tutils.ExpectNoError(wait.Poll(tutils.DefaultPollPeriod, tutils.SinglePodTimeout, func() (done bool, err error) {
```
Crumby, line 1496: You should also check restartCount==1 and pod is ready for pod that was hit by OOM. Other thing that would be nice to verify after this is condition is met is that no cache is in degraded state

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1450,61 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
...5stfuBQC0vOmLvpZrhVk5sgBqlFqAFHBkIndGR9ZNEGSD2nSZHMEgMfuU32VOLXqMn8QycaLKosaX4VUlkXDJZ2f0qR8LiBEOvn6MRoggayIeLiQO3A1VnoAV8ifbANrSriuRrPTOJMqkKollYXMfyJ3oIdXedDIyFDTpOX5oWhcw0rBOr6f68lFghVSCjwq4QnyAaA6WDFZ2HuretkhbAxQizmeCICBXhfn2fU4tOGxUjh7YQxqBL3XvEnqZcZJhYD0f5iustMTrVrOcHMnvf38PmmT3uK0KmHs00wJ4cbob2Muh8GoNsQWaunrdI1oJRyiI8R6UwCbR6KdXqPy3Vc67GuGJ6Am2lI2KbnZSdjN6lAlvZjuzF1w31jbY1r5PDgBVnPWz6ZUL8zoOg0rfsdXkWLMaxjil9YmMAc5bu7YmY38x05hYf7ck9eaxAelYbOllHHZ8SS6EKoofdx7FwQ0ih5uBfUgcHSryLSG4088r7LCVdqkrUmirGICXZFYh50wkwnEBVuF7Vq9OkJqEzQOehwtbZ7aC9idhypfJalhq4xEAI5FlaoQgpkzmSfinLEz2nt8ktTom7aKZ5QcM75RZwGLfQYfeXKvNSE5D44EsKEZObXvf78sBpveBwJkT6G94kP4rttWEcntpsowx2mDRb9HbC8iJYpYXa5rt90elD5mbu16Pj7XMkhSJckPtJhLwcgzxvmyIx4hA6Ewt3l9uh9BrTvYPwkEeh7bxGrWsZ8Ftokwyc7vi55CFjVvAHKoP1tQ0qcHYqIWzAKVxhozDuOfBk4nQ5UqbwIamT6Au5ZbHtnKrmpAeQvWxJWNy3ehRkQbso3JvE0cw6kWh4EkkSyiH8Jh677U36GKxEs9ngFNaoOdTFpCQWrfCtITUEt2eFGoHllcEQ6hpexKZkgbPGGb1R6OV1vPgGqeJ7Mms2g9Z4Fd2vLjxxsVykuHlWJPxueKTvBRhp7GEonjhLTgfWHOAVJqW7jxfLCVfJpsECAUE6duS43ZIBxfXTrN9IiV75wWdDnJl197BM3w8uxdAwMe50qTuEWF6ey8uR40L94mfLi1MnwWbJeJnPyLvhU0DkLMhmJGdTNPxd3OSekKhUtGgt65j7xU37YEaF3aBiL45tSASJ7jZg6CWQ2GdVhnRlSQRXW15uEY6hzv26eZMYs2VX6TwxFjYHDHydCi0PL"
```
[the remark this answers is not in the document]
fabriziosta: Alright, I've added a function to generate random strings based on a charset

test/e2e/main/main_test.go
fabriziosta: Lowered memory to 256, it goes in OOM faster

test/e2e/main/main_test.go
fabriziosta: With the new code, when it hits an OOM it breaks the loop instead of put all the occurences

test/e2e/main/main_test.go
fabriziosta: I have changed this part. Now it polls immediately after the OOM; whether the cluster is able to bootstrap again the test ends successfully. In every other case it fails

ryanemerson:
@fabriziosta Can rebase so only your JDG-3991 is present. We won't want to include the merge commit on `main`.

Can you create a GH issue with a brief summary of the issue you're testing and use the GH issue number in your commit message instead of the downstream JDG id.

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1451,63 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+	testKube.CreateInfinispan(ispn, tutils.Namespace)
+	defer testKube.CleanNamespaceAndLogOnPanic(tutils.Namespace, ispn.Labels)
+	testKube.WaitForInfinispanPods(int(ispn.Spec.Replicas), tutils.SinglePodTimeout, ispn.Name, tutils.Namespace)
+	testKube.WaitForInfinispanCondition(ispn.Name, ispn.Namespace, ispnv1.ConditionWellFormed)
+
+	//Creating cache
+	cacheName := "failover-cache"
+	template := `<infinispan><cache-container><replicated-cache name ="` + cacheName +
+		`"><encoding media-type="text/plain"/></replicated-cache></cache-container></infinispan>`
+	veryLongValue := GenerateStringWithCharset(6000)
```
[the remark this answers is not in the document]
Crumby: Default cache was only a way to discover underlying issue but it doesn't matter in reality. The purpose of this test is to verify that cluster will survive OOM node failure

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1451,63 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+		_, error := client.Post(keyURL, veryLongValue, headers)
+
+		if error != nil {
+			fmt.Printf("ERROR for key=%d, Description=%s", key, error)
+			break
+		}
+	}
+
+	//Check if all pods are running and they are not degraded
+	testKube.WaitForInfinispanPods(int(ispn.Spec.Replicas), tutils.SinglePodTimeout, ispn.Name, tutils.Namespace)
```
ryanemerson, line 1493: It would be good if we first asserted that the pods had indeed failed over. Assuming that the very large keys somehow didn't cause the OOM, the test would still pass with these checks.

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1451,63 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+func TestPodDegradationAfterOOM(t *testing.T) {
+	t.Parallel()
+
+	//Creating Infinispan cluster
+	ispn := tutils.DefaultSpec(testKube)
+	ispn.Name = strcase.ToKebab(t.Name())
+	ispn.Labels = map[string]string{"test-name": t.Name()}
+	ispn.Spec.Replicas = 2
+	ispn.Spec.Container.Memory = "256Mi"
+	ispn.Spec.Security.EndpointAuthentication = pointer.BoolPtr(false)
```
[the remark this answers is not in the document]
fabriziosta: Actually, I think it is not necessary, let me take care of that line, I'll get rid of that

test/e2e/main/main_test.go
fabriziosta: In the updated version I've added an if statement to check whether the pod restarts for an OOM or not.

fabriziosta:
> @fabriziosta Can rebase so only your JDG-3991 is present. We won't want to include the merge commit on `main`.
> 
> Can you create a GH issue with a brief summary of the issue you're testing and use the GH issue number in your commit message instead of the downstream JDG id.

Let me know if it's better now

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1452,81 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+			break
+		}
+	}
+
+	//Check if all pods are running and they are not degraded
+	testKube.WaitForInfinispanPods(int(ispn.Spec.Replicas), tutils.SinglePodTimeout, ispn.Name, tutils.Namespace)
+	testKube.WaitForInfinispanCondition(ispn.Name, ispn.Namespace, ispnv1.ConditionWellFormed)
+
+	//Verify whether the pod restarted for an OOM exception
+	hasOOMhappened := false
```
ryanemerson, line 1497: Shouldn't this check occur before we call `WaitForInfinispanPods` and `WaitForInfinispanCondition`?

Also, I would expect that the test should fail if `!hasOOMhappened`, because it's not possible to test pod recovery from OOM if we can't confirm it has indeed occurred.

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1452,81 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+				}
+				fmt.Printf("\nExitCode='%d' Reason='%s' Message='%s'\n", terminatedPod.ExitCode, terminatedPod.Reason, terminatedPod.Message)
+			}
+		}
+	}
+
+	if kube.AreAllPodsReady(podList) && hasOOMhappened {
+		fmt.Printf("\nAll pods are ready\n")
+	} else {
+		panic("\nOne of the pods is degraded\n")
```
ryanemerson, line 1519: No need for the `\n` chars in a `panic` string

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1452,81 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+				if terminatedPod.Reason == "OOMKilled" {
+					hasOOMhappened = true
+				}
+				fmt.Printf("\nExitCode='%d' Reason='%s' Message='%s'\n", terminatedPod.ExitCode, terminatedPod.Reason, terminatedPod.Message)
+			}
+		}
+	}
+
+	if kube.AreAllPodsReady(podList) && hasOOMhappened {
+		fmt.Printf("\nAll pods are ready\n")
```
ryanemerson, line 1517: `Println` without `\n` should suffice

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1452,81 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+	createCacheWithXMLTemplate(cacheName, hostAddr, template, client)
+
+	//Generate tons of random entries
+	for key := 1; key < 50000; key++ {
+		keyURL := fmt.Sprintf("%v/%v", cacheURL(cacheName, hostAddr), key)
+		headers := map[string]string{"Content-Type": "text/plain"}
+		_, error := client.Post(keyURL, veryLongValue, headers)
+
+		if error != nil {
+			fmt.Printf("\nERROR for key=%d, Description=%s\n", key, error)
```
ryanemerson, line 1487: There shouldn't be a need for `\n` at the start of the string

test/e2e/main/main_test.go
```diff
@@ -1450,3 +1452,81 @@ func createCacheWithCR(cacheName string, nameSpace string, clusterName string) *
...
+		status := pod.Status.ContainerStatuses
+
+		for _, containerStatuses := range status {
+			if containerStatuses.LastTerminationState.Terminated != nil {
+				terminatedPod := containerStatuses.LastTerminationState.Terminated
+
+				if terminatedPod.Reason == "OOMKilled" {
+					hasOOMhappened = true
+				}
+				fmt.Printf("\nExitCode='%d' Reason='%s' Message='%s'\n", terminatedPod.ExitCode, terminatedPod.Reason, terminatedPod.Message)
```
ryanemerson, line 1511: There shouldn't be a need for `\n` at the start of the string

test/e2e/main/main_test.go
```diff
@@ -301,6 +302,7 @@ func TestNodeStartup(t *testing.T) {
 // and match them with the labels map provided by the caller
 func areOperatorLabelsPropagated(namespace, varName string, labels map[string]string) bool {
 	podList := &corev1.PodList{}
+
```
ryanemerson, line 305: This change can be removed.

test/e2e/main/main_test.go
fabriziosta: I am moving the check before "WaitForInfinispanPods" and "WaitForInfinispanCondition", I don't get the second part though.
The condition checks whether "kube.AreAllPodsReady(podList) && hasOOMhappened", so I am excepting that if !hasOOMhappened the test panic. Am I losing something else?

fabriziosta: I am assuming that with that memory and with that values it will always go in OOM (it breaks with only 2500 occurrences of 50000). Whether I don't find "hasOOMhappened" to true the test is going to fail

ryanemerson:
@fabriziosta Force pushing automatically starts a CI run :slightly_smiling_face: 

Your code changes are failing the linter now:

```
15:16:52  + make lint
15:16:52  go: creating new go.mod: module tmp
15:16:52  Downloading github.com/golangci/golangci-lint/cmd/golangci-lint@v1.39.0
15:16:52  go: found github.com/golangci/golangci-lint/cmd/golangci-lint in github.com/golangci/golangci-lint v1.39.0
15:17:02  /home/infinispan/workspace/Infinispan_Operator_PR-1251/bin/golangci-lint run --enable errorlint
15:17:35  test/e2e/main/main_test.go:1498:35: Error return value of `testKube.Kubernetes.ResourcesList` is not checked (errcheck)
15:17:35  	testKube.Kubernetes.ResourcesList(tutils.Namespace, controllers.PodLabels(ispn.Name), podList, context.TODO())
15:17:35  	                                 ^
15:17:35  make: *** [Makefile:42: lint] Error 1
```

You can reproduce locally by executing `make lint`

PR #1252, opened 2021-09-27
Increase Pod memory for XSite tests

pruivo:


test/e2e/xsite/xsite_test.go
```diff
@@ -53,6 +53,9 @@ func crossSiteSpec(name string, replicas int32, primarySite, backupSite, siteNam
 					},
 				},
 			},
+			Container: ispnv1.InfinispanContainerSpec{
+				Memory: "1Gi",
```
dmvolod, line 57: @pruivo it would be nice to make it configurable via env vars the same as for other tests
https://github.com/infinispan/infinispan-operator/blob/13cc14b728f5f9d1cb7add2ac2b59423fc6ae5b4/test/e2e/utils/constants.go#L32

pruivo: Do you want another env variable name? or can I use the value from contansts.go ?

dmvolod: It could be the same as tests run sequentially

pruivo: done. I changed the Jenkinsfile to increase the memory for xsite tests only.
Thanks @dmvolod !

ryanemerson:
Thanks @pruivo

PR #1256, opened 2021-10-01
[E2E] Clean-up ConfigMap in TestBatchConfigMap

Crumby:
Make `TestBatchConfigMap` clean-up `ConfigMap` after it's executed otherwise second run of the test will fail on already existing `ConfigMap`.


ryanemerson (approved):

ryanemerson:
Thanks @Crumby

PR #1257, opened 2021-10-01
JDG-4595 Make locations required if site is set

pruivo:
https://issues.redhat.com/browse/JDG-4595


ryanemerson (commented):
@pruivo The code changes LGTM. Can you create a GH issue and update the commit message, otherwise theirs no upstream visibility.

pruivo:
done :+1:

Crumby:
@pruivo Wouldn't it be better if it worked rather other way around? If the site is set then enable cross site but with only cluster in it and no remote locations? User could start separate clusters and once the they know their locations they could add them to the locations

pruivo:
the JGroups configuration is static. To add more sites, you need to stop all pods, update the Jgroups configuration and start them again.

pruivo:
> But if I'd wanted to configure manually xsite location using AWS LoadBalancer then that's impossible

are you suggesting creating and exposing another service instead of using the one created by the Operator? I'm afraid it does not work. I tried and the traffic from the other service is rejected.

If you don't have or don't want to configure  another location, you can do something like 

```yaml
locations: 
  - name: undefined
    url: infinsipan+xsite://127.0.0.1:11111
```

and the operator creates the pods just fine with cross-site enabled. You can even boot a server in your laptop and, if configured to use the "-site" service, it can join the k8s cluster.

pruivo:
they can use and update it when they have the hostnames.
```yaml
locations: 
  - name: undefined
    url: infinsipan+xsite://127.0.0.1:11111
```

PR #1260, opened 2021-10-04
Rolling Upgrade Orchestration

gustavonalle:
~~Opening PR to let it pass through CI. Not 100% done yet, but I don't expect any fundamental changes to it. Feel free to review~~


controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		// Modify the DNS Ping name in the config
+		currentConfig, err := config.FromYaml(configMapObject.Data[consts.ServerConfigFilename])
+		if err != nil {
+			return err
+		}
+		currentConfig.Infinispan.ClusterName = r.targetName
+		currentConfig.JGroups.DNSPing.Query = fmt.Sprintf("%s-ping.%s.svc.cluster.local", r.targetName, namespace)
+		yaml, err := currentConfig.Yaml()
+		if err != nil {
+			r.reqLogger.Error(err, "error converting current config to Yaml")
```
ryanemerson, line 610: We should return the `err` here as we can't recover from this.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+}
+
+// reconcileNewPingService Reconcile a specific ping service for the pods of the new statefulSet
+func (r *LiveUpdateRequest) reconcileNewPingService() error {
+	namespace := r.infinispan.GetNamespace()
+	pingService := &corev1.Service{}
+	newPingServiceName := fmt.Sprintf("%s-ping", r.targetName)
+	err := r.Get(r.ctx, types.NamespacedName{Namespace: namespace, Name: newPingServiceName}, pingService)
+
+	if err != nil && errors.IsNotFound(err) {
```
ryanemerson, line 552: Same as `reconcileNewConfigMap` `!errors.IsNotFound(err)` needs to be handled.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		if err != nil {
+			r.reqLogger.Error(err, fmt.Sprintf("Could not find ping service '%s'", r.infinispan.GetPingServiceName()))
+			return err
+		}
+		// Clone existing service changing the DNS Query
+		newPingService := pingService.DeepCopy()
+		labels := newPingService.Labels
+		newPingService.ObjectMeta.ResourceVersion = ""
+		newPingService.Name = newPingServiceName
+		newPingService.Labels = labels
```
ryanemerson, line 565: I'm not sure why this is necessary? It seems that `labels` is already `newPingService.Labels`. Is this supposed to be `labels := pingService.Labels`?

No need for `labels` var, assignment can be direct to `newPingService.Labels = pingService.Labels`.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		selector := pingService.Spec.Selector
```
ryanemerson, line 566: Redundant var

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		return reconcile.Result{}, err
+	}
+
+	// Obtain existing statefulSet
+	namespace := r.infinispan.Namespace
+	statefulSetName := r.infinispan.GetStatefulSetName()
+	sourceStatefulSet := &appsv1.StatefulSet{}
+	err := r.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: statefulSetName}, sourceStatefulSet)
+
+	if err != nil && errors.IsNotFound(err) {
```
ryanemerson, line 160: `!errors.IsNotFound(err)` needs to be handled.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		r.reqLogger.Error(err, "Could not find existing statefulSet'%s'", statefulSetName)
+		return reconcile.Result{}, err
+	}
+
+	// Create a new statefulSet based on the existing one
+	targetStatefulSet := &appsv1.StatefulSet{}
+	err = r.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: r.targetName}, targetStatefulSet)
+	if err != nil && errors.IsNotFound(err) {
```
ryanemerson, line 168: `!errors.IsNotFound(err)` needs to be handled.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		targetStatefulset := sourceStatefulSet.DeepCopy()
+		volumes := targetStatefulset.Spec.Template.Spec.Volumes
+		// Change the config name
+		for _, volume := range volumes {
+			if volume.Name == ConfigVolumeName {
+				volume.ConfigMap.Name = r.newConfigName
+			}
+		}
+		targetStatefulset.Name = r.targetName
+		targetStatefulset.ObjectMeta.ResourceVersion = ""
```
ryanemerson, line 178: I think we probably need to wipe everything inside `ObjectMeta` except the annotations, labels, name and namespace to be on the safe side. For example we don't want the old UUID field to be retrained. So it might be easier to just create a new ObjectMeta struct with ^ values from the old StatefulSet.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		container.Image = consts.DefaultImageName
+		env := container.Env
+		for o, envVar := range env {
+			if envVar.Name == "DEFAULT_IMAGE" {
+				a := corev1.EnvVar{
+					Name:      envVar.Name,
+					Value:     consts.DefaultImageName,
+					ValueFrom: envVar.ValueFrom,
+				}
+				env[o] = a
```
ryanemerson, line 190: No need for `a` var.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+				break
+			}
+		}
+		container.Env = env
```
ryanemerson, line 194: If we use `container.Env` directly then there's no need for this reassignment or the `env` var.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+
+		err = r.Create(r.ctx, targetStatefulset)
+		if err != nil {
+			r.reqLogger.Error(err, "failed to create new statefulSet '%s'", r.targetName)
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{RequeueAfter: consts.DefaultWaitClusterNotWellFormed}, err
```
ryanemerson, line 201: `DefaultWaitOnCluster`?

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+	// If the new statefulSet was already created, check if the pods are ready
+	podList, err := r.podList(targetStatefulSet)
```
ryanemerson, line 204: Can't we use `PodListStatefulSet` and remove the `podList` func from this controller?

controllers/infinispan_controller.go
```diff
@@ -848,9 +863,36 @@ func IsUpgradeRequired(infinispan *infinispanv1.Infinispan, kube *kube.Kubernete
 	return upgradeRequired(infinispan, podList)
 }
 
-func PodList(infinispan *infinispanv1.Infinispan, kube *kube.Kubernetes, ctx context.Context) (*corev1.PodList, error) {
+func PodListStatefulSet(statefulSetName, namespace, infinispanName string, kube *kube.Kubernetes, ctx context.Context) (*corev1.PodList, error) {
```
ryanemerson, line 866: Lets move this and `PodList` to `pod.go` to reduce the clutter in `infinispan_controller.go`.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+			return reconcile.Result{RequeueAfter: consts.DefaultWaitOnCluster}, err
+		}
+		// Add a "version" label to pods
+		_, err = controllerutil.CreateOrUpdate(r.ctx, r.Client, &pod, func() error {
+			if pod.CreationTimestamp.IsZero() {
+				return errors.NewNotFound(corev1.Resource(""), pod.Name)
+			}
+			labels := pod.ObjectMeta.Labels
+			labels["version"] = strings.Split(consts.DefaultImageName, ":")[1]
+			pod.Labels = labels
```
ryanemerson, line 222: Can be a one liner `pod.ObjectMeta.Labels["version"] = `

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+			return nil
+		})
+	}
+
+	// Move to the next stage
+	return r.updateStage(ispnv1.LiveUpgradeStart)
```
ryanemerson, line 228: We never check that the number of ready pods is equal to the number of replicas in the StatefulSet spec. As the StatefulSet spec brings up the cluster a pod at a time, I think if the timing is wrong it's possible for replica = 2, but `len(podList.Items) == 1`, resulting in the controller progressing to `LiveUpgradeStart` prematurely.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+	if err != nil {
+		r.reqLogger.Error(err, "Failed to obtain pods from source cluster")
+		return reconcile.Result{}, err
+	}
+	podNameSource := podListSource.Items[0].Name
+
+	// Obtain the podName from the new stateFul to invoke requests
+	targetStatefulSet := &appsv1.StatefulSet{}
+	err = r.Get(r.ctx, types.NamespacedName{Namespace: r.infinispan.GetNamespace(), Name: r.targetName}, targetStatefulSet)
+	if err != nil && errors.IsNotFound(err) {
```
ryanemerson, line 244: `!errors.IsNotFound(err)` needs to be handled.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+	// Move to next stage
+	return r.updateStage(ispnv1.LiveUpgradeRouteRedirected)
+}
+
+// syncData Copies data from the source cluster to the destination statefulSet
+func (r *LiveUpdateRequest) syncData() (ctrl.Result, error) {
+	// Obtain the podName from the new stateFul to invoke requests
+	targetStatefulSet := &appsv1.StatefulSet{}
+	err := r.Get(r.ctx, types.NamespacedName{Namespace: r.infinispan.GetNamespace(), Name: r.targetName}, targetStatefulSet)
+	if err != nil && errors.IsNotFound(err) {
```
ryanemerson, line 363: `!errors.IsNotFound(err)` needs to be handled.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		if err != nil {
+			r.reqLogger.Error(err, "Failed to update service")
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Remove version label from the pods
+	targetStatefulSet := &appsv1.StatefulSet{}
+	err := r.Get(r.ctx, types.NamespacedName{Namespace: r.infinispan.GetNamespace(), Name: r.targetName}, targetStatefulSet)
+	if err != nil && errors.IsNotFound(err) {
```
ryanemerson, line 432: `!errors.IsNotFound(err)` needs to be handled.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+	if err != nil {
+		r.reqLogger.Error(err, "Error Deleting source ping service")
+		return reconcile.Result{}, err
+	}
+
+	// Delete old statefulSet
+	sourceStatefulSet := &appsv1.StatefulSet{}
+	err = r.Get(context.TODO(), types.NamespacedName{Namespace: ispn.Namespace, Name: ispn.Name}, sourceStatefulSet)
+
+	if err != nil && errors.IsNotFound(err) {
```
ryanemerson, line 490: `!errors.IsNotFound(err)` needs to be handled.

ryanemerson (changes requested):

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+					return reconcile.Result{}, err
+				}
+				err = r.cluster.CreateCache(cacheName, configuration, podName)
+				if err != nil {
+					r.reqLogger.Error(err, fmt.Sprintf("failed to create cache '%s'", cacheName))
+					return reconcile.Result{}, err
+				}
+				r.reqLogger.Info(fmt.Sprintf("Cache '%s' created", cacheName))
+			} else {
+				r.reqLogger.Info(fmt.Sprintf("Cache '%s' already existed", cacheName))
```
ryanemerson, line 295: "already exists"

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+			sourceClusterService.Spec.Selector = selector
+			return nil
+		})
+		if err != nil {
+			r.reqLogger.Error(err, "Failed to update service")
+			return reconcile.Result{}, err
+		}
+	}
+
+	//TODO: Redirect other kind of services
```
ryanemerson, line 352: Why services did you have in mind?

controllers/liveupdate_controller.go
ryanemerson: Can we use `kube.LookupResource` here? I suspect there will be other cases where this is possible as well.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+	// Obtain the admin service in the source cluster to create the remote store config
+	sourceClusterService := &corev1.Service{}
+	res, err := kube.LookupResource(r.infinispan.GetAdminServiceName(), r.infinispan.GetNamespace(), sourceClusterService, r.Client, r.reqLogger, r.eventRec, r.ctx)
+
+	if res != nil {
+		r.reqLogger.Error(err, "failed to get service admin to connect to source cluster")
+		return reconcile.Result{}, err
+	}
+
+	for _, cacheName := range names {
```
ryanemerson, line 274: The logic from here on in the method concerns interactions with Infinispan, not k8s resources. Maybe a more appropriate place for this logic would be a method(s) in `remote_store.go`. This func could take `r.cluster`, `sourceClusterService` and `podName` as args and then perform the required actions there. That way we're separating the k8s and Infinispan logic. It also means that we don't have to return `reconcile.Result{}, err` etc for every stage of the upgrade.

The same applies for most of the cache logic in the `syncData` function.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+func (r *LiveUpdateRequest) destroy() (ctrl.Result, error) {
+	// Remove version label from the ping service
+	sourceClusterService := &corev1.Service{}
+	ispn := r.infinispan
+	if result, err := kube.LookupResource(ispn.Name, ispn.Namespace, sourceClusterService, r, r.log, r.eventRec, r.ctx); result != nil {
+		return *result, err
+	}
+
+	selector := sourceClusterService.Spec.Selector
+	if selector["version"] != "" {
```
ryanemerson, line 414: Why do we need to update the `sourceClusterService` labels here if we're about to delete the service?

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+			return nil
+		})
+		if err != nil {
+			r.reqLogger.Error(err, "Failed to update service")
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Remove version label from the pods
+	targetStatefulSet := &appsv1.StatefulSet{}
```
ryanemerson, line 430: Can we remove the need for this step if we update the logic so that instead of checking if `labels["version"] != ""` we instead check that if `labels["version"] == $currentVersion`.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+	sourceStatefulSet := &appsv1.StatefulSet{}
+	err = r.Get(context.TODO(), types.NamespacedName{Namespace: ispn.Namespace, Name: ispn.Name}, sourceStatefulSet)
+
+	if err != nil && errors.IsNotFound(err) {
+		r.reqLogger.Error(err, "source stateful set not found")
+		return reconcile.Result{}, err
+	}
+
+	err = r.Delete(r.ctx, sourceStatefulSet)
+	if err != nil {
```
ryanemerson, line 496: There's no need to do both `GET` and `DELETE` here. We can just call `DELETE` and if  `errors.IsNotFound` continue.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+		r.reqLogger.Error(err, "Error Deleting statefulSet")
+		return reconcile.Result{}, err
+	}
+
+	// Remove any LiveUpdate status from the Infinispan CR
+	_, err = controllerutil.CreateOrPatch(r.ctx, r, ispn, func() error {
+		ispn.Status.LiveUpgradeStage = ""
```
ryanemerson, line 503: There's multiple updates to the `ispn.Status` in this method. I think we need to either return on the first call and requeue another reconcilliation, or add the `ispn.Status.StatefulSetName = r.targetName` here.

pkg/infinispan/cluster.go
```diff
@@ -62,6 +70,8 @@ type ClusterInterface interface {
 	ExistsCache(cacheName, podName string) (bool, error)
 	CreateCacheWithTemplate(cacheName, cacheXML, podName string) error
 	CreateCacheWithTemplateName(cacheName, templateName, podName string) error
+	CreateCache(cacheName, jsonConfig, podName string) error
+	GetCacheConfig(cacheName, podName string) (cacheConfig string, err error)
```
ryanemerson, line 74: We need to make sure that all of the new methods are added to this interface so that we can access via `infinispan.ClusterInterface`.

controllers/liveupdate_controller.go
```diff
@@ -0,0 +1,653 @@
...
+	kubernetes *kube.Kubernetes
+	eventRec   record.EventRecorder
+}
+
+type LiveUpdateRequest struct {
+	*LiveUpdateReconciler
+	ctx           context.Context
+	infinispan    *ispnv1.Infinispan
+	reqLogger     logr.Logger
+	cluster       *infinispan.Cluster
```
ryanemerson, line 44: This should be `infinispan.ClusterInterface`

pkg/infinispan/cluster.go
```diff
@@ -241,6 +251,99 @@ func (c Cluster) CreateCacheWithTemplateName(cacheName, templateName, podName st
...
+	}()
+	var body json.RawMessage
+	if err = json.NewDecoder(rsp.Body).Decode(&body); err != nil {
+		return "", fmt.Errorf("unable to decode: %w", err)
+	}
+	return string(body), nil
+}
+
+// AddRemoteStore connects the target cluster to the source cluster using a Remote Store
+func (c Cluster) AddRemoteStore(cacheName, jsonConfig, podName string) error {
```
ryanemerson, line 282: `AddRemoteStore`, `IsSourceConnected`, `DisconnectSource` and `SyncData` are all very specific Rolling upgrades. It would be nice if we had another interface/implementation specific for Rolling upgrades to separate concerns better (something we haven't been great with so far).

gustavonalle:
Added documentation @oraNod

documentation/asciidoc/topics/ref_upgrades.adoc
```diff
@@ -26,3 +26,25 @@ ifdef::downstream[]
 If you upgrade {brandname} clusters manually and have upgraded the channel for your {ispn_operator} subscription from **8.1.x** to **8.2.x** you should apply the upgrade for the latest {brandname} 8.2.x version as soon as possible to avoid potential data loss that can result from an issue in 8.2.0.
 For more information, see link:{relnotes_openshift}[{ispn_operator} {ProductVersion} Release Notes].
 endif::downstream[]
+
+[discrete]
+== Upgrade types
+//Community Only
```
oraNod, line 32: @gustavonalle Should all of this be community only, including the 2nd level header?

I think it would be worth putting this into a standalone file `ref_upgrade_types.adoc` then you can wrap the includes directive in the assembly like so:

```
ifdef::community[]
include::{topics}/ref_upgrade_types.adoc[]
endif::community[]
```

documentation/asciidoc/topics/ref_upgrades.adoc
```diff
@@ -26,3 +26,25 @@ ifdef::downstream[]
...
+ifdef::community[]
+The Infinispan CR path ```spec.upgrades.type``` defines how the upgrades are done:
```
oraNod, line 34: The `spec.upgrades.type` field controls how {ispn_operator} upgrades your {brandname} cluster when new versions become available.

documentation/asciidoc/topics/ref_upgrades.adoc
```diff
@@ -26,3 +26,25 @@ ifdef::downstream[]
...
+
+[%header,cols=2*]
+|===
+|Value
+|Description
+
+|`HotRodRolling`
+|Upgrades the cluster with no downtime using Hot Rod Rolling Upgrades. The operator first creates a new statefulSet with pods in the new version, that use existing pods as a remote store; then redirects all clients to the new pods while copying all the data from the old cluster, disposing the original pods in the end.
```
oraNod, line 42: Should we say something here about the fact this requires twice the capacity as you'll basically have two clusters?

oraNod, line 42: "Upgrades the cluster with no downtime using Hot Rod Rolling Upgrades. The operator first creates a new statefulSet with pods in the new version, that use existing pods as a remote store; then redirects all clients to the new pods while copying all the data from the old cluster, disposing the original pods in the end."

Upgrades {brandname} clusters without downtime. {ispn_operator} creates a new {brandname} cluster with the target version and uses a remote store to copy data from the existing cluster. During the upgrade {ispn_operator} also redirects all clients to the new cluster. When it completes the upgrade process, {ispn_operator} removes the old {brandname} cluster.

documentation/asciidoc/topics/ref_upgrades.adoc
```diff
@@ -26,3 +26,25 @@ ifdef::downstream[]
...
+
+|`Shutdown`
+|This is the default migration type. The operator issues an orderly shutdown of the cluster, and pods are started with the new version.
```
oraNod, line 45: Upgrades {brandname} clusters with downtime. {ispn_operator} shuts down the existing cluster before removing it, then creates a new cluster with the target version. This is the default upgrade type.

documentation/asciidoc/topics/ref_upgrades.adoc
```diff
@@ -26,3 +26,25 @@ ifdef::downstream[]
...
+
+|`spec.service.type`
```
oraNod, line 47: Should we remove this field from the table?

documentation/asciidoc/topics/ref_upgrades.adoc
```diff
@@ -26,3 +26,25 @@ ifdef::downstream[]
...
+|Configures the type {brandname} service. A value of `Cache` creates a cluster with {cacheservice} pods.
+|===
+endif::community[]
```
oraNod, line 50: Should we mention anything about data loss/persistence with upgrades? For HR rolling upgrades the operator migrates data across so you don't need a cache store. For offline upgrades, do you need to first create a cache store and back up data to the persistent volume?

oraNod (changes requested):

ryanemerson:
> Test failures are also happening on `main`

The upgrade failures aren't happening on `main`. Looking at the CI logs, I see:

```
14:40:01  E1105 14:38:03.462991       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
```

controllers/hotrod_upgrade_controller.go
```diff
@@ -0,0 +1,702 @@
...
+	}
+
+	// Skip for x-site
+	if ispn.HasSites() {
+		return reconcile.Result{}, errors.New("hot rod rolling upgrades not supported for clusters with X-Site")
+	}
+
+	// Skip for non-datagrid
+	if ispn.Spec.Service.Type != ispnv1.ServiceTypeDataGrid {
+		return reconcile.Result{}, errors.New("hot rod rolling upgrades only supported for ServiceTypeDataGrid")
```
ryanemerson, line 99: I see in the CI logs this is being output several times in tests that don't use `HotRodRolling`: https://ci.infinispan.org/job/Infinispan%20Operator/job/PR-1260/103/consoleFull

documentation/asciidoc/topics/ref_upgrades.adoc
gustavonalle: ops

documentation/asciidoc/topics/ref_upgrades.adoc
gustavonalle: ok

gustavonalle:
@oraNod I committed ```ref_upgrade_types.adoc``` but the  underlying structure of the doc changed and I apparently there is no upgrade section anymore, I will let you wire it later :smile:

oraNod:
> @oraNod I committed `ref_upgrade_types.adoc` but the underlying structure of the doc changed and apparently there is no upgrade section anymore, I will let you wire it later smile

@gustavonalle That sounds good to me, thanks. I did remove the upgrading topic in a recent PR because it was a bit useless for the 2.2.1 branch. Cheers. :beers:

oraNod (approved):

controllers/hotrod_upgrade_controller.go
gustavonalle: that's weird, looking...

gustavonalle: It is a race condition where the upgrade field is still ```nil``` and ```ApplyDefaults()``` hasn't been called yet. Will fix it

controllers/pods.go
```diff
@@ -223,3 +227,43 @@ func AddSecretVolume(secretName, volumeName, mountPath string, spec *corev1.PodS
...
+		return podList, nil
+	}
+	// Obtain pod list associated with the cluster
+	err = kube.ResourcesList(namespace, PodLabels(infinispan.GetName()), podList, ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter out pods not owned by the statefulSet
+	ownerId := stateFulSet.GetUID()
```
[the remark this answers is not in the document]
gustavonalle: I think so, this is a leftover from previous incarnations where I was using the version as label

gustavonalle:
> > Test failures are also happening on `main`
> 
> The upgrade failures aren't happening on `main`. Looking at the CI logs, I see:
> 
> ```
> 14:40:01  E1105 14:38:03.462991       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
> ```

It passed yesterday, are you sure it was introduced by this PR? Also, I can't run the tests locally as they always time out in the start

ryanemerson:
> I can't run `make upgrade-test` locally as it always time out at the start

How are you installing `kind`? You need todo:

1. `./scripts/ci/kind-with-olm.sh `
2. `./scripts/ci/install-catalog-source.sh`
3. `make upgrade-test`

ryanemerson:
> Apparently I need to setup a private registry first

Unfortunately yes, as we need to create a new image for the `CatalogSource` that includes the latest code in a OLM bundle. That `CatalogSource` is then used by OLM for the Operator installation and upgrading.

gustavonalle:
> > Apparently I need to setup a private registry first
> 
> Unfortunately yes, as we need to create a new image for the `CatalogSource` that includes the latest code in a OLM bundle. That `CatalogSource` is then used by OLM for the Operator installation and upgrading.

@ryanemerson I installed a private registry, ```./scripts/ci/install-catalog-source.sh```run successfully, but the test fails for me with a ```CrashLoopBackOff``` 
```
namespace-for-testing   12s         Warning   Failed                                                      pod/test-upgrade-1                                                    Error: failed to create containerd task: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "-Dinfinispan.zero-capacity-node=false": executable file not found in $PATH: unknown
```

Note: this is running on ```main```

What am I missing?

gustavonalle:
@ryanemerson Trying out the test here locally

controllers/pods.go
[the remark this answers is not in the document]
gustavonalle: I will need to revert back from your suggestion. Unfortunately we can't expect the label to be always present, e.g., when the cluster was created by an older operator version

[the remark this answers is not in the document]
gustavonalle: This was causing failures in the upgrade-test

gustavonalle:
@ryanemerson I believe the failure should be fixed now, let's wait and see

ryanemerson:
Thanks @gustavonalle

PR #1262, opened 2021-10-05
Add version label to pods and services

### LINKED ISSUE #1261: Pods and service labels should contain the version
During the upcoming rolling upgrade feature https://github.com/infinispan/infinispan-operator/pull/1260, a new statefulSet will be created and will live in parallel to the existing one, for the duration of the data migration.

To be able to correctly target pods from those 2 statefulSets, it's necessary that the services and the pods carry an extra label "version"

gustavonalle:
Fix #1261


controllers/labels.go
```diff
@@ -17,6 +27,7 @@ func ServiceLabels(name string) map[string]string {
 	return map[string]string{
 		"clusterName": name,
 		"app":         "infinispan-pod",
+		"version":     version(),
```
ryanemerson, line 30: I think we should be consistent here with the label names and use "infinispan_version".

dmvolod: label name should conform k8s recommended rules https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ and could be like
`infinispan.org/version`

api/v1/types_util.go
```diff
@@ -520,6 +520,10 @@ func (ispn *Infinispan) AddLabelsForPods(uMap map[string]string) {
 	addLabelsFor(ispn, PodTargetLabels, uMap)
 }
 
+func (ispn *Infinispan) AddVersionLabelForPods(uMap map[string]string) {
+	uMap["infinispan_version"] = strings.Split(consts.DefaultImageName, ":")[1]
```
dmvolod, line 524: label name should conform k8s recommended rules https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/ and could be like
`infinispan.org/version`

ryanemerson: That's true @dmvolod but unfortunately none of our existing labels follow this strategy atm. I'll create a GH issue to address this in a different PR.

ryanemerson: The additional labels can be handled in another PR https://github.com/infinispan/infinispan-operator/issues/597

dmvolod: coordinator is living out his last days and will be removed in #1207 
Others need to conform, but for Pod labels selectors it's a bit dangerous for Rolling Upgrade, as controllers are working in non determined sequence (the same as golang routines) service could be updated early than Pod labels and kube-dns could disable network connectivity for some moments due to inconsistent labels matching

PR #1268, opened 2021-10-14
Use statefulSet name instead of version to label pods

gustavocoding:
Some deloyments use image name as a SHA string and it overflows the maximum label size.


gustavonalle:
Let's try a third one

PR #1279, opened 2021-10-20
removing nested assemblies

oraNod:
Cannot have nested assemblies downstream so need to adjust the cross-site assembly.


oraNod:
@pruivo Some minor doc changes here that compact some of the details for setting up service account tokens. Doc only PR but feel free to comment. Thanks.

ryanemerson:
Thanks @oraNod

PR #1283, opened 2021-10-21
Replace panic with assert and require

### LINKED ISSUE #1217: Testsuite: Utilise testing.Fail instead of panic
Currently all of our tests are failed by using `panic()`. Instead, we should leverage the `.Fail()` and `.FailNow()`[0] commands. This will allow subsequent tests to continue executing after a failure.

[0] https://pkg.go.dev/testing#B.Fail
https://ieftimov.com/post/testing-in-go-failing-tests/

Crumby:
Resolves #1217 by using [assert](https://pkg.go.dev/github.com/stretchr/testify/assert) and [require](https://pkg.go.dev/github.com/stretchr/testify/require) packages which call `fail()` or `failNow()` instead of `Panic()`. As a result test suite execution shouldn't be halted by failed test.


Crumby:
Closing in favor of #1376.

PR #1294, opened 2021-10-26
Add support for cross-site communication using Openshift Routes

pruivo:
Enable Openshift Route for cross-site communication. 

Currently, I'm unable to add tests for CI. :/

### DIFF
--- a/config/crd/bases/infinispan.org_infinispans.yaml
+++ b/config/crd/bases/infinispan.org_infinispans.yaml
@@ -921,6 +921,10 @@ spec:
                               port:
                                 format: int32
                                 type: integer
+                              routeHostName:
+                                description: RouteHostName optionally, specifies a
+                                  custom hostname to be used by Openshift Route.
+                                type: string
                               type:
                                 description: Type specifies different exposition methods
                                   for data grid
@@ -928,6 +932,7 @@ spec:
                                 - NodePort
                                 - LoadBalancer
                                 - ClusterIP
+                                - Route
                                 type: string
                             required:
                             - type
--- a/controllers/infinispan_service_controller.go
+++ b/controllers/infinispan_service_controller.go
@@ -142,8 +142,24 @@ func (reconciler *ServiceReconciler) Reconcile(ctx context.Context, request reco
 	}
 
 	if s.infinispan.HasSites() {
-		if err := s.reconcileResource(computeSiteService(s.infinispan)); err != nil {
-			return reconcile.Result{}, err
+		switch s.infinispan.GetCrossSiteExposeType() {
+		case ispnv1.CrossSiteExposeTypeClusterIP, ispnv1.CrossSiteExposeTypeNodePort, ispnv1.CrossSiteExposeTypeLoadBalancer:
+			// just a single external service
+			if err := s.reconcileResource(computeSiteService(s.infinispan, s.infinispan.GetCrossSiteExposeType())); err != nil {
+				return reconcile.Result{}, err
+			}
+		case ispnv1.CrossSiteExposeTypeRoute:
+			if s.isTypeSupported(consts.ExternalTypeRoute) {
+				// create internal service
+				if err := s.reconcileResource(computeSiteService(s.infinispan, ispnv1.CrossSiteExposeTypeClusterIP)); err != nil {
+					return reconcile.Result{}, err
+				}
+				if err := s.reconcileResource(computeSiteRoute(s.infinispan)); err != nil {
+					return reconcile.Result{}, err
+				}
+			} else {
+				return reconcile.Result{}, fmt.Errorf("route cross-site expose type is not supported")
+			}
 		}
 	} else {
 		siteService := &corev1.Service{
@@ -480,14 +496,14 @@ func computeServiceExternal(ispn *ispnv1.Infinispan) *corev1.Service {
 }
 
 // computeSiteService compute the XSite service
-func computeSiteService(ispn *ispnv1.Infinispan) *corev1.Service {
+func computeSiteService(ispn *ispnv1.Infinispan, exposeType ispnv1.CrossSiteExposeType) *corev1.Service {
 	lsPodSelector := GossipRouterPodLabels(ispn.Name)
 
 	exposeSpec := corev1.ServiceSpec{}
 	exposeConf := ispn.Spec.Service.Sites.Local.Expose
 	exposeSpec.Selector = lsPodSelector
 
-	switch exposeConf.Type {
+	switch exposeType {
 	case ispnv1.CrossSiteExposeTypeNodePort:
 		exposeSpec.Type = corev1.ServiceTypeNodePort
 		exposeSpec.Ports = []corev1.ServicePort{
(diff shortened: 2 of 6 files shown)


controllers/xsite_config.go
```diff
@@ -112,18 +139,40 @@ func appendRemoteLocation(ctx context.Context, infinispan *ispnv1.Infinispan, re
...
+	if err == nil {
+		host, port, err = getCrossSiteServiceHostPort(siteService, remoteKubernetes, logger, eventRec, "XSiteRemoteServiceUnsupported", ctx)
+		if err != nil {
+			logger.Error(err, "error retrieving remote x-site service information")
+			return err
+		}
+	} else {
+		// Try searching for the Route
+		siteRoute := &routev1.Route{}
+		err = remoteKubernetes.Client.Get(context.TODO(), types.NamespacedName{Name: remoteRouteName, Namespace: remoteNamespace}, siteRoute)
```
ryanemerson, line 169: We need to use the `ctx` provided as a parameter.

Makefile
```diff
@@ -134,7 +134,7 @@ generate: controller-gen rice
 .PHONY: operator-build
 ## Build the operator image
 operator-build: manager
-	$(CONTAINER_TOOL) build -t $(IMG) .
+	$(CONTAINER_TOOL) build --build-arg VERSION=$(VERSION) -t $(IMG) .
```
ryanemerson, line 137: Why is this necessary?

ryanemerson (changes requested):

ryanemerson:
> Currently, I'm unable to add tests for CI. :/

You could add a test that works with Openshift. We can then perform a quick check at the start of the test method to determine if the k8s flavour supports Routes and if not calls `t.Skip()`.

Makefile
pruivo: There is a `VERSION` argument in the Dockerfile that is not set and it does not have a default value. 
Isn't an issue?

Podman does not like it
```
$ make operator-build CONTAINER_TOOL=podman
/home/pedro/Work/infinispan-operator/bin/controller-gen object paths="./..."
/home/pedro/Work/infinispan-operator/bin/rice embed-go -i controllers/dependencies.go -i controllers/grafana.go
find . -type f -name 'rice-box.go' -exec sed -i "s|time.Unix(.*, 0)|time.Unix(1620137619, 0)|" {} \;
go fmt ./...
go vet ./...
go build -o bin/manager main.go
podman build -t quay.io/infinispan/operator:latest .
WARN[0001] missing "VERSION" build argument. Try adding "--build-arg VERSION=<VALUE>" to the command line 
```

pruivo: actually, `VERSION` is set in the base image.
Adding `RUN echo ${VERSION}` I got (with docker)
```
Step 1/11 : FROM registry.access.redhat.com/ubi8/go-toolset:1.15.14 AS build
...
Step 3/11 : RUN echo ${VERSION}
 ---> Running in 6aeb06d29666
1.15.14
```
Is it intentional?

ryanemerson: > There is a `VERSION` argument in the Dockerfile that is not set and it does not have a default value.
> Isn't an issue?

No it is, good spot :+1: 

> actually, `VERSION` is set in the base image. Adding `RUN echo ${VERSION}` I got (with docker)
> Is it intentional?

Nope :sweat_smile:

controllers/xsite_config.go
pruivo: fixed!

pruivo:
> We can then perform a quick check at the start of the test method to determine if the k8s flavour supports Routes

I need time to investigate how to do it :)

ryanemerson:
> I need time to investigate how to do it :)

We already have a method to determine if a GVK is supported on the k8s platform: https://github.com/infinispan/infinispan-operator/blob/main/pkg/kubernetes/kubernetes.go#L72-L92

ryanemerson:
@pruivo I knew I had already implemented something similar for the testsuite, but couldn't find it ... you can use the `tutils.SkipForExposeType` to ensure a test is only executed when the `EXPOSE_SERVICE_TYPE=Route`.

https://github.com/ryanemerson/infinispan-operator/commit/c8219e311f827d226a5a34c53427441e398f3833

pruivo:
it looks like that commit is not available in the operator. Do you mind if I cherry-pick it to this PR?

ryanemerson:
@pruivo That explains why I couldn't find the code :upside_down_face:.  Go for it!

config/crd/bases/infinispan.org_infinispans.yaml
```diff
@@ -921,13 +921,18 @@ spec:
                               port:
                                 format: int32
                                 type: integer
+                              routerHostName:
```
Crumby, line 924: Here is inconsistency between field name and the description. While the field is `routerHostName` the description mentions `RouteHostName` (route vs router), probably should be `route`

controllers/infinispan_service_controller.go
```diff
@@ -538,6 +554,39 @@ func computeSiteService(ispn *ispnv1.Infinispan) *corev1.Service {
...
+			APIVersion: "route.openshift.io/v1",
+			Kind:       "Route",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ispn.GetSiteRouteName(),
+			Namespace: ispn.Namespace,
+			Labels:    LabelsResource(ispn.Name, "infinispan-route-xsite"),
+		},
+		Spec: routev1.RouteSpec{
+			Host: host,
```
Crumby, line 570: What if the `host` is empty?

test/e2e/xsite/xsite_test.go
```diff
@@ -231,6 +231,16 @@ func TestDefaultTLSLoadBalancerWithPort(t *testing.T) {
 	testCrossSiteView(t, true, ispnv1.CrossSiteSchemeTypeOpenShift, ispnv1.CrossSiteExposeTypeLoadBalancer, 1443, 1, DefaultTLS, nil)
 }
 
+func TestCrossSiteViewOpenshiftRoute(t *testing.T) {
+	tutils.SkipForExposeType(t, ispnv1.ExposeTypeNodePort, ispnv1.ExposeTypeLoadBalancer)
```
Crumby, line 235: Maybe skipping based on whether `Route` API is available would be more consistent with the rest of the test rather then on test suite configuration? (We are not skipping LoadBalancer tests just because NodePort expose type is configured)

test/e2e/xsite/xsite_test.go
```diff
@@ -231,6 +231,16 @@ func TestDefaultTLSLoadBalancerWithPort(t *testing.T) {
...
+	testCrossSiteView(t, true, ispnv1.CrossSiteSchemeTypeOpenShift, ispnv1.CrossSiteExposeTypeRoute, 0, 1, NoTLS, nil)
```
Crumby, line 236: `NoTLS` option is confusing. Cross site over Route won't work without TLS

test/e2e/xsite/xsite_test.go
```diff
@@ -231,6 +231,16 @@ func TestDefaultTLSLoadBalancerWithPort(t *testing.T) {
 	testCrossSiteView(t, true, ispnv1.CrossSiteSchemeTypeOpenShift, ispnv1.CrossSiteExposeTypeLoadBalancer, 1443, 1, DefaultTLS, nil)
 }
 
+func TestCrossSiteViewOpenshiftRoute(t *testing.T) {
```
Crumby, line 234: Having internal version of the test for smoke testing on a single cluster would be nice

pruivo: that is useless. Cross-site in a single cluster uses the cluster IP directly and not the Routes (same for NodePort & LoadBalancer)

test/e2e/xsite/xsite_test.go
pruivo: I'm going to remove this test. It does not make sense to me :rofl:

test/e2e/xsite/xsite_test.go
pruivo: any pointers how to do that?

ryanemerson: We can make use of the `IsGroupVersionSupported` method like [here](https://github.com/infinispan/infinispan-operator/blob/main/test/e2e/utils/common.go#L216) to see if it's supported and if not call skip on the test.

test/e2e/xsite/xsite_test.go
Crumby: Right, I incorrectly though we have "single cluster" version of the tests for NodePort and LoadBalancer too.

controllers/infinispan_service_controller.go
Crumby: Right, I was more wondering whether OpenShift won't complain if you send a request with explicit `host: nil`, compared to "not mentioning" `host` at all. But if it works then good, it's definitely nicer then having if block.

pruivo:
Updated!
* Fixed `TestCrossSiteViewKubernetesLoadBalancer`
* Tested the skip() method
```
=== RUN   TestDefaultTLSOpenshiftRoute
    xsite_test.go:297: Route not available. Skipping test
--- SKIP: TestDefaultTLSOpenshiftRoute (6.09s)
```

pruivo:
green :partying_face:

ryanemerson:
Thanks @pruivo

PR #1297, opened 2021-10-26
Rename Gossip Router deployment name

pruivo:
* "router pod" sound better instead of "tunnel pod"

### DIFF
--- a/api/v1/types_util.go
+++ b/api/v1/types_util.go
@@ -50,7 +50,7 @@ const (
 	SiteRouteNameSuffix     = "-route-site"
 	SiteServiceFQNTemplate  = "%s.%s.svc.cluster.local"
 
-	GossipRouterDeploymentNameTemplate = "%s-tunnel"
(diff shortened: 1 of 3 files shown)


ryanemerson (approved):
Changes are fine by me. Just waiting on another CI run as the xsite tests failed on the last one

ryanemerson:
@pruivo On an upgrade won't this leave the old `-tunnel` deployment orphaned? What will the consequences be if there's both the old and new deployment running concurrently?

pruivo:
I'm not sure but I would say yes.
Is there any function to perform deployments renaming?

pruivo:
@ryanemerson added code to remove the old deployment.

ryanemerson:
> Is there any function to perform deployments renaming?

Not that I'm aware of. I think the code you've added is the best we can do.

ryanemerson:
Thanks @pruivo

Issue #747, opened 2021-01-18
Cache CR Reconciliation


ryanemerson:
The Cache CR allows Infinispan caches to be created by other k8 resources, e.g. other operator. However in it's current form a Cache CR does not reflect the current status of a cache on the server:

  - A cache can be removed via a Infinispan client (CLI, Console, HotRod, Rest) and the CR will remain
  - Two-way mapping of caches and CRs requires polling of Infinispan state by the operator which does not scale as the number of caches increases

## Proposal
Reconcile Cache CR state by consuming REST events [ISPN-12606](https://issues.redhat.com/browse/ISPN-12606) and updating/creating Cache CRs. This allows CRs to be updated when CRUD operations are performed on a cache's state without polling.

Creation of new caches via any of the Infinispan clients will result in a corresponding Cache CR being created. Similarly, the removal of any caches will result in their CR being removed (or their status updated?).

## Implementation
Deploy a stateless "config-reconciler" pod per Infinispan CR. This pod is responsible for consuming events from the Infinispan service and updating/creating a corresponding Cache CR.  On startup/restart/failover, this pod must set `includeCurrentState=true` when connecting to the events endpoint to ensure that the latest state is reconciled.

Utilising an independent pod has the following advantages:
  - Consuming events directly in the operator does not scale in multi-namespace environments where multiple Infinispan clusters are deployed simultaneously (potentially wtih many caches per cluster).
  - It has no impact on horizontal scaling of clusters. Utilising a sidecar in the existing Infinispan pod's would require consensus amongst the pod's to determine who was responsible for reconciling configuration state.

### Config Reconciler Pod
The config-reconciler should have as small a footprint as possible, so we should deploy this as a natively compiled executable based on a scratch image.

### Reconciler per namespace
In order to reduce the total number of pods required by Infinispan, it could be possible to allow a "config-reconciler" pod to consume events from multiple Infinispan CRs in the same namespace. This could be configurable in the Infinispan CR. If enabled, it would be necessary for the config-reconciler pod spec to be updated and the pod restarted in order for the new cluster to be watched.

## Future Work
We should extend the Operator and config-reconciler to support the following types:

- templates
- counters
- schemas
- tasks

dmvolod:
> The config-reconciler should have as small a footprint as possible, so we should deploy this as a natively compiled executable based on a scratch image.

@ryanemerson, can we utilize golang for this implementation?

ryanemerson:
[...]

PR #1302, opened 2021-10-29
Split core testsuite so that all Cache CR tests belong in test/e2e/cache

ryanemerson:
This is required by the Cache reconciliation work https://github.com/infinispan/infinispan-operator/issues/747. Creating as an initial PR to reduce the size of the final PR.


ryanemerson:
Closing as more changes to come in this area, so I will just introduce as the one PR

PR #1303, opened 2021-11-02
Minor fixes and nits

leo8a:
Signed-off-by: Leo Ochoa lochoa@redhat.com


ryanemerson (approved):

pkg/apis/integreatly/v1alpha1/grafanadashboard_types.go
```diff
@@ -34,7 +34,7 @@ type GrafanaDashboardDatasource struct {
 	DatasourceName string `json:"datasourceName"`
 }
 
-// Used to keep a dashboard reference without having access to the dashboard
+// GrafanaDashboardRef Used to keep a dashboard reference without having access to the dashboard
```
dmvolod, line 37: Not sure that this fix make sense as all `pkg/apis/integreatly/v1alpha1` structures copied from the grafana operator and need to be fixed there before. If we will need to bump or replace it again, those changes will be overwritten.

dmvolod (changes requested):

pkg/apis/integreatly/v1alpha1/grafanadashboard_types.go
ryanemerson: Good spot @dmvolod. @leo8a Can you remove all of the files under ` pkg/apis/integreatly` from your commit.

leo8a: Sure, thanks both! I went ahead and made a PR fixing it upstream (see [grafana-operator#598](https://github.com/grafana-operator/grafana-operator/pull/598)), let's see how it goes.

ryanemerson:
Thanks @leo8a

PR #1308, opened 2021-11-03
[test-integration] Infinispan 13.0 updates

Crumby:
Updates client version and and stats parser in the integrations tests


ryanemerson:
Thanks @Crumby

PR #1311, opened 2021-11-10
Explicitly create ServiceAccount with `Make deploy`

ryanemerson:
This will need to be backported to `2.2.x`.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -22,7 +22,7 @@ spec:
       labels:
         control-plane: controller-manager
     spec:
-      serviceAccountName: infinispan-operator
+      serviceAccountName: controller-manager
       containers:
       - name: manager
         image: controller:latest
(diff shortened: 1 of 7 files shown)


pruivo (approved):

pruivo:
LGTM. I'm just wondering if `controller-manager` is a good name for the service account. Isn't possible for other operators to create a service account with the same name?
What would happen in that case?

ryanemerson:
> LGTM. I'm just wondering if `controller-manager` is a good name for the service account. Isn't possible for other operators to create a service account with the same name?

The created `ServiceAccount` is actually called `infinispan-operator-controller-manager`, which is the same as the operator deployment. The `infinispan-operator-` name prefix is added by [Kustomize](https://kustomize.io/) in `config/default/kustomization.yaml`.

pruivo:
:+1: nice! 
I don't have write access :) so I guess you or @rigazilla can press the button

PR #1315, opened 2021-11-15
Use pointer for optional cross-site encryption

pruivo:


ryanemerson:
@pruivo Does this need backporting for `2.2.x`?

PR #1316, opened 2021-11-15
Multi version client

ryanemerson:
Based on top of https://github.com/infinispan/infinispan-operator/pull/1310

Tasks:

- [x] Introduce Infinispan API for all operator->operand interactions: `pkg/infinispan/client/api`
- [x] Create Infinispan implementation for Infinispan 13
- [x] Convert controllers to use Infinispan API
- [x] Move `pkg/infinispan/http` -> `pkg/http`
- [x] Convert testsuite to use Infinispan API with `HttpClient` implementation for external routes.
- [x] Green CI


ryanemerson:
PR is ready for review. The only CI failures are xsite which already exists on `main`.

controllers/infinispan_controller.go
```diff
@@ -672,19 +680,20 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
 	if infinispan.HasSites() {
-		crossSiteViewCondition, err := r.GetCrossSiteViewCondition(podList, infinispan.GetSiteLocationsName(), cluster)
+		crossSiteViewCondition, err := r.GetCrossSiteViewCondition(podList, infinispan.GetSiteLocationsName(), curl)
 		if err != nil {
 			return ctrl.Result{}, err
 		}
 		// ISPN-13116 If xsite view has been formed, then we must perform state-transfer to all sites if a SFS recovery has occurred
 		if crossSiteViewCondition.Status == metav1.ConditionTrue {
 			podName := podList.Items[0].Name
+			curl.SetPodname(podName)
```
rigazilla, line 690: I don't like a lot this way of configuring the target pod for the ispnClient, looks a kind of 'configuring by side effect' to me...
but I don't have a better option, though.

rigazilla (changes requested):
Looks like a great work @ryanemerson! Very clean design.
I'm requesting changes only because I really think interfaces need to be documented.
Just minor comments the rest

pkg/infinispan/configuration/empty_package.go
```diff
@@ -0,0 +1 @@
+package configuration
```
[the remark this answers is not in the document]
ryanemerson: Hmm at some point I needed this for it to compile, but it seems to work fine now :shrug:.

pkg/infinispan/client/v13/infinispan.go
```diff
@@ -0,0 +1,42 @@
...
+
+type infinispan struct {
+	http.HttpClient
+}
+
+func New(client http.HttpClient) api.Infinispan {
+	return &infinispan{client}
+}
+
+func (i *infinispan) Cache(name string) api.Cache {
```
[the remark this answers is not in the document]
ryanemerson: > well you can actually do something like: return v13.New().Cache()

That's what I was thinking. To save doing `v13.New()` every time, we can also store the older client in the v14 struct that is created when `v14.New(...)` is called, i.e.:

```go
package v14
...
type infinispan struct {
	http.HttpClient
	ispn13 api.Infinispan
}

func New(client http.HttpClient) api.Infinispan {
	return &infinispan{
		HttpClient: client,
		ispn13: v13.New(client),
	}
}

func (i *infinispan) Cache(name string) api.Cache {
	return i.ispn13.Cache(name)
}
```

controllers/infinispan_controller.go
ryanemerson: Yeah I don't like it either! I was trying to avoid the need todo `ispnClient.New()` every time a pod changed as that will eventually have some logic to determine the required client version, so would result in unnecessary rest calls. I'll have a think this afternoon and see if I can improve this.

ryanemerson: I've replaced it with a new helper method `InfinispanForPod` which clones an existing curl client and updates it's podname to the required value

pkg/infinispan/client/v13/infinispan_v13.go
```diff
@@ -0,0 +1,43 @@
+// Package v13 implenents a client for interacting with Infinispan 13.x servers
```
rigazilla, line 1:
```suggestion
// Package v13 implements a client for interacting with Infinispan 13.x servers
```

rigazilla (approved):
LGTM

ryanemerson:
Thanks for reviewing @rigazilla

PR #1345, opened 2021-12-06
Ensure poll methods don't return err instance when IsNotFound

ryanemerson:


ryanemerson:
Merging. Xsite failure caused by MetalLB binding to `192.168.x.y.`

Issue #1154, opened 2021-06-18
Graceful Shutdown Refactoring


ryanemerson:
Graceful shutdown is used when a user sets `spec.replicas = 0` or when an upgrade is performed. The ultimate goal is to scale the cluster from `x` pods to 0, with each pod terminating normally so that all data is preserved when the pods are recreated. 

# Current Behaviour
1. A REST call is made to `/rest/v2/cluster?action=stop` on the first Infinispan pod marked as READY
2. Wait for all pods in the cluster to be NOT READY
3. Set the StatefulSet replicas to 0

## Limitations
- Once each Infinispan pod shutsdown in 1. they will always be restarted in order to satisfy the number of StatefulSet replicas
- It's possible that with a very large number of pods that one or more pods readinessProbe may pass and enter the READY state preventing conditions in stage 2 every being satisfied. Ultimately causing the cluster to not scale down.
- Verbose code required to handle steps 1-3.

# Proposal
Make use of StatefulSet semantics to scale down the cluster by setting `statefulSet.Spec.Replicas = 0`.

On Graceful shutdown request, do the following:

1. Disable cluster rebalancing via REST

This ensures that when scaling down from `{N-1..0}` cache entries aren't slowly migrated to pod 0.

2. Set `statefulSet.Spec.Replicas = 0`

The StatefulSet will start to scale down the cluster pods in the order `{N-1..0}`. Pods are shutdown gracefully as long as they don't exceed their configured `terminationGracePeriodSeconds` (30 seconds by default). We will need to experiment and ensure this is configured correctly.

3. On cluster scale up enable rebalancing (might not be necessary if this is config is not persisted, need to clarify).

# Missing
- Rebalancing can be disabled, however it is not yet exposed via the REST api.

ryanemerson:
@rigazilla @dmvolod @Crumby These are just some thoughts after working on a upgrade issue today. Maybe I have missed some obvious limitations with this approach, please let me know what you think.

dmvolod:
Very good proposal, @ryanemerson 
Once we are refactoring a graceful shutdown, it would be nice to skip step 1 in proposal for the ephemeral clusters, as graceful shutdown doesn't make sense for them

dmvolod:
[...]

PR #1347, opened 2021-12-09
[2.2.x] Graceful Shutdown Refactoring #1154

ryanemerson:
No longer explicitly delete pods on GracefulShutdown, instead scale the StatefulSet once all of the caches have been shutdown on the server.

### DIFF
--- a/pkg/infinispan/cluster.go
+++ b/pkg/infinispan/cluster.go
@@ -112,7 +112,7 @@ func (c Cluster) GetClusterSize(podName string) (int, error) {
 
 // GracefulShutdown performs clean cluster shutdown
 func (c Cluster) GracefulShutdown(podName string) error {
-	rsp, err, reason := c.Client.Post(podName, consts.ServerHTTPClusterStop, "", nil)
+	rsp, err, reason := c.Client.Post(podName, consts.ServerHTTPContainerShutdown, "", nil)
 	return validateResponse(rsp, reason, err, "during graceful shutdown", http.StatusNoContent)
 }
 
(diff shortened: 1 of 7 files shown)


rigazilla (approved):
lgtm

pkg/infinispan/cluster.go
```diff
@@ -127,18 +127,23 @@ func (c Cluster) GracefulShutdownTask(podName string) error {
...
+	   stdErr.println("Encountered error trying to determine cache shutdown order: " + err);
+	}
 
-	var internalCaches = new HashSet(icr.getInternalCacheNames());
-	/* The ___script_cache is included in both getCacheNames() and getInternalCacheNames so prevent repeated shutdown calls */
-	internalCaches.removeAll(cacheNames);
-	shutdown(internalCaches);
+	cachesToShutdown.addAll(icr.getInternalCacheNames());
+	cachesToShutdown.addAll(cacheManager.getCacheNames());
+	shutdown(cachesToShutdown);
```
danberindei, line 146: Shouldn't we define the shutdown function before invoking it?

Crumby:
@ryanemerson Would it make sense to update gracefulShutdown test to use 2 nodes? The test itself would uncover the issue if executed on OpenShift 4.9+

ryanemerson:
> Would it make sense to update gracefulShutdown test to use 2 nodes? The test itself would uncover the issue if executed on OpenShift 4.9+

I've increased the number of nodes to 2, but the test won't fail unless we add a write to one of the caches that encounters [the exception](https://issues.redhat.com/browse/ISPN-13121?focusedCommentId=19442375&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-19442375). Obviously I can't add that test to the CI until we have a new server image with the fix, so I think I'll create a follow up PR with this change

config/manager/manager.yaml
```diff
@@ -37,7 +37,7 @@ spec:
               fieldRef:
                 fieldPath: metadata.namespace
           - name: RELATED_IMAGE_OPENJDK
-            value: "quay.io/infinispan/server:13.0.1.Final"
+            value: "quay.io/infinispan/server:13.0.2.Final"
```
Crumby, line 40: Shouldn't this be 13.0.3?

ryanemerson: :+1: It wasn't available when I created the PR and was required for CI. I'll be changing this to 13.0.3.Final as part of the release commit.

PR #1349, opened 2021-12-10
Test fixes/enhancements

Crumby:
* Add missing deferred deletions for created ConfigMaps by custom config tests
* Add missing deferred deletion for encryption secret in encryption update test
* Fix the namespace where OperatorGroup is created by upgrade test and skip the creation if it's global namespace
* Add `encoding` and `fetch-state` for upgrade cache
* Increase values size from 6kB to 100kB for OOM test to speed up its execution


ryanemerson:
Thanks @Crumby

PR #1351, opened 2021-12-14
Remove find storage class

### LINKED ISSUE #1005: Utilize kube.LookupResource instead of separate kube.FindStorageClass
We should use `kube.LookupResource` instead of separate `kube.FindStorageClass` and safely remove it for reduce code base

leo8a:
Fixes #1005 

Signed-off-by: Leo Ochoa <lochoa@redhat.com>


controllers/infinispan_controller.go
```diff
@@ -1280,7 +1281,8 @@ func (r *infinispanRequest) statefulSetForInfinispan(adminSecret, userSecret, ke
 		pvc.OwnerReferences[0].BlockOwnerDeletion = pointer.BoolPtr(false)
 		// Set a storage class if it specified
 		if storageClassName := ispn.StorageClassName(); storageClassName != "" {
-			if _, err := kube.FindStorageClass(storageClassName, r.Client, r.ctx); err != nil {
+			storageClass := &storagev1.StorageClass{}
+			if _,err := kube.LookupResource(storageClassName, ispn.Namespace, storageClass, ispn, r.Client, reqLogger, r.eventRec, r.ctx); err != nil {
```
dmvolod, line 1285:
```suggestion
			if _,err := kube.LookupResource(storageClassName, ispn.Namespace,  &storagev1.StorageClass{}, ispn, r.Client, reqLogger, r.eventRec, r.ctx); err != nil {
```

test/e2e/main/main_test.go
```diff
@@ -303,6 +287,33 @@ func TestNodeWithEphemeralStorage(t *testing.T) {
 	}
 }
 
+// Test if single node with a storage class
+func TestNodeWithStorageClass(t *testing.T) {
+	t.Parallel()
+	kindDefaultStorageClass := "standard"
```
dmvolod, line 293: This shouldn't work outside the Kind, however QA's are testing on non-Kind envs.
You could adopt this script to get default StorageClass and then test again it.

.gitignore
```diff
@@ -89,10 +88,12 @@ tags
 .vscode/*
 .history
 openshift.local.clusterup/*
-build/_output
 vendor/
+### JetBrains ###
 .idea/
-*.iml
+### Vagrant ###
```
dmvolod, line 94: Why this needed for the project?

controllers/infinispan_controller.go
leo8a: +1, thanks.

.gitignore
leo8a: initially started with a Vagrant env for dev but now I'm just kind :). It's not needed anymore, will remove it.

test/e2e/main/main_test.go
leo8a: kk, want me to add a [local storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/#local) to the cluster in case the get default `StorageClass` returns nil? Or could we assume that we'll always have a default `StorageClass` in QA?

dmvolod: It's probably you need to get a StorageClass with `storageclass.kubernetes.io/is-default-class: "true"` annotation and play around it. This should work on any platform including OCP/OKD/Kind/etc.

ryanemerson:
@leo8a When you get a chance to push the changes can you make sure to rebase on `main` first so that the CI will pass.

pkg/kubernetes/kubernetes.go
```diff
@@ -101,6 +101,16 @@ func (k Kubernetes) GetSecret(secretName, namespace string, ctx context.Context)
 	return secret, err
 }
 
+// GetPVC returns a PVC bound with given PVC name
+func (k Kubernetes) GetPVC(pvcName, namespace string, ctx context.Context) (*corev1.PersistentVolumeClaim, error) {
+	pvc := &corev1.PersistentVolumeClaim{}
+	err := k.Client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: pvcName}, pvc)
+	if err != nil {
```
ryanemerson, line 108: As we don't need this method in the operator itself, only the testsuite, I would move this to `test/e2e/utils/kubernetes.go`.

We can then simplify the method to something like:

```go
func (k TestKubernetes) GetPVC (pvcName, namespace string) *corev1.PersistentVolumeClaim {
   pvc := &corev1.PersistentVolumeClaim{}
   err := k.Client.Get(ctx, types.NamespacedName{Namespace: namespace, Name: pvcName}, pvc)
   ExpectNoError(err)
   return pvc
}
```

controllers/infinispan_controller.go
ryanemerson, line 1285: It wasn't the original aim of this PR, but I think we can remove this lookup altogether. If the StorageClass doesn't exist, then the StatefulSet creation will fail.

This is better from a user perspective, as a user who is able to create a Infinispan CR will also be able to inspect created resources in their namespace. However, it's not guaranteed that they will have access to the operator log if it is installed cluster wide and they will not know why the Infinispan cluster cannot be created.

test/e2e/main/main_test.go
```diff
@@ -303,6 +287,33 @@ func TestNodeWithEphemeralStorage(t *testing.T) {
...
+	// Register it
+	spec.Labels = map[string]string{"test-name": t.Name()}
+	testKube.CreateInfinispan(spec, tutils.Namespace)
+	defer testKube.CleanNamespaceAndLogOnPanic(tutils.Namespace, spec.Labels)
+	testKube.WaitForInfinispanPods(1, tutils.SinglePodTimeout, spec.Name, tutils.Namespace)
+	testKube.WaitForInfinispanCondition(spec.Name, spec.Namespace, ispnv1.ConditionWellFormed)
+
+	// Ensure a PVCs is bound from the storageClass
+	pvcName := "data-volume-test-node-with-storage-class-0"
+	_, err := testKube.Kubernetes.GetPVC(pvcName, spec.Namespace, context.TODO())
```
ryanemerson, line 310: If you update the `GetPVC` method as suggested, then it's possible to simplify this simply to:

```
testKube.GetPVC("data-volume-test-node-with-storage-class-0", spec.Namespace)
```

All of the subsequent checks can then be removed.

controllers/infinispan_controller.go
ryanemerson: @leo8a On second thought I think we should keep the `LookupResource` call. @dmvolod Makes an excellent point that the user will be notified via the event logger that the `StorageClass` is not available, so we can avoid trying to create the StatefulSet with an invalid config and the user will be notified.

leo8a: agree.

pkg/kubernetes/kubernetes.go
leo8a: k, makes sense.

test/e2e/main/main_test.go
leo8a: added it.

leo8a:
thanks for the good feedback @ryanemerson and @dmvolod. I just pushed some changes based on ^^ suggestions, will add some more once I got the bits for properly getting the StorageClass on any k8s-based platform.

test/e2e/main/main_test.go
leo8a: Indeed, I just got the `StorageClass` name of the one with the `storageclass.kubernetes.io/is-default-class: "true"` annotation, as suggested. This should now be compatible with any k8s-powered platform (see [k8s official docs](https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/)), thanks!

.gitignore
```diff
@@ -89,10 +88,9 @@ tags
 .vscode/*
 .history
 openshift.local.clusterup/*
-build/_output
```
ryanemerson, line 92: I don't think the changes to this file are necessary. We want to ignore the files/dirs you have removed and the only addition is the Jetbrains comment.

test/e2e/main/main_test.go
```diff
@@ -893,7 +901,7 @@ func TestExternalServiceWithAuth(t *testing.T) {
 	tutils.ExpectNoError(err)
 
 	// Sleep for a while to be sure that the old pods are gone
-	// The restart is ongoing and it would that more than 10 sec
+	// The restart is ongoing, and it would that more than 10 sec,
```
ryanemerson, line 904: The comment doesn't really make sense in the first place so I think we can remove it entirely.

test/e2e/main/main_test.go
```diff
@@ -1287,7 +1295,7 @@ func TestPodDegradationAfterOOM(t *testing.T) {
 
 func GenerateStringWithCharset(length int) string {
 	const charset = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
-	var seededRand *rand.Rand = rand.New(rand.NewSource(time.Now().UnixNano()))
+	var seededRand = rand.New(rand.NewSource(time.Now().UnixNano()))
```
ryanemerson, line 1298: If we're changing this we can just do `seededRand := rand.New(...)`

test/e2e/utils/kubernetes.go
```diff
@@ -210,7 +210,7 @@ func (k TestKubernetes) PrintAllResources(namespace string, list runtime.Object,
 	unstructuredResourceList.SetUnstructuredContent(unstructuredResource)
 
 	for _, item := range unstructuredResourceList.Items {
-		yaml, err := yaml.Marshal(item)
+		yaml_, err := yaml.Marshal(item)
```
ryanemerson, line 213: Any reason for the var name change?

test/e2e/utils/kubernetes.go
```diff
@@ -486,6 +486,38 @@ func (k TestKubernetes) WaitForExternalService(ispn *ispnv1.Infinispan, timeout
...
+	ExpectNoError(err)
+
+	// Identify the default StorageClass in the cluster. There's only one that contains the
+	// following annotation {"storageclass.kubernetes.io/is-default-class":"true"}, for more info
+	// check https://kubernetes.io/docs/tasks/administer-cluster/change-default-storage-class/.
+	for _, sc := range scList.Items {
+		for key, value := range sc.Annotations {
+			if key == "storageclass.kubernetes.io/is-default-class" && value == "true" {
+				defaultStorageClassName = sc.Name
+				return defaultStorageClassName
```
ryanemerson, line 514: We can just return `sc.Name` directly here.

test/e2e/utils/kubernetes.go
```diff
@@ -486,6 +486,38 @@ func (k TestKubernetes) WaitForExternalService(ispn *ispnv1.Infinispan, timeout
...
+func (k TestKubernetes) GetPVC(pvcName, namespace string) *corev1.PersistentVolumeClaim {
+	pvc := &corev1.PersistentVolumeClaim{}
+	err := k.Kubernetes.Client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: pvcName}, pvc)
+	ExpectNoError(err)
+	return pvc
+}
+
+// GetSCList returns the default StorageClasses name in cluster.
+func (k TestKubernetes) GetSCList() string {
+	var defaultStorageClassName string
```
ryanemerson, line 499: Redundant var if we just return `""` at the end of the method.

test/e2e/utils/kubernetes.go
```diff
@@ -486,6 +486,38 @@ func (k TestKubernetes) WaitForExternalService(ispn *ispnv1.Infinispan, timeout
...
+// GetPVC returns a PVC bound with given PVC name
+func (k TestKubernetes) GetPVC(pvcName, namespace string) *corev1.PersistentVolumeClaim {
+	pvc := &corev1.PersistentVolumeClaim{}
+	err := k.Kubernetes.Client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: pvcName}, pvc)
+	ExpectNoError(err)
+	return pvc
+}
+
+// GetSCList returns the default StorageClasses name in cluster.
+func (k TestKubernetes) GetSCList() string {
```
ryanemerson, line 498: `GetDefaultStorageClass`?

test/e2e/utils/kubernetes.go
```diff
@@ -486,6 +486,38 @@ func (k TestKubernetes) WaitForExternalService(ispn *ispnv1.Infinispan, timeout
...
+	var defaultStorageClassName string
+
+	// Get the list of StorageClasses in cluster.
+	scList := &storagev1.StorageClassList{}
+	listOps := &client.ListOptions{}
+	err := k.Kubernetes.Client.List(context.TODO(), scList, listOps)
```
ryanemerson, line 504: As `ListOps` is never used again, I would just do: `err := k.Kubernetes.Client.List(context.TODO(), scList,  &client.ListOptions{})`

test/e2e/utils/kubernetes.go
```diff
@@ -886,7 +918,7 @@ func (k *TestKubernetes) AssertK8ResourceExists(name, namespace string, obj clie
 	client := k.Kubernetes.Client
 	key := types.NamespacedName{
 		Name:      name,
-		Namespace: Namespace,
+		Namespace: namespace,
```
ryanemerson, line 921: Good spot :+1:

test/e2e/main/main_test.go
```diff
@@ -303,6 +287,30 @@ func TestNodeWithEphemeralStorage(t *testing.T) {
...
+
+	// Register above created resource
+	spec.Labels = map[string]string{"test-name": t.Name()}
+	testKube.CreateInfinispan(spec, tutils.Namespace)
+	defer testKube.CleanNamespaceAndLogOnPanic(tutils.Namespace, spec.Labels)
+	testKube.WaitForInfinispanPods(1, tutils.SinglePodTimeout, spec.Name, tutils.Namespace)
+	testKube.WaitForInfinispanCondition(spec.Name, spec.Namespace, ispnv1.ConditionWellFormed)
+
+	// Ensure a PVCs is bound from the storageClass
+	testKube.GetPVC("data-volume-test-node-with-storage-class-0", spec.Namespace)
```
ryanemerson, line 311: Shouldn't we assert that the PVC spec is using the `defaultStorageClass` here?

ryanemerson (changes requested):

.gitignore
leo8a: k, agree. This's repeated in [L4](https://github.com/infinispan/infinispan-operator/blob/main/.gitignore#L4), hence the removal.

test/e2e/utils/kubernetes.go
leo8a: Just changed the name of that var internally to avoid any collision with the imported package name... I thought it was a nit, but if you prefer we may keep its previous name.

test/e2e/utils/kubernetes.go
leo8a: totally agree, thanks.

test/e2e/utils/kubernetes.go
ryanemerson: No that's fine, makes sense :+1:

test/e2e/main/main_test.go
```diff
@@ -274,13 +275,17 @@ func areOperatorLabelsPropagated(namespace, varName string, labels map[string]st
...
 	tutils.ExpectNoError(err)
 
-	_, err = kube.FindStorageClass("not-present-storage-class", serviceAccountKube.Kubernetes.Client, context.TODO())
-	if !k8sErrors.IsNotFound(err) {
+	// Create a resource without passing any config
+	spec := tutils.DefaultSpec(testKube)
+
+	//_, err = kube.FindStorageClass("not-present-storage-class", serviceAccountKube.Kubernetes.Client, context.TODO())
+	if _, err := kube.LookupResource("not-present-storage-class", spec.Namespace, &storagev1.StorageClass{}, spec,
+		serviceAccountKube.Kubernetes.Client, log, r.eventRec, context.TODO()); !k8sErrors.IsNotFound(err) {
```
leo8a, line 288: hey @ryanemerson, after rebasing I started the refactoring of the `TestRolesSynthetic`, given that this PR is removing `FindStorageClass` as well. However, not sure here how to best approach the param `record.EventRecorder` from within the main e2e tests?

ryanemerson: In our tests instead of using `kube.LookupResource` it's simpler to just perform a `Get` to retrieve a resource using the kube client i.e. `client.Get(ctx, types.NamespacedName{Name: name}, storageClass)`


However, in this case we actually want to remove `TestRolesSynthetic` so we don't even have to do that :slightly_smiling_face:. You had done this before, so I think it was accidentally reintroduced when rebasing.

leo8a: 🙂 kk! fixed now, thanks.

test/e2e/main/main_test.go
```diff
@@ -168,13 +165,19 @@ func TestNodeStartup(t *testing.T) {
 	spec.Annotations[v1.TargetLabels] = "my-svc-label"
 	spec.Labels = make(map[string]string)
 	spec.Labels["my-svc-label"] = "my-svc-value"
-	os.Setenv(v1.OperatorTargetLabelsEnvVarName, "{\"operator-svc-label\":\"operator-svc-value\"}")
-	os.Setenv(v1.OperatorTargetLabelsEnvVarName, "{\"operator-svc-label\":\"operator-svc-value\"}")
+	if err := os.Setenv(v1.OperatorTargetLabelsEnvVarName, "{\"operator-svc-label\":\"operator-svc-value\"}"); err != nil {
```
ryanemerson, line 165: `ExpectNoError` already checks if the passed param is nil, so we can just do: `ExpectNoError(os.Setenv(v1.OperatorTargetLabelsEnvVarName, "{\"operator-svc-label\":\"operator-svc-value\"}"))`

As this is test code and the test will fail if the env is not set anyway, I don't think checking err is really important here.

ryanemerson:
Thanks @leo8a

Issue #1326, opened 2021-11-18
Support multiple Operand versions


ryanemerson:
# Goal
The operator should be able to reconcile Infinispan CRs for multiple Infinispan versions at any one time. It should be possible for users to define the Infinispan version that they want to use when creating a Infinispan CR. Similarly, it should be possible for the version to be updated at runtime in order to trigger a server upgrade.

# Motivation
An organisation's k8 cluster is host to many different microservices, each of which are have their own dev lifecycle and are owned by independent teams. In such scenarios, it's not possible for all teams to converge on a single Infinispan version.

# Existing Solutions
The operator currently has a one-to-one mapping with the Infinispan server, currently it's possible to deploy different versions doing one of the following:

## Custom Image
A user can provide a custom image per Infinispan CR spec

Disadvantages:
- Upgrades not supported
- No guarantee that the configured server image is compatible with the installed operator

## Namespace and Multi-namespace Scoped Operators
It's possible to install the Infinispan operator per namespace, or in multi-namespace mode,  so that Infinspan CRs in different namespaces can be controlled by distinct operators.

- CRDs are installed cluster-wide
  - Backwards incompatible CRDs changes can break older operator versions when a new Operator version installed
  - Operator removal order can remove CRD

- Different Infinispan versions can't be reconciled in the same namespace

# Design
Allow Infinispan server version to be configured in the Infinispan CR spec:

```yaml
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: example-infinispan
spec:
  version: 13.0.0 # Dictates the Infinispan server version
  replicas: 2
```
Server upgrades/downgrades triggered by changes to the `spec.version` field
- Omit field to utilise latest server version supported by installed Operator version on initial CR creation. Subsequent upgrades triggered by manual update of `spec.version`
- Allows staggered upgrades to reduce blast radius
  - Failed HotRodRolling upgrades can be rolledback by resetting the `spec.version` field

## Supported Server Versions
The supported server versions should be defined by a meta file. For each server version, we must define a stack of available images containing the image SHA and a flag to indicate whether the image is a CVE respin that does not impact server code.

For example:

```json
{
  "13.0.1": [
    {
      "image": "quay.io/infinispan/server@sha256:6db20af2f227d93b1b91d7ffe78dbad386680c120e59f87b3b3f2db516b0cf87",
      "cve": false
    }
  ],
  "13.0.0": [
    {
[...]

PR #1369, opened 2022-01-07
Versioned server configuration

ryanemerson:


rigazilla (commented):
looks good to me.
nice clean up of the Reconcile funcs! 

Versioning implementation in the operator looks neat, @ryanemerson do we have a description of how the Version will be set up? will it be configured in the CR? or computed from the server image?

ryanemerson:
> Versioning implementation in the operator looks neat, @ryanemerson do we have a description of how the Version will be set up? will it be configured in the CR? or computed from the server image?

I was thinking that we would have a `spec.version` field. More details about how this can be set/modfied are on https://github.com/infinispan/infinispan-operator/issues/1326

PR #1376, opened 2022-01-27
Do not panic on test failure

### LINKED ISSUE #1217: Testsuite: Utilise testing.Fail instead of panic

Crumby:
Resolves #1217 by catching up panic code and changing test state to fail by calling `t.Fatal()` in general `CleanNamespaceAndLogOnPanic` function. Other test changes introduced:
* `DefaultsSpec` now accepts `t *testing.T` and generates Infinispan name and labels based on that (duplicity reduction)
* All owned CRDs are now expected to be labeled with `"test-name": t.Name()` so they can be properly caught up by clean up function (consistency)
* `testKube.CleanNamespaceAndLogOnPanic` is hooked up rather at the begging of test code

### DIFF
--- a/test/e2e/backup-restore/backup_restore_test.go
+++ b/test/e2e/backup-restore/backup_restore_test.go
@@ -24,7 +24,7 @@ import (
 
 var testKube = tutils.NewTestKubernetes(os.Getenv("TESTING_CONTEXT"))
 
-type clusterSpec func(name, namespace string, clusterSize int) *v1.Infinispan
+type clusterSpec func(t *testing.T, name, namespace string, clusterSize int) *v1.Infinispan
 
 func TestMain(m *testing.M) {
 	tutils.RunOperator(m, testKube)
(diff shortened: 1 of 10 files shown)


ryanemerson:
@Crumby Looks good! I'm just re-running Ci to make sure the upgrade failure is not related.

ryanemerson:
@Crumby The upgrade test seems to be consistently failing with the changes

Crumby:
Think I found the issue. Batch created by `CreateBatch` in `batch_helper.go` labeled the Batches with `strcase.ToKebab(t.Name())` instead of `t.Name()` as the rest of the tests and thus it wasn't caught by cleanup delete function.

Crumby:
@ryanemerson Upgrade test is fixed. XSite tests running with LoadBalancer are now failing but the PR doesn't seem to be root cause of the failures.

ryanemerson:
Thanks @Crumby

PR #1388, opened 2022-02-03
Update probe period

ryanemerson:


Makefile
```diff
@@ -247,7 +247,7 @@ endif
 .PHONY: catalog-build
 ## Build a catalog image by adding bundle images to an empty catalog using the operator package manager tool, 'opm'.
 catalog-build: opm ## Build a catalog image.
-	$(OPM) index add --container-tool $(CONTAINER_TOOL) --mode replaces --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
+	sudo $(OPM) index add --container-tool $(CONTAINER_TOOL) --mode replaces --tag $(CATALOG_IMG) --bundles $(BUNDLE_IMGS) $(FROM_INDEX_OPT)
```
pruivo, line 250: sudo?

controllers/pods.go
```diff
@@ -36,16 +36,16 @@ func PodPortsWithXsite(i *infinispanv1.Infinispan) []corev1.ContainerPort {
...
 func PodReadinessProbe() *corev1.Probe {
-	return probe(5, 10, 10, 1, 80)
+	return probe(5, 0, 10, 1, 80)
 }
 
 func PodStartupProbe() *corev1.Probe {
-	// Maximum 10 minutes (60 * 10s) to finish startup
-	return probe(60, 10, 10, 1, 80)
+	// Maximum 10 minutes (600 * 1s) to finish startup
+	return probe(600, 10, 1, 1, 80)
```
pruivo, line 48: we discussed the initial delay to be 1 second too.

scripts/ci/install-catalog-source.sh
```diff
@@ -14,7 +14,7 @@ BUNDLE_IMG_NAME=infinispan-operator-bundle
 export IMG=${IMG_REGISTRY}/infinispan-operator
 export BUNDLE_IMG=${IMG_REGISTRY}/${BUNDLE_IMG_NAME}:v${VERSION}
 export CATALOG_IMG=${IMG_REGISTRY}/infinispan-test-catalog
-export CATALOG_BASE_IMG=${CATALOG_BASE_IMG-"quay.io/operatorhubio/catalog_tmp"}
+export CATALOG_BASE_IMG=${CATALOG_BASE_IMG-"quay.io/operatorhubio/catalog:tmp_latest_sql"}
```
pruivo, line 17: don't know what this is but I'll trust you :)

Makefile
ryanemerson: Merged as part of https://github.com/infinispan/infinispan-operator/pull/1382.

Unfortunately `sudo` is now required with their multi-arch images due to the way they have set the permissions. Previously we were using a single-arch tag, but that's no longer available so we have to work around this way.

scripts/ci/install-catalog-source.sh
ryanemerson: That's part of https://github.com/infinispan/infinispan-operator/pull/1382 as well.

controllers/pods.go
ryanemerson: Whoops, good catch!

ryanemerson:
@pruivo Do we want to reconsider the GossipRouter probe values as well?

pruivo:
> @pruivo Do we want to reconsider the GossipRouter probe values as well?

probably :) 
the startup could be 1 sec initial delay and 1 sec period (I believe the gossip router starts in under 10 ms)
the liveness/readiness can have their initial delay to 0.

ryanemerson:
Updated

ryanemerson:
CI looks good :slightly_smiling_face:

pruivo:
merged! thanks @ryanemerson !

PR #1389, opened 2022-02-03
Update default limits

ryanemerson:


api/v1/types_util.go
```diff
@@ -434,8 +430,12 @@ func (spec *InfinispanContainerSpec) GetMemoryResources() (requests resource.Qua
 }
 
 func getRequestLimits(str string) (requests resource.Quantity, limits resource.Quantity, err error) {
-	parts := strings.Split(str, ":")
+	if str == "" {
+		err = fmt.Errorf("resource string cannot be empty")
```
Crumby, line 434: Shouldn't this be handled by CRD definition?

ryanemerson: A CPU value of `""` is valid for the CR as that's the default, i.e no limit. This is simply a sanity check to ensure that nowhere in the code do we accidentally pass `""`. Calls to `GetCpuResources()` should only be made if there's a non-empty value.

pruivo (approved):
LGTM!

rigazilla (approved):
lgtm

Crumby (approved):

PR #1392, opened 2022-02-15
[2.2.x] OCP4.10 test fix

Crumby:
Use different label to retrieve prometheus pod as OpenShift 4.10 removed `app: prometheus` label from it.


ryanemerson:
Thanks @Crumby

PR #1397, opened 2022-02-16
Single operand image. Fixes #1396

### LINKED ISSUE #1396: Single operand image
Since an openjdk/ubi8 image is available for ppc64le and s390x there's no more need to refer different operand images in: https://github.com/infinispan/infinispan-operator/blob/dbaf3a6fc98eea26c8c3a604f055dc73f17697ab/controllers/constants/image.go#L18

see also https://issues.redhat.com/browse/JDG-5110

rigazilla:
Fixes #1396

### DIFF
--- a/controllers/constants/image.go
+++ b/controllers/constants/image.go
@@ -1,33 +1,20 @@
 package constants
 
-import "runtime"
+import "os"
 
 const (
 	// OperandImageOpenJDK envvar name for OpenJDK operand image
 	OperandImageOpenJDK = "RELATED_IMAGE_OPENJDK"
-	// OperandImageOpenJ9 envvar name for OpenJ9 operand image
-	OperandImageOpenJ9 = "RELATED_IMAGE_OPENJ9"
 
 	// DefaultOperandImageOpenJDK default image for OpenJDK stack
 	DefaultOperandImageOpenJDK = "quay.io/infinispan/server:13.0"
-	// DefaultOperandImageOpenJ9 default image for OpenJ9 stack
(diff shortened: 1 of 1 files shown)


controllers/constants/image.go
```diff
@@ -1,33 +1,18 @@
...
-		return GetEnvWithDefault(OperandImageOpenJ9, DefaultOperandImageOpenJ9)
-	default:
-		// Try new name RELATED_IMAGE_OPENJDK first
-		retVal := GetEnvWithDefault(OperandImageOpenJDK, "")
-		if retVal != "" {
-			return retVal
-		}
-		// If RELATED_IMAGE_OPENJDK is not defined fallback on old name for backward comp
-		return GetEnvWithDefault("DEFAULT_IMAGE", DefaultOperandImageOpenJDK)
+	if retVal := GetEnvWithDefault(OperandImageOpenJDK, ""); retVal != "" {
```
ryanemerson, line 13: As we're not using a default value here, we can just do:

```go
	if retVal, exists := os.LookupEnv(OperandImageOpenJ9); exists {
		return retVal
	}
	return GetEnvWithDefault("DEFAULT_IMAGE", DefaultOperandImageOpenJDK)
```

ryanemerson: Ignore that comment :slightly_smiling_face:. `os.LookupEnv` will return true if the env is set but empty, which is different to `GetEnvWithDefault`.

rigazilla: ok, done

ryanemerson:
Thanks @rigazilla

Issue #1381, opened 2022-02-02
Implement test filtering mechanism to allow running smoke/integration tests


Crumby:
Currently it's possible to filter test only by either package/file name or using test regex (`-run ${regex}`). Neither of these mechanisms is practical to filter out tests based on the reason why they are being executed such as smoke testing or integration testing with new K8s/OCP versions.

There are several possible solutions to overcome this issue:

**1. Using build tags**
Would be reasonably usable if we had single test per file.

_Pro:_ Simplifies tests execution by running the tests with `-tags integration/smoke` and it'll automatically pick the right tests.
_Con:_ It's usable only at file level, it's not possible to mark only some of the tests from the file.

**2. Using test/file naming conventions**
Tests that should be part of the particular test suite would have it's name in it. Eg. TestFeatureIntegration

_Pro:_ Overcomes issue of using build tags and makes running tests with regex possible systematically (`-run .*Integration/Smoke.*`)
_Con:_ Test naming could be confusing. Eg. with tree CustomConfig tests for yaml,json,xml, only one would be ending with `Integration/Smoke`

**3. Wrapping the tests in each file and selecting the tests to be executed based on env var.**
There would be a master test in each file that would pick tests to executed based on env var input
```
SomeTest(t *testing.T) {
  if(env.Integration) {
    t.run(t, test1)
  } else {
    t.run(t, test1)
    t.run(t, test2)
  }
}
```
_Pro:_ Fine grained solution.
_Con:_ Tests and where they belong needs to be manually maintained in each file

**4. Implementing `skipIf()` method used at the beginning of each test**
Each test would call `skipIf() ` at the beginning that would decide whether test should be executed or not based on env var.
```
TestSomeFeature(t *testing.T) {
  skipIf(t, env.Integration)
}

skipIf(t *testing.T, envName string) {
   if(env.Integration == true) {
    t.skip()
   }
}
```

_Pro_ Fine grained solution
_Con_ If the call of `skipIf ` is forgottent then the may be executed even though it's no desired.

Crumby:
@ryanemerson @rigazilla I guess the closest to ideal solution is using the build tags even though that would mean having single test per file but I'd like to hear your opinion and I'm all ears for any other possible solutions. Ideally those that are rather marking which tests should be executed rather then other way around.

ryanemerson:
> I guess the closest to ideal solution is using the build tags even though that would mean having single test per file

[...]

PR #1403, opened 2022-02-17
Split main_test.go into separate files

### LINKED ISSUE #1117: Split main E2E testsuite
**Issue:** There is currently more then 20 tests running different deployments in main e2e test suite. When executing all the tests outside of CI (eg. OpenShift certification) any test failure will immediately halt rest of the test execution. With that many tests in single test file it's hard to tell which were executed and re-execute only those that were not.

**Proposal:** Split the main e2e test suite in several, logically grouped tests.

It'd be ideal if golang wouldn't halt test execution on failure but I haven't find a way to achieve this behavior yet.

Crumby:
Resolves #1117 by splitting tests in the `main_test.go` file in separate `*_test.go` files grouped by feature topic.


ryanemerson:
@Crumby What's the command required to execute the different category of test?

Crumby:
@ryanemerson TBH I'm not sure that's reasonably possible in our case. Theoretically it should be achievable by running `go test test/e2e/main/${group}_test.go` but I haven't tried it out and might not work in our case due to need of having `setup_test.go` which contains `TestMain()`. Also go doesn't see multiple files, it only sees functions, vars, etc. as if all the tests were in the single file (thus it's possible to have `TestMain()` and `testKube` in separate file and it works nicely).

I would say that best should would be with `-run ${regex}` option and some convention in test naming.

ryanemerson:
@Crumby I thought the idea was to utilise build tags once the tests has been separated out into different files? https://github.com/infinispan/infinispan-operator/issues/1381

Crumby:
@ryanemerson This PR is addressing #1117 only and the changes are purely cosmetic. The idea here is to improve navigation between tests and visibility in what is being testing in relation to a particular feature rather then having big single test file.

> I thought the idea was to utilise build tags once the tests has been separated out into different files? https://github.com/infinispan/infinispan-operator/issues/1381

I'll address #1381 in the different PR with a bit different more flexible solution that the build tags provide.

ryanemerson:
Thanks @Crumby

Issue #1364, opened 2021-12-22
Allow the user to provide custom log4j.xml configurations


ryanemerson:
It should be possible for users to provide their own log4j.xml config file via the same config map that is used for [custom server configuration](https://infinispan.org/docs/infinispan-operator/main/operator.html#applying-custom-configuration_configuring-clusters). This allows users with specific use-cases to configure additional Appenders etc as required.

PR #1417, opened 2022-02-24
[feature] Allow end-users to provide custom `log4j.xml` configuration

leo8a:
This PR aims to solve [#1364](https://github.com/infinispan/infinispan-operator/issues/1364).

Additionally, it contains code for testing and documenting the new feature.

Signed-off-by: Leo Ochoa <lochoa@redhat.com>

### DIFF
--- a/controllers/infinispan_controller.go
+++ b/controllers/infinispan_controller.go
@@ -302,20 +302,22 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 	// If an overlay configuration is specified wait for the related ConfigMap
 	overlayConfigMap := &corev1.ConfigMap{}
 	var overlayConfigMapKey string
+	var overlayLog4jConfig bool
 	if infinispan.Spec.ConfigMapName != "" {
 		if result, err := kube.LookupResource(infinispan.Spec.ConfigMapName, infinispan.Namespace, overlayConfigMap, infinispan, r.Client, reqLogger, r.eventRec, r.ctx); result != nil {
 			return *result, err
 		}
-		var foundKey bool
-		// Loop through the data looking for something like xml,json or yaml
+		// Loop through the data looking for something like xml, json or yaml
 		for overlayConfigMapKey = range overlayConfigMap.Data {
 			if overlayConfigMapKey == "infinispan-config.xml" || overlayConfigMapKey == "infinispan-config.json" || overlayConfigMapKey == "infinispan-config.yaml" {
-				foundKey = true
 				break
 			}
 		}
-		if !foundKey {
-			err := fmt.Errorf("infinispan-copnfig.[xml|yaml|json] configuration not found in ConfigMap: %s", overlayConfigMap.Name)
+		// Check if the user added a custom log4j.xml config
+		_, overlayLog4jConfig = overlayConfigMap.Data["log4j.xml"]
+
+		if overlayConfigMapKey == "" && !overlayLog4jConfig {
+			err := fmt.Errorf("infinispan-config.[xml|yaml|json] or log4j.xml configuration not found in ConfigMap: %s", overlayConfigMap.Name)
 			return reconcile.Result{}, err
 		}
 	}
@@ -439,7 +441,7 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 		reqLogger.Info("Configuring the StatefulSet")
 
 		// Define a new StatefulSet
-		statefulSet, err = r.statefulSetForInfinispan(adminSecret, userSecret, keystoreSecret, trustSecret, configMap, overlayConfigMap, overlayConfigMapKey)
+		statefulSet, err = r.statefulSetForInfinispan(adminSecret, userSecret, keystoreSecret, trustSecret, configMap, overlayConfigMap, overlayConfigMapKey, overlayLog4jConfig)
 		if err != nil {
 			reqLogger.Error(err, "failed to configure new StatefulSet")
 			return ctrl.Result{}, err
@@ -539,7 +541,7 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 
 	// Here where to reconcile with spec updates that reflect into
 	// changes to statefulset.spec.container.
-	res, err = r.reconcileContainerConf(statefulSet, configMap, overlayConfigMap, overlayConfigMapKey, adminSecret, userSecret, keystoreSecret, trustSecret)
+	res, err = r.reconcileContainerConf(statefulSet, configMap, overlayConfigMap, overlayConfigMapKey, overlayLog4jConfig, adminSecret, userSecret, keystoreSecret, trustSecret)
 	if res != nil {
 		return *res, err
 	}
@@ -1125,7 +1127,7 @@ func (r *infinispanRequest) updatePodsLabels(podList *corev1.PodList) error {
 
 // statefulSetForInfinispan returns an infinispan StatefulSet object
 func (r *infinispanRequest) statefulSetForInfinispan(adminSecret, userSecret, keystoreSecret, trustSecret *corev1.Secret,
-	configMap, overlayConfigMap *corev1.ConfigMap, overlayConfigMapKey string) (*appsv1.StatefulSet, error) {
+	configMap, overlayConfigMap *corev1.ConfigMap, overlayConfigMapKey string, overlayLog4jConfig bool) (*appsv1.StatefulSet, error) {
 	ispn := r.infinispan
 	reqLogger := r.log.WithValues("Request.Namespace", ispn.Namespace, "Request.Name", ispn.Name)
 	lsPod := PodLabels(ispn.Name)
@@ -1242,7 +1244,7 @@ func (r *infinispanRequest) statefulSetForInfinispan(adminSecret, userSecret, ke
 						StartupProbe:   PodStartupProbe(),
 						Resources:      *podResources,
 						VolumeMounts:   volumeMounts,
-						Args:           buildStartupArgs(overlayConfigMapKey, "false"),
+						Args:           buildStartupArgs(overlayConfigMapKey, overlayLog4jConfig, "false"),
 					}},
 					Volumes: volumes,
 				},
@@ -1527,7 +1529,7 @@ func (r *infinispanRequest) gracefulShutdownReq(podList *corev1.PodList, logger 
 }
 
 // reconcileContainerConf reconcile the .Container struct is changed in .Spec. This needs a cluster restart
-func (r *infinispanRequest) reconcileContainerConf(statefulSet *appsv1.StatefulSet, configMap, overlayConfigMap *corev1.ConfigMap, overlayConfigMapKey string, adminSecret,
+func (r *infinispanRequest) reconcileContainerConf(statefulSet *appsv1.StatefulSet, configMap, overlayConfigMap *corev1.ConfigMap, overlayConfigMapKey string, overlayLog4jConfig bool, adminSecret,
 	userSecret, keystoreSecret, trustSecret *corev1.Secret) (*ctrl.Result, error) {
 	ispn := r.infinispan
 	updateNeeded := false
@@ -1587,7 +1589,7 @@ func (r *infinispanRequest) reconcileContainerConf(statefulSet *appsv1.StatefulS
 	updateNeeded = updateStatefulSetEnv(statefulSet, "CONFIG_HASH", hash.HashString(configMap.Data[consts.ServerConfigFilename])) || updateNeeded
 	updateNeeded = updateStatefulSetEnv(statefulSet, "ADMIN_IDENTITIES_HASH", hash.HashByte(adminSecret.Data[consts.ServerIdentitiesFilename])) || updateNeeded
 
-	if updateCmdArgs, err := updateStartupArgs(statefulSet, overlayConfigMapKey, "false"); err != nil {
+	if updateCmdArgs, err := updateStartupArgs(statefulSet, overlayConfigMapKey, overlayLog4jConfig, "false"); err != nil {
 		return &ctrl.Result{}, err
 	} else {
 		updateNeeded = updateCmdArgs || updateNeeded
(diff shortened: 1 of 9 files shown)


controllers/infinispan_controller.go
```diff
@@ -300,21 +300,26 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 
 	// If an overlay configuration is specified wait for the related ConfigMap
 	overlayConfigMap := &corev1.ConfigMap{}
-	var overlayConfigMapKey string
+	var overlayConfigMapKey, overlayLog4jConfigMapKey string
```
ryanemerson, line 303: There's no need for the `overlayLog4jConfigMapKey` here as the filename is always `log4j.xml`. The reason we require `overlayConfigMapKey` is because we support three different names for the server configuration.

controllers/infinispan_controller.go
```diff
@@ -300,21 +300,26 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
+			case "log4j.xml":
+				overlayLog4jConfigMapKey = configMapKey
+			}
+			if overlayConfigMapKey != "" && overlayLog4jConfigMapKey != "" {
 				break
 			}
 		}
-		if !foundKey {
-			err := fmt.Errorf("infinispan-config.[xml|yaml|json] configuration not found in ConfigMap: %s", overlayConfigMap.Name)
+		if overlayConfigMapKey == "" && overlayLog4jConfigMapKey == "" {
```
ryanemerson, line 321: We can simplify this to:

```go
	overlayConfigMap := &corev1.ConfigMap{}
	var overlayConfigMapKey string
	if infinispan.Spec.ConfigMapName != "" {
		if result, err := kube.LookupResource(infinispan.Spec.ConfigMapName, infinispan.Namespace, overlayConfigMap, infinispan, r.Client, reqLogger, r.eventRec, r.ctx); result != nil {
			return *result, err
		}
		// Loop through the data looking for something like xml, json or yaml
		for overlayConfigMapKey = range overlayConfigMap.Data {
			if overlayConfigMapKey == "infinispan-config.xml" || overlayConfigMapKey == "infinispan-config.json" || overlayConfigMapKey == "infinispan-config.yaml" {
				break
			}
		}
		_, logFileExists := overlayConfigMap.Data["log4j.xml"]
		if overlayConfigMapKey == "" && !logFileExists{
			err := fmt.Errorf("infinispan-config.[xml|yaml|json] or log4j.xml configuration not found in ConfigMap: %s", overlayConfigMap.Name)
		}
	}
```

controllers/infinispan_controller.go
```diff
@@ -1760,18 +1765,23 @@ func GossipRouterPodList(infinispan *infinispanv1.Infinispan, kube *kube.Kuberne
 	return podList, kube.ResourcesList(infinispan.Namespace, GossipRouterPodLabels(infinispan.Name), podList, ctx)
 }
 
-func buildStartupArgs(overlayConfigMapKey string, zeroCapacity string) []string {
-	var args []string
+func buildStartupArgs(overlayConfigMapKey, overlayLog4jConfigMapKey, zeroCapacity string) []string {
```
ryanemerson, line 1768: As the file must always be named "log4j.xml` we can remove this parameter and just use the filename directly when building the args string.

controllers/infinispan_controller.go
```diff
@@ -1760,18 +1765,23 @@ func GossipRouterPodList(infinispan *infinispanv1.Infinispan, kube *kube.Kuberne
...
+		args = append(args, "-l", "user/"+overlayLog4jConfigMapKey)
+	} else {
+		args = append(args, "-l", OperatorConfMountPath+"/log4j.xml")
+	}
 	if overlayConfigMapKey != "" {
-		args = []string{"-Dinfinispan.zero-capacity-node=" + zeroCapacity, "-l", OperatorConfMountPath + "/log4j.xml", "-c", "user/" + overlayConfigMapKey, "-c", "operator/infinispan.xml"}
+		args = append(args, "-c", "user/"+overlayConfigMapKey, "-c", "operator/infinispan.xml")
 	} else {
-		args = []string{"-Dinfinispan.zero-capacity-node=" + zeroCapacity, "-l", OperatorConfMountPath + "/log4j.xml", "-c", "operator/infinispan.xml"}
+		args = append(args, "-c", "operator/infinispan.xml")
```
ryanemerson, line 1778: We can make this method more efficient and easier to read by using a [strings.Builder](https://pkg.go.dev/strings#Builder)

test/e2e/main/custom_configuration_test.go
```diff
@@ -200,3 +207,32 @@ xmlns:server="urn:infinispan:server:13.0">
...
+    <Loggers>
+        <Root level="INFO">
+            <AppenderRef ref="STDOUT" level="{{ .Logging.Console.Level | UpperCase }}"/>
+        </Root>
+
+        {{ range $key, $value := .Logging.Categories }}
+        <Logger name="{{ $key }}" level="{{ $value | UpperCase }}"/>
+        {{ end }}
+    </Loggers>
+</Configuration>`
```
ryanemerson, line 232: This xml contains go template logic which won't be processed before the file is mounted to the server pod, resulting in an invalid logging configuration. You should remove all content in the `{{  }}` brackets and replace with the actual values to be used.

The xml below is valid and it configures the server so that all infinispan packages are logged at the trace level.

```xml

<?xml version="1.0" encoding="UTF-8"?>
<Configuration name="InfinispanServerConfig" monitorInterval="60" shutdownHook="disable">
    <Appenders>
        <!-- Colored output on the console -->
        <Console name="STDOUT">
            <PatternLayout pattern="%d{HH:mm:ss,SSS} %-5p (%t) [%c] %m%throwable%n"/>
        </Console>
    </Appenders>

    <Loggers>
        <Root level="INFO">
            <AppenderRef ref="STDOUT" level="TRACE"/>
        </Root>
        <Logger name="org.infinispan" level="TRACE"/>
    </Loggers>
</Configuration>
```

test/e2e/main/custom_configuration_test.go
```diff
@@ -31,6 +31,13 @@ func TestUserJsonCustomConfig(t *testing.T) {
 	testCustomConfig(t, configMap)
 }
 
+func TestUserLog4jCustomConfig(t *testing.T) {
+	t.Parallel()
+	configMap := newCustomConfigMap(t.Name(), "xml")
+	configMap = addLog4jToCustomConfigMap(configMap)
+	testCustomConfig(t, configMap)
```
ryanemerson, line 38: Instead of adding a new test, I would update `testCustomConfig` so that the custom xml is added for all tests. We can then update `testCustomConfig` so that we retrieve the logs of the server pod and check to see if there's a TRACE statement to ensure that the custom logging has been applied as expected.

ryanemerson (changes requested):
@leo8a Thanks for the PR :slightly_smiling_face: 

Overall the approach is right, just a few suggestions to improve some details.

I would suggest removing the first commit though. While it does clean some parts of the code, there are lots of unrelated changes and such commits often make rebasing harder as well as making the history of the code less easy to trace as `git blame` is no longer as effective.

controllers/infinispan_controller.go
leo8a: ok, looking into it.

test/e2e/main/custom_configuration_test.go
leo8a: cool, updated the config with the suggested sample, thanks!

test/e2e/main/custom_configuration_test.go
leo8a: I like the idea, will update the test as commented.

controllers/infinispan_controller.go
leo8a: In this case, how could we then specify the correct path to the logging configuration file? It looks like the path in the `args` changes depending on the user providing a custom config file.
```
when user-provided --> user/log4j.xml
default --> /conf/operator/log4j.xml
```

ryanemerson: I was thinking that instead of passing `overlayLog4jConfigMapKey` we could pass the `overlayConfigMap` from `statefulSetForInfinispan` and check if the file exists in the CM. However, that doesn't take into account `updateStartupArgs` ... so I think it's better to stick with your original approach :+1:

leo8a: ok, updating the PR as discussed and will push it again when ready (for another round 🙂)

ryanemerson:
@leo8a I just merged some PRs that slightly conflict with this, so you will need to rebase on the latest `main` before pushing your changes.

controllers/infinispan_controller.go
```diff
@@ -1760,18 +1763,32 @@ func GossipRouterPodList(infinispan *infinispanv1.Infinispan, kube *kube.Kuberne
...
 }
 
-func buildStartupArgs(overlayConfigMapKey string, zeroCapacity string) []string {
-	var args []string
+func buildStartupArgs(overlayConfigMapKey, overlayLog4jConfigMapKey, zeroCapacity string) []string {
+	var args strings.Builder
+
+	// Preallocate a buffer of 10 bytes to speed up string building (saves code from growing the memory dynamically)
+	args.Grow(110)
+	args.WriteString("-Dinfinispan.zero-capacity-node=" + zeroCapacity)
```
ryanemerson, line 1771: This should be two different calls to `WriteString`. When using the builder you shouldn't use `+` at all.

e.g.

```go
args.WriteStritng("-Dinfinispan.zero-capacity-node=")
args.WriteString(zeroCapacity)
```

controllers/infinispan_controller.go
```diff
@@ -1760,18 +1763,32 @@ func GossipRouterPodList(infinispan *infinispanv1.Infinispan, kube *kube.Kuberne
...
+		args.WriteString(" -l" + OperatorConfMountPath + "/log4j.xml")
+	}
+
+	// Check if the user defined an overlay operator config
 	if overlayConfigMapKey != "" {
-		args = []string{"-Dinfinispan.zero-capacity-node=" + zeroCapacity, "-l", OperatorConfMountPath + "/log4j.xml", "-c", "user/" + overlayConfigMapKey, "-c", "operator/infinispan.xml"}
+		args.WriteString(" -c user/" + overlayConfigMapKey + " -c operator/infinispan.xml")
 	} else {
-		args = []string{"-Dinfinispan.zero-capacity-node=" + zeroCapacity, "-l", OperatorConfMountPath + "/log4j.xml", "-c", "operator/infinispan.xml"}
+		args.WriteString(" -c operator/infinispan.xml")
```
ryanemerson, line 1784: An advantage of using the builder is we can simplify ^ logic to something like:

```go
	args.WriteString(" -l ")
	// Check if the user defined a custom log4j config
	if overlayLog4jConfigMapKey != "" {
		args.WriteString("user/log4j.xml")
	} else {
		args.WriteString(OperatorConfMountPath)
		args.WriteString("/log4j.xml")
	}

	// Check if the user defined an overlay operator config
	if overlayConfigMapKey != "" {
		args.WriteString("-c user/")
		args.WriteString(overlayConfigMapKey)
	}
	args.WriteString(" -c operator/infinispan.xml")
```

controllers/infinispan_controller.go
```diff
@@ -1762,18 +1765,36 @@ func GossipRouterPodList(infinispan *infinispanv1.Infinispan, kube *kube.Kuberne
...
 	}
-	return args
+
+	// Check if the user defined an overlay operator config
+	args.WriteString(" -c ")
+	if overlayConfigMapKey != "" {
+		args.WriteString("user/")
+		args.WriteString(overlayConfigMapKey)
+	}
+	args.WriteString("operator/infinispan.xml")
```
ryanemerson, line 1791: We still need to provide the " -c " here, so `" -c operator/infinispan.xml"`

leo8a: done, thanks for pointing it out.

leo8a: Had to also add the overlay `log4j.xml` snipped to the `ConfigMaps` on both tests `TestUserCustomConfigUpdateOnNameChange` and `TestUserCustomConfigWithAuthUpdate`.

ryanemerson (approved):
LGTM, just waiting on CI

controllers/infinispan_controller.go
```diff
@@ -301,22 +301,25 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
-				foundKey = true
 				break
 			}
 		}
-		if !foundKey {
-			err := fmt.Errorf("infinispan-copnfig.[xml|yaml|json] configuration not found in ConfigMap: %s", overlayConfigMap.Name)
+		// Check if the user added a custom log4j.xml config
+		_, logFileExists := overlayConfigMap.Data["log4j.xml"]
+
+		if overlayConfigMapKey == "" && !logFileExists {
```
rigazilla, line 318: This confuses me. If I'm not wrong `overlayConfigMapKey` will be the last key in the `.Data` map when the `for` loops over the whole map.

also is `&&` the correct boolean op here? I mean if logFileExists and a configuration is not found the `else` branch is executed

ryanemerson (changes requested):
As discussed F2F, needs updating so that the `overlayLog4jConfigMapKey` is not always set.

controllers/infinispan_controller.go
leo8a: great observation @rigazilla! I'm currently updating it and will push it shortly, thanks!

leo8a:
Thanks @ryanemerson! I've pushed it with the discussed changes.

controllers/infinispan_controller.go
```diff
@@ -301,21 +301,22 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
 		for overlayConfigMapKey = range overlayConfigMap.Data {
 			if overlayConfigMapKey == "infinispan-config.xml" || overlayConfigMapKey == "infinispan-config.json" || overlayConfigMapKey == "infinispan-config.yaml" {
-				foundKey = true
 				break
 			}
 		}
-		if !foundKey {
-			err := fmt.Errorf("infinispan-copnfig.[xml|yaml|json] configuration not found in ConfigMap: %s", overlayConfigMap.Name)
+		// Check if the user added a custom log4j.xml config
+		overlayLog4jConfigMapKey = overlayConfigMap.Data["log4j.xml"]
```
ryanemerson, line 316: No we're storing the contents of the log4j.xml file, so the var name `overlayLog4jConfigMapKey` doesn't make sense.

I would change the var to something like `var userLog4jConfig bool` , then you can just do: `_, userLog4jConfig = overlayConfigMap.Data["log4j.xml"]`.

The `buildStartupArgs` and other method signatures will then need to be updated to take a `bool` instead of a `string`.

leo8a: ack! Updated the code accordingly, thxs.

test/e2e/main/custom_configuration_test.go
```diff
@@ -200,3 +208,28 @@ xmlns:server="urn:infinispan:server:13.0">
 		Data: map[string]string{"infinispan-config." + format: userCacheContainer},
 	}
 }
+
+func addLog4jToCustomConfigMap(customConfigMap *corev1.ConfigMap) *corev1.ConfigMap {
```
ryanemerson, line 212: Last comment, I promise :slightly_smiling_face: 

The return type of this method is never used, so we can remove `*corev1.ConfigMap` from the signature and just do `customConfigMap.Data["log4j.xml"] = ...`

ryanemerson:
Thanks @leo8a, great work!

controllers/infinispan_controller.go
```diff
@@ -302,20 +302,22 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
...
 				break
 			}
 		}
-		if !foundKey {
-			err := fmt.Errorf("infinispan-copnfig.[xml|yaml|json] configuration not found in ConfigMap: %s", overlayConfigMap.Name)
+		// Check if the user added a custom log4j.xml config
+		_, overlayLog4jConfig = overlayConfigMap.Data["log4j.xml"]
+
+		if overlayConfigMapKey == "" && !overlayLog4jConfig {
+			err := fmt.Errorf("infinispan-config.[xml|yaml|json] or log4j.xml configuration not found in ConfigMap: %s", overlayConfigMap.Name)
```
rigazilla, line 320: @ryanemerson , @leo8a please double check this snip. Still sounds weird to me. Maybe I'm reading wrong, but:
- it supposes that if `infinispan-config` is missing then `overlayConfigMapKey == ""` but it seems to me that it will be equal to the last key iterated in the map;
- plus looks like the error is returned only if both config AND log4j are missing.

ryanemerson: > plus looks like the error is returned only if both config AND log4j are missing.

This is expected. It should be possible for the user to provide 1 . Custom server config, 2. log4j.xml. 3. Both

> it supposes that if infinispan-config is missing then overlayConfigMapKey == "" but it seems to me that it will be equal to the last key iterated in the map;

Good spot! The easiest thing is to reintroduce the `foundKey` var, or we can update the `for` to use a different var name and then explicitly set `overlayConfigMapKey` if a key is found.

@leo8a can you create a follow up PR?

rigazilla: > This is expected. It should be possible for the user to provide 1 . Custom server config, 2. log4j.xml. 3. Both

oh ok, I would fix the `or` in the error message then

ryanemerson: Yeah maybe we can be more explicit and say something like:

"One of infinispan-config.[xml|yaml|json] or  log4j.xml must be present in the provided ConfigMap"

leo8a: sure, this is being fixed by [PR#1451](https://github.com/infinispan/infinispan-operator/pull/1451), thanks.

PR #1367, opened 2021-12-24
CRDs should have a separate setting for requests and limits under res…

ryanemerson:

PR #1419, opened 2022-02-28
#1367 update service yaml

oraNod:
Update to service yaml examples with cpu and mem resource req:limit.


ryanemerson:
Thanks @oraNod

Issue #1428, opened 2022-03-02
Infinispan CR foreground deletion hangs indefinitely


ryanemerson:
The following command will hang indefinitely:

`kubectl delete infinispan example-infinispan --cascade=foreground`

This is because the various controllers don't take into account the Infinispan CRs deletion timestamp when attempting to reconcile resources. Consequently, after the Infinispan CR has been deleted, the controllers will continue to provision resources (statefulset, secrets etc) and set their controller reference to the deleted CR. This results in the foreground deletion never completing, as the Infinispan CR keeps gaining new resources.

We should update the various Infinispan CR controllers to not proceed with reconciliation if the CRs deletion timestamp has been set.

PR #1429, opened 2022-03-02
Infinispan CR foreground deletion hangs indefinitely #1428

ryanemerson:


controllers/infinispan_config_controller.go
```diff
@@ -82,6 +82,12 @@ func (reconciler *ConfigReconciler) Reconcile(ctx context.Context, request recon
 		return reconcile.Result{}, fmt.Errorf("unable to fetch Infinispan CR %w", err)
 	}
 
+	// Don't reconcile Infinispan CRs marked for deletion
+	if infinispan.GetDeletionTimestamp() != nil {
```
rigazilla, line 86: Should we check if deletion timestamp is in the past? Seems it could be a future value

rigazilla: Right... I should have read after the first sentence of the docs :sweat_smile:

rigazilla (approved):
lgtm!

PR #1432, opened 2022-03-02
Jenkins export test failures

ryanemerson:


scripts/run-tests.sh
```diff
@@ -12,7 +12,12 @@ GO_LDFLAGS="-X github.com/infinispan/infinispan-operator/launcher.Version=${VERS
 
 go clean -testcache ./test/e2e
 if [ -z "${TEST_NAME}" ]; then
-  go test -v ./test/e2e/"${TEST_BUNDLE}" -timeout 90m -ldflags "${GO_LDFLAGS}" -parallel "${PARALLEL_COUNT}"
+  go test -v ./test/e2e/"${TEST_BUNDLE}" -timeout 90m -ldflags "${GO_LDFLAGS}" -parallel "${PARALLEL_COUNT}" 2>&1 | tee /tmp/testOutput
+  if [ ! -z ${TEST_REPORT_DIR} ]; then
+    mkdir -p ${TEST_REPORT_DIR}
+    cat /tmp/testOutput | ${GO_JUNIT_REPORT} > ${TEST_REPORT_DIR}/${TEST_BUNDLE}.xml
```
pruivo, line 18: shouldn't be this outside the `if`?

ryanemerson: We only want to create the xml report if `TEST_REPORT_DIR` is defined. If it's not defined, then the normal testsuite output has already been displayed via `tee` on line 15.

pruivo: :+1:

scripts/run-tests.sh
```diff
@@ -12,8 +13,15 @@ GO_LDFLAGS="-X github.com/infinispan/infinispan-operator/launcher.Version=${VERS
...
 if [ -z "${TEST_NAME}" ]; then
-  go test -v ./test/e2e/"${TEST_BUNDLE}" -timeout 90m -ldflags "${GO_LDFLAGS}" -parallel "${PARALLEL_COUNT}"
+  go test -v ./test/e2e/"${TEST_BUNDLE}" -timeout ${TIMEOUT} -ldflags "${GO_LDFLAGS}" -parallel "${PARALLEL_COUNT}" 2>&1 | tee /tmp/testOutput
+  testExitCode=${PIPESTATUS[0]}
+  if [ ! -z ${TEST_REPORT_DIR} ]; then
+    mkdir -p ${TEST_REPORT_DIR}
+    cat /tmp/testOutput | ${GO_JUNIT_REPORT} > ${TEST_REPORT_DIR}/${TEST_BUNDLE}.xml
+  fi
+  rm /tmp/testOutput
+  exit $testExitCode
```
pruivo, line 23: I'm not sure if this is a good idea. if a test fails in `main`, for example, no other stages are run.

ryanemerson: That's already the case. I was maintaining this behaviour as it allows the CI to fail-fast and free up one of the two agents for other operator builds.

Jenkinsfile
```diff
@@ -91,51 +93,67 @@ pipeline {
...
                     }
                 }
 
                 stage('Xsite') {
                     steps {
-                        sh 'scripts/ci/configure-xsite.sh'
-                        sh 'INFINISPAN_MEMORY="1Gi" go test -v ./test/e2e/xsite/ -timeout 45m'
+                        catchError (buildResult: 'FAILURE', stageResult: 'FAILURE') {
+                            sh 'scripts/ci/configure-xsite.sh'
+                            sh 'INFINISPAN_MEMORY="1Gi" ./scripts/run-tests.sh xsite 45m'
```
pruivo, line 155: I think `GO_JUNIT_REPORT` is not visible outside the `Makefile` commands and that is causing the "issues" in Jenkins. Maybe try to make a new target to run the xsite test suite.

Jenkinsfile
```diff
@@ -144,6 +162,12 @@ pipeline {
                         }
                     }
                 }
+
+                stage('Publish test results') {
+                    steps {
+                        junit 'test/reports/*.xml'
```
pruivo, line 168: maybe try https://stackoverflow.com/questions/67162746/how-to-get-rid-of-noisy-warning-no-suitable-checks-publisher-found to make this stage green?

pruivo (approved):

pruivo:
merged! thanks @ryanemerson !

Issue #1433, opened 2022-03-02
Utilise Ephemeral storage in testsuite


ryanemerson:
Utilising Ephemeral storage as the default in the test suite greatly reduces the execution time of the tests as it speeds up the provisioning of cluster pods.

Initial tests by @Crumby show the benefits of using ephemeral storage:

With PVC:
```
--- PASS: TestClusterFormationWithTLS (53.02s)
--- PASS: TestClusterFormation (56.78s)
```

With EphemeralStorage:
```
--- PASS: TestClusterFormation (28.68s)
--- PASS: TestClusterFormationWithTLS (29.63s)
```

PR #1434, opened 2022-03-02
Utilise Ephemeral storage in testsuite #1433

ryanemerson:


Crumby:
I would say that `TestPodDegradationAfterOOM` and `TestNodeWithStorageClass` should have ephemeral set to `false` too

Crumby:
@ryanemerson I wondering whether some of the "field update" tests should have ephemeral storage set to true. They are effectively restarting the pods or whole cluster with different configuration. Question is whether the update change somehow affects anything that is stored on the persistent storage.

Eg. `TestCPUUpdateWithTwoReplicas` uses two replicas only because there was an issue with global state file in the initial release and updating CPU caused some issue when the pods started cycling through update.

Crumby:
Cross site tests could leverage ephemeral storage too. No need to have PVs for the tests we currently have

ryanemerson:
> @ryanemerson I wondering whether some of the "field update" tests should have ephemeral storage set to true. They are effectively restarting the pods or whole cluster with different configuration. Question is whether the update change somehow affects anything that is stored on the persistent storage.

I think `TestContainerCPUUpdateWithTwoReplicas` is the only update test that doesn't use the default spec, so is the only update test I'm aware of that doesn't have Ephemeral storage.

> Eg. `TestCPUUpdateWithTwoReplicas` uses two replicas only because there was an issue with global state file in the initial release and updating CPU caused some issue when the pods started cycling through update.

I think it would be good to keep `TestCPUUpdateWithTwoReplicas` with actual storage so we have some coverage for StatefulSet in place pod updates with global state.

ryanemerson:
> Cross site tests could leverage ephemeral storage too. No need to have PVs for the tests we currently have

It seems they already are :slightly_smiling_face:

ryanemerson:
The results aren't dramatic, but we seemed to have shaven a few minutes of the total testsuite time.

PR #1440, opened 2022-03-03
[e2e] Remove test redundancies

### LINKED ISSUE #1404: E2E tests redundancy
`main` e2e tests currently contains tests that can be considered redundant and add up to the total execution time while providing minimal to zero added value.

**TLS:**
* `TestTLSWithExistingKeystore`
* `TestClusterFormationWithTLS`
Difference between the two above is one running 1 replica while the other runs 2 replicas. The latter is a stronger case of the first one. If the first one fails then the second one fails too. Thus having `TestClusterFormationWithTLS` should enough.

**Graceful Shutdown**
* `TestPermanentCache`
* `TestCheckDataSurviveToShutdown`
The two tests are testing basically the same thing. One is verifying the cache exists and is usable after graceful shutdown when created with `PERNAMENT` (which is default for some time now). The other checks that the cache exists and data were not lost with persistence configured. In this case the later is again stronger case and there is no added value by first one.

**Basic**
* `TestNodeStartup`
* `TestClusterFormation`
While there is a subtle difference where the first one tests also label propagation it's not clearly communicated out in the test name and if the node startup fails then all the tests fails. This two could be merged in a single 'SmokeTest' with a single deployment that would execute subtests on that deployment (verify clustering, label and other possible defaults).

**External service**
* `TestExternalService`
* `TestExternalServiceWithAuth`
All the tests are exposing external service and thus test it. Added value in those tests may be explicit configuration of user secrets by the later one but the test should be renamed. Default user verification could be included in the `SmokeTest` deployment above.

Crumby:
* Resolves #1404

### DIFF
--- a/test/e2e/main/scaling_test.go
+++ b/test/e2e/main/scaling_test.go
@@ -9,81 +9,61 @@ import (
 	tutils "github.com/infinispan/infinispan-operator/test/e2e/utils"
 )
 
-// TestPermanentCache creates a permanent cache the stop/start
-// the cluster and checks that the cache is still there
-func TestPermanentCache(t *testing.T) {
+// TestGracefulShutdownWithTwoReplicas creates a permanent cache with file-store and any entry,
+// shutdowns the cluster and checks that the cache and the data are still there
+func TestGracefulShutdownWithTwoReplicas(t *testing.T) {
 	t.Parallel()
 	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
 
-	cacheName := "test"
-	// Define function for the generic stop/start test procedure
-	var createPermanentCache = func(ispn *ispnv1.Infinispan) {
-		client_ := tutils.HTTPClientForCluster(ispn, testKube)
-		tutils.NewCacheHelper(cacheName, client_).CreateWithDefault("PERMANENT")
-	}
+	// Cache definitions
+	volatileCache := "volatile-cache"
+	volatileCacheConfig := `<distributed-cache name="` + volatileCache + `"/>`
+	filestoreCache := "filestore-cache"
+	filestoreCacheConfig := `<distributed-cache name ="` + filestoreCache + `"><persistence><file-store/></persistence></distributed-cache>`
 
-	var usePermanentCache = func(ispn *ispnv1.Infinispan) {
-		client_ := tutils.HTTPClientForCluster(ispn, testKube)
-		key := "test"
-		value := "test-operator"
-		cacheHelper := tutils.NewCacheHelper(cacheName, client_)
-		cacheHelper.TestBasicUsage(key, value)
-		cacheHelper.Delete()
-	}
-
-	genericTestForGracefulShutdown(t, createPermanentCache, usePermanentCache)
-}
-
-// TestCheckDataSurviveToShutdown creates a cache with file-store the stop/start
-// the cluster and checks that the cache and the data are still there
-func TestCheckDataSurviveToShutdown(t *testing.T) {
-	t.Parallel()
-	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
-
-	cacheName := "test"
-	template := `<infinispan><cache-container><distributed-cache name ="` + cacheName +
-		`"><persistence><file-store/></persistence></distributed-cache></cache-container></infinispan>`
-	key := "test"
-	value := "test-operator"
-
-	// Define function for the generic stop/start test procedure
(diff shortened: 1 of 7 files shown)


test/e2e/main/scaling_test.go
```diff
@@ -9,79 +9,62 @@ import (
...
 
-	cacheName := "test"
-	// Define function for the generic stop/start test procedure
-	var createPermanentCache = func(ispn *ispnv1.Infinispan) {
-		client_ := tutils.HTTPClientForCluster(ispn, testKube)
-		tutils.NewCacheHelper(cacheName, client_).CreateWithDefault("PERMANENT")
-	}
+	// Cache definitions
+	defaultCache := "default-cache"
+	defaultCacheConfig := `<distributed-cache name="` + defaultCache + `" mode="SYNC" statistics="true">
```
ryanemerson, line 20: For ease of reading I think we should simplify this to just `<distributed-cache name=" + defaultCache + "/>"

test/e2e/main/scaling_test.go
```diff
@@ -9,79 +9,62 @@ import (
...
+	defaultKey := "defaultKey"
+	defaultValue := "defaultValue"
 
-	// Delete the cluster
-	testKube.GracefulShutdownInfinispan(spec)
-	testKube.GracefulRestartInfinispan(spec, 1, tutils.SinglePodTimeout)
+	defaultCacheHelper.TestBasicUsage(defaultKey, defaultValue)
+
+	// Server may return 500 upon deleting cache after graceful shutdown of a cluster
+	// defaultCacheHelper.Delete()
```
ryanemerson, line 60: This doesn't sound right, any idea why a 500 response is being returned?

test/e2e/main/smoke_test.go
```diff
@@ -18,83 +22,79 @@ import (
 func TestNodeStartup(t *testing.T) {
 	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
 
-	// Create a resource without passing any config
+	// Infinispan cluster defintion with extra labels to be propagated to the service and pods
+	replicas := 2
```
ryanemerson, line 26: I think we should keep the `TestClusterFormation` name instead of `TestNodeStarupt` as the former implies pod startup and cluster creation works.

test/e2e/main/scaling_test.go
Crumby: No idea, it doesn't happen always but often enough. There is an issue reported already.

Between keeping `deleteCache` and increasing number of cluster nodes to two I came to a conclusion that the later is more important to be covered by tests.

ryanemerson: > There is an issue reported already.

Do you have a link? I can't find it

test/e2e/main/scaling_test.go
Crumby: Unfortunately I'm not following up. Could you elaborate?

test/e2e/main/smoke_test.go
Crumby: Could be, I was actually thinking on who to name this one as it doesn't test only cluster formation but default authentication and labels too. Maybe `TestDefaultClusterFeatures`?

test/e2e/main/scaling_test.go
ryanemerson: I got the markdown wrong, causing the config to be hidden. Hopefully the original comment is a bit clearer now.

tldr; Lets remove the unnecessary attributes and elements from the cache definition to make the test as simple as possible.

test/e2e/main/scaling_test.go
ryanemerson: [ISPN-13740](https://issues.redhat.com/browse/ISPN-13740)

test/e2e/main/smoke_test.go
ryanemerson: Sounds good.

test/e2e/main/scaling_test.go
Crumby: Replaced as suggested. Even though it would be nice to have option to define caches in some friendly way in the code. Using xml in any code is just painful. But so is that the cache doesn't have defaults for some features it nags you to have defined (encoding, fetch-state for persistent caches).

test/e2e/main/scaling_test.go
Crumby: Added Jira ID to the comment

Crumby:
@ryanemerson I also updated number of entries used by backup-restore tests. That can be technically considered redundancy too.

ryanemerson:
Thanks @Crumby

PR #1442, opened 2022-03-04
Create kind network with custom subnet

pruivo:
Trying to fix
```
[2022-03-02T19:32:35.500Z] Error response from daemon: error while removing network: network kind id d7500cdfe653399fac9b407c7b04786f4a8def5a5095bf0035f12d0e2be3c791 has active endpoints
[2022-03-02T19:32:35.500Z] Error response from daemon: network with name kind already exists
```


ryanemerson:
Two Green runs in a row :tada:

Thanks @pruivo

Issue #1449, opened 2022-03-08
Update Cache CR examples to use YAML


oraNod:
https://github.com/infinispan/infinispan-operator/blob/main/documentation/asciidoc/topics/proc_creating_caches_xml.adoc

PR #1450, opened 2022-03-08
issue #1449 create cache examples

oraNod:
issue #1449 

Needs backport to 2.2.x


documentation/asciidoc/topics/proc_creating_caches.adoc
```diff
@@ -0,0 +1,41 @@
...
+.Procedure
+
+. Create a `Cache` CR with a unique value in the `metadata.name` field.
+. Specify the target {brandname} cluster with the `spec.clusterName` field.
+. Name your cache with the `spec.name` field.
++
+[NOTE]
+====
+The `name` attribute in the cache configuration is ignored.
+Only the `spec.name` field applies to the resulting cache.
```
ryanemerson, line 16: This isn't strictly true. The cache is named using `spec.name`, but if this is omitted `meta.name` is used instead. Is this covered elsewhere?

documentation/asciidoc/topics/xml/infinispan_caches.xml
```diff
@@ -0,0 +1,27 @@
+<infinispan
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xsi:schemaLocation="urn:infinispan:config:{schemaversion} https://infinispan.org/schemas/infinispan-config-{schemaversion}.xsd
+                          urn:infinispan:server:{schemaversion} https://infinispan.org/schemas/infinispan-server-{schemaversion}.xsd"
+      xmlns="urn:infinispan:config:{schemaversion}"
+      xmlns:server="urn:infinispan:server:{schemaversion}">
+  <cache-container name="default"
+                   statistics="true">
+    <transport stack="tcp"/>
```
ryanemerson, line 9: `transport` definitions will be ignored as they are overridden by the operator configuration. Without this behaviour it would be possible for the user to break the cluster communication. 

I suggest we remove `transport` from all of our examples to avoid confusion.

ryanemerson (changes requested):
Just a couple of small points to address and your branch requires a rebase on latest `main`.

documentation/asciidoc/topics/proc_creating_caches.adoc
oraNod: @ryanemerson If you omit `spec.name` then the cache gets the same name as the Cache CR supplied with `metadata.name`?

Does the name in the cache configuration ever take effect when you create via Cache CR? I'll try this out.

ryanemerson: >If you omit spec.name then the cache gets the same name as the Cache CR supplied with metadata.name?

Yep

> Does the name in the cache configuration ever take effect when you create via Cache CR?

It shouldn't!

oraNod: I've updated with this. Thanks for pointing that out about the `metadata.name` field.

ryanemerson:
Thanks @oraNod

PR #1452, opened 2022-03-09
Improve xsite tests time

pruivo:
Initialize secrets before running any tests
Parallel tests are broken (to fix later)


ryanemerson:
CI time actually seems to be longer than on `main` :octopus:

Issue #1275, opened 2021-10-19
Allow to edit rollover policies for logging to the end user via infinispan CR


durgeshanaokar:
Infinispan CR doesn't  provide flexibility to update the rollover policies in the log4j2.xml, which controls the logging for Infinispan cluster.
As of now we can only update the logging categories and logging level. However, with default logging policies, if a pod is generating more than 100MB or say 300MB of logs daily. It is possible to face issues with ephemeral storage and may lead to shortage of space for pod.

Can we provide a functionality or any changes in specs, which could allow user to add either new policies or update the default log4j2.xml and add DefaultRolloverStrategy so at the log files will be under controls during long run of the pods.

ryanemerson:
Will be resolved by https://github.com/infinispan/infinispan-operator/issues/1364

ryanemerson:
It's now possible to define custom logging policies by providing a custom log4j.xml file  https://github.com/infinispan/infinispan-operator/pull/1417

PR #1453, opened 2022-03-09
doc updates for logging config

oraNod:


documentation/asciidoc/topics/proc_applying_custom_configuration.adoc
```diff
@@ -19,22 +19,20 @@ Changing this configuration can result in error and result in service downtime f
...
-
 .Prerequisites
 
 * Have valid {brandname} configuration in XML, YAML, or JSON format.
 
 .Procedure
 
-. Add your {brandname} configuration to the `data` field of a `ConfigMap` as `infinispan-config.[xml|yaml|json]`, as in the following example:
+. Add your {brandname} configuration to the `data` field of a `ConfigMap` as `infinispan-config.[xml|yaml|json]`.
+. Optionally include `log4j.xml` logging configuration.
```
ryanemerson, line 29: Should we omit this here and just have it in the logging section? It's valid to provide either log4j, server configuration or both.

documentation/asciidoc/topics/ref_infinispan_config_logging.adoc
```diff
@@ -0,0 +1,14 @@
+[id='infinispan-logging-configuration_{context}']
+= {brandname} logging configuration
+
+{brandname} uses Apache Log4j to provide configurable logging mechanisms that capture details about the environment and record cache operations for troubleshooting purposes and root cause analysis.
```
ryanemerson, line 4: I'm not sure we need to explain the advantage of logging here, just that's it's possible to provide a full log4j logging configuration if required.

Maybe we should point the user to the `spec.logging` configuration in the CR as the preferred method and explain this approach should be used if advantage logging options, such as file-based logging are required?

e.g https://github.com/infinispan/infinispan-operator/issues/1275

ryanemerson:
Thanks @oraNod

PR #1463, opened 2022-03-22
Use dedicated zero-capacity server configuration

ryanemerson:


ryanemerson:
One downside of this change is that if any of the service selectors incorrectly target the zero capacity pods, then user operations will fail. I need to verify the selectors are 100% correct before merging.

ryanemerson:
Service selectors appear correct :slightly_smiling_face:

Issue #1464, opened 2022-03-23
Make internal admin service headless


ryanemerson:
The addition of one k8s service to the platform creates ip rules proportional to the following equation:

`(2 rules + 3 rules * number of pods) * number of ports`

However, if the service is headless it does not create any iptable rule.

A headless service can be used by the internal controllers without any negative impact on it's capabilities, therefore we should adopt such a service.

PR #1465, opened 2022-03-23
Make internal admin service headless #1464

ryanemerson:


rigazilla (approved):
LGMT.
I'm wondering if we really need a service at all at this point. Maybe not interesting but we could save DNS entries if not.

ryanemerson:
> I'm wondering if we really need a service at all at this point. Maybe not interesting but we could save DNS entries if not.

I thought the same thing :slightly_smiling_face: 

Currently we use the service directly for the batch `cli.properties` so that we don't have to worry about recreating the batch Job if cluster pods go up/down.

PR #1492, opened 2022-03-29
Fix Openshift TestNodeWithStorageClass failures

ryanemerson:
@Crumby Let me know if you want this backported to 2.2.x

### DIFF
--- a/test/e2e/infinispan/storage_test.go
+++ b/test/e2e/infinispan/storage_test.go
@@ -1,11 +1,11 @@
 package infinispan
 
 import (
-	"fmt"
 	"testing"
 
 	ispnv1 "github.com/infinispan/infinispan-operator/api/v1"
 	tutils "github.com/infinispan/infinispan-operator/test/e2e/utils"
+	testifyRequire "github.com/stretchr/testify/require"
 )
 
 // Test if single node with a storage class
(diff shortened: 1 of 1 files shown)


test/e2e/infinispan/storage_test.go
```diff
@@ -28,8 +27,9 @@ func TestNodeWithStorageClass(t *testing.T) {
 
 	// Ensure a PVCs is bound from defaultStorageClass
 	pvcName := "data-volume-test-node-with-storage-class-0"
-	if *testKube.GetPVC(pvcName, spec.Namespace).Spec.StorageClassName != defaultStorageClass {
-		tutils.ExpectNoError(fmt.Errorf("persistent volume claim (%s) was created, but not bound to the cluster's default storage class (%s)",
-			pvcName, defaultStorageClass))
+	storageClassName := testKube.GetPVC(pvcName, spec.Namespace).Spec.StorageClassName
+	if (defaultStorageClass == "" && storageClassName != nil) || (defaultStorageClass != "" && *storageClassName != defaultStorageClass) {
```
Crumby, line 31: Just an idea, might be easier to read:
```
if (defaultStorageClass == "") {
  require.Nil(storageClassName, "StorageClassName should be empty")
} else {
  require.Equal(defaultStorageClass, storageClassName, "StorageClassName should use default storage class")
}
```
Maybe we could go even step further by using  only require.Equal() but I'm not sure what happens on `""` vs `nil` comparison

Crumby:
> @Crumby Let me know if you want this backported to 2.2.x

I haven't had issues with the test but it seems like the test was misbehaving and should be backported.

ryanemerson:
Updated and backported: https://github.com/infinispan/infinispan-operator/pull/1495

PR #1495, opened 2022-03-31
[2.2.x] Fix Openshift TestNodeWithStorageClass failures

ryanemerson:


test/e2e/infinispan/storage_test.go
```diff
@@ -28,8 +28,12 @@ func TestNodeWithStorageClass(t *testing.T) {
 
 	// Ensure a PVCs is bound from defaultStorageClass
 	pvcName := "data-volume-test-node-with-storage-class-0"
-	if *testKube.GetPVC(pvcName, spec.Namespace).Spec.StorageClassName != defaultStorageClass {
-		tutils.ExpectNoError(fmt.Errorf("persistent volume claim (%s) was created, but not bound to the cluster's default storage class (%s)",
-			pvcName, defaultStorageClass))
+	storageClassName := testKube.GetPVC(pvcName, spec.Namespace).Spec.StorageClassName
+
+	require := testifyRequire.New(t)
```
Crumby, line 33: This line shouldn't be necessary given later usage. It's to init an instance that has `t` already provided so you don't need to pass (t) every time. But given we have only single `require` in each branch there is no point of doing so.
```
require := testifyRequire.New(t)
require.Nil(storageClassName, "StorageClassName should be empty")
```
vs
```
testifyRequire.Nil(t, storageClassName, "StorageClassName should be empty")

```

ryanemerson: Argh I see, thanks for explaining. I'll update both PRs

PR #1524, opened 2022-04-12
Cross-site should check the Route object first

### LINKED ISSUE #1522: Cross-site: operator should search for Route first
The operator should search for the Route in the remote cluster first and, if not found, try to search for the cross-site service and fail if not found.
Also, add more logging to this part of the code.

pruivo:
Closes #1522

### DIFF
--- a/controllers/xsite_config.go
+++ b/controllers/xsite_config.go
@@ -120,44 +120,60 @@ func appendKubernetesRemoteLocation(ctx context.Context, infinispan *ispnv1.Infi
 	remoteServiceName := infinispan.GetRemoteSiteServiceName(remoteLocationName)
 	remoteRouteName := infinispan.GetRemoteSiteRouteName(remoteLocationName)
 
-	var host string
-	var port int32
-	var err error
+	routeObj := &reconcileType{
+		ObjectType:   &routev1.Route{},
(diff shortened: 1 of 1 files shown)


pruivo:
needs backport to 2.2.x

controllers/xsite_config.go
```diff
@@ -120,44 +120,60 @@ func appendKubernetesRemoteLocation(ctx context.Context, infinispan *ispnv1.Infi
...
-	// service found
-	if err == nil {
-		host, port, err = getCrossSiteServiceHostPort(siteService, remoteKubernetes, logger, eventRec, "XSiteRemoteServiceUnsupported", ctx)
-		if err != nil {
-			logger.Error(err, "error retrieving remote x-site service information")
-			return err
-		}
-	} else {
-		// Try searching for the Route
+	if routeSupported {
```
Crumby, line 133: Wouldn't it be a bit clearer if we checked that expose type for cross site is set to `Route` before blindly attempting to get the hostname just because Route is supported?

pruivo: in theory, you can have one cluster exposing cross-site as a Route and another as a LoadBalancer (or NodePort) service.

ryanemerson:
Thanks @pruivo

PR #1525, opened 2022-04-12
Fix JGroups cluster name

pruivo:


ryanemerson:
Thanks @pruivo. What's the consequence of the cluster name being empty? Do we need to backport this?

pruivo:
> Thanks @pruivo. What's the consequence of the cluster name being empty? Do we need to backport this?

It is only problematic if we have 2 clusters running on the same network with conflicting configurations.So there is no consequence for k8s

PR #1526, opened 2022-04-13
doc updates for bidirectional reconciliation

oraNod:
Doc updates related to #747


documentation/asciidoc/topics/con_cache_cr.adoc
```diff
@@ -9,8 +9,7 @@ When using `Cache` CRs, the following rules apply:
 * `Cache` CRs apply to {datagridservice} pods only.
 * You can create a single cache for each `Cache` CR.
 * If your `Cache` CR contains a template and a cache configuration, {ispn_operator} uses the template.
-* If you edit caches in the {osweb}, the changes are reflected through the user interface but do not take effect on the {brandname} cluster. You cannot edit caches. To change cache configuration, you must first delete the cache through the console or CLI and then re-create the cache.
-* Deleting `Cache` CRs in the {osweb} does not remove caches from {brandname} clusters. You must delete caches through the console or CLI.
+* You cannot edit caches after you create them. To change cache configuration, you must first delete the cache and then re-create it.
```
Crumby, line 12: This is not entirely true, you can change the Caches through CR as well as you could through REST api.

oraNod: Thanks, Pavel. I've removed this one.

documentation/asciidoc/topics/con_caches.adoc
```diff
@@ -13,3 +13,17 @@ To create caches with {brandname} clusters running on {k8s}, you should:
...
 You can use Hot Rod or HTTP clients but {brandname} recommends `Cache` CR or `Batch` CR unless your specific use case requires programmatic remote cache creation.
+
+[discrete]
+== Cache CRs
+
+When using `Cache` CRs, the following rules apply:
+
+* `Cache` CRs apply to {datagridservice} pods only.
+* You can create a single cache for each `Cache` CR.
+* If your `Cache` CR contains a template and a cache configuration, {ispn_operator} uses the template.
```
Crumby, line 24: This line might be confusing. Cache CR contains `template` and `templateName`, where the first should be cache configuration and the second one refers to an existing `template` on the server. I blame poor choice of the Cache CR fields in here.

oraNod: What do you think if we drop this as well as the procedure for creating caches from templates? I think we're moving away from the cache templates and they don't serve much purpose. I can't see anyone creating a cache from a template via the CR. Maybe from the console for a demo or something.

ryanemerson (commented):
It's probably beyond the scope of this PR, but we should detail how the ConfigListener creates Cache CRs and the rules that apply to them.

With https://github.com/infinispan/infinispan-operator/pull/1519 Cache CRs now transform the server cache name into a format that complies with k8s naming policy and also adds a random suffix to the created resource to avoid any naming conflicts.

ryanemerson:
Thanks @oraNod

PR #1531, opened 2022-04-14
[2.2.x] Cross-site should check the Route object first

### LINKED ISSUE #1522: Cross-site: operator should search for Route first

pruivo:
Closes #1522


ryanemerson:
Thanks @pruivo

PR #1539, opened 2022-04-28
[2.2.x] Test fixes for OLM deployed Operator/IT

Crumby:
Update integration suite:
* Update labels
* Disable broken autoscaling test

E2E
* Update label for selecting Operator pod
* Add missing `defer` delete for config map
* Shorten cluster suffix in `TestSameCacheNameInMultipleClusters` test by a bit otherwise it's possible that Operator will fail to create second route due truncating of too long name (depends also on base domain and namespace)


ryanemerson:
Thanks @Crumby

PR #1545, opened 2022-05-19
Support multiple Operand versions #1326

ryanemerson:
Builds upon https://github.com/infinispan/infinispan-operator/pull/1542

Users can now define the Operand version they require for their cluster via the `spec.version` field. For example to create a 13.0.10.Final cluster:

```yaml
spec:
  replicas: 1
  version: 13.0.10
```

An Operator's supported Operand versions are defined as a JSON array via the `INFINISPAN_OPERAND_VERSIONS` env var that is defined as part of the CSV bundle. e.g.:

```json
[{
    "downstream-version": "8.3.0"
    "upstream-version": "13.0.6",
    "image": "quay.io/infinispan/server:13.0.6.Final"
}, {
    "downstream-version": "8.3.1",
    "upstream-version": "13.0.10",
    "image": "quay.io/infinispan/server:13.0.10.Final"
}]
```

The `upstream-version` and `image` fields are mandatory. However, `downstream-version` is optional and as the name suggest is only required downstream. If `downstream-version` is set, this is the value that is used to populate `spec.Version`. For example, configuring the downstream cluster:

```yaml
spec:
  replicas: 1
  version: 8.3.1
```

Internally, the operator always uses the `upstream-version` to determine version specific configuration of resources.

# TODO
- [x] Support StatefulSet rolling upgrades for CVE images that don't contain server changes
- [x] Docs


ryanemerson:
@Crumby You may want to take a look at the tests I have put in place and how they can be updated to be used downstream. The key is that it's necessary for the `INFINISPAN_OPERAND_VERSIONS` to be set in order for the supported Operand versions to be available to the Operator. This value will normally be set in the CSV bundle, but it's also possible to set this when executing tests locally.

documentation/asciidoc/stories/assembly_upgrading_clusters.adoc
```diff
@@ -5,6 +5,7 @@ ifdef::context[:parent-context: {context}]
 
 [role="_abstract"]
 {ispn_operator} handles {brandname} cluster upgrades when new versions become available.
```
ryanemerson, line 7: Cluster upgrades are no longer automatic. Once the user upgrades to an operator that has a new server version available, they must manually set the `spec.version` field to the server version they require. Once this has been updated, the cluster upgrade will proceed.

documentation/asciidoc/topics/proc_creating_clusters.adoc
```diff
@@ -20,6 +20,13 @@ For example, create a {brandname} cluster with two pods as follows:
 {kube_client} infinispan create cluster --replicas=3 -Pservice.type=DataGrid infinispan
 ----
 +
+[TIP]
+====
+Add the `--version` argument to control the {brandname} version of your cluster.
```
ryanemerson, line 25: You have spotted a missing feature here :slightly_smiling_face:

I need to add the `--version` arg to the CLI, I've created [ISPN-13936](https://issues.redhat.com/browse/ISPN-13936) to track this

We should leave this in the docs though, as this will be available in the next Operator/Server release.

ryanemerson:
@dvagnero It's now possible for failed Hot Rod Rolling Upgrades to be rolled back on failure. We should update the docs to reflect this. The process is as follows:

1. Create cluster with `spec.version=x`
2. Initiate HR upgrade by changing `spec.version=y`
3. Failure occurs (optional)
4. Update `spec.version=x` to initiate rollback. If the version supplied here is not equal to the original cluster version in 1, then rollback will fail.

A rollback can only be initiated when a rolling upgrade is in progress. If the `status.hotRodRollingUpgradeStatus` field is not present, then rollback is not possible.

Functionally, a rollback simply deletes the newly created Infinispan cluster before the original cluster has been removed and the services switched.

pkg/infinispan/version/version.go
```diff
@@ -1,25 +1,135 @@
...
+
+		key := o.Ref()
+		if _, exists := manager.operandMap[key]; exists {
+			return nil, fmt.Errorf("multiple operands have the same version reference '%s'", key)
+		}
+		manager.operandMap[key] = o
+		manager.Operands[i] = o
+	}
+	sort.Slice(manager.Operands, func(i, j int) bool {
+		return manager.Operands[i].UpstreamVersion.LT(*manager.Operands[j].UpstreamVersion)
```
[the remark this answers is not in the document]
ryanemerson: Good spot! I don't think I had added the LT method at the Operand level when I wrote the sort function.

pkg/reconcile/pipeline/infinispan/context/context.go
```diff
@@ -58,12 +60,29 @@ type contextImpl struct {
...
+	resources      map[string]client.Object
+}
+
+func (c *contextImpl) Operand() version.Operand {
+	v := c.Infinispan.Spec.Version
+	if v == "" {
+		return c.versionManager.Latest()
+	}
+	if operand, err := c.VersionManager.WithRef(v); err != nil {
+		panic(fmt.Errorf("spec.Version '%s' doesn't exist", v))
```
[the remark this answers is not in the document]
ryanemerson: I used a `panic` here instead of returning `(version.Operand, error)` for the following reasons:

- It shouldn't be possible for `spec.Version` to be an invalid version due to the Webhook checks
- It simplifies the handler logic so we don't have to handle the returned `error` on every invocation of `Context#Operand` for a state that should never happen
- If a `panic` is thrown by a handler, then it's caught and logged by the pipeline, so the Operator will continue to function.

rigazilla (approved):

dvagnero:
@ryanemerson docs are in progress. I'll send updates later today

documentation/asciidoc/stories/assembly_upgrading_clusters.adoc
dvagnero: @ryanemerson and do users get notified when new Infinispan versions become available?

ryanemerson: > @ryanemerson and do users get notified when new Infinispan versions become available?

I think we just have to document this in the release notes. Similarly, it's important that we include a supported Operand table in the docs similar to what strimzi do at https://strimzi.io/downloads/

> Wouldn't it be possible to keep them automatic if the user wouldn't set operand version?

@Crumby Yes it's possible but I'm not sure how desirable automatic cluster upgrades are in practice, especially when the default GracefulShutdown upgrades are configured. Automatic upgrades with GracefulShutdown results in cluster downtime at the point of Operator upgrade (assuming a new Operand version is available), which we have seen is not desired by users and is why the Manual upgrade strategy has been desired by users so far. 

Automatic upgrading is something that we can easily add in the future if there's user demand.

N.B. Now that Operator -> Operand upgrading is decoupled, I would hope that more users will trust the Automatic OLM upgrading of the Operator.

dvagnero: Thanks for the info Ryan. I'll add the table for the upstream docs and will follow up with downstream later.

Crumby:
> @Crumby You may want to take a look at the tests I have put in place and how they can be updated to be used downstream. The key is that it's necessary for the INFINISPAN_OPERAND_VERSIONS to be set in order for the supported Operand versions to be available to the Operator. This value will normally be set in the CSV bundle, but it's also possible to set this when executing tests locally.

@ryanemerson I'll probably need to get hands on some build with this PR in but I don't see an issue adapting tests after it's merged. There are some requirements that current may not meet:
* Ideally test should be executable with any supported version, it looks like current implementation picks always the latest
* All supported operands need to have entries under `spec.relatedImages` in CSV. It's used to discover images required by Operator and mirroring all the images for disconnected cluster. While the section is somewhat created automatically I'm not sure it'll work with `INFINISPAN_OPERAND_VERSIONS` format.

It might be enough to list only latest released image in `relatedImages` section though. We'll still need new bundle with each server release and as a result all the images should be mirrored.

One question I have is how does the functionality work with CVE respins? Let's say we have respin of x.y.z release. How does the user triggers the update in this case?

ryanemerson:
+1 to updating tests as we go, just some comments below:

> * Ideally test should be executable with any supported version, it looks like current implementation picks always the latest

That's true of `TestUpgrade` in `test/e2e/upgrades`, but we also have the following tests in `test/e2e/infinispan`:

`TestOperandUpgrade` follows the entire update graph, upgrading one Operand at a time, skipping CVE releases.
`TestOperandCVEUpgrade` upgrades from the oldest release to the latest, with the expectation that the latest release has to have CVE==true.

Any ideas on how we can update `TestUpgrade` to allow specific versions to be installed per Operator version?


> * All supported operands need to have entries under `spec.relatedImages` in CSV. It's used to discover images required by Operator and mirroring all the images for disconnected cluster. While the section is somewhat created automatically I'm not sure it'll work with `INFINISPAN_OPERAND_VERSIONS` format.

If `spec.relatedImages` is manually defined, then the digest pinning that previously occurred on `RELATED_IMAGE_OPENJDK` does not take effect. My plan was that we will need to manually enter the digest for the latest Operand when releasing a CSV and add a new entry to the `relatedImages` array. The digest for old Operand versions can never change, so each CSV release should only require the manual definition of a single Operand.

> One question I have is how does the functionality work with CVE respins? Let's say we have respin of x.y.z release. How does the user triggers the update in this case?

A new Operator release is required in order for the CVE image to be consumable by the user. The `spec.version` field supports versions with qualifiers, so we can simply do something like below but with appropriate downstream versions:

```json
                [{
			"upstream-version": "13.0.9-1",
			"image": "quay.io/infinispan/server@sha256...."
		},{
			"upstream-version": "13.0.9-2",
			"image": "quay.io/infinispan/server@sha256....l",
			"cve": true
		}]
```
The user must then manually change the `spec.version=13.0.9-2` in order for the upgrade to occur. I toyed with the idea of making CVE upgrades automatic as they use StatefulSet rolling upgrades, however in the case of `spec.replicas=1` that would still result in cluster downtime.

Crumby:
> That's true of TestUpgrade in test/e2e/upgrades

I had all the functional tests on mind this time.

> Any ideas on how we can update TestUpgrade to allow specific versions to be installed per Operator version?

Nothing reasonably simple. I'd go with latest/latest by default. The ability to pick server version separately from the Operator version gives so many possibilities that maybe it would be best to have soak tests testing random upgrade paths. But that's out of scope for CI

> I toyed with the idea of making CVE upgrades automatic as they use StatefulSet rolling upgrades, however in the case of spec.replicas=1 that would still result in cluster downtime.

I'd ignore this scenario. Nobody who wishes to have HA and no cluster downtime should be running single replica. I wonder whether we could ignore CVE versioning in such case. (CVE release would replace the cluster automatically on Operator upgrade)

rigazilla:
> My experiences with customer's so far, lends me to believe that they want as much control over the upgrade process as possible so that the can schedule upgrades during off-peak hours etc. We could include a `spec.upgrades.rollingCve` flag that allows users to specify whether automatic CVE upgrades should be enabled, with the default value being true. That way it's at least possible for the user to disable the behaviour if they want more control or if it causes issues with their cache configurations. WDYT @rigazilla @Crumby?

I think it's ok to let the user the freedom to choose. As a user I'd prefer to have it defaulted to false, I'm of those that want control on my systems.

Crumby:
> We could include a spec.upgrades.rollingCve flag that allows users to specify whether automatic CVE upgrades should be enabled, with the default value being true.

Sure, maybe we don't need to rush with the option immediately and wait for the user feedback if anybody would be interested. Otherwise we have other option that we'd need to include in testing

ryanemerson:
> Sure, maybe we don't need to rush with the option immediately and wait for the user feedback if anybody would be interested.

That sounds like a good plan to me. Let's keep the current behaviour of not automatically applying the CVE upgrade and evolve the feature in the future based upon user feedback.

documentation/asciidoc/topics/proc_recovering_rolling_upgrades.adoc
```diff
@@ -0,0 +1,23 @@
+[id='recovering_rolling_upgrades_{context}']
+= Recovering from a failed Hot Rod rolling upgrade
+
+[role="_abstract"]
+You can roll back failed Hot Rod rolling upgrade to a previous version if the original cluster is still present.
```
ryanemerson, line 5: You can only rollback to the version that was installed prior to the upgrade being initiated, so maybe we should have:

"You can rollback a failed Hot Rod rolling upgrade to the previous version ..."

documentation/asciidoc/topics/proc_recovering_rolling_upgrades.adoc
```diff
@@ -0,0 +1,23 @@
...
+. Ensure the Hot Rod rolling upgrade is in progress.
++
+[source,options="nowrap",subs=attributes+]
+----
+{oc_get_infinispan} <cr_name> -o yaml
+----
++
+The `status.hotRodRollingUpgradeStatus` field must be present.
++
+. Update `spec.version` field of your `Infinispan CR` to initiate rollback to the original cluster version.
```
ryanemerson, line 21: We should be more explicit here. We need to update `spec.version` to the original `spec.version` prior to the failed upgrade, this version is recorded in the `status.hotRodRollingUpgradeStatus.SourceVersion` field. An invalid value will result in the Infinispan CR update being rejected by kubernetes.

ryanemerson: So maybe the following:

"Update the `spec.version` field of your Infinispan CR to the original cluster version defined in `status.hotRodRollingUpgradeStatus.SourceVersion`."

PR #1551, opened 2022-05-27
Testing log output

ryanemerson:
@Crumby The reason your PR wasn't working is due to the following from the `archiveArtifacts` [docs](https://www.jenkins.io/doc/pipeline/steps/core/):

```qoute
The base directory is the workspace. You can only archive files that are located in your workspace.
```

I've added an additional commit to make it so the log output dir is configurable via an env var. The default will be to use `os.TempDir()+/infinispan-operator`, but in Jenkins we use `$WORKSPACE/log`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,6 +25,7 @@ pipeline {
         GO111MODULE = 'on'
         KUBECONFIG = "$WORKSPACE/kind-kube-config.yaml"
         TESTING_NAMESPACE = 'namespace-for-testing'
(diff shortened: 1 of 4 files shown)


Crumby:
@ryanemerson Ah, nice find, I was scratching my head over this

ryanemerson:
Thanks @Crumby. You can see the archived artifacts [here](https://ci.infinispan.org/job/Infinispan%20Operator/job/PR-1551/6/)

PR #1552, opened 2022-05-27
[2.2.x] Write logs and yaml to files on panic

ryanemerson:


test/e2e/utils/kubernetes.go
```diff
@@ -155,20 +156,30 @@ func (k TestKubernetes) CleanNamespaceAndLogOnPanic(t *testing.T, namespace stri
...
-		k.PrintAllResources(namespace, &corev1.PodList{}, map[string]string{"app": "infinispan-pod"})
-		k.PrintAllResources(namespace, &corev1.PodList{}, map[string]string{"app": "infinispan-batch-pod"})
-		k.PrintAllResources(namespace, &appsv1.StatefulSetList{}, map[string]string{})
-		k.PrintAllResources(namespace, &ispnv1.InfinispanList{}, map[string]string{})
-		k.PrintAllResources(namespace, &ispnv2.BackupList{}, map[string]string{})
-		k.PrintAllResources(namespace, &ispnv2.RestoreList{}, map[string]string{})
-		k.PrintAllResources(namespace, &ispnv2.BatchList{}, map[string]string{})
-		k.PrintAllResources(namespace, &ispnv2.CacheList{}, map[string]string{})
+	specLabel := make(map[string]string)
+	if t != nil {
```
Crumby, line 160: I wonder in what use cases the `t` is nil?

ryanemerson: There's not a specific case in the `2.2.x` code base atm where we pass a nil `t`, but I figured backporting the check now has no consequences and will make subsequent backports related to this code simpler as this check is need on `main`.

Crumby:
@ryanemerson MRs LGTM thanks for looking at this

PR #1555, opened 2022-06-07
[2.2.x] Graceful shutdown does not work with xsite

### LINKED ISSUE #1554: Graceful shutdown does not work with cross-site replication
When `spec.replicas` is set to zero, the operator performs a graceful shutdown in the Infinispan cluster and also stops the gossip router pod.

However, there is a small code that sets the `GossipRouterReady` condition and, when it does not find the gossip router pod, it assumes it is not ready and reschedules the reconciliation.

It is incorrect and the code should let the operator continue with the graceful shutdown. The effect is `GracefulShutdown` condition is never created/set and `Stopping` is never set to `false`.

It affects branch 2.2.x. I didn't test in other versions (edit: tested locally with `kind`)

pruivo:
When cross-site is enabled, the Gossip Router pod condition blocks the
process.

closes #1554

### DIFF
--- a/controllers/infinispan_controller.go
+++ b/controllers/infinispan_controller.go
@@ -357,80 +357,9 @@ func (reconciler *InfinispanReconciler) Reconcile(ctx context.Context, ctrlReque
 	}
 
 	if infinispan.HasSites() {
-		var gossipRouterTLSSecret *corev1.Secret
-		if infinispan.IsSiteTLSEnabled() {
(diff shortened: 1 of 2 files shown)


ryanemerson:
LGTM, just running the CI again to make sure the upgrade failure isn't related

ryanemerson:
Thanks @pruivo

PR #1577, opened 2022-06-28
add istio-proxy support

vladimir22:
Hi Team, 
We are successfully using your `infinispan-operator (v.1.x.x, v.2.1.x, v.2.2.x)` product which is covered by [Istio service mesh](https://istio.io/latest/about/service-mesh/) in our k8s clusters. 
It requires minor updates inside the `infinispan-operator` sources.

I would be very appreciated if you review this PR and include new "Istio Support" feature into your next release.

Thanks in advance,
Volodymyr.


ryanemerson (commented):
@vladimir22 Thanks for the contribution :slightly_smiling_face: 

All new features need to be contributed to the `main` branch, as `2.2.x` is restricted to bug fixes only.

With regards to the contribution itself, Istio support is something that we're definitely interested in providing officially. However, there are a few things that need to be addressed:

- The StatefulSet annotation should be applied to existing StatefulSets on spec updates as well as new StatefulSets.
- Automated testing (this is something I can look into)

pkg/kubernetes/kubernetes.go
```diff
@@ -206,7 +206,14 @@ func (k Kubernetes) GetKubernetesRESTConfig(masterURL, secretName, namespace str
 		logger.Error(err, "unable to find Secret", "secret name", secretName)
 		return nil, err
 	}
-
+	
+	// We want to connect the opposite site using "infinispan-operator" secret with token value inside
+	if token, ok := secret.Data["token"]; ok {
```
ryanemerson, line 211: Can you explain why this change is required?

vladimir22: It is not related to the "Istio Feature", but anyway, this logic was taken from the previous  [infinispan-operator 1.1.x](https://github.com/infinispan/infinispan-operator/blob/1.1.x/pkg/controller/infinispan/infinispan_controller.go#L1826) version
and we are still successfully using that because working with token value (infinispan-operator's namespace scope) from the backend side is more safety than using the _whole_  admin.conf credentials (Cluster scope).
Please advise me if you have a better idea.
Thanks.

ryanemerson: You've added this code to the `GetKubernetesRESTConfig` which shouldn't be called when the Openshift url scheme is used for backup site Urls. The `token` value is still part of `GetOpenShiftRESTConfig` and is used with the Openshift scheme.

vladimir22: Yes, you right, I have changed schema to `openshift://` and it works well with token value on Vanilla Kubernetes clusters :)

controllers/constants/constants.go
```diff
@@ -114,6 +114,11 @@ const (
 			</distributed-cache>
 		</cache-container>
 	</infinispan>`
+
+	InfinispanSslKeyExchangePort        = 2157
```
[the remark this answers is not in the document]
vladimir22: Thanks for the note, I see that it is not needed anymore.
It was needed for previous infinispan-operator version for SSL_KEY_EXCHANGE (config example in the [issue](https://github.com/infinispan/infinispan-images/issues/31)).
I enabled this block using flag "jgroups.encrypt = true" inside the infinispan.yaml

vladimir22:
@ryanemerson,
Thanks for your review, and let me share my thoughts about your comment:

_All new features need to be contributed to the main branch, as 2.2.x is restricted to bug fixes only._
> Ok, I will create additional PR against the `main` branch when we come to conclusion with this PR.

_However, there are a few things that need to be addressed:
The StatefulSet annotation should be applied to existing StatefulSets on spec updates as well as new StatefulSets._
> It is not clear to me why we have to update existing StatefulSets? 
Currently, I just need to add [special annotations](https://github.com/vladimir22/infinispan-operator/blob/b507222e98c02184b76fe5aaf824e1d275995629/controllers/infinispan_controller.go#L1653-L1654) into StatefulSet if I want to use x-site connection (exclude these ports from the [Istio Mesh](https://istio.io/latest/docs/reference/config/annotations/) because backup site is out of the mesh). These annotations are working only for newly created PODs to configure [istio-proxy](https://github.com/istio/proxy#istio-proxy) sidecar container.

_Automated testing (this is something I can look into)_
> It is because I have added a [custom logic](https://github.com/vladimir22/infinispan-operator/blob/b507222e98c02184b76fe5aaf824e1d275995629/controllers/infinispan_controller.go#L1641) to have a chance to add custom StatefulSet annotations through the "Infisinpan" resource and I suspect it has conflicts with existing tests.


Let me summarize it, 

We have to implement the next if we want to use "Istio Support" feature:
 - [Check ALL container statuses](https://github.com/vladimir22/infinispan-operator/blob/b507222e98c02184b76fe5aaf824e1d275995629/pkg/kubernetes/pods.go#L44-L49) inside the POD because of using istio-proxy
 - Have a chance to inject custom annotations into the StatefulSet. (Currently, I am using Infinispan [annotations](https://github.com/vladimir22/infinispan-operator/blob/b507222e98c02184b76fe5aaf824e1d275995629/controllers/infinispan_controller.go#L1652) custom resource field, but probably you have a better idea).
 - Have a chance to inject custom annotations into [Gossip Router Deployment](https://github.com/vladimir22/infinispan-operator/blob/b507222e98c02184b76fe5aaf824e1d275995629/controllers/xsite.go#L119-L122) (Currently, it is hardcoded, that is not good)

What do you think about that?

Thanks,
Volodymyr.

Issue #1618, opened 2022-07-11
Provide greater flexibility for custom user configuration


ryanemerson:
Currently custom Infinispan configurations provided via `spec.configMapName` are provided as an initial configuration to the server and the generated Operator configuration is applied as an overlay. Consequently, any security-realm (and other) configurations defined by the user are overwritten by the Operator config.

To allow the user the freedom to configure more aspects of the server, we should split the configurations into 3 overlays:

1. `infinispan-base.xml` - Contains all user impacting configurations that are generated by the Infinispan CR spec. For example, cache-container authorization, the default security-realm and default endpoint definitions.
2. `infinispan-config.xml` - Configuration provided by the user
3. `infinispan-admin.xml` - Only essential configuration that is required in order for the Operator to function. For example, the "admin" endpoint and security-realm.

This will allow the user the freedom to modify any of the content generated in 1, allowing features such as LDAP, which is currently not possible with the 2 file overlay.

PR #1619, opened 2022-07-11
Provide greater flexibility for custom user configuration #1618

ryanemerson:


rigazilla (commented):
the approach looks good to me. Is this PR covered by the current testsuite?

ryanemerson:
`test/e2e/infinispan/custom_configuration_test.go` provides some coverage, but I'm working on a new test that enables LDAP to prove that the 3 way approach allows security-realm overwriting as expected.

ryanemerson:
Merging now as the test will be added in a follow up PR.

PR #1633, opened 2022-07-14
[2.2.x] Webhook prevents Cache CRs being created with the same spec.Name for …

ryanemerson:


api/v2alpha1/cache_webhook.go
```diff
@@ -126,8 +126,8 @@ func (cv *cacheValidator) Create(ctx context.Context, c *Cache) error {
 		allErrs = append(allErrs, field.InternalError(field.NewPath("spec").Child("name"), err))
 	} else {
 		for _, cache := range list.Items {
-			if c.Spec.Name == cache.Spec.Name {
-				msg := fmt.Sprintf("Cache CR already exists with spec.Name '%s'", c.Spec.Name)
+			if c.Spec.Name == cache.Spec.Name && c.Spec.ClusterName == cache.Spec.ClusterName {
```
Crumby, line 129: Does the code differentiate between create and update operations? Just curios whether it's possible that this part of code would refuse to update the CR on cache config update

ryanemerson: Yes this only applies to create operations. For updates `spec.Name` and `spec.clusterName` are now immutable, so it's not possible to change the value after resource creation.

PR #1637, opened 2022-07-20
Infinispan 14 support

ryanemerson:
`Dev03` image to be replaced with any subsequent dev releases and then `14.0.0.Final` before next Operator release.


ryanemerson:
The upgrade test will continue to fail until ISPN-14023 is resolved and available in the next Infinispan 14 release.

pruivo:
merged! thanks @ryanemerson !

PR #1655, opened 2022-07-27
Improve HR Upgrade test

Crumby:
* Don't explicitly configure expose type. It's inherited from `DefaultSpec` and configured by env var.
* Increase cluster number with each upgrade


ryanemerson:
Thanks @Crumby

ryanemerson:
I'll integrate this into https://github.com/infinispan/infinispan-operator/pull/1654 for `main`.

Issue #1657, opened 2022-08-02
Document new fields for jvm options routerExtraJvmOpts and cliExtraJvmOpts


dvagnero:
Documents #1643

PR #1661, opened 2022-08-03
1657#/cli_jvm_options

dvagnero:
#1657 
Docs updates for routerExtraJvmOpts and cliExtraJvmOpts
Needs backport to 2.2.x

### DIFF
--- a/documentation/asciidoc/topics/yaml/container_extra_jvm_opts.yaml
+++ b/documentation/asciidoc/topics/yaml/container_extra_jvm_opts.yaml
@@ -1,5 +1,5 @@
 spec:
   container:
-    extraJvmOpts: "-XX:NativeMemoryTracking=summary"
-    gossipRouterExtraJvmOpts: ""
-    cliExtraJvmOpts: ""
+    extraJvmOpts: "-<option>=<value>"
+    routerExtraJvmOpts: "-<option>=<value>"
+    cliExtraJvmOpts: "-<option>=<value>"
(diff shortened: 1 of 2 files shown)


documentation/asciidoc/topics/yaml/container_extra_jvm_opts.yaml
```diff
@@ -1,3 +1,5 @@
 spec:
   container:
     extraJvmOpts: "-XX:NativeMemoryTracking=summary"
+    gossipRouterExtraJvmOpts: ""
```
dvagnero, line 4: @ryanemerson Hi, could you please think about values for these two fields in this example? I'm looking for something more generic then disabling fips

ryanemerson: The field name is `routerExtraJvmOpts` not `gossipRouterExtraJvmOpts`.

The current `extraJvmOpts` value is valid for both the `cliExtraJvmOpts` and `routerExtraJvmOpts` fields as it's a generic JVM arg, so we can keep it simple and set it for all three.

Users may want to tune the JVM used by these components, so generic JVM args such as `-Xms4` and `-Xmx` can be specified, however the only specific use-case example we have is FIPs.

dvagnero: Thank you. I went with placeholders

ryanemerson:
Thanks @dvagnero

PR #1676, opened 2022-09-06
Use TLSv1.3 as default for Cross Site

pruivo:


pruivo:
> Can't we just set `TLS` and let the handshake figure the best ?

Sure; I had the idea that we want to force TLS 1.3 everywhere.
If both the server and gossip-router support TLS 1.2 and TLS 1.3, which version will the protocol choose?

pruivo:
keeping this on hold until we have a server image with [JGRP-2638](https://issues.redhat.com/browse/JGRP-2638)

ryanemerson:
@pruivo Is this included in 14.0.0.Final?

pruivo:
@ryanemerson no, it isn't. Part of the fix is in JGroups code and we didn't upgrade yet.

ryanemerson:
Thanks for the update. Just let me know when the time is right :+1:

PR #1690, opened 2022-09-30
Document running Infinispan on a FIPS-enabled OCP cluster

dfitzmau:
ISPN: https://issues.redhat.com/browse/ISPN-14185

### DIFF
--- a/documentation/asciidoc/topics/con_fips_mode_cr.adoc
+++ b/documentation/asciidoc/topics/con_fips_mode_cr.adoc
@@ -1,0 +1,16 @@
+[id="fips-mode-CR_{context}"]
+= FIPS mode for your `Infinispan` CR
+
+[role="_abstract"]
+The {openshiftplatform} can use certain Federal Information Processing Standards (FIPS) components that ensure {openshiftshort} clusters meet the requirements of a FIPS compliance audit.
+
+If you enabled FIPS mode on your {openshiftshort} cluster then the {ispn_operator} automatically enables FIPS mode for your `{brandname}` CR.
+
+[IMPORTANT]
+====
+Client certificate authentication is not currently supported with FIPS mode. Attempts to create {brandname} CR with `spec.security.endpointEncryption.clientCert` set to a value other than `None` will fail.
+====
+
+[role="_additional-resources"]
+.Additional resources
+* link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html/installing/installing-fips#doc-wrapper[Support for FIPS cryptography {openshiftplatform}]
--- a/documentation/asciidoc/stories/assembly_creating_services.adoc
+++ b/documentation/asciidoc/stories/assembly_creating_services.adoc
@@ -18,6 +18,11 @@ include::{topics}/ref_persistent_cache_store.adoc[leveloffset=+2]
 include::{topics}/proc_allocating_cpu_memory.adoc[leveloffset=+1]
 include::{topics}/proc_setting_jvm_options.adoc[leveloffset=+1]
 
+// Downstream content
+ifdef::downstream[]
+include::{topics}/con_fips_mode_cr.adoc[leveloffset=+1]
+endif::downstream[]
+
 //Logging
 include::{topics}/proc_configuring_logging.adoc[leveloffset=+1]
 include::{topics}/ref_logging.adoc[leveloffset=+2]


ryanemerson (changes requested):
@dfitzmau I think there's some confusion here. In `main` and the next downstream release, FIPS mode IS supported for all use-cases except [client cert authentication](https://infinispan.org/docs/infinispan-operator/main/operator.html#client-certificates). The user does not have to configure their Infinispan CR any differently to when running in a cluster without FIPS enabled.

We should add the FIPS section, but it only needs to outline that client-cert auth is not supported with FIPS.

It's also be worth mentioning that on upgrading from a 2.2.5 (DG 8.3.7) Operator to 2.3.0 (DG 8.4), the previously required `-Dcom.redhat.fips=false` properties are automatically removed from the Infinispan CR.

dfitzmau:
Hi @ryanemerson 

Just to confirm that the Infini Operator disables FIPS mode for an Infinispan CR by default? One can enable FIPS mode is they want to? If so, why would they want to enable it and should I document how they do this?

ryanemerson:
> Just to confirm that the Infini Operator disables FIPS mode for an Infinispan CR by default?

No, Infinispan 14.0.0 is now able to operate with FIPS mode (previously  it couldn't), so there's no need for the user to explicitly disable it with `-Dcom.redhat.fips=false`.

> One can enable FIPS mode is they want to? If so, why would they want to enable it and should I document how they do this?

FIPS mode  is enabled in Infinispan if it has already been enabled for the Openshift cluster. There's no need for the user to perform any additional actions with the Operator and Infinispan 14.0.0.


> It's also be worth mentioning that on upgrading from a 2.2.5 (DG 8.3.7) Operator to 2.3.0 (DG 8.4), the previously required `-Dcom.redhat.fips=false` properties are automatically removed from the Infinispan CR.

I should clarify my statement above as it's incorrect. FIPS is still not supported in Infinispan 13.0.10.Final (DG 8.3.1) operand, so if the user uses this operand version then they still have to set `-Dcom.redhat.fips=false`. When the user upgrades the operand from `spec.version=13.0.10` -> `spec.version=14.0.0` (8.3.1 -> 8.4.0) on a FIPS enabled cluster, the `-Dcom.redhat.fips=false` configuration is automatically removed.

documentation/asciidoc/topics/con_fips_mode_cr.adoc
```diff
@@ -0,0 +1,22 @@
...
+[role="_abstract"]
+The {openshiftplatform} can use certain Federal Information Processing Standards (FIPS) components that ensure {openshiftshort} clusters meet the requirements of a FIPS compliance audit.
+
+If you enabled FIPS mode on your {openshiftshort} cluster then the {ispn_operator} automatically enables FIPS mode for your `{brandname}` CR.
+
+[IMPORTANT]
+====
+The client-certificate implementation does not support FIPS mode.
+For a FIPS-mode enabled `{brandname}` CR, you cannot configure the client-certificate implementation to secure an application that runs inside your `{brandname}` cluster.
+You can resolve this issue by setting `extraJvmOpts: "-Dcom.redhat.fips=false"` in the `spec.container` field in your `Infinispan` CR.
```
ryanemerson, line 13: This won't work, as the Infinispan webhook will prevent the Infinispan CR being created if Client cert is configured.

How about:

```
[IMPORTANT]
====
Client certificate authentication is not currently supported with FIPS mode. Attempts to create an Infinispan CR with `spec.security.endpointEncryption.clientCert` set to a value other than `None` will fail.
====
```

dfitzmau: Yes, this works. I updated the note as per your suggestion. 

Are you OK to approve the changes or do you see other changes?

documentation/asciidoc/topics/con_fips_mode_cr.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+[IMPORTANT]
+====
+Client certificate authentication is not currently supported with FIPS mode. Attempts to create {brandname} CR with `spec.security.endpointEncryption.clientCert` set to a value other than `None` will fail.
+====
+
+[role="_additional-resources"]
+.Additional resources
+* link:https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html/installing/installing-fips#doc-wrapper[Support for FIPS cryptography {openshiftplatform}]
+* link:https://access.redhat.com/documentation/en-us/red_hat_data_grid/{ispn_version}}/html-single/data_grid_operator_guide/index#setting-jvm-options_services[Setting JVM options]
```
ryanemerson, line 17: I don't think we need this link any more.

dfitzmau: I removed the "Setting JVM options" link.

ryanemerson:
Thanks @dfitzmau

PR #1698, opened 2022-10-13
Enable version selection in tests + fixes

Crumby:
Includes also some fixes in relation to Infinispan 14 and new OpenShift versions

Changes summary:
* Enable server version selection in integration tests by property
* Enable server version selection in Go E2E tests (default to none unless Operator runs locally)
* Bump up Infinispan version in integration tests
* Retrieving grafana-service-account token needs update in order to work with OCP4.11
* Fix num of onwers parser to work with Infinispan 13 and 14 in integration tests
* Randomize serial number for certificates generated by E2E tests


test/e2e/utils/tls.go
```diff
@@ -29,7 +31,7 @@ const (
 	tmpDir             = "/tmp/infinispan/operator/tls"
 )
 
-var serialNumber int64 = 1
+var serialNumber int64 = mrand.Int63n(math.MaxInt64)
```
ryanemerson, line 34: You need to either seed  mrand or use the "crypto/rand" package here. Otherwise the randomness is deterministic on each process start, which we don't want in this case as we're trying to avoid two simultaneous  test runs having the same `serialNumber`.

https://stackoverflow.com/questions/63655205/random-integer-in-go-language-are-predictable-in-each-runtime

test/e2e/utils/common.go
```diff
@@ -173,10 +172,18 @@ func DefaultSpec(t *testing.T, testKube *TestKubernetes, initializer func(*ispnv
 	if initializer != nil {
 		initializer(infinispan)
 	}
+
+	if OperandVersion != "" {
+		infinispan.Spec.Version = OperandVersion
+	}
```
ryanemerson, line 178: If we add this after the `if RunLocalOperator == TRUE` block, then we can overwrite in both modes of operation.

Crumby: Good point

test/e2e/utils/tls.go
Crumby: Fixed, hopefully correctly

ryanemerson:
There's quite a few failures

Crumby:
> There's quite a few failures

The issue was in the `initializer` running before setting up version, which Operand upgrade tests did not like. After switching the execution Operand upgrades tests are happy but lot of other started to be unhappy

Crumby:
@ryanemerson There is only one failure atm at it seem to be rather random as it did not appear in previous runs:
```
[2022-10-20T08:33:18.338Z] === CONT  TestClientCertValidate
[2022-10-20T08:33:18.338Z]     kubernetes.go:208: Operation cannot be fulfilled on infinispans.infinispan.org "test-client-cert-validate": the object has been modified; please apply your changes to the latest version and try again 
[2022-10-20T08:33:18.339Z]         
[2022-10-20T08:33:18.339Z] --- FAIL: TestClientCertValidate (22.24s)
```

I also updated upgrade tests so the version manager fetches the the json from the CSV if the Operator is deployed via OLM

test/e2e/utils/common.go
```diff
@@ -30,7 +31,25 @@ const (
 	truststoreSecretSuffix = "truststore"
 )
 
-var VersionManager = func() *version.Manager {
+// var VersionManager = func() *version.Manager {
+func VersionManager(testKube *TestKubernetes) *version.Manager {
+	// If Operator is running via OLM, fetch the envs
+	if RunLocalOperator == "FALSE" {
```
ryanemerson, line 37: This won't work if we want to run the tests against an Operator deployed via `skaffold run`.

test/e2e/utils/olm.go
```diff
@@ -259,6 +263,18 @@ func (k TestKubernetes) InstalledCSVEnv(envName string, sub *coreos.Subscription
...
+func (k TestKubernetes) SetRelatedImagesEnvs(sub *coreos.Subscription) {
+	csv, err := k.InstalledCSV(sub)
+	ExpectNoError(err)
+
+	envVars := csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs[0].Spec.Template.Spec.Containers[0].Env
+	for _, envVar := range envVars {
+		if strings.HasPrefix(envVar.Name, "RELATED") {
+			os.Setenv(envVar.Name, envVar.Value)
+		}
+	}
```
ryanemerson, line 275: I'm confused why we need to set these Env vars on the node running the tests, what am I missing?

Crumby: The VersionManager will crash with downstream since they refer env. But those envs aren't available on host that runs the tests.

test/e2e/utils/common.go
Crumby: So should we change the env to be rather enum then true false? Is skaffold something we are using at the moment?

ryanemerson: > So should we change the env to be rather enum then true false?

Tbh I'm not sure this should exist upstream. We already have the `INFINISPAN_OPERAND_VERSIONS` env variable that allows you to set the versions used by the tests. Can't you execute the equivalent of the `SetRelatedImagesEnv` in a script on Jenkins before executing the downstream testsuite?

> Is skaffold something we are using at the moment?

Yes, it greatly simplifies deploying and remote debugging the operator in development. Basic instructions are on the README.

test/e2e/infinispan/upgrade_operand_test.go
```diff
@@ -114,8 +124,12 @@ func TestOperandCVERollingUpgrade(t *testing.T) {
 // the currently installed operand, result in a GracefulShutdown upgrade
 func TestOperandCVEGracefulShutdown(t *testing.T) {
 	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
+	versionManager := tutils.VersionManager()
+
+	if !versionManager.Latest().CVE {
+		t.Skip("Latest release is non-cve, skipping test")
+	}
```
ryanemerson, line 131: If we're going to skip the test, then we can remove the `if !cveOperand.CVE {` check on L146-150.

test/e2e/infinispan/upgrade_operand_test.go
```diff
@@ -78,7 +84,11 @@ func TestOperandUpgrade(t *testing.T) {
 // installed operand, only result in a StatefulSet rolling upgrade
 func TestOperandCVERollingUpgrade(t *testing.T) {
 	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
-	versionManager := tutils.VersionManager
+	versionManager := tutils.VersionManager()
+
+	if !versionManager.Latest().CVE {
+		t.Skip("Latest release is non-cve, skipping test")
+	}
```
ryanemerson, line 91: If we're going to skip the test, then we can remove the `if !cveOperand.CVE {` check on L95-99.

test/e2e/infinispan/upgrade_operand_test.go
Crumby: Removed

test/e2e/utils/olm.go
```diff
@@ -56,8 +56,8 @@ func (k TestKubernetes) OLMTestEnv() OLMEnv {
 	env := &OLMEnv{
 		CatalogSource:          constants.GetEnvWithDefault("SUBSCRIPTION_CATALOG_SOURCE", "test-catalog"),
 		CatalogSourceNamespace: constants.GetEnvWithDefault("SUBSCRIPTION_CATALOG_SOURCE_NAMESPACE", Namespace),
-		SubName:                constants.GetEnvWithDefault("SUBSCRIPTION_NAME", "infinispan-operator"),
-		SubNamespace:           constants.GetEnvWithDefault("SUBSCRIPTION_NAMESPACE", Namespace),
+		SubName:                SubscriptionName,
+		SubNamespace:           SubscriptionNamespace,
```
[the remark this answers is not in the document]
Crumby: Nope, brought back the original code

test/e2e/infinispan/upgrade_operand_test.go
Crumby: Removed

ryanemerson:
Thanks @Crumby. I included the  PR summary in the commit so that we don't lose the history

PR #1701, opened 2022-10-17
updates GH actions to sync from 14.0.x

dvagnero:


oraNod:
@dvagnero Thanks for this. I believe we need to wait until @ryanemerson branches to 2.3.x.

When 2.3.x is cut we need to add it here: https://github.com/infinispan/infinispan-operator/blob/main/.github/workflows/sync_docs.yaml#L6

Send the first commit against the main branch then backport to 2.3.x and 2.2.x. All workflows should match the same ref/heads pattern for the action to run.

dvagnero:
Hi @oraNod thanks for the review! I checked with Ryan and he will continue dev on the main for some time

ryanemerson:
@oraNod We're holding off on branching for now as everything going into `main` is planned for `2.3.x`. This save us the pain of backporting everything.

ryanemerson:
Thanks @dvagnero

PR #1711, opened 2022-11-02
Update ServiceAccount token lookup logic

Crumby:
Token secret is no longer listed between ServiceAccount secrets in OCP4.11 making the multi cluster cross site tests fail while looking for the SA token. This PR updates the logic to find the secret by listing the secrets and filtering on Operator SA name + token and taking the first result.

### DIFF
--- a/pkg/kubernetes/controllerutil.go
+++ b/pkg/kubernetes/controllerutil.go
@@ -183,43 +183,6 @@ func LookupResource(name, namespace string, resource, caller client.Object, clie
 	return nil, nil
 }
 
-func LookupServiceAccountTokenSecret(name, namespace string, client client.Client, ctx context.Context) (*corev1.Secret, error) {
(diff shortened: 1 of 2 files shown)


ryanemerson:
The linter is failing because the `isServiceAccount` function is no longer used.

`LookupServiceAccountTokenSecret` is only used by `xsite_test.go` now. Can you move their? I feel like this method is something we shouldn't be relying on in the operator itself, so moving it to the test package will make that more explicit.

ryanemerson:
I have no idea why the CI isn't working with the linter for this PR. I tried cleaning up the workspace etc on the CI agents, but still the same problem. @Crumby As a workaround could you try to create a new PR with the changes and see if we have any luck?

ryanemerson:
Thanks @Crumby. I'm closing this PR as I cherry-picked and modified your commit as part of https://github.com/infinispan/infinispan-operator/pull/1892.

Issue #1715, opened 2022-11-16
Docs: Add operand version attribute


dvagnero:
We need to introduce a new attribute. The `spec.version` field in product requires server version and the tag.

ryanemerson:
@dvagnero @dfitzmau In the future don't create an issue if there's no description, this doesn't add any value to the community and is a waste of your time. 

Of course, if you can provide a meaningful description for your work beyond the commit message please create an issue as it's useful to have a history of what motivated the changes.

dvagnero:
Solved

PR #1716, opened 2022-11-16
add attribute for operand version #1715

dvagnero:
Add new attribute for the operand version

### DIFF
--- a/documentation/asciidoc/topics/attributes/community-attributes.adoc
+++ b/documentation/asciidoc/topics/attributes/community-attributes.adoc
@@ -23,6 +23,7 @@
 :backup_path: /opt/infinispan/backups
 :server_image: quay.io/infinispan/server
 :server_image_version: 14.0.1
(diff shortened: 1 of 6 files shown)


documentation/asciidoc/topics/attributes/community-attributes.adoc
```diff
@@ -23,6 +23,7 @@
 :backup_path: /opt/infinispan/backups
 :server_image: quay.io/infinispan/server
 :server_image_version: 14.0.1
+:operand_version: 14.0.1
```
ryanemerson, line 26: Why do we need this as well as `server_image_version`?

dvagnero: Hi @ryanemerson , the versions are the same in community but not in the product. There we need 8.4.0-1. I understand it's another version field to maintain but that's the solution I came up with for now.

PR #1717, opened 2022-11-29
upd go get to go install

rigazilla:
"go get" is removed in go 1.18. replaced with "go install"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,12 +210,8 @@ PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
 define go-get-tool
 @[ -f $(1) ] || { \
 set -e ;\
-TMP_DIR=$$(mktemp -d) ;\
-cd $$TMP_DIR ;\
-go mod init tmp ;\
 echo "Downloading $(2)" ;\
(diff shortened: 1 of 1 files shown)


controllers/operatorconfig_controller.go
```diff
@@ -32,7 +33,7 @@ type ReconcileOperatorConfig struct {
 func (r *ReconcileOperatorConfig) SetupWithManager(mgr ctrl.Manager) error {
 	name := "config"
 	r.Client = mgr.GetClient()
-	r.log = ctrl.Log.WithName("controllers").WithName(strings.Title(name))
+	r.log = ctrl.Log.WithName("controllers").WithName(cases.Title(language.Und).String(name))
```
ryanemerson, line 36: Is this necessary given that we baseline on 1.17?

rigazilla: CI seems running 1.18. [fail](https://ci.infinispan.org/blue/organizations/jenkins/Infinispan%20Operator/detail/PR-1717/2/pipeline/22)
The change seems to be safe, I would consider to keep it since 1.17 is eof [lifetime](https://endoflife.date/go)

ryanemerson:
Thanks @rigazilla

PR #1731, opened 2023-01-12
Allow users to define credential-store entries. Resolves #1730

### LINKED ISSUE #1730: Allow users to define credential-store entries
In order to allow users to securely define custom server configurations requiring credentials, e.g. LDAP configurations, it's necessary for us to provide a way for users to define entries in the default credential-store.

We can optionally allow the user to specify a Secret containing the required credential-store references:

```yaml
spec:
  security:
    credentialStoreSecretName: user-secret
```

ryanemerson:


rigazilla (approved):
lgtm!

PR #1741, opened 2023-01-18
Use OpenSSL to create the truststore

pruivo:


pruivo:
Java is unable to read the truststore created from openssl.

PR #1757, opened 2023-01-31
CredentialStore only created when TLS configured. Fixes #1756

### LINKED ISSUE #1756: CredentialStore only created when TLS configured
The server Credential Store is only added to the server configuration when TLS with an existing key or truststore is provided. Since #1730 was added, it's necessary for the store to always be defined.

ryanemerson:


test/e2e/infinispan/credential_store_test.go
```diff
@@ -13,6 +13,36 @@ import (
...
+                        urn:org:jgroups http://www.jgroups.org/schema/jgroups-5.2.xsd"
+    xmlns="urn:infinispan:config:14.0"
+    xmlns:server="urn:infinispan:server:14.0">
+
+    <server xmlns="urn:infinispan:server:14.0">
+        <data-sources>
+          <data-source name="postgres" jndi-name="jdbc/postgres">
+            <connection-factory driver="org.postgresql.Driver"
+                                username="dbuser"
+                                url="${org.infinispan.server.test.postgres.jdbcUrl}">
```
Crumby, line 30: Isn't this going to cause issue given the test does not deploy database?

pkg/templates/templates/infinispan-base-14.xml
```diff
@@ -37,13 +37,11 @@
         <socket-binding name="default" port="${infinispan.bind.port:11222}"/>
     </socket-bindings>
     <security>
-        {{ if or .Keystore.Password .Truststore.Path }}
```
Crumby, line 40: So it's always enabled even if user does not define any credential store?

test/e2e/infinispan/credential_store_test.go
ryanemerson: Datasources configurations are validated on startup, but instances of the datasource are accessed lazily by caches. So we'll be fine as long as we don't create a Cache referencing the datasource in our test.

AFAICT all of the places in the server configuration that allow credential-store references are related to configuring connections to external resources, so this was the simplest way I could think to test without having to deploy anything else.

pkg/templates/templates/infinispan-base-14.xml
ryanemerson: I've updated this so that we now create two credential-stores as discussed on Zulip. The user credential-store will only be created if the user has provided one or more credentials.

ryanemerson:
`TestCacheWithServerListener` unrelated to changes https://github.com/infinispan/infinispan-operator/pull/1759

PR #1778, opened 2023-03-06
update labels on operator restart. Fixes #JDG-5931 JDG-5936

rigazilla:
see #1780 
currently labels are updated only if PreliminaryCheckCondition==false.
This prevents labels update on operator restart.
Also selector template in statefulset has been changed to allow label update.

Label are merged in the Infinispan CR (using ApplyOperatorMeta) while are replaced in Statefulset and Pod.
replace= all the old keys are lost
merge= updated affects only label present in the new label map

I would prefer to use merge approach also for Statefulset and Pod, since I don't like the idea to delete user labels.
wdyt @ryanemerson ?

### DIFF
--- a/pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
@@ -14,11 +14,28 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+func subMapOf(f, s map[string]string) bool {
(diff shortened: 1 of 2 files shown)


ryanemerson:
The changes LGTM, but the upgrade test is failing. I've requested another CI build, but can you verify locally as well.

Can you create a GH issue explaining the issue of labels/annotations not being updated on  upgrade. The JDG Jiras are not available to the community, so we should avoid using them in the git history.

ryanemerson:
Thanks @rigazilla

PR #1779, opened 2023-03-06
fixed go.mod inconsistencies

rigazilla:
I got some error with the build starting with an empty go root:

- tags on github.com/openshift/api has been deleted (v3.9.0 doesn't exists)
- -github.com/mikefarah/yaml/v2 must be referred as gopkg.in/yaml.v2
-


ryanemerson:
Thanks @rigazilla

Issue #1782, opened 2023-03-08
GetPodMemoryLimitBytes fails with cgroups v2


rigazilla:
the method relay on file `/sys/fs/cgroup/memory/memory.limit_in_bytes` with cgroups v2 (fedora 37) should be `/sys/fs/cgroup/memory.max`

fix should be something like https://github.com/oracle/docker-images/pull/1963

PR #1783, opened 2023-03-08
fix memory limit for cgroups2 #1782

rigazilla:
proposal solution for  #1782

### DIFF
--- a/pkg/kubernetes/pods.go
+++ b/pkg/kubernetes/pods.go
@@ -156,7 +156,7 @@ func GetContainer(name string, spec *corev1.PodSpec) *corev1.Container {
 func GetPodMemoryLimitBytes(container, podName, namespace string, kube *Kubernetes) (uint64, error) {
 	execOut, err := kube.ExecWithOptions(ExecOptions{
 		Container: container,
-		Command:   []string{"cat", "/sys/fs/cgroup/memory/memory.limit_in_bytes"},
+		Command:   []string{"/bin/bash", "-c", "cat /sys/fs/cgroup/memory/memory.limit_in_bytes || cat /sys/fs/cgroup/memory.max"},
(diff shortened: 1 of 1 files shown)


pkg/kubernetes/pods.go
```diff
@@ -156,7 +156,7 @@ func GetContainer(name string, spec *corev1.PodSpec) *corev1.Container {
 func GetPodMemoryLimitBytes(container, podName, namespace string, kube *Kubernetes) (uint64, error) {
 	execOut, err := kube.ExecWithOptions(ExecOptions{
 		Container: container,
-		Command:   []string{"cat", "/sys/fs/cgroup/memory/memory.limit_in_bytes"},
+		Command:   []string{"/bin/bash", "-c", "cat /sys/fs/cgroup/memory/memory.limit_in_bytes || cat /sys/fs/cgroup/memory.max"},
```
ryanemerson, line 159: Won't this result in the output containing something like:

```
cat: /sys/fs/cgroup/memory/memory.limit_in_bytes :No such file or directory
```

Followed by the value in the second file?

rigazilla: mmm probably yes... It worked locally but let me check

rigazilla: No, it works, only the stdout is returned, stderr if error happens is wrapper in an error instance:
https://github.com/infinispan/infinispan-operator/blob/bdfc2db793ee21e33f89a0d99be491d7c0a95bf6/pkg/kubernetes/kubernetes.go#L150-L159

ryanemerson: O nice, thanks for double checking :+1:

ryanemerson:
Thanks @rigazilla

Issue #1781, opened 2023-03-08
TestCacheWithServerLifecycle occasionally fails


rigazilla:
Seems that cache creation produces several update-event for the listener.
Something this result in the following sequence that generates a test failure:
```
2023-03-08T09:07:14.569Z	DEBUG	ConfigListener received event 'create-cache':
---
{
  "infinispan" : {
    "cache-container" : {
      "caches" : {
        "test/cache/with/server/lifecycle" : {
          "local-cache" : {
            "memory" : {
              "max-count" : "100"
            }
          }
        }
      }
    }
  }
}
---

2023-03-08T09:07:14.571Z	DEBUG	No existing Cache CR found for Cache=test/cache/with/server/lifecycle, Cluster=test-cache-with-server-lifecycle
2023-03-08T09:07:14.632Z	INFO	Creating Cache CR for 'test/cache/with/server/lifecycle'
localCache: 
  memory: 
    maxCount: "100"

2023-03-08T09:07:14.636Z	INFO	Cache CR 'test-cache-with-server-lifecycle-2h2s5' created
2023-03-08T09:07:14.939Z	DEBUG	ConfigListener received event 'update-cache':
---
{
  "infinispan" : {
    "cache-container" : {
      "caches" : {
        "test/cache/with/server/lifecycle" : {
          "local-cache" : {
            "memory" : {
              "max-count" : "100"
            }
          }
        }
      }
    }
  }
}
---

2023-03-08T09:07:14.941Z	DEBUG	An existing Cache CR 'test-cache-with-server-lifecycle-2h2s5' was found for Cache=test/cache/with/server/lifecycle, Cluster=test-cache-with-server-lifecycle
2023-03-08T09:07:15.060Z	DEBUG	Cache 'test-cache-with-server-lifecycle-2h2s5' configuration on update has not changed, ignoring update
2023-03-08T09:07:15.287Z	DEBUG	ConfigListener received event 'update-cache':
---
{
  "infinispan" : {
    "cache-container" : {
      "caches" : {
        "test/cache/with/server/lifecycle" : {
          "local-cache" : {
            "memory" : {
              "max-count" : "100"
            }
          }
        }
      }
    }
  }
}
---

2023-03-08T09:07:15.290Z	DEBUG	An existing Cache CR 'test-cache-with-server-lifecycle-2h2s5' was found for Cache=test/cache/with/server/lifecycle, Cluster=test-cache-with-server-lifecycle
[...]

PR #1784, opened 2023-03-08
stabilize TestCacheWithServerLifecycle #1781

rigazilla:
workaround fix for #1781

### DIFF
--- a/test/e2e/cache/cache_test.go
+++ b/test/e2e/cache/cache_test.go
@@ -272,6 +272,8 @@ func TestCacheWithServerLifecycle(t *testing.T) {
 	testifyAssert.True(t, kubernetes.IsOwnedBy(cr, ispn), "Cache has unexpected owner reference")
 
 	// Update cache configuration via REST
+	// Await all creation related updates are gone
+	time.Sleep(10 * time.Second)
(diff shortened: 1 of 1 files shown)


test/e2e/cache/cache_test.go
```diff
@@ -272,6 +272,8 @@ func TestCacheWithServerLifecycle(t *testing.T) {
 	testifyAssert.True(t, kubernetes.IsOwnedBy(cr, ispn), "Cache has unexpected owner reference")
 
 	// Update cache configuration via REST
+	// Await all creation related updates are gone
+	time.Sleep(10 * time.Second)
```
ryanemerson, line 276: Instead of always waiting 10 seconds we could retry `cacheHelper.CacheClient.UpdateConfig(..)` until it succeeds or times out after 10 seconds.

rigazilla: > Instead of always waiting 10 seconds we could retry `cacheHelper.CacheClient.UpdateConfig(..)` until it succeeds or times out after 10 seconds.

you mean an ad hoc WaitForTemplateUpdate with the above policy?

ryanemerson: Yeah exactly

ryanemerson:
Thanks @rigazilla

PR #1791, opened 2023-03-14
Test HotRod Rolling upgrades against older Operator releases

ryanemerson:


test/e2e/hotrod-rolling-upgrade/hotrod_rolling_upgrade_test.go
```diff
@@ -124,7 +124,7 @@ func TestRollingUpgrade(t *testing.T) {
 			if !isRollingUpgrade {
 				clusterCounter++
 				currentStatefulSetName := newStatefulSetName
-				newStatefulSetName := fmt.Sprintf("%s-%d", spec.Name, clusterCounter)
+				newStatefulSetName = fmt.Sprintf("%s-%d", spec.Name, clusterCounter)
```
ryanemerson, line 127: @Crumby As this originally had `:=` it was only declaring `newStatefulSetName` in the scope of the if statement causing the subsequent `testKube.WaitForStateFulSet(newStatefulSetName, tutils.Namespace)` to timeout as it was waiting for pods from the old statefulSet still.

PR #1792, opened 2023-03-15
add_router_annotations

vladimir22:


ryanemerson:
Thanks for the PR @vladimir22. Do you require annotations specific to just the GossipRouter pods or just those specified via `infinispan.org/podTargetAnnotations` to be propagated to the GossipRouter?

If it's the latter, then there's no need to introduce `infinispan.org/routerAnnotations`. Instead we can just apply the existing `podTargetAnnotations` to the Deployment spec as it's exclsuion is an oversight/bug.

vladimir22:
Ok, got it, will do, but 
actually, it is very good questions :)

Currently, I can do that only with `infinispan.org/podTargetAnnotations`,

but what if I need to specify the same annotations but with different values, 
example:

- GossipRouter Deployment : 
```
traffic.sidecar.istio.io/excludeInboundPorts: 7900
traffic.sidecar.istio.io/includeInboundPorts 7900
```

- Infinispan Statefulset: 
```
traffic.sidecar.istio.io/excludeInboundPorts: 11223
traffic.sidecar.istio.io/includeInboundPorts 11223
```

It would be nice to have an option for this case too.

ryanemerson:
That's a good point!

I think the code is good to go. Can you do me a favour though and add an example of the annotation to this file so that other users are aware of this capability:

https://github.com/infinispan/infinispan-operator/blob/d291909d9e8ed6fdd019116fcdb4db2fecd58dd5/documentation/asciidoc/topics/yaml/custom_annotations.yaml

That will mean that we have an example [here in the docs](https://infinispan.org/docs/infinispan-operator/main/operator.html#adding-custom-labels_services).

vladimir22:
Done, I have added the next value in the [custom_annotations.yaml](https://github.com/infinispan/infinispan-operator/pull/1792/commits/a0e7bdceb5b02893feab1fae826f66da71b9c268):
```
  annotations:
    infinispan.org/targetAnnotations: service-annotation1, service-annotation2
    infinispan.org/podTargetAnnotations: pod-annotation1, pod-annotation2
    infinispan.org/routerAnnotations: router-annotation1, router-annotation2

    service-annotation1: value
    service-annotation2: value
    pod-annotation1: value
    pod-annotation2: value
    router-annotation1: value
    router-annotation2: value
```

but again, let me explain that we still having the issue above.

I *can't* use the same annotation with different values for different resources, example

```yaml
  annotations:
    infinispan.org/podTargetAnnotations: traffic.sidecar.istio.io/excludeInboundPorts, traffic.sidecar.istio.io/includeInboundPorts
    infinispan.org/routerAnnotations: traffic.sidecar.istio.io/excludeInboundPorts, traffic.sidecar.istio.io/includeInboundPorts

    ## Exclude port for Infinispan PODs
    traffic.sidecar.istio.io/excludeInboundPorts: 11223 
    traffic.sidecar.istio.io/includeInboundPorts: 11223

    ## Exclude port for GossipRouter POD
    traffic.sidecar.istio.io/excludeInboundPorts: 7900  ## This value will override value above, GossipRouter POD will have 11223 port
    traffic.sidecar.istio.io/includeInboundPorts: 7900

```

It would be nice to fix that in the next versions by adding extra logic into `addAnnotationsFor` function.
For example like that:

```yaml
  annotations:
    infinispan.org/podTargetAnnotations: traffic.sidecar.istio.io/excludeInboundPorts, traffic.sidecar.istio.io/includeInboundPorts

    ## Exclude port by default
    traffic.sidecar.istio.io/excludeInboundPorts: 11223 
    traffic.sidecar.istio.io/includeInboundPorts: 11223

    ## Apply annotations for GossipRouter with 'overridden' values
    infinispan.org/routerAnnotations: traffic.sidecar.istio.io/excludeInboundPorts=7900, traffic.sidecar.istio.io/includeInboundPorts=7900

```

ryanemerson:
> I _can't_ use the same annotation with different values for different resources, example

Right, I understand now. 


> It would be nice to fix that in the next versions by adding extra logic into [addAnnotationsFor](https://github.com/infinispan/infinispan-operator/blob/6a84cdaea1e3c3a876f88da24d001fb0dca94696/api/v1/types_util.go#L684) function. For example like that:
> 
> ```yaml
>   annotations:
>     infinispan.org/podTargetAnnotations: traffic.sidecar.istio.io/excludeInboundPorts, traffic.sidecar.istio.io/includeInboundPorts
> 
>     ## Exclude port by default
>     traffic.sidecar.istio.io/excludeInboundPorts: 11223 
>     traffic.sidecar.istio.io/includeInboundPorts: 11223
> 
>     ## Apply annotations for GossipRouter with 'overridden' values
>     infinispan.org/routerAnnotations: traffic.sidecar.istio.io/excludeInboundPorts=7900, traffic.sidecar.istio.
> ```

That seems reasonable to me. Obviously we can't take this approach with labels due to the k8s restrictions on values, but I think this is fine if we consistently apply this approach to the various annotations.

Are you able to contribute this as part of the PR? If not, I might get a chance to add this in the near future.

vladimir22:
I would propose merging the current PR, and after creating a new one to discuss implementation details.

PR #1800, opened 2023-03-27
Update label only if operand is upgraded. Fixes #1798

### LINKED ISSUE #1798: Pods are disengaged from SS after upgrade
Pods labels for metering aren't updated on statefulset when operator upgrade doesn't requires pods restart.
This results in a disconnection between statefulset and pods.

Proposal solution
When the operator is upgraded with new metering labels:
- propagate new labels on new infinispan CRs
- propagate new labes on existing Infinispan CR and related statefulset/pods ONLY if they are restarted and upgraded with new operand version.

rigazilla:
Metering labels are updated only if operand is upgraded
fixes #1798

### DIFF
--- a/pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
@@ -14,26 +14,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func subMapOf(f, s map[string]string) bool {
-	for k := range f {
-		if v, ok := s[k]; !ok || f[k] != v {
-			return false
-		}
-	}
-	return true
-}
-
 func PreliminaryChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
-	// Ensure that CR has all the operator required labels
-	if !subMapOf(ctx.DefaultLabels(), i.ObjectMeta.Labels) || !subMapOf(ctx.DefaultAnnotations(), i.ObjectMeta.Annotations) {
-		if err := ctx.UpdateInfinispan(func() {
-			ctx.Log().Info("Updating Labels", "Labels: ", ctx.DefaultLabels())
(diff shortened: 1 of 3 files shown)


pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
```diff
@@ -50,6 +40,11 @@ func PreliminaryChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
 				if ctx.IsTypeSupported(pipeline.ServiceMonitorGVK) {
 					i.ApplyMonitoringAnnotation()
 				}
+				// Ensure that CR has all the operator required labels
+				if !subMapOf(ctx.DefaultLabels(), i.ObjectMeta.Labels) || !subMapOf(ctx.DefaultAnnotations(), i.ObjectMeta.Annotations) {
```
ryanemerson, line 44: Do we need the `subMapOf` checks here given that we're only applying to new Infinispan CRs?

Can't we just revert this part of the code to how it was before your previous labels commit? i.e. just

```go
i.ApplyOperatorMeta(ctx.DefaultLabels(), ctx.DefaultAnnotations())
```

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -204,10 +204,15 @@ func GracefulShutdownUpgrade(i *ispnv1.Infinispan, ctx pipeline.Context) {
 
 		ctx.Requeue(
 			ctx.UpdateInfinispan(func() {
+				// Ensure that CR has all the operator required labels
+				if !subMapOf(ctx.DefaultLabels(), i.ObjectMeta.Labels) || !subMapOf(ctx.DefaultAnnotations(), i.ObjectMeta.Annotations) {
```
ryanemerson, line 208: Do we need the `subMapOf` checks, can't we just always call `i.ApplyOperatorMeta(ctx.DefaultLabels(), ctx.DefaultAnnotations())`?

ryanemerson:
Thanks @rigazilla

PR #1802, opened 2023-03-29
Gossip Router improvements

pruivo:
* Increase the probe timeout
* Allow setting the CPU and Memory resources

---

This PR improves the cross-site in the following way

* startup probe increased to 15 sec
* liveness and readiness probes increased to 50 sec 

Since there are multiple Gossip Router pods by default, increasing the probe timeouts shouldn't have any impact.

The second "big" change is the possibility to set the CPU and Memory limit/request for the Gossip Router. And finally, a small improvement in the code, skipping searching for the trust store secret if none is configured.


pkg/infinispan/client/client.go
```diff
@@ -16,6 +16,7 @@ To prevent duplicated code, newer api implementation packages should use composi
 function as expected on newer server versions.
 
 For example, if v14 can reuse a new `api.Cache` implementation but requires a new `api.Caches` implementation:
+
```
ryanemerson, line 19: Why is this and [pkg/apis/integreatly/v1alpha1/grafanadashboard_types.go](https://github.com/infinispan/infinispan-operator/pull/1802/files#diff-73f5e8039ea36620af1fb7969676000b6c82e3bfe2c3dc70c0220e2870893d8e) change required?

pkg/reconcile/pipeline/infinispan/handler/provision/gossiprouter.go
```diff
@@ -129,33 +129,43 @@ func GossipRouter(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+			Args:           args,
+			Env:            []corev1.EnvVar{{Name: "ROUTER_JAVA_OPTIONS", Value: i.Spec.Container.RouterExtraJvmOpts}},
+			Ports:          containerPorts,
+			LivenessProbe:  TcpProbe(probePort, 10, 5, 0, 1, 60),
+			ReadinessProbe: TcpProbe(probePort, 10, 5, 0, 1, 60),
+			StartupProbe:   TcpProbe(probePort, 15, 1, 1, 1, 60),
+		}
+
+		if podResources, err := gossipRouterPodResources(i.Spec.Service.Sites.Local.Discovery); err != nil {
+			ctx.Requeue(fmt.Errorf("unable to calculate Gossip Router pod resources: %w", err))
```
ryanemerson, line 146: As this is in the mutateFn, should we just be returning the error here so that the `CreateOrUpdate` call fails with "Failed to configure Cross-Site Deployment..."

pkg/reconcile/pipeline/infinispan/handler/provision/gossiprouter.go
```diff
@@ -129,33 +129,43 @@ func GossipRouter(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+
+		container := &corev1.Container{
+			Name:           GossipRouterContainer,
+			Image:          i.ImageName(),
+			Command:        []string{"/opt/gossiprouter/bin/launch.sh"},
+			Args:           args,
+			Env:            []corev1.EnvVar{{Name: "ROUTER_JAVA_OPTIONS", Value: i.Spec.Container.RouterExtraJvmOpts}},
+			Ports:          containerPorts,
+			LivenessProbe:  TcpProbe(probePort, 10, 5, 0, 1, 60),
+			ReadinessProbe: TcpProbe(probePort, 10, 5, 0, 1, 60),
```
ryanemerson, line 141: The default value for `Probe.PeriodSeconds` is 10 seconds if the configured value is < 1. This seems high for the readiness probe, should it be smaller?

pkg/reconcile/pipeline/infinispan/handler/provision/gossiprouter.go
```diff
@@ -203,3 +213,32 @@ func GossipRouter(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+	if spec.CPU == "" && spec.Memory == "" {
+		return nil, nil
+	}
+
+	req := &corev1.ResourceRequirements{
+		Limits:   corev1.ResourceList{},
+		Requests: corev1.ResourceList{},
+	}
+	if spec.Memory != "" {
+		memRequests, memLimits, err := spec.MemoryResources()
```
ryanemerson, line 227: If you validate the Memory and CPU values in the webhook, it's safe to ignore the returned `error` here as it will never be thrown. That being said, if you want to leave this, it can't hurt :slightly_smiling_face: .

ryanemerson (changes requested):
Overall looks good to me, however you should validate the user enters valid Memory and CPU values in the Infinispan webhook. You can use the container configuration as an example: https://github.com/infinispan/infinispan-operator/blob/d5663f0e43a5ce6ea209b279a21c9795d8f30eb7/api/v1/infinispan_webhook.go#L219-L229.

You should also add a test to `infinispan_webhook_test.go`.

pkg/infinispan/client/client.go
pruivo: probably some wrong setting in my Go installation. When I run `make manager`, the format is changed in a couple of files and I probably miss tho revert those 2.

api/v1/infinispan_types.go
```diff
@@ -175,6 +175,12 @@ type DiscoverySiteSpec struct {
 	// Enables (default) or disables the Gossip Router pod and cross-site services
 	// +optional
 	LaunchGossipRouter *bool `json:"launchGossipRouter,omitempty"`
+	// Memory resource request for Gossip Router (if LaunchGossipRouter == true)
```
ryanemerson, line 178: I would replace the "( if Launch.." with "...request for Gossip Router if enabled". Same applies for CPU.

dvagnero:
@pruivo docs review completed. I moved all the content related to the GR in one section. Let me know what else can be improved

pruivo:
@dvagnero looks good to me. Thanks!

ryanemerson:
Thanks @pruivo

PR #1815, opened 2023-05-10
marks service monitor labels as community only

dvagnero:
This content should be available only in the community


ryanemerson:
Thanks @dvagnero

PR #1824, opened 2023-05-18
Remove leftover IT resources upon finished execution

Crumby:
Update integration test so they remove all the leftover resources after the execution. Leaving some of them can end up in other tests instability due to lack of k8s cluster resources.


ryanemerson:
Thanks @Crumby

PR #1828, opened 2023-05-25
Clean up events after test execution

Crumby:
Events in the testing namespace are not cleaned up upon test execution. This may results in unrelated events being reported in case of other test failure.


ryanemerson:
Thanks @Crumby

PR #1837, opened 2023-07-21
Allow JMX endpoint to be exposed

ryanemerson:


api/v1/infinispan_webhook.go
```diff
@@ -184,6 +188,10 @@ func (i *Infinispan) ValidateUpdate(oldRuntimeObj runtime.Object) error {
 			}
 		}
 	}
+
+	if old.Spec.Jmx != nil && old.Spec.Jmx.Enabled != i.Spec.Jmx.Enabled {
```
Crumby, line 192: Should we even consider nil, here? `Nil` effectively means `jmx.enabled: false` so if it's not defined we should initiate the value and work with it as always defined

ryanemerson: If we don't check `nil` here then it will panic with a nil pointer dereference.

PR #1845, opened 2023-08-10
Env vars#1520

### LINKED ISSUE #1520: Env vars in operator subscription
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: my-infinispan
  namespace: namespace-for-testing
spec:
  channel: 2.2.x
  name: infinispan
  source: operatorhubio-catalog
  sourceNamespace: olm
  config:
    env:
      - name: ADDITIONAL_VARS
        value: "[\"DEBUG\"]"
      - name: DEBUG
        value: true
```

each pod in the cluster's StatefulSet will have the DEBUG env variable set to true

This will apply to ALL Infinispan CRs created by this Operator.

dvagnero:
Resolves #1520 
Add procedure how to define env var using the ocp console and command line

### DIFF
--- a/documentation/asciidoc/topics/attributes/community-attributes.adoc
+++ b/documentation/asciidoc/topics/attributes/community-attributes.adoc
@@ -17,8 +17,10 @@
 
 :example_crd_name: infinispan
 :example_namespace: ispn-namespace
+:subscription_name: infinispan
 :subscription_namespace: operators
-:operator_source: community operators
+:source_namespace: olm
+:operator_source: operatorhubio-catalog
 :pv_path: /opt/infinispan/server/data
 :conf_path: /opt/infinispan/server/conf
 :lib_path: /opt/infinispan/server/lib
@@ -33,7 +35,6 @@
 :schemaversion: 14.0
 :brandname: Infinispan
 :fullbrandname: Infinispan
-:subscription_name: infinispan
 :datagridservice: Data Grid Service
 :cacheservice: Cache Service
 
--- a/documentation/asciidoc/topics/attributes/community-attributes.adoc
+++ b/documentation/asciidoc/topics/attributes/community-attributes.adoc
@@ -17,6 +17,8 @@
 
 :example_crd_name: infinispan
 :example_namespace: ispn-namespace
+:subscription_namespace: operators
+:operator_source: community operators
 :pv_path: /opt/infinispan/server/data
 :conf_path: /opt/infinispan/server/conf
 :lib_path: /opt/infinispan/server/lib
@@ -27,7 +29,7 @@
 
 :ispn_operator: Infinispan Operator
 :ispn_operator_version: 2.3
-:operator_channel: 2.3
+:operator_channel: 2.3.x
 :schemaversion: 14.0
 :brandname: Infinispan
 :fullbrandname: Infinispan
--- a/documentation/asciidoc/topics/attributes/downstream-attributes.adoc
+++ b/documentation/asciidoc/topics/attributes/downstream-attributes.adoc
@@ -14,6 +14,7 @@
 :example_crd_name: infinispan
 :example_namespace: rhdg-namespace
 :subscription_namespace: openshift-operators
+:source_namespace: openshift-marketplace
 :subscription_name: datagrid
 :operator_source: redhat-operators
 :pv_path: /opt/infinispan/server/data
--- a/documentation/asciidoc/topics/attributes/downstream-attributes.adoc
+++ b/documentation/asciidoc/topics/attributes/downstream-attributes.adoc
@@ -13,6 +13,9 @@
 
 :example_crd_name: infinispan
 :example_namespace: rhdg-namespace
+:subscription_namespace: openshift-operators
+:subscription_name: datagrid
+:operator_source: redhat-operators
 :pv_path: /opt/infinispan/server/data
 :conf_path: /opt/infinispan/server/conf
 :backup_path: /opt/infinispan/backups
--- a/documentation/asciidoc/topics/proc_adding_labels_annotations_env.adoc
+++ b/documentation/asciidoc/topics/proc_adding_labels_annotations_env.adoc
@@ -12,7 +12,7 @@ Add labels and annotations to your {ispn_operator} subscription with the `spec.c
 +
 [source,options="nowrap",subs=attributes+]
 ----
-{oc_edit} subscription {subscription_name} -n {example_namespace}
+{oc_edit} subscription {subscription_name} -n {subscription_namespace}
 ----
 
 * Use the {openshift} Console.
--- a/documentation/asciidoc/topics/proc_adding_labels_annotations_env.adoc
+++ b/documentation/asciidoc/topics/proc_adding_labels_annotations_env.adoc
@@ -12,12 +12,12 @@ Add labels and annotations to your {ispn_operator} subscription with the `spec.c
 +
 [source,options="nowrap",subs=attributes+]
 ----
-{oc_edit} subscription {subscription_name} -n openshift-operators
+{oc_edit} subscription {subscription_name} -n {example_namespace}
 ----
 
 * Use the {openshift} Console.
 +
-. Navigate to **Operators** > **Installed Operators** > **Subscription**.
+. Navigate to **Operators** > **Installed Operators** > click the *{ispn_operator}*.
 . Select **Edit Subscription** from the **Actions** menu.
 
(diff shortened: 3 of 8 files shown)


documentation/asciidoc/topics/proc_adding_labels_annotations_env.adoc
```diff
@@ -12,12 +12,12 @@ Add labels and annotations to your {ispn_operator} subscription with the `spec.c
 +
 [source,options="nowrap",subs=attributes+]
 ----
-{oc_edit} subscription {subscription_name} -n openshift-operators
+{oc_edit} subscription {subscription_name} -n {example_namespace}
```
dvagnero, line 15: Atm {example_namespace} resolves in the community to ispn-namespace and downstream to rhdg-namespace but on the openshift console I see  openshift-operators.
Does it make sense to use openshift-operators as a example namespace in the downstream docs?

documentation/asciidoc/topics/proc_defining_environment_var.adoc
```diff
@@ -0,0 +1,51 @@
+[id='defining-environment-variables_{context}']
+= Defining environment variables in the {ispn_operator} subscription
+
+[role="_abstract"]
+Define environment variables in your {ispn_operator} subscription that automatically propagate to each {brandname} po and service managed by the corresponding Operator.
+
+`spec.config.env` field is an array of environment variable definitions.
+You can add one or more environment variable names and corresponding values.
```
dvagnero, line 8: Do you have any recommendations here? What the environment variables should not be used for? Best practices etc? 
For example, Kubernetes docs mention " Environment variables may reference each other, however ordering is important. Variables making use of others defined in the same context must come later in the list. Similarly, avoid circular references." & "The environment variables set using the env or envFrom field override any environment variables specified in the container image."

documentation/asciidoc/topics/proc_defining_environment_var.adoc
```diff
@@ -0,0 +1,51 @@
...
+
+* Retrieve the environment variables from the subscription YAML:
+[source,options="nowrap",subs=attributes+]
+----
+{oc} get subscription {subscription_name} -n {example_namespace} -o jsonpath='{.spec.config.env[*].name}'
+----
+
+[discrete]
+== Reference
+//add specific example
```
dvagnero, line 51: Do you have an idea for a relevant example of env vars? Otherwise, I could use the one with timezones

documentation/asciidoc/topics/proc_adding_labels_annotations_env.adoc
ryanemerson: Downstream we should be using `openshift-operators` and upstream just `operators` when describing interactions with the Operator Subscription.

documentation/asciidoc/topics/proc_defining_environment_var.adoc
```diff
@@ -0,0 +1,51 @@
+[id='defining-environment-variables_{context}']
+= Defining environment variables in the {ispn_operator} subscription
+
+[role="_abstract"]
+Define environment variables in your {ispn_operator} subscription that automatically propagate to each {brandname} po and service managed by the corresponding Operator.
```
ryanemerson, line 5: `s/po/pod`

The environment variables will not be set on a Service as there's no environment associated with this.

documentation/asciidoc/topics/proc_defining_environment_var.adoc
ryanemerson: > Do you have any recommendations here?

I think the `TZ` environment variable in the Jira is a good one as we have users who actively need to change the timezone associated with their deployments.


> What the environment variables should not be used for? Best practices etc?

I think it's sufficient to just state that the order of the variables is important if one env var references another.

documentation/asciidoc/topics/proc_defining_environment_var.adoc
ryanemerson: Timezones is good.

documentation/asciidoc/topics/proc_defining_environment_var.adoc
```diff
@@ -0,0 +1,51 @@
...
+You can add one or more environment variable names and corresponding values.
+
+.Prerequisites
+* Ensure the Operator Lifecycle Manager (OLM) is installed.
+
+.Procedure
+
+Define environment variables in your {ispn_operator} subscription using the `spec.config.env` field:
+
+. Use the `{oc_edit} subscription` command.
```
ryanemerson, line 17: I think it would be better if we show creating a `Subscription` with the variables instead of editing an existing one. So `oc apply` instead of  `oc edit`. The example yaml will need to be updated to include the rest of the Subscription fields.

ryanemerson: I see that we have edit for the console as well. Maybe it's best to have an apply workflow for the CLI and edit for the console? Or have edit for both and just apply for CLI, WDYT?

dvagnero: Thanks for the comments! I agree that we should add apply workflow. My proposal is to have a follow up step showing how to edit the existing subscription. I'll try and see how it looks

dvagnero:
@ryanemerson ready for your review. If the content looks good I can squash the commits

documentation/asciidoc/topics/attributes/community-attributes.adoc
```diff
@@ -17,6 +17,8 @@
 
 :example_crd_name: infinispan
 :example_namespace: ispn-namespace
+:subscription_namespace: operators
+:operator_source: community operators
```
ryanemerson, line 21: This needs to be `operatorhubio-catalog`

documentation/asciidoc/topics/yaml/env_var_subscription.yaml
```diff
@@ -0,0 +1,18 @@
...
+metadata:
+  name: {subscription_name}
+  namespace: {subscription_namespace}
+spec:
+  channel: {operator_channel}
+    operator_channel: null
+  installPlanApproval: Automatic
+  name: {subscription_name}
+  source: {operator_source}
+  sourceNamespace: openshift-marketplace
```
ryanemerson, line 12: This needs to be `olm` for upstream.

documentation/asciidoc/topics/yaml/env_var_subscription.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: {subscription_name}
+  namespace: {subscription_namespace}
+spec:
+  channel: {operator_channel}
+    operator_channel: null
```
ryanemerson, line 8: This line can be removed.

documentation/asciidoc/topics/attributes/downstream-attributes.adoc
```diff
@@ -13,6 +13,9 @@
 
 :example_crd_name: infinispan
 :example_namespace: rhdg-namespace
+:subscription_namespace: openshift-operators
+:subscription_name: datagrid
+:operator_source: redhat-operators
```
ryanemerson, line 18: We need to add `operator_channel` here.

documentation/asciidoc/topics/attributes/community-attributes.adoc
```diff
@@ -17,6 +17,8 @@
 
 :example_crd_name: infinispan
 :example_namespace: ispn-namespace
+:subscription_namespace: operators
```
ryanemerson, line 20: `{subscription_name}` missing.

ryanemerson (changes requested):

documentation/asciidoc/topics/attributes/downstream-attributes.adoc
dvagnero: Hey! We do have another attribute file downstream 
https://gitlab.cee.redhat.com/red-hat-jboss-data-grid-documentation/rhdg-documentation/-/blob/8.4.x/rhdg-common/attributes/rhdg-attributes.adoc?plain=1#L62
It's a bit confusing but for now I'd like to keep it this way. When the downstream docs are build, the attribute is available.
I plan to do a follow-up review of the attributes and document the use of them

dvagnero:
Thank you for your review! It helps with the attributes a lot

ryanemerson:
Thanks @dvagnero

PR #1851, opened 2023-08-11
Allow StatefulSet PriorityClass to be configured. Resolves #1850

### LINKED ISSUE #1850: Allow StatefulSet PriorityClass to be configured
It should be possible for a user to configure the `PriorityClassName` used by the cluster StatefulSet via the Infinispan CR:

```yaml
kind: Infinispan
...
spec: 
  statefulset: 
    affinity: 
      ...
    priorityClassName: ""
    # Possible future configurations
    tolerations: 
      ...
    topologySpreadConstraints: 
      ...
```

Utilising a nested `spec.statefulset` structure provides increased flexibility that we can leverage for future enhancements such as https://github.com/infinispan/infinispan-operator/issues/1801 and https://github.com/infinispan/infinispan-operator/issues/1803

ryanemerson:


pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
```diff
@@ -127,6 +127,7 @@ func ClusterStatefulSetSpec(statefulSetName string, i *ispnv1.Infinispan, ctx pi
 							MountPath: DataMountPath,
 						}},
 					}},
+					PriorityClassName: i.PriorityClassName(),
```
Crumby, line 130: Would it be possible to write some kind of unit test for the value propagation to StatefulSet? Unfortunately functional test is not reasonably applicable here.

ryanemerson: Good point, I have added a unit test.

ryanemerson:
Updated to use `spec.scheduling` instead of `spec.statefulSet`.

PR #1858, opened 2023-08-23
[test-integration] Mark cache-service default cache tests as unstable

Crumby:
Sometimes default cache is not created (on slower k8s clusters) and it'll result in test failure. Given cache service is deprecated mark those tests as unstable so they are not executed as part of interop.


ryanemerson:
Thanks @Crumby

PR #1860, opened 2023-08-23
[stable][test-integration] Missed unstable tag

Crumby:
Missed tests to be marked as unstable with https://github.com/infinispan/infinispan-operator/pull/1858


ryanemerson:
Thanks @Crumby

PR #1865, opened 2023-08-29
Docs fix: Update statement about Cache CR

domiborges:
needs backport to 2.2.x


ryanemerson:
Thanks @domiborges. I don't think there's a need to backport to 2.2.x as the next upstream and downstream release will be built from `main`.

domiborges:
Thanks for a quick action @ryanemerson

PR #1882, opened 2023-10-12
Fixes JDG-6584 wrong tls secret

rigazilla:
In case it's not possible to create a keystore for the user provided `spec.security.endpointEncryption.certSecretName` deployment is halted and a failure status condition (`TLSSecretValid`) is added to the Infinispan CR.

Fixes JDG-6584

### DIFF
--- a/pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
@@ -9,6 +9,7 @@ import (
 	"github.com/infinispan/infinispan-operator/pkg/infinispan/security"
 	pipeline "github.com/infinispan/infinispan-operator/pkg/reconcile/pipeline/infinispan"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -48,8 +49,18 @@ func Keystore(i *ispnv1.Infinispan, ctx pipeline.Context) {
 		} else if isUserProvidedPrivateKey() {
 			keystore.Path = consts.ServerOperatorSecurity + "/" + EncryptPemKeystoreName
 			keystore.PemFile = append(keystoreSecret.Data["tls.key"], keystoreSecret.Data["tls.crt"]...)
+		} else {
+			errMsg := fmt.Sprintf("Failed to setup TLS keystore from secret %s", keystoreSecret.Name)
+			_ = ctx.UpdateInfinispan(func() {
+				i.SetCondition(ispnv1.ConditionTLSSecretValid, metav1.ConditionFalse, errMsg)
+			})
+			ctx.Requeue(fmt.Errorf(errMsg))
+			return
 		}
 	}
+	_ = ctx.UpdateInfinispan(func() {
+		i.SetCondition(ispnv1.ConditionTLSSecretValid, metav1.ConditionTrue, "")
+	})
 	ctx.ConfigFiles().Keystore = keystore
 }
 
--- a/pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
@@ -54,7 +54,7 @@ func Keystore(i *ispnv1.Infinispan, ctx pipeline.Context) {
 			_ = ctx.UpdateInfinispan(func() {
 				i.SetCondition(ispnv1.ConditionTLSSecretValid, metav1.ConditionFalse, errMsg)
 			})
-			ctx.Requeue(fmt.Errorf(errMsg))
+			ctx.Requeue(nil)
 			return
 		}
 	}
--- a/pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
@@ -50,7 +50,7 @@ func Keystore(i *ispnv1.Infinispan, ctx pipeline.Context) {
 			keystore.Path = consts.ServerOperatorSecurity + "/" + EncryptPemKeystoreName
 			keystore.PemFile = append(keystoreSecret.Data["tls.key"], keystoreSecret.Data["tls.crt"]...)
 		} else {
-			errMsg := fmt.Sprintf("Failed to setup TLS keystore from secret %s", keystoreSecret.Name)
+			errMsg := fmt.Sprintf("Failed to setup TLS. Secret %s must contain a keystore.p12 or a tls.key/tls.crt pair", keystoreSecret.Name)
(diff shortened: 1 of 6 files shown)


pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
```diff
@@ -48,8 +49,18 @@ func Keystore(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+			errMsg := fmt.Sprintf("Failed to setup TLS keystore from secret %s", keystoreSecret.Name)
+			_ = ctx.UpdateInfinispan(func() {
+				i.SetCondition(ispnv1.ConditionTLSSecretValid, metav1.ConditionFalse, errMsg)
+			})
+			ctx.Requeue(fmt.Errorf(errMsg))
+			return
 		}
 	}
+	_ = ctx.UpdateInfinispan(func() {
+		i.RemoveCondition(ispnv1.ConditionTLSSecretValid)
```
ryanemerson, line 62: Instead of removing the condition we should set it to `ConditionTrue`

ryanemerson (changes requested):
@rigazilla Please create a GH issue for this and update the commit message to reference it, so that we're able to track the work upstream.

ryanemerson:
Can you add a new test to `encryption_test.go` to verify that the `TLSSecretValid` condition is correctly set to false when the secret is misconfigured? For `TLSSecretValid=true` I think it's fine just to add a one-liner to `TestTLSWithExistingKeystore`.

pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
```diff
@@ -48,8 +49,18 @@ func Keystore(i *ispnv1.Infinispan, ctx pipeline.Context) {
 		} else if isUserProvidedPrivateKey() {
 			keystore.Path = consts.ServerOperatorSecurity + "/" + EncryptPemKeystoreName
 			keystore.PemFile = append(keystoreSecret.Data["tls.key"], keystoreSecret.Data["tls.crt"]...)
+		} else {
+			errMsg := fmt.Sprintf("Failed to setup TLS keystore from secret %s", keystoreSecret.Name)
+			_ = ctx.UpdateInfinispan(func() {
+				i.SetCondition(ispnv1.ConditionTLSSecretValid, metav1.ConditionFalse, errMsg)
+			})
+			ctx.Requeue(fmt.Errorf(errMsg))
```
ryanemerson, line 57: I don't think we need to return the `errMsg` here as this isn't an Operator error and we're conveying the message to the user via the Infinispan condition. So `ctx.Requeue(nil)` should be sufficient to avoid excessive error logging.

pkg/reconcile/pipeline/infinispan/handler/configure/tls.go
```diff
@@ -48,8 +49,18 @@ func Keystore(i *ispnv1.Infinispan, ctx pipeline.Context) {
 		} else if isUserProvidedPrivateKey() {
 			keystore.Path = consts.ServerOperatorSecurity + "/" + EncryptPemKeystoreName
 			keystore.PemFile = append(keystoreSecret.Data["tls.key"], keystoreSecret.Data["tls.crt"]...)
+		} else {
+			errMsg := fmt.Sprintf("Failed to setup TLS keystore from secret %s", keystoreSecret.Name)
```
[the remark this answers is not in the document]
rigazilla: Is it possible to identify which one is missing? code just looks for the p12 and then fallback on tls.*

test/e2e/infinispan/encryption_test.go
```diff
@@ -45,6 +46,31 @@ func TestTLSWithExistingKeystore(t *testing.T) {
...
 }
 
+func TestTLSConditionWithBadKeystore(t *testing.T) {
+	t.Parallel()
+	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
+
+	// Create a resource without passing any config
+	replicas := 2
+	spec := tutils.DefaultSpec(t, testKube, func(i *ispnv1.Infinispan) {
+		i.Spec.Replicas = int32(replicas)
```
ryanemerson, line 56: There's no need to explicitly define this as the test will work with the default value of 1.

test/e2e/utils/kubernetes.go
```diff
@@ -638,7 +638,16 @@ func (k TestKubernetes) WaitForPods(required int, timeout time.Duration, listOps
...
+func (k TestKubernetes) WaitForInfinispanConditionWithTimeout(name, namespace string, condition ispnv1.ConditionType, timeout time.Duration, value ...string) *ispnv1.Infinispan {
+	if len(value) > 0 && value[0] == string(metav1.ConditionFalse) {
+		return k.WaitForInfinispanStateWithTimeout(name, namespace, timeout, func(i *ispnv1.Infinispan) bool {
+			if i.IsConditionFalse(condition) {
+				log.Info("infinispan condition met", "condition", condition, "status", metav1.ConditionTrue)
+				return true
+			}
+			return false
+		})
+	}
```
ryanemerson, line 650: Instead of having the `if` logic in this function we can just add the logic directly to the `WaitForInfinispanConditionFalse` function:

```go
func (k TestKubernetes) WaitForInfinispanConditionFalse(name, namespace string, condition ispnv1.ConditionType) *ispnv1.Infinispan {
	return k.WaitForInfinispanStateWithTimeout(name, namespace, timeout, func(i *ispnv1.Infinispan) bool {
		if i.IsConditionFalse(condition) {
			log.Info("infinispan condition met", "condition", condition, "status", metav1.ConditionFalse)
			return true
		}
		return false
	})
}
```

test/e2e/infinispan/encryption_test.go
```diff
@@ -45,6 +46,29 @@ func TestTLSWithExistingKeystore(t *testing.T) {
...
+	spec := tutils.DefaultSpec(t, testKube, func(i *ispnv1.Infinispan) {
+		i.Spec.Security = ispnv1.InfinispanSecurity{
+			EndpointEncryption: tutils.EndpointEncryption(i.Name),
+		}
+	})
+
+	// Create secret
+	serverName := tutils.GetServerName(spec)
+	keystore, _ := tutils.CreateKeystore(serverName)
+	secret := tutils.EncryptionBadSecretKeystore(spec.Name, tutils.Namespace, keystore)
```
ryanemerson, line 63: Do you think we'll need this in another test? If not I would just declare inline to keep things simpler.

ryanemerson (commented):
LGTM, just a couple of nitpicks

test/e2e/infinispan/encryption_test.go
rigazilla: there are private func inside. I changed the code to corrupt the right secret

ryanemerson:
Thanks @rigazilla

PR #1902, opened 2023-11-07
XSite NodePort value has no effect. Fixes #1874

### LINKED ISSUE #1874: XSite NodePort value has no effect
Based on the documentation, should be able to set a custom NodePort value for Cross Site (especially in cases where firewall rules are being setup to allow traffic between sites)

I have set a section as:
```
    sites:
      local:
        discovery:
          launchGossipRouter: true
          type: gossiprouter
        expose:
          nodePort: 31314
          type: NodePort
```

But the NodePort is assigned randomly to the "site" service for the router.

Looking at the code, it would appear that NodePort is not handled in the function
https://github.com/infinispan/infinispan-operator/blob/8443d1ee715d07e2623d566a9fe699180c68e960/pkg/reconcile/pipeline/infinispan/handler/provision/services.go#L321

ryanemerson:


pruivo (approved):

pruivo:
merged! thanks @ryanemerson !

PR #1905, opened 2023-11-16
rename include to fix docs build #1881

domiborges:
follow up on #1881 
Apologies for the noise @ryanemerson but I forgot to rename the include


ryanemerson:
No worries @domiborges, thanks.

PR #1910, opened 2023-11-23
[test-integration] Reflect recent (Infinispan|OpenShift|Java) releases

Crumby:
Several fixes for the integration testsuite:
* Update Infinispan to 14.0.21.Final
  * Using `sniHostname` is now mandatory
*  Update Lombok to 1.18.30 (Required for Java 17 compatibility)
* Stabilize DataGridServiceIT#serviceMonitorTest by implementing condition waiter
  * Starting with OpenShift 4.13 it seems that Prometheus config reloader takes more time to (re)load service monitor config which results in test failure


ryanemerson:
Thanks @Crumby

PR #1920, opened 2023-12-05
Cross-site stack missing from zero capacity pod

### LINKED ISSUE #1915: Cross-site stack missing from Infinispan zero capacity pod
If cross-site is configured, the Infinispan zero capacity pod is unable to start caches because `RELA2` JGroups protocol is missing from the stack. This is a bug and it needs to be fixed. It also requires an additional attribute, `can_become_site_master=false` so the zero capacity pod isn't promoted to the `SiteMaster` role.

pruivo:
Fixes #1915


test/e2e/xsite/xsite_test.go
```diff
@@ -274,69 +276,69 @@ func TestDefaultTLSOpenshiftRoute(t *testing.T) {
 
 func TestCrossSiteGracefulShutdown(t *testing.T) {
 	testName := tutils.TestName(t)
-	tesKubes := map[string]*crossSiteKubernetes{"xsite1": {}, "xsite2": {}}
+	testKubes := map[string]*crossSiteKubernetes{"xsite1": {}, "xsite2": {}}
```
ryanemerson, line 279: I've been so tempted to do this on many occasions :slightly_smiling_face:

PR #1921, opened 2023-12-06
E2E tests stabilizations

Crumby:
Fixes several E2E tests to reduce noise in the test results:
* **TestJmxEnabled**
  * Skip the test if the test is executed against Infinispan 13
* **TestUpdatePodTargetLabels**
  * The test removed the 'testname' labels used to clean up resources after the execution
* **BackupRestoreTests**
  * The test occasionally fails with ISPN-15173 when running against older Infinispan versions. Given the issue won't be fixed, just skip the test when the error with this message pops up
* **TestGracefulShutdownWithTwoReplicas**
  * It's possible for the test to fail the PUT request with EOF right after the cluster was brought back up. The issue randomly appears only when the test is executed as part of the whole testsuite. Debugging showed up that in the occasions the EOF happens the request does not reach the cluster which suggests that there is some race condition between routing network in the OpenShift and executing the request. Fixed by retrying the request.

### DIFF
--- a/test/e2e/backup-restore/backup_restore_test.go
+++ b/test/e2e/backup-restore/backup_restore_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/iancoleman/strcase"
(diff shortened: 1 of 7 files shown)


test/e2e/backup-restore/backup_restore_test.go
```diff
@@ -159,7 +160,14 @@ func testBackupRestore(t *testing.T, clusterSpec clusterSpec, clusterSize, numEn
 	testKube.Create(restoreSpec)
 
 	// Ensure the restore pod has joined the cluster
-	testKube.WaitForValidRestorePhase(restoreName, namespace, v2.RestoreSucceeded)
+	err = testKube.WaitForValidRestorePhase(restoreName, namespace, v2.RestoreSucceeded)
+	if err != nil {
+		// Known issue with older operands that won't be fixed, skip the test to reduce noise in the test results
+		if strings.Contains(err.Error(), "ISPN-15173") {
+			tutils.SkipPriorTo(t, "14.0.21", "Known issue: "+err.Error())
```
ryanemerson, line 167: Nitpick. This should be the same as the ISPN-15173 fix version "14.0.18" as upstream we support "14.0.19", "14.0.20" and we may want to execute the test with said versions at some point.

PR #1963, opened 2023-12-22
Change the default timestamp format in the operator logging

### LINKED ISSUE #1962: Change the default timestamp format in the operator logging
The following options are available:

https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.12.3/pkg/log/zap#Options

It looks like we need to configure the `TimeEncoder` in https://github.com/infinispan/infinispan-operator/blob/main/main.go#L20
```
// TimeEncoder specifies the encoder for the timestamps in log messages.
// Defaults to EpochTimeEncoder as this is the default in Zap currently.
```

pruivo:
Fixes #1962

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -31,6 +31,8 @@ spec:
           args:
             - operator
             - --leader-elect
+            - --zap-time-encoding
+            - rfc3339
           imagePullPolicy: Always
           livenessProbe:
             httpGet:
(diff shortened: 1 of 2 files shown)


ryanemerson:
@pruivo You need to rebase on the latest `main` so that the `PROJECT` file is compatible with operator-sdk `v1.24.1`.

ryanemerson:
Resolved by https://github.com/infinispan/infinispan-operator/pull/2025

PR #1964, opened 2024-01-02
Updated instructions to wait for Batch CR to complete

domiborges:
Updated instructions to wait for Batch CR to complete
follow up to #1922


ryanemerson:
Thanks @domiborges

PR #1978, opened 2024-01-23
Specify TZ environment variable in POSIX format

### LINKED ISSUE #1977: Specify TZ environment variable in POSIX format
The expression("Asia/Tokyo") does not affect the local timezone in [1]. It needs to be specified in Posix format like "JST-9".
Sample:

```
kind: Subscription
spec:
  ...
  config:
    env:
    - name: ADDITIONAL_VARS
      value: "[\"TZ\"]"
    - name: TZ
      value: "JST-9"

```

[1] https://infinispan.org/docs/infinispan-operator/main/operator.html#defining-environment-variables_services

theashiot:
Update TZ to POSIX format


ryanemerson:
Thanks @theashiot

PR #1982, opened 2024-02-05
[2.3.x] Configure Gossip Route idle connection timeout

### LINKED ISSUE #1981: Configure Gossip Route idle connection timeout
Use the current heartbeat configuration (interval & timeout) to configure the idle connection timeout in the Gossip Router. It will allow the Gossip Router to force close orphan TCP connection in case of an Infinispan pod crash.

pruivo:
Fixes #1981


ryanemerson:
Thanks @pruivo

PR #1993, opened 2024-02-19
Make trust store required for cross-site connections

### LINKED ISSUE #1990: Truststore should be required
If the TLS trust store is not set, it will use the JVM trust store by default. In theory, any certificate signed by those CAs should be valid however, it is difficult to test this behavior. 

We should change the code so trust store is required for TLS encryption of cross-site communication.

pruivo:
Fixes #1990

### DIFF
--- a/api/v1/infinispan_webhook.go
+++ b/api/v1/infinispan_webhook.go
@@ -406,6 +406,13 @@ func (i *Infinispan) validate() error {
 				}
 			}
 		}
+
+		// print a warning if the truststore is not configured
+		if i.IsSiteTLSEnabled() && i.Spec.Service.Sites.Local.Encryption.TrustStore == nil {
+			errMsg := "The Trust Store for Cross-Site Encryption is recommended but it is not configured. It will fallback to the JVM default Trust Store."
+			eventRec.Event(i, corev1.EventTypeWarning, "CrossSiteTrustStoreMissing", errMsg)
+			log.Info(errMsg, "Request.Namespace", i.Namespace, "Request.Name", i.Name)
+		}
 	}
 
 	if i.Spec.CloudEvents != nil && operand.UpstreamVersion.GTE(semver.Version{Major: 15}) {
--- a/documentation/asciidoc/topics/proc_securing_cross_site_connections.adoc
+++ b/documentation/asciidoc/topics/proc_securing_cross_site_connections.adoc
@@ -20,15 +20,13 @@ You must provide a keystore for relay pods and router pods to secure cross-site 
 The keystore can be the same for relay pods and router pods or you can provide separate keystores for each. +
 You can also use the same keystore for each {brandname} cluster or a unique keystore for each cluster.
 
-* Optionally have a trust store that contains part of the certificate chain or root CA certificate that verifies public certificates for {brandname} relay pods and router pods.
-+
-By default, {brandname} uses the Java trust store to verify public certificates.
+* Have a PKCS12 trust store that contains part of the certificate chain or root CA certificate that verifies public certificates for {brandname} relay pods and router pods.
 
 .Procedure
 
 . Create cross-site encryption secrets.
 .. Create keystore secrets.
-.. Create trust store secrets if you do not want to use the default Java trust store.
+.. Create trust store secrets.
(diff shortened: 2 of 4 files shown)


Crumby:
@pruivo What happens with existing deployments that do not have truststore set?

pruivo:
@Crumby they have to add it. I'm happy to keep this out of 2.3.x and introduce this in the new version.
Do we have a test without Truststore? I'm getting a lot of questions about it asking why it does not work without a trust store.

Crumby:
If there isn't a test without a trustore in the xsite suite then no. Whether we decide to make it mandatory now or later we should test what happens on upgrade. I recall some kind of issues when old deployments did not meet new requirements. Might be unrelated though.

api/v1/infinispan_types.go
```diff
@@ -633,11 +633,10 @@ const (
 // EncryptionSiteSpec enables TLS for cross-site replication
 type EncryptionSiteSpec struct {
 	// +optional
-	Protocol          TLSProtocol       `json:"protocol,omitempty"`
-	TransportKeyStore CrossSiteKeyStore `json:"transportKeyStore"`
-	RouterKeyStore    CrossSiteKeyStore `json:"routerKeyStore"`
-	// +optional
-	TrustStore *CrossSiteTrustStore `json:"trustStore,omitempty"`
```
ryanemerson, line 640: I don't think we can make this change without breaking CRD compatibility :disappointed:

pruivo:
> If there isn't a test without a trustore in the xsite suite then no. Whether we decide to make it mandatory now or later we should test what happens on upgrade. I recall some kind of issues when old deployments did not meet new requirements. Might be unrelated though.

We do not have any testing. I don't have an answer for you but I hope nobody is using the trust store from JVM. It was idiot on my side to try to support it and, more importantly, is not secure.

api/v1/infinispan_types.go
pruivo: How? It is the same type and name, just not optional.

pruivo: I hope it fails because it isn't secure. See my comment: https://github.com/infinispan/infinispan-operator/pull/1993#issuecomment-1953827521

ryanemerson: We need to verify this is the case. On upgrade I *think* the webhooks will be invoked for the Infinispan CR. If we verify this is in fact the case and we keep the field as a pointer, then we can potentially return an error to the user informing them what change is required.

pruivo:
@ryanemerson @Crumby Should I revert the code changes and just print a warning that is recommended to set up a trust store, similar to https://github.com/infinispan/infinispan-operator/pull/2000/files#diff-cb655519d40d767a433c6eded34583bb0ffded9a892fc3434387260abe1a534cR411 ?

I would keep the docs changes saying it is required... I assume there is no way to make it required in the code.

ryanemerson:
@pruivo +1 to log a warning and issue a k8s event. I think that's the simplest and safest solution.

config/crd/bases/infinispan.org_infinispans.yaml
```diff
@@ -886,8 +886,8 @@ spec:
                 - minReplicas
                 type: object
               cloudEvents:
-                description: InfinispanCloudEvents describes how Infinispan is connected
-                  with Cloud Event, see Kafka docs for more info
+                description: Deprecated. Has no effect starting with Infinispan 15.0.0
```
pruivo, line 889: @ryanemerson this line changed when I compiled the code. I'm not sure if I should revert it or not.

pruivo:
Reverted the code changes and kept the documentation changes to make it clear that is required. A warning is shown if not configured:
```
LAST SEEN   TYPE      REASON                       OBJECT                      MESSAGE
6s          Warning   EphemeralStorageEnables      infinispan/tls-infinispan   Ephemeral storage configured. All data will be lost on cluster shutdown and restart.
6s          Warning   CrossSiteTrustStoreMissing   infinispan/tls-infinispan   The Trust Store for Cross-Site Encryption is recommended but it is not configured. It will fallback to the JVM default Trust Store.
```

config/crd/bases/infinispan.org_infinispans.yaml
ryanemerson: I've pushed the update to `main` :+1:

ryanemerson:
Thanks @pruivo

PR #1996, opened 2024-02-20
Improve security in Gossip Router

### LINKED ISSUE #1995: Improve security in Gossip Router
Configure TLS client authentication so it rejects connections from JGroups instances with invalid certificates.

pruivo:
Configures TLS client authentication

Fixes #1995


ryanemerson:
Thanks @pruivo. Does this need to be backported?

pruivo:
"Need"? Not really, but it would be nice. Would it be ok to add it?

ryanemerson:
@pruivo Definitely. Please create a PR  :+1:

PR #2009, opened 2024-02-27
Utilise 'caches?action=compare' endpoint in ConfigListener

ryanemerson:


ryanemerson:
Requires either a new DevXX release or 15.0.0.Final in order for `TestCacheWithServerLifecycle` to pass

PR #2013, opened 2024-02-29
Cleanup JGroups stacks

pruivo:
* use kubernetes stack shipped with Infinispan
* remove attributes with the same value as the default
* remove extra stack with cross-site

---

Just trying to clean up the stacks. The stack `xsite` was removed and `RELAY2` was added to the existing `image-tcp` stack.


ryanemerson:
LGTM, just waiting on the GH ugprade test to re-run. I've removed the "2.3.x" label as it will fail due to the `*15.xml` files not being present on the downstream branch.

ryanemerson:
Thanks @pruivo

PR #2015, opened 2024-03-01
[2.3.x] Cleanup JGroups stacks

pruivo:
* use kubernetes stack shipped with Infinispan
* remove attributes with the same value as the default
* remove extra stack with cross-site


ryanemerson:
Thanks @pruivo

PR #2020, opened 2024-03-04
Allow Readiness, Liveness and Startup probe values to be configured. …

ryanemerson:


pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
```diff
@@ -84,6 +84,18 @@ func StatefulSetRollingUpgrade(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+	probedChanged := func(current, new *corev1.Probe) bool {
+		if reflect.DeepEqual(current, new) {
+			return false
+		}
+		*current = *new
+		return true
+	}
+	updateNeeded = probedChanged(container.LivenessProbe, provision.PodLivenessProbe(i))
+	updateNeeded = probedChanged(container.ReadinessProbe, provision.PodLivenessProbe(i))
+	updateNeeded = probedChanged(container.StartupProbe, provision.PodLivenessProbe(i))
```
pruivo, line 97: 2 concernts
* all methods are using `provision.PodLivenessProbe(i)`?
* shouldn't be "or", I mean, if the startup probe is not changed, `updateNeeded` will be set to `false` independently of the other probes

documentation/asciidoc/topics/proc_configuring_pod_probes.adoc
```diff
@@ -0,0 +1,26 @@
+[id='configuring-pod-probes{context}']
```
pruivo, line 1: file needs to be linked somewhere, right?

ryanemerson: It would definitely help :slightly_smiling_face:

ryanemerson:
@pruivo I've re-worked the defaulting logic so that default values are applied on a per field basis. This means that if a user does not provide a value explicitly, then the Operator default will always be applied.

I think this is less confusing than the previous behaviour, where configuring a single field for a probe type (e.g. readiness probe) would overwrite all default values for that probe.

api/v1/infinispan_types.go
```diff
@@ -106,6 +106,45 @@ type InfinispanServiceContainerSpec struct {
...
+	StartupProbe ContainerProbeSpec `json:"startupProbe,omitempty"`
+}
+
+type ContainerProbeSpec struct {
+	// Number of seconds after the container has started before liveness probes are initiated.
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+	// +optional
+	InitialDelaySeconds *int32 `json:"initialDelaySeconds,omitempty"`
+	// Number of seconds after which the probe times out.
+	// Defaults to 1 second. Minimum value is 1.
```
pruivo, line 132: I would remove the `Defaults to ...` from all docs here since each type of probe has different timeouts.

api/v1/infinispan_types.go
```diff
@@ -106,6 +106,45 @@ type InfinispanServiceContainerSpec struct {
...
+	// More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+	// +optional
+	TimeoutSeconds *int32 `json:"timeoutSeconds,omitempty"`
+	// How often (in seconds) to perform the probe.
+	// Default to 10 seconds. Minimum value is 1.
+	// +optional
+	PeriodSeconds *int32 `json:"periodSeconds,omitempty"`
+	// Minimum consecutive successes for the probe to be considered successful after having failed.
+	// Defaults to 1. Must be 1 for liveness and startup. Minimum value is 1.
```
pruivo, line 141: Do we want to enforce `Must be 1 for liveness and startup`? I'm guessing, probably not.

ryanemerson: > Do we want to enforce `Must be 1 for liveness and startup`?

We're already doing this in the webhook.

api/v1/infinispan_types.go
ryanemerson: Agreed. I copy and  pasted these from the k8s go libraries originally before I changed the approach, thanks for spotting.

api/v1/infinispan_webhook.go
```diff
@@ -421,6 +428,24 @@ func (i *Infinispan) validate() error {
 		log.Info(errMsg, "Request.Namespace", i.Namespace, "Request.Name", i.Name)
 	}
 
+	validateProbes := func(c *ContainerProbeSpec, path *field.Path) {
+		checkMinVal := func(val, min int32, path *field.Path) {
+			if val < min {
+				msg := fmt.Sprintf("Probe value must be greater than or equal to '%d'", val)
```
pruivo, line 434:
```suggestion
				msg := fmt.Sprintf("Probe value must be greater than or equal to '%d'", min)
```

api/v1/infinispan_types.go
pruivo: From my understanding, webhook is only setting defaults. If the user configures it in the CR, the user can set a value greater than 1.

ryanemerson: Argh I see what you mean now. I have updated our webhook to prevent values != 1 for liveness and startup probes.

pruivo (approved):

ryanemerson:
Thanks for the review @pruivo

PR #2036, opened 2024-03-25
Test 2.3.7 -> 2.4.0 upgrades only to isolate restore failure

ryanemerson:
@jabolina FYI


ryanemerson:
Argh, that's happening on the `main` branch as well. I'll take a closer look

PR #2043, opened 2024-04-04
Emit event and log warning if service is Cache. Fixes #2042 and JDG-6081

### LINKED ISSUE #2042: Create a warning for Deprecated service type cache in Infinispan CR
Emit a warning event and log when a Infinispan CR is of type CacheService

rigazilla:
fixes #2042


Crumby:
Isn't there a way to emit a warning without having to store Operator name in the Infinispan CR status? From why understood it's there so we are able to recognize that Operator has been restarted, which seems bit weird. Isn't there a code that runs on startup and checks all the CRs and their states?

Other question I have is regarding how often the Warning will be emitted. Usually warnings are emitted periodically as long as they persist but this does not seem to be case here

ryanemerson:
> Isn't there a way to emit a warning without having to store Operator name in the Infinispan CR status? From why understood it's there so we are able to recognize that Operator has been restarted, which seems bit weird. Isn't there a code that runs on startup and checks all the CRs and their states?

A reconciliation event for every CR will be triggered on Operator pod startup. However, without storing the Operator name in the status it's not possible to ensure the log and event is only triggered once. Instead, every time the CR is reconciled the event/log would be raised, which will be very verbose in some cases as we watch for updates to other resources in order to queue reconcilliations.

> Other question I have is regarding how often the Warning will be emitted. Usually warnings are emitted periodically as long as they persist but this does not seem to be case here

An event will be issued every time a CR with the cache service is updated, but not every reconciliation event.

ryanemerson:
Thanks @rigazilla @Crumby

PR #2049, opened 2024-04-10
OOM failover test fix

Crumby:
`TestPodDegradationAfterOOM` on OpenShift 4.15 no longer seems to fail with `OOMKilled` but rather provides exit code 137 which indicates it was abruptly terminated by external signal.


ryanemerson:
Thanks @Crumby

PR #2054, opened 2024-04-11
Add deprecation note for cloud events in doc #2053

theashiot:


documentation/asciidoc/stories/assembly_eventing.adoc
```diff
@@ -14,6 +14,12 @@ Sending cloud events with {openshift} Serverless is available as a technology pr
 ====
 
 include::{topics}/ref_tech_preview.adoc[leveloffset=+1]
+
+[IMPORTANT]
+====
+This feature is deprecated and planned for removal in a future release.
```
ryanemerson, line 20: I think we need to be stronger. We should say something like:

"This feature is deprecated and has no effect when utilising Infinispan 15 or higher."

theashiot: Thanks, @ryanemerson, updated!

ryanemerson:
Thanks @theashiot

PR #2060, opened 2024-04-17
[test-integration] CustomLibsIT Multi-Attach error fix

Crumby:
Mounting a PV with ReadWriteOnce access mode is not possible from multiple k8s nodes. In the context of test it may cause an issue if the uploading pod is taking most of the space on the node not allowing for Infinispan node to be scheduled on the same node and result in Multi-Attach error as a result.

**Fix:** Delete the pod created to upload the libs for the cluster after the upload is complete.


ryanemerson:
Thanks @Crumby

Issue #2072, opened 2024-05-10
Forbid new CR to be of type CacheService


rigazilla:
Operator shouldn't allow new deployments of type CacheService.
Old CacheService CR once upgraded will be marked as WellFormed=False

ref. https://issues.redhat.com/browse/JDG-6191

PR #2074, opened 2024-05-13
Remove Cache Service Support #2072

rigazilla:
This PR implements the following policy:

CacheService CR are no more allowed with Infinispan 15+

see #2072 
and https://issues.redhat.com/browse/JDG-6191


api/v1/infinispan_webhook.go
```diff
@@ -285,7 +285,9 @@ func (i *Infinispan) validate() error {
 		}
 	}
 
-	if err := i.validateCacheService(); err != nil {
+	if i.IsCache() {
+		msg := "CacheService is no more supported."
```
ryanemerson, line 289: `s/no more/no longer`

api/v1/infinispan_webhook_test.go
```diff
@@ -55,34 +55,6 @@ var _ = Describe("Infinispan Webhooks", func() {
 	})
 
 	Context("Infinispan", func() {
-		It("Should initiate Cache Service defaults", func() {
```
ryanemerson, line 58: We should add a new test which ensures that `FieldValueForbidden` is returned when a user tries to create a CR with the CacheService type.

pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
```diff
@@ -41,22 +41,21 @@ func PreliminaryChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+
 func OperatorStatusChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
 	operatorPod := kube.GetOperatorPodName()
 	// Pod name is changed, means operator restarted
 	if i.Status.Operator.Pod != operatorPod {
-		if i.IsCache() {
-			errMsg := "ServiceType Cache configured is deprecated and will be removed in future releases. ServiceType DataGrid is recomended."
-			ctx.EventRecorder().Event(i, corev1.EventTypeWarning, "DeprecatedCacheService", errMsg)
-			ctx.Log().Info(errMsg)
+		if i.IsCache() && ctx.Operand().BansCacheService() {
```
ryanemerson, line 50: I think this should be outside of the pod if check as we always want to set `WellFormed=False` if the cache service is present. Furthermore, we need to ensure that if cache service is present that we don't continue with the reconcilliation pipeline by calling `ctx.Stop(nil)`

pkg/infinispan/version/version.go
```diff
@@ -70,6 +70,13 @@ func (o Operand) LT(other Operand) bool {
 	return o.UpstreamVersion.LT(*other.UpstreamVersion)
 }
 
+// Hardcoding first version that bans cacheservice
+var minVerBansCacheService = semver.MustParse("15.0.0")
+
+func (o Operand) BansCacheService() bool {
```
ryanemerson, line 76: There's no need for us to check the Upstream Operand version as we're removing the cache service functionality from the Operator completely, so it won't work with any of the Operand versions.

pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
```diff
@@ -115,9 +114,18 @@ func AwaitWellFormedCondition(i *ispnv1.Infinispan, ctx pipeline.Context) {
 }
 
 func wellFormedCondition(i *ispnv1.Infinispan, ctx pipeline.Context, podList *corev1.PodList) ispnv1.InfinispanCondition {
+	// Avoid computation if CR requires CacheService when is forbidden
+	wellFormed := ispnv1.InfinispanCondition{Type: ispnv1.ConditionWellFormed}
```
ryanemerson, line 118: If we call `ctx.Stop` in `OperatorStatusChecks` then there's no need for this logic as it will never be reached.

test/e2e/infinispan/smoke_test.go
```diff
@@ -88,23 +88,3 @@ func verifyDefaultAuthention(require *require.Assertions, ispn *v1.Infinispan) {
 
 	testAuthentication(ispn, schema, user, pass)
 }
-
-func TestCacheService(t *testing.T) {
```
ryanemerson, line 92: We should update this test to ensure that WellFormed=False is set. As webhooks are not in place when executing this test, it should still be possible to configure the service type.

ryanemerson (changes requested):

ryanemerson (commented):
My understanding was that we were going to remove the Cache service implementation from the Operator itself,  in which case we can also remove the various pipeline handler's associated with this service type.

rigazilla:
> My understanding was that we were going to remove the Cache service implementation from the Operator itself, in which case we can also remove the various pipeline handler's associated with this service type.

uhm ok, let me see

controllers/cache_controller.go
```diff
@@ -274,50 +273,13 @@ func (r *cacheRequest) ispnCreateOrUpdate() (*ctrl.Result, error) {
 		return &ctrl.Result{}, err
 	}
 
-	if r.infinispan.IsDataGrid() {
```
ryanemerson, line 277: I think there's a small window when a user upgrades to the latest Operator where we can have WellFormed=True with the cache service, as the cache_controller and infinispan_controller are independent. We can guard against this by adding a check after the `if !infinispan.IsWellFormed() {` block on L190, something like:

```yaml
if infinispan.IsCache() {
   log.info("Ignoring Cache CR as cluster is marked as CacheService")
   return
}
```

pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
```diff
@@ -41,22 +41,27 @@ func PreliminaryChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
 	}
 }
 
+var cacheServiceForbiddenMsg = "Please change to service type DataGrid. CacheService is no longer supported."
```
ryanemerson, line 44: If we want this to be a separate var, than we should mark this as a `const`.

Personally I would just declare this inline as it's only used once in the function, e.g.:

```go
i.SetCondition(
  ispnv1.ConditionWellFormed,
  metav1.ConditionFalse,
  "Please change to service type DataGrid. CacheService is no longer supported."
)
```

That's your call though :slightly_smiling_face:

rigazilla: ok to remove, it was just a relic from a previous version thanks

ryanemerson:
Thanks @rigazilla

Issue #2075, opened 2024-05-15
Remove Infinispan 13.0.10


ryanemerson:
The 13.0.10 server is no longer supported and no security updates are provided, so we should not longer bundle this with the Operator. This version will be marked as deprecated in 2.4.2.CSV release and removed in 2.4.3.CSV.

PR #2076, opened 2024-05-15
Deprecate Infinispan 13.0.10 Operand for removal in 2.4.3.CSV

ryanemerson:
@rigazilla I realised that we still have 13.0.10 in our list of Operands, I have marked this as deprecated so that we can remove this in the release after next. The 13.0.10 server is no longer supported and no security updates are provided, so we should not longer bundle this with the Operator #2075.


rigazilla:
merged thanks @ryanemerson

PR #2081, opened 2024-05-16
Avoid spec.autoscale to be set. Fixes #2079

### LINKED ISSUE #2079: Block the activation of spec.autoscale
User should not be allowed to set `spec.autoscale!=nil` since Cache Service type is no longer available.

See https://issues.redhat.com/browse/JDG-5989

rigazilla:
Block setting of spec.autoscale on new cr and emit error log during operator upgrade if spec.autoscale is not nil

#2079

### DIFF
--- a/pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
@@ -54,6 +54,11 @@ func OperatorStatusChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
 	}
 	// Pod name is changed, means operator restarted
 	if i.Status.Operator.Pod != operatorPod {
+		if i.Spec.Autoscale != nil {
+			errMsg := "Autoscale is no longer supported. Please remove spec.autoscale field."
+			ctx.EventRecorder().Event(i, corev1.EventTypeWarning, "AutoscaleNotSupported", errMsg)
+			ctx.Log().Error(fmt.Errorf("AutoscaleNotSupported"), errMsg)
+		}
 		ctx.Requeue(
 			ctx.UpdateInfinispan(func() {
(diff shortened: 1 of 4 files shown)


pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
```diff
@@ -54,6 +54,9 @@ func OperatorStatusChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
 	}
 	// Pod name is changed, means operator restarted
 	if i.Status.Operator.Pod != operatorPod {
+		if i.Spec.Autoscale != nil {
+			ctx.Log().Info("Autoscale is no longer supported. Please remove spec.autoscale field.", "Request.Namespace", i.Namespace, "Request.Name", i.Name)
```
ryanemerson, line 58: Should this be an error so that it has more visibility?

rigazilla: should we do as we did for [cacheservice](https://github.com/infinispan/infinispan-operator/commit/5b121f826f4499a4b8bfb3417de360e9e07a54eb), warn + event?

ryanemerson: Yes, that's a good idea :+1:

ryanemerson: The current code is still logging an info message. I know the API doesn't expose a `.Warn(..)` method, so we either stick with `info()` and risk the log being missed or go with `.Error(..)` to increase it's exposure.

rigazilla: ok for error

ryanemerson:
Thanks @rigazilla

PR #2085, opened 2024-05-16
Include Operand deprecation status in Infinispan CR .status. Resolves…

ryanemerson:


pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
```diff
@@ -247,8 +247,9 @@ func OperandStatus(i *ispnv1.Infinispan, phase ispnv1.OperandPhase, operand vers
 	}
 
 	return ispnv1.OperandStatus{
-		Image:   img,
-		Phase:   phase,
-		Version: operand.Ref(),
+		Deprecated: operand.Deprecated,
```
Crumby, line 250: I presume the status is refreshed on Operator respin or regularly?

ryanemerson: Good question, before it was only actually set for new Infinispan CRs. I have added a new check that will update the status if the existing value is not equal to that contained in the metadata.

rigazilla (approved):
Looks good to me. 
Quick note: 
https://github.com/ryanemerson/infinispan-operator/blob/6659be1c4711f0a9e44fd95e00486961121b914f/pkg/infinispan/version/version.go#L99-L100
could now return deprecated version? could this generate side effects?

PR #2087, opened 2024-05-16
Nil pointer error on upgrade from dropped Operand version. Fixes #2086

### LINKED ISSUE #2086: Nil pointer error on upgrade from dropped Operand version
If an operand version specified in an Infinispan CR is removed from the INFINISPAN_OPERAND_VERSIONS json in a subsequent Operator release, then this will result in the Infinispan CR being ignored by the Operator and the following being logged:

```qoute
"unable to continue reconcilliation. Operand spec.version='%s' is not supported by this Operator version"
```

If the user then tries to update `spec.version`` to use a supported version, [this code](https://github.com/infinispan/infinispan-operator/blob/dc8d4c03b78d980a204b8fdcda019ec6f354d19e/pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go#L68-L76) will throw a nil pointer dereference error as the version specified in the `.status.operand.version` is now unknown.

We should update the code so that the upgrade is able to proceed if an unknown version is specified in the `.status.operand.version` field

ryanemerson:


rigazilla:
merged thanks @ryanemerson

PR #2090, opened 2024-05-21
Deprecate operand warning

ryanemerson:


pkg/reconcile/pipeline/infinispan/handler/manage/conditions.go
```diff
@@ -54,6 +54,12 @@ func OperatorStatusChecks(i *ispnv1.Infinispan, ctx pipeline.Context) {
 	}
 	// Pod name is changed, means operator restarted
 	if i.Status.Operator.Pod != operatorPod {
+		if ctx.Operand().Deprecated {
+			msg := fmt.Sprintf("Infinispan version '%s' will be removed in a subsequent Operator release. You must upgrade to a non-deprecated release before upgrading the Operator.", i.Spec.Version)
+			ctx.EventRecorder().Event(i, corev1.EventTypeWarning, "DeprecatedOperandVersion", msg)
+			ctx.Log().Error(nil, msg)
```
Crumby, line 60: I'd say that error indicates invalid state so I'd go with Warning here.

ryanemerson: The log interface does not have a Warn category, so the options are INFO or ERROR. Elsewhere, see Autoscale no longer supported below, we have also used Error to ensure that the message is not missed by users.

Crumby (approved):

rigazilla (approved):

rigazilla:
@ryanemerson can I squash the commits?

ryanemerson:
@rigazilla I've kept the commits separate as they're two distinct changes and the automatic backporting ignores the squashed commit

PR #2095, opened 2024-05-22
Operator is using deprecated cache-managers endpoint for health check…

ryanemerson:
This PR introduces a PathResolver interface which allows the urls of the various server resources to be abstracted so that we don't need to duplicate resource logic between versions when the path has changed but the semantics remain the same.


rigazilla (commented):
seems good to me, but ci is complaining

PR #2096, opened 2024-05-23
Resources spec for Batch. Fixes #2093

### LINKED ISSUE #2093: It should be possible for users to specify Batch resource request/limits
Currently the Batch Job created by the Operator does not define CPU or Memory request/limits. Consequently, customers that have quotas set which require limits/requests set as mandatory field will not be able to utilise Batch CRs.

rigazilla:
fixes #2093 

see also https://issues.redhat.com/browse/JDG-7031


controllers/batch_controller.go
```diff
@@ -318,3 +320,32 @@ func batchLabels(name string) map[string]string {
 		"app":              "infinispan-batch-pod",
 	}
 }
+
+func BatchResources(spec *v2.BatchContainerSpec) (*corev1.ResourceRequirements, error) {
```
ryanemerson, line 324: We always ignore the returned `error`, so we can simplify this function to also ignore the error return values from `MemoryResources` and `CpuResources`. This should be safe as the webhook always ensures that the values are in the correct format.

PR #2098, opened 2024-05-29
[test-integration] Remove Cache Service related tests

Crumby:
* Remove Cache Service related tests
* Repurpose original Cache Service deployment test to OpenShift certificate test


ryanemerson:
Thanks @Crumby

PR #2108, opened 2024-06-07
Do not assume latest Operand in tests

Crumby:
Testsuite assumes that tests are always executed with latest Operand using latest API. That's causing test failures when executed against older operand versions.


ryanemerson:
Thanks @Crumby

PR #2110, opened 2024-06-11
Drop support for Infinispan 13

ryanemerson:


Crumby:
I presume we don't want to drop https://github.com/infinispan/infinispan-operator/tree/main/pkg/infinispan/client/v13 as it contains base for v14 and v15

ryanemerson:
@Crumby We do want to drop that, I was just rushing when I created the PR and forgot to include the actual code changes :smile: 


@pruivo Can you take a look at the xsite changes?

pruivo:
@ryanemerson the cross-site changes look fine to me :+1:

PR #2118, opened 2024-06-28
Remove cloud events documentation from docs #2117

theashiot:


theashiot:
Hi @ryanemerson,

Can you please review this PR for removing Cloud Events docs?

thanks,
ashwin

Crumby:
Should we remove the parts on the configuration as well?
* https://github.com/infinispan/infinispan-operator/blob/main/documentation/asciidoc/topics/proc_enabling_cloud_events.adoc
* https://github.com/infinispan/infinispan-operator/blob/main/documentation/asciidoc/topics/con_cloud_events.adoc
* https://github.com/infinispan/infinispan-operator/blob/main/documentation/asciidoc/stories/assembly_eventing.adoc

Or maybe just keep the info about having the configuration having no effect and remove all the configuration because at this point it's pretty pointless to keep around

theashiot:
Hi @Crumby, do you mean we should delete the files as well? I just removed the  assembly_eventing.adoc include so that info about cloud events is not visible in the guide.

thanks,
ashwin

Crumby:
> I don't think we want to remove any content. Cloud Events will still work with older Operands

ryanemerson:
> Right, is that something we want though? If I'm not mistaken eg. Cache Service won't work with older operands.

The difference between cloud events and the Cache Service is that the former is a server-side capability, whereas the Cache Service configuration is provided by the Operator. So IMO it still makes sense for CloudEvents configuration to be possilbe on 14.0.x servers via the CR, as the user could still configure this with a custom user configuration, so why make it harder for them?

theashiot:
Perhaps we can merge this as-is right now for downstream, and revisit this via a a new GH issue on whether to reinstate/modify the docs?

ryanemerson:
> I wonder who'd be the target audience for information on how to set it up. Anyone who'd be setting up a new cluster would use Infinispan 15 which doesn't have cloud events. Anyone coming from older releases has the feature already set up (we are removing just the documentation here) or can find the information in Infinispan 14 docs.

What about people maintaining existing deployments?

> Generally I wonder if it makes sense to keep the documentation/feature in the Operator for a corresponding server release where it gets removed.

The purpose of adding multi-operand support to the Operator was so that there wasn't a 1-1 coupling between the server and the Operator. I don't think it makes sense to impose this on the documentation.

theashiot:
I have removed the files. WDYT?

ryanemerson:
Thanks @theashiot

PR #2130, opened 2024-07-23
Update go-redis to v9.6.0

Crumby:
There is no go-redis v9.5.3. Attempt to vendor repository fails with unknown revision.


ryanemerson:
Hmm that's very strange, I wonder where that disappeared to :thinking:

PR #2131, opened 2024-07-23
Remove deprecated operand check

Crumby:
`TestOperandUpgrade` expects the oldest operand version to be deprecated which may not be true.


ryanemerson:
@Crumby We need this to make sure that the deprecated field in the status is set correctly. Upstream we ensure that the oldest version in the testing Operand list is deprecated specifically for this check. Instead of removing it, could you surround the assertion with an if statement checking whether the oldest Operand is deprecated, and if not skip?

Crumby:
@ryanemerson Fixed. I didn't realize it's actually checking for status field

ryanemerson:
Thanks @Crumby

PR #2133, opened 2024-07-29
Exposing scale subresource to enable HPA's to modify the replicas of Infinispan CR

ZeidH:
Enhanced the `spec.subresources` field for the `Infinispan` CRD to map scaling
```
      scale:
        specReplicasPath: .spec.replicas
        statusReplicasPath: .status.replicas
```

This enables APIs like the HorizontalPodAutoscaler and KEDA to modify the replica count of the Cache.


ryanemerson (changes requested):
Thanks for the contribution @ZeidH, however we generate our CRD content from code, so it's not sufficient to update the static resource files.

Looking at the [Artemis PR](https://github.com/rh-messaging/activemq-artemis-operator/commit/3ba7a87485f5ac9efa221360afe3900da1fce6cb) you linked on the issue, it seems like we can generate this by adding the following annotation above the [Infinispan struct](https://github.com/infinispan/infinispan-operator/blob/main/api/v1/infinispan_types.go#L664-L675):

```
//+kubebuilder:subresource:scale:specpath=.spec.replicas,statuspath=.status.replicas
```

You should then be able to generate the static files by doing `make generate`.

ZeidH:
Thanks for the clear instructions @ryanemerson!, I'm rather new to this so appreciate the help.
Encountered [this issue](https://github.com/kubernetes-sigs/controller-tools/issues/888) with controller-gen on my version of go (1.22.5) so I also updated it to latest (1.15.0).

ryanemerson:
Thanks for updating @ZeidH. It seems like we need to increase our baseline go version in order to use that version of controller-gen. I have created a PR to baseline on 1.21 (1.20 is now EOL) and update the controller-gen version https://github.com/infinispan/infinispan-operator/pull/2134. I'll let you know once that PR has been merged so that you can rebase this PR.

ZeidH:
I've tested the autoscaling with this setup:
```
./scripts/ci/kind-with-olm.sh
make deploy-cert-manager
helm install keda kedacore/keda --namespace keda --create-namespace
skaffold dev
```
and a `ScaledObject` that targets the Infinispan CR
```
apiVersion: keda.sh/v1alpha1
kind: ScaledObject
metadata:
  name: infinispan-scaler
  namespace: infinispan-operator-system
spec:
  minReplicaCount: 0
  scaleTargetRef:
    apiVersion: infinispan.org/v1
    kind: Infinispan
    name: test-cache
  triggers:
    - metadata:
        desiredReplicas: '2'
        start: 52 8 * * *
        end: 59 8 * * *
        timezone: UTC
      type: cron
---
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  annotations:
    infinispan.org/monitoring: 'true'
    infinispan.org/operatorPodTargetLabels: 'rht.comp,rht.comp_ver,rht.prod_name,rht.prod_ver,rht.subcomp_t'
  name: test-cache
  namespace: infinispan-operator-system
spec:
  service:
    container:
      livenessProbe:
        failureThreshold: 5
        initialDelaySeconds: 0
        periodSeconds: 10
        successThreshold: 1
        timeoutSeconds: 1
      readinessProbe:
        failureThreshold: 5
        initialDelaySeconds: 0
        periodSeconds: 10
        successThreshold: 1
        timeoutSeconds: 1
      startupProbe:
        failureThreshold: 600
        initialDelaySeconds: 3
        periodSeconds: 1
        successThreshold: 1
        timeoutSeconds: 1
    type: DataGrid
  jmx: {}
  configListener:
    enabled: true
    logging:
      level: info
  upgrades:
    type: Shutdown
  replicas: 1
  ```

The Statefulset was able to scale from zero to n and back but I can imagine that for some configurations scaling to zero is not possible

File: api/v1/infinispan_types.go [hunk 1]
```diff
@@ -611,6 +611,9 @@ type InfinispanStatus struct {
 	// +optional
 	// +operator-sdk:csv:customresourcedefinitions:type=status,displayName="Operator Status"
 	Operator Operator `json:"operator,omitempty"`
+	// The Operator Selector
+	// +optional
+	Selector string  `json:"selector"`
```
ryanemerson, line 616: Why do we need this additional field?

ryanemerson:
Apologies for the delay @ZeidH, but the [go 1.21 baseline](https://github.com/infinispan/infinispan-operator/commit/94f48320233e63a24a277c1e9abaa3d87a043a59) has been merged to main, can you please rebase on the latest code?

ryanemerson:
> but I can imagine that for some configurations scaling to zero is not possible

Can you elaborate on the configurations you have in mind?

api/v1/infinispan_types.go
ZeidH: The HPA that KEDA generates requires the target CR to have a selector.
> The HPA will use the scale endpoint to retrieve the labelselectors of the child pods, as well as the status and spec of replicas for scaling.

I basically followed this article: https://medium.com/@thescott111/autoscaling-kubernetes-custom-resource-using-the-hpa-957d00bb7993

config/manifests/bases/infinispan-operator.clusterserviceversion.yaml
```diff
@@ -136,6 +136,8 @@ spec:
         path: replicas
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:podCount
+      - displayName: Replicas
+        path: Replicas
```
ryanemerson, line 140: This array entry can be removed as `Replica` is already defined above.

api/v1/infinispan_types.go
ryanemerson: Thanks for linking the article, that was really helpful :+1:

In that case I'm happy to add the additional field. Can you update the description of the field to be more informative, maybe:

"The Selector used to identify Infinispan cluster pods"

ryanemerson (changes requested):
Thanks @ZeidH, just a couple of minor comments and then I think we're good.

ZeidH:
> > but I can imagine that for some configurations scaling to zero is not possible
> 
> Can you explain which configurations you think would cause issues?

I can't think of any specific to Infinispan, the amount of available [configurations](https://infinispan.org/docs/stable/titles/configuring/configuring.html) is so immense it would take a long time to figure out what config can be autoscaled, and what cannot. I think after learning a bit on what Infinispan does the past few months, the configs where autoscaling doesn't work well also most likely wouldn't make much sense to use autoscaling/scale to zero, like clustered caches where you (almost)always want to have a hot instance. 

But from experience there can be issues with autoscaling when using specific types of storage configurations, or as @rigazilla mentioned in the GH Issue: about the operator sometimes needs to explicitly set the `.spec.Replicas`, i.e. in [upgrade](https://github.com/infinispan/infinispan-operator/blob/c9bdd7f5d8e25b95b03c804e73f4fb0f7dc3c869/pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go#L62)

What we attempted to do is to dynamically scale a replicated cache with the amount of nodes (1 cache per node), essentially making Infinispan behave like a `DaemonSet`. Although we're not done with it just yet, so far it's been working out well.

ryanemerson:
Thanks @ZeidH and sorry for the delay. I'll create a follow up PR with some documentation. Let me know if you need this feature soon via OLM and I can expedite the next release.

ZeidH:
Appreciate the feedback and the completion of this issue! It will help us greatly! We're not in a rush but we'd be very happy if this was released sometime in coming 2 weeks!

ryanemerson:
@ZeidH 2.4.5 has been released now.

PR #2144, opened 2024-09-04
RESP and HotRodMigration test fixes

Crumby:
Several test fixes to adjust for branding versions, feature availability and different default behavior on OpenShift:
* TestRollingUpgrade compared a given version to an upstream one, which is incorrectly evaluated in case of brand versioning
* Redis is available by default starting from 15.0.0
* Running the tests on the OpenShift makes the cluster encrypted by default but Redis client does not account for that
* Fix HostAndPort to really have a port in case of Route expose type. Redis will complain on missing port otherwise


ryanemerson:
@Crumby There are CI failures, please address.

ryanemerson:
Thanks  @Crumby

PR #2145, opened 2024-09-09
 Updates to spec.image on existing Infinispan CR have no effect

### LINKED ISSUE #1223: Changing `spec.image` does not cause reconciling
https://github.com/infinispan/infinispan-operator/blob/aec1a7e179bceed9f9df74336df4c3b856595515/pkg/controller/infinispan/infinispan_controller.go#L809

`upgradeRequired` required only looks at `DEFAULT_IMAGE` which is derived from the operator configuration. `spec.image` is completely ignored.

ryanemerson:
Fixes #1223

### DIFF
--- a/api/v1/infinispan_types.go
+++ b/api/v1/infinispan_types.go
@@ -623,6 +623,9 @@ const (
 )
 
 type OperandStatus struct {
+	// Whether the Operand installed/pending is using a custom image
+	// +optional
+	CustomImage bool `json:"customImage,omitempty"`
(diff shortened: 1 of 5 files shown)


Crumby:
@ryanemerson Have you tried going from a custom image back to a newer version (remove `spec.image` and use newer version)

ryanemerson:
> Have you tried going from a custom image back to a newer version (remove `spec.image` and use newer version)

@Crumby  Good catch. I updated the test to include this scenario and the initial fix didn't work. In order to handle this scenario, it was necessary to add an additional ".status.operand.customImage" field.

PR #2149, opened 2024-09-13
Fix creation of custom image clusters and correctly handles upgrades …

ryanemerson:


pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -89,22 +89,24 @@ func UpgradeRequired(i *ispnv1.Infinispan, ctx pipeline.Context) bool {
 			return false
 		}
 
-		// If the Operand is marked as a CVE base-image release, then we perform the upgrade as a StatefulSet rolling upgrade
-		// as the server components are not changed.
-		customImage := i.Status.Operand.CustomImage
-		if !customImage && requestedOperand.CVE && installedOperand.UpstreamVersion.EQ(*requestedOperand.UpstreamVersion) {
+		// If the Operand is marked as a CVE base-image release and no custom image is installed/requested, then we perform the
```
rigazilla, line 92: I'm missing something here?
the comment says 'perform upgrade' but the return value is `false`

rigazilla (approved):
seems ok to me.
custom image is stable
I've tested successfully this upgrade case:
custom->custom->default

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
ryanemerson: This method returns true if a Graceful/HR Rolling upgrade is required. In this case we turn false, as the `statefulset_updates.go` handler will trigger a StatefulSet rolling upgrade instead.

PR #2152, opened 2024-09-25
Cross-site fails with 3 or more sites

### LINKED ISSUE #1834: XSite fails with 3 or more sites on the same cluster
```yaml
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: keycloak-infinispan
spec:
  replicas: 3
  version: 14.0.9
  service:
    type: DataGrid
    sites:
      local:
        name: EUROPE
        expose:
          type: ClusterIP
        maxRelayNodes: 100
      locations:
        - name: ASIA
          url: infinispan+xsite://infinispan-cs.keycloak-asia.svc.cluster.local:7900
        - name: US
          url: infinispan+xsite://infinispan-cs.keycloak-us.svc.cluster.local:7900
  security:
    authorization:
      enabled: true
    endpointSecretName: connect-secret
    endpointEncryption:
      type: None
  logging:
    categories:
      org.infinispan: trace
      org.jgroups.protocols.TCP: error
      org.jgroups.protocols.relay.RELAY2: error
  container:
    cpu: 2000m:100m
    extraJvmOpts: -XX:NativeMemoryTracking=summary
    memory: 2Gi:256Mi
```

The user was able to workaround the issue by downgrading to the 2.2.x Operator,. Could this related to the reconciliation of the GossipRouter?

[operator.tar.gz](https://github.com/infinispan/infinispan-operator/files/12111431/operator.tar.gz)
[infinispan.tar.gz](https://github.com/infinispan/infinispan-operator/files/12111432/infinispan.tar.gz)

pruivo:
Sort by site name when creating the configuration XML file to ensure a stable hashing

Closes #1834

### DIFF
--- a/test/e2e/xsite/xsite_test.go
+++ b/test/e2e/xsite/xsite_test.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/iancoleman/strcase"
 	ispnv1 "github.com/infinispan/infinispan-operator/api/v1"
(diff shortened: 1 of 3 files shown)


test/e2e/xsite/xsite_test.go
```diff
@@ -467,6 +468,87 @@ func TestSuspectAndHearbeatConfig(t *testing.T) {
...
+
+	version := kube.GetStatefulSet(infinispan.Name, infinispan.Namespace).ObjectMeta.ResourceVersion
+
+	time.Sleep(10 * time.Second)
+
+	kube.WaitForInfinispanPods(3, tutils.SinglePodTimeout, infinispan.Name, namespace)
+	kube.WaitForInfinispanCondition(infinispan.Name, namespace, ispnv1.ConditionWellFormed)
+
+	otherVersion := kube.GetStatefulSet(infinispan.Name, infinispan.Namespace).ObjectMeta.ResourceVersion
+	assert.Equal(t, version, otherVersion)
```
pruivo, line 549: @ryanemerson is there a better way to test it?

test/e2e/xsite/xsite_test.go
```diff
@@ -467,6 +468,87 @@ func TestSuspectAndHearbeatConfig(t *testing.T) {
...
+func TestMultipleSites(t *testing.T) {
+	// single cluster cross-site is enough.
+	testName := tutils.TestName(t)
+	clientConfig := clientcmd.GetConfigFromFileOrDie(kube.FindKubeConfig())
+	kube := tutils.NewTestKubernetes(clientConfig.CurrentContext)
+	namespace := fmt.Sprintf("%s-xsite2", tutils.Namespace)
+
+	defer kube.CleanNamespaceAndLogOnPanic(t, namespace)
+
+	infinispan := &ispnv1.Infinispan{
```
ryanemerson, line 480: We can reduce the verbosity a fair bit here by using `tutils.DefaultSpec(t, testKube, func(i *ispnv1.Infinispan) {` to define the CR.

test/e2e/xsite/xsite_test.go
ryanemerson: I think you should check the StatefulSet `ObservedGeneration` is 1 as the StatefulSet should never be modified once the initial cluster has been created. Take a look at `spec_update_test` for an idea of how to retrieve and check this.

test/e2e/xsite/xsite_test.go
```diff
@@ -467,6 +468,87 @@ func TestSuspectAndHearbeatConfig(t *testing.T) {
 	expectHeartBeatConfiguration(t, testKubes["xsite2"], true, hbInterval, hbTimeout)
 }
 
+func TestMultipleSites(t *testing.T) {
```
ryanemerson, line 471: As we're not actually waiting for the xsite view to form, maybe we should link the issue as a comment and update the name to `TestMultipleSiteDefinitions`

test/e2e/xsite/xsite_test.go
pruivo: I'll try the generation... it is failing randomly locally :cry:

test/e2e/xsite/xsite_test.go
pruivo: I had to set `RUN_LOCAL_OPERATOR=false` locally to make it work with that spec. Let's see if I need to update the GHA too.

pruivo: Added to the GHA too. It was failing without it.

ryanemerson: I think that's because with the `DefaultSpec` function and `RUN_LOCAL_OPERATOR` we're apply the webhook defaults. I'm surprised setting `!RUN_LOCAL_OPERATOR` works though, as that should prevent the Operator from being started so I wouldn't expect any reconciliation in the test :thinking:

pruivo: Am I doing something wrong? The cross-site tests deploy the operator in the `kind` clusters: 
https://github.com/infinispan/infinispan-operator/blob/afdfd22dbb978e4590639115169e765414187f01/scripts/ci/configure-xsite.sh#L43

ryanemerson: Nope, you're doing exactly the right thing ... I have just forgot that detail :smile:

ryanemerson:
Thanks @pruivo

PR #2157, opened 2024-09-30
ISPN-16696 Configurable logging pattern

karesti:
https://issues.redhat.com/browse/ISPN-16696


karesti:
@ryanemerson pr to know if I'm in the good direction here

api/v1/infinispan_types.go
```diff
@@ -283,6 +283,7 @@ const (
 )
 
 type InfinispanLoggingSpec struct {
+	Pattern    string                      `json:"pattern"`
```
ryanemerson, line 286: We need `omitempty` so the `Pattern` element is only included in the resource output if not `""`

api/v1/types_util.go
```diff
@@ -531,6 +533,15 @@ func (ispn *Infinispan) GetLogCategoriesForConfig() map[string]string {
 	return copied
 }
 
+// GetLogPatternForConfig returns the log pattern or empty string
+func (ispn *Infinispan) GetLogPatternForConfig() string {
+	var pattern = DefaultLoggingPattern
+	if ispn.Spec.Logging != nil {
+		pattern = ispn.Spec.Logging.Pattern
+	}
+	return pattern
```
ryanemerson, line 542: No need for the `pattern` variable, we can just return the values directly.

api/v1/types_util.go
```diff
@@ -531,6 +533,15 @@ func (ispn *Infinispan) GetLogCategoriesForConfig() map[string]string {
 	return copied
 }
 
+// GetLogPatternForConfig returns the log pattern or empty string
+func (ispn *Infinispan) GetLogPatternForConfig() string {
+	var pattern = DefaultLoggingPattern
+	if ispn.Spec.Logging != nil {
```
ryanemerson, line 539: You also need to check that `ispn.Spec.Logging.Pattern != ""` as the current code will return an empty string if the user provides:

```yaml
  logging:
    categories:
      org.infinispan: debug
      org.jgroups: debug
```

api/v1/infinispan_types.go
ryanemerson: Please add the `// +optional` comment to this field.

ryanemerson (commented):
Good start @karesti :tada: 

As well as the changes, the next step is to create a test that ensures a custom pattern is applied to the ConfigMap. It would consist of:

1. Create Infinispan CR with 1 replica and a custom log pattern
2. Wait for the cluster to form (see TestBaseFunctionality)
3. Retrieve the `i.GetConfigName()` and check that the `log4j.xml` contains the expected log pattern.

karesti:
@ryanemerson I reported your changes and added docs and tests

test/e2e/infinispan/smoke_test.go
```diff
@@ -76,6 +77,32 @@ func TestBaseFunctionality(t *testing.T) {
 	}
 }
 
+// Test if the ConfigMap is applying the custom logging configuration
+func TestCustomLogging(t *testing.T) {
+	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
+
+	// Infinispan cluster definition with extra labels to be propagated to the service and pods
```
ryanemerson, line 84: Comment does not apply to this test can be removed.

test/e2e/infinispan/smoke_test.go
```diff
@@ -76,6 +77,32 @@ func TestBaseFunctionality(t *testing.T) {
...
+	replicas := 1
+	spec := tutils.DefaultSpec(t, testKube, func(i *ispnv1.Infinispan) {
+		i.Spec.Logging.Pattern = "custom_log_pattern_for_test"
+	})
+
+	// Create the cluster
```
ryanemerson, line 90: I think we can remove this, the following line of code is clear.

test/e2e/infinispan/smoke_test.go
```diff
@@ -76,6 +77,32 @@ func TestBaseFunctionality(t *testing.T) {
 	}
 }
 
+// Test if the ConfigMap is applying the custom logging configuration
+func TestCustomLogging(t *testing.T) {
+	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
+
+	// Infinispan cluster definition with extra labels to be propagated to the service and pods
+	replicas := 1
```
ryanemerson, line 85: The default value is 1 and we're only using this in the `WaitForInfinispanPods` check, so less verbose if we remove this variable and just do `WaitForInfinispanPods(1, ...)`

ryanemerson (commented):
Great work @karesti, just a few very minor comments.

test/e2e/infinispan/smoke_test.go
karesti: ops

karesti:
@ryanemerson I think I reviewed all the comments

ryanemerson:
I didn't notice the ISPN- ticket in the commit message before. We use GH issues for upstream operator work, so I created #2162 to track this and have updated the commit message accordingly.

Thanks for your first contribution to the Operator @karesti :tada:

PR #2158, opened 2024-09-30
TestBatchWithResources doesn't cleanup after itself

Crumby:
Test leaves running cluster after it ends


ryanemerson:
@Crumby Can you rebase on the latest `main` so that CI can run, thanks

ryanemerson:
Thanks @Crumby

PR #2169, opened 2024-10-09
spec.service.container.storage immutable. Fixes #2168

### LINKED ISSUE #2168: spec.service.container.storage should be immutable
Reconciliation for storage is not provided, so this field should be immutable

rigazilla:
Web hook checks for immutable spec.service.container.storage

solves #2168


ryanemerson:
Thanks @rigazilla

PR #2176, opened 2024-11-08
Fix status check expectation in TestSpecImage with explicit version

Crumby:
TestSpecImage expects to have latestVersion in the status but that fails when test is executed with older version


rigazilla:
thanks @Crumby !

PR #2177, opened 2024-11-12
Use cluster rather then single node in upgrade test

Crumby:
https://github.com/infinispan/infinispan-operator/pull/2147 changed the replicas count to of UpgradeTest to 1. This prevents discovery any of the upgrade issue that may be related to upgrading a cluster.

There's also an issue when upgrading from older Infinispan/Operator version that the CR may not enter the Pending state (CVE upgrade) which halts the tests. The issue doesn't happen with cluster upgrade and is specific to a particular version.


ryanemerson:
Thanks for fixing @Crumby, I merged this change by accident.

PR #2187, opened 2024-12-12
Next version 2.4.7

fax4ever:
I tested locally:

1. skaffold dev => the operator is deployed and seems to be working correctly with manual tests
2. make test = passes locally
3. make infinispan-test =>  passes locally

My knowledge and experience on these technologies are very limited at the moment, but I can say that in general I found everything very stable and solid.


ryanemerson:
Thanks for the PR.

The important thing to test when transitioning to a new version is that the [OLM](https://olm.operatorframework.io/) integration is working with the last released CSV (2.4.6) and the new version which we're transitioning to (2.4.7). The following E2E tests are executed with OLM:

- hotrod-rolling-upgrade
- upgrade
- webhook

So if you want to do a smoke test locally, the quickest method is to execute the webhook tests. Tbh though I usually just create a PR and wait for the actions as it's rare that this commit fails.

In this case the initial run of the upgrades failed, however it was related to a much older Operator version, so we know the root cause isn't related to this new version and so it's safe to simply retry the job. I'll merge this PR if the second attempt is green.

PR #2200, opened 2025-01-07
[#2199] Allow operand version to be explicitly set when calling test_upgrades workflow

### LINKED ISSUE #2199: Allow operand version to be explicitly set when calling test_upgrade workflow
Currently the `test_upgrades` workflow assumes that the Operand to be tested is always the latest. However, in order to allow different Infinispan branches to be tested correctly, we need to allow the Operand version to be explicitly passed so that the Operand is ordered correctly in `INFINISPAN_OPERAND_VERSIONS`.

ryanemerson:
…upgrades workflow

Closes #2199

### DIFF
--- a/.github/workflows/test_upgrades.yml
+++ b/.github/workflows/test_upgrades.yml
@@ -11,6 +11,10 @@ on:
         description: 'The name of the GH artifact containing the Operand image'
         required: false
         type: string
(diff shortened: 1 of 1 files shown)


fax4ever:
Merged, thanks!

PR #2204, opened 2025-01-09
[#2203] Run kind without olm if not needed

### LINKED ISSUE #2203: Kind GH action always installs kind with OLM
If `input.olm==false` we still call `kind-with-olm.sh`, whereas we should call `kind.sh`. Fixing this should save some time on the individual jobs that do not require OLM.

fax4ever:
Closes #2203


ryanemerson:
Thanks @fax4ever

fax4ever:
Thank you!

PR #2205, opened 2025-01-10
Document skip operands 15.1.0 and 15.1.1

fax4ever:


documentation/asciidoc/topics/ref_supported_versions.adoc
```diff
@@ -6,3 +6,11 @@ Since version 2.3.0 {ispn_operator} supports multiple {brandname} Server version
 You can upgrade the version of your cluster between supported {brandname} versions:
 
 include::supported_operands/operand_table.adoc[]
+
+[NOTE]
+====
+Operand versions `15.1.0` and `15.1.1` should be not used, using instead the version `15.1.2` or later if invalidation caches are present.
```
ryanemerson, line 12: `15.1.2` doesn't really exist, the release messed up, so the user should be going to `15.1.3.Final`.

I also think this effects users even if they weren't using invalidation caches explicitly, as they were previously contained within the server template file, so backup/restores can be problematic.

How about: "Operand versions `15.1.0` and `15.1.1` should be skipped when upgrading from the `15.0.x` stream, due to https://github.com/infinispan/infinispan/issues/13519[#13519] which may cause issues if invalidation caches are present or if attempting to restore from backups created with prior Infinispan versions."

documentation/asciidoc/topics/ref_supported_versions.adoc
```diff
@@ -6,3 +6,11 @@ Since version 2.3.0 {ispn_operator} supports multiple {brandname} Server version
 You can upgrade the version of your cluster between supported {brandname} versions:
 
 include::supported_operands/operand_table.adoc[]
+
+[NOTE]
```
ryanemerson, line 10: `[IMPORTANT]`?

documentation/asciidoc/topics/ref_supported_versions.adoc
fax4ever: Sure!

fax4ever:
Thanks for the review!

ryanemerson:
Thanks @fax4ever

fax4ever:
Thank you.

PR #2206, opened 2025-01-13
Next Version 2.4.10

fax4ever:
~~draft => still need https://quay.io/repository/operatorhubio/infinispan?tab=tags&tag=v2.4.9~~ done!


fax4ever:
I made the PR too early. The errors are related to the fact that the bundle is not already present:

https://quay.io/repository/operatorhubio/infinispan?tab=tags&tag=v2.4.9

I re-run the CI as soon as we have it.

test/e2e/utils/asserts.go
```diff
@@ -17,6 +17,7 @@ func LogError(err error) {
 // ExpectNoError checks if an error exists, and if so halts execution
 func ExpectNoError(err error) {
 	if err != nil {
+		LogError(err)
```
ryanemerson, line 20: I think this is redundant as we already log the error in `kubernetes.go#CleanNamespaceAndLogWithPanic` after the panic is thrown.

fax4ever: You're right. In fact, I have the same results. I'm removing it. Thanks

ryanemerson:
Thanks @fax4ever

fax4ever:
Thanks

Issue #2209, opened 2025-01-24
Automatically add release labels to PRs


ryanemerson:
#Description

Currently operator issues are not tagged with an Operator release that they are fixed in. When a PR is merged we should automatically label the associated issue it with the next Operator version that will contain the fix. Release labels  should use the following format `release/x.y.z`.

# Implementation ideas

Use the same approach as in the Infinspan repository: https://github.com/infinispan/infinispan/pull/13518

PR #2214, opened 2025-02-20
[#2209] Add label from version on merge

fax4ever:
This version is simplified wrt the Infinispan one, since we don't need to address merge other than main here.
Moreover the version is taken from the `version.txt`.

Labeled issue on my fork: https://github.com/fax4ever/infinispan-operator/issues/14

### DIFF
--- a/.github/workflows/label.yml
+++ b/.github/workflows/label.yml
@@ -1,0 +1,24 @@
+name: Labeller
+
+on:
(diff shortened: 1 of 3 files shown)


ryanemerson:
Thanks @fax4ever

fax4ever:
Thanks Ryan!

PR #2217, opened 2025-03-04
Set source channel to target channel in operand upgrades

Crumby:
The test assumes using latest Operator release but if source channel configuration differs from target channel then the test will fail on unknown CSV.


ryanemerson (approved):

ryanemerson:
Thanks @Crumby

PR #2226, opened 2025-03-13
Fail upgrade tests if one or more caches become DEGRADED before attempting a Backup or Restore

ryanemerson:


fax4ever:
This really helps! Thank you. Merged

PR #2231, opened 2025-03-18
Log client config error

fax4ever:
This helps to understand the problem in case the Kube client file cannot be used.


pkg/kubernetes/kubernetes.go
```diff
@@ -162,7 +163,8 @@ func (k Kubernetes) ExecWithOptions(options ExecOptions) (bytes.Buffer, error) {
 
 // FindKubeConfig returns local Kubernetes configuration
 func FindKubeConfig() string {
```
ryanemerson, line 165: This function is only ever called from test code, so we can move this to `test/e2e/utils/kubernetes.go`. It's also the only consumer of `DefaultKubeConfig`, so I would just make the method:

```go
// FindKubeConfig returns local Kubernetes configuration
func FindKubeConfig() string {
	config, exists := os.LookupEnv("KUBECONFIG")
	if exists {
		return config
	}
	homedir, err := os.UserHomeDir()
	ExpectNoError(err)
	return homedir + "/.kube/config"
}
```

I avoided using `consts.GetEnvWithDefault` as the above code will start pass if the tests are run in an environment where the home dir is not configured.

fax4ever: great idea thanks

fax4ever: It seems that `ExpectNoError(err)` is a test call. Isn't it?

ryanemerson: Yes, but if we move this function to `test/e2e/utils/kubernetes.go` that's not a problem

fax4ever: According to the call hierarchy it seems that the function is called only from the test code... maybe we can move it to there.

fax4ever: I said something wrong... it is used also my the main code. Maybe we can safely ignore an error of the call of `os.UserHomeDir()`, since with this implementation we only run it in case the `KUBECONFIG` system property is not set.

ryanemerson: > I said something wrong... it is used also my the main code.

Where? The only reference I see in production code is from the function's own comment

fax4ever: You're right... moving it :P

ryanemerson:
Thanks @fax4ever

fax4ever:
Thank you @ryanemerson

Issue #2230, opened 2025-03-14
GracefulShutdown upgrade may get stuck in between versions


Crumby:
There's a slight chance for 'TestOperandUpgrades' to fail in between individual version upgrades due to new StatefulSet version not being created.

**Operator log:**
[infinispan-operator-controller-manager-5776c64f76-l889k.log](https://github.com/user-attachments/files/19253863/infinispan-operator-controller-manager-5776c64f76-l889k.log)

**Test log:**
```
2025/03/14 19:22:43 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/caches/volatileCache?action=size
2025/03/14 19:23:06 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/container/health/status
2025/03/14 19:23:07 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/caches/persistentCache?action=size
2025/03/14 19:23:07 INFO Next Operand 15.0.3
2025/03/14 19:23:09 INFO Upgrading Operand to 15.0.3
2025/03/14 19:23:38 INFO Upgrade complete
2025/03/14 19:23:42 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/container/health/status
2025/03/14 19:23:42 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/caches/persistentCache?action=size
2025/03/14 19:23:55 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/caches/volatileCache?action=size
2025/03/14 19:24:14 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/container/health/status
2025/03/14 19:24:14 INFO GET: https://test-operand-upgrades-external-datagrid.apps.paris.rhdg-qe.rhocf-dev.net:443/rest/v2/caches/persistentCache?action=size
2025/03/14 19:24:14 INFO Next Operand 15.0.4
2025/03/14 19:24:16 INFO Upgrading Operand to 15.0.4
[...]

PR #2233, opened 2025-03-20
[#2230] Reconcile on unamanaged secret created 

Crumby:
It seems that when executing `TestOperandUpgrades` against OpenShift it's possible to get k8s (unreasonable) exponential back off for requeues beyond test time out when checking for service `*-cert-secret` generated by OpenShift.

The reconciliation eventually happens but it suddenly goes from seconds/minute to ~15 minutes.

This change also seems to halve upgrade time on OpenShift and overall test execution which is quite substantial for this particular test.

### DIFF
--- a/controllers/infinispan.go
+++ b/controllers/infinispan.go
@@ -11,6 +11,7 @@ import (
 	pipelineBuilder "github.com/infinispan/infinispan-operator/pkg/reconcile/pipeline/infinispan/pipeline"
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
@@ -110,15 +111,19 @@ func (r *InfinispanReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&infinispanv1.Infinispan{}).
 		Owns(&corev1.ConfigMap{}).
-		Owns(&corev1.Secret{}).
+		Owns(&corev1.Secret{},
+			builder.WithPredicates(
+				predicate.Funcs{
+					CreateFunc: func(ce event.CreateEvent) bool { return false },
+				},
+			),
+		).
 		Owns(&appsv1.Deployment{}).
 		Owns(&appsv1.StatefulSet{}).
 		WithEventFilter(predicate.Funcs{
 			CreateFunc: func(e event.CreateEvent) bool {
 				switch e.Object.(type) {
 				case *corev1.ConfigMap:
-					return false
-				case *corev1.Secret:
 					return false
 				case *appsv1.StatefulSet:
 					return false


pkg/reconcile/pipeline/infinispan/context/resources.go
```diff
@@ -128,7 +129,7 @@ func (r resources) load(name string, obj client.Object, load func() error, opts
 
 		if config.RetryOnErr {
 			// Set NotFound errors to nil so that the Operator logs are not populated with unuseful NotFound stacktraces
-			r.Requeue(client.IgnoreNotFound(err))
+			r.RequeueAfter(time.Second, client.IgnoreNotFound(err))
```
ryanemerson, line 132: I don't think it's a good idea to remove the exponential back off entirely as this could cause other problems. However, I looked at the `controller-runtime` code and it's possible to configure the max delay of the algorithm. The default is `1000s` :scream:

We can update `infinispan.go` to provide the following when defining the controller:

```go
return ctrl.NewControllerManagedBy(mgr).
		WithOptions(controller.Options{
			RateLimiter: workqueue.NewItemExponentialFailureRateLimiter(time.Millisecond, 10*time.Second),
		})...
```

IMO 1 second is probably too low as the max value, so I have gone with 10s in the above code. WDYT?

Crumby: While this should work as well, and I'll play around with this option, I can't get rid of the feeling that there's a logic behind the RateLimiters we don't fully understand and probably should.

[Default rate limiter](https://github.com/kubernetes/client-go/blob/v0.24.17/util/workqueue/default_rate_limiters.go#L39-L45) defines two limiters, the exponential one and bucket one. The later should limit amount of parallel request being processed and using the config above we'd drop that one.

What does not add up is the reconciler suddenly going straight for the 1000s delay as it should be increasing the delay exponentially but in if you check the logs there are no 1/2/5 minutes wait. And how does the delay being counted is a bit mystery to me but it seems as it should be doubling after each requeue per item but that could get pretty high during upgrades as we are constantly requeueing?

I'd say that those 10s depend on our proper understanding of the rate limiter in place and whether it's not too low in case you'd have a lot of cluster deployments where a lot of them could get into a invalid state and neverending reconcilation loop which could cause issues.

ryanemerson: > While this should work as well, and I'll play around with this option, I can't get rid of the feeling that there's a logic behind the RateLimiters we don't fully understand and probably should.

+1 that's very true, and something we need to explore.

So the upgrade tests executed on Openshift are automatically using the Openshift service based TLS, do we know if this backoff/wait issues happens to other resources, or is it only the `-cert-secret` created by Openshift?

If it's the latter, we could workaround this in a less invasive way by just defining a specific delay when waiting on the cert resource, i.e. update the `Keystore` function in `tls.go` to be:

```go
	if err := ctx.Resources().Load(i.GetKeystoreSecretName(), keystoreSecret); err != nil {
		ctx.RequeueAfter(time.Second, err)
		return
	}
```

> [Default rate limiter](https://github.com/kubernetes/client-go/blob/v0.24.17/util/workqueue/default_rate_limiters.go#L39-L45) defines two limiters, the exponential one and bucket one. The later should limit amount of parallel request being processed and using the config above we'd drop that one.

Good catch, if we do change the default RateLimiter we should probably include both e.g. something like:

```go
		WithOptions(controller.Options{
			RateLimiter: workqueue.NewMaxOfRateLimiter(
				workqueue.NewItemExponentialFailureRateLimiter(5*time.Millisecond, 10*time.Second),
				&workqueue.BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
			),
		}).
```

ryanemerson: You make some excellent points @Crumby and the more I think about it, the more hesitant I am to change the default `RateLimiter` as this is obviously used by many other Operator in production. We definitely shouldn't rush into this.

Crumby: One thing that makes debugging tricky is that there is multiple concurrent reconciliation events happening at the same for the same CR. They are not processed concurrently but even when there's a delay applied another event starts processing the CR through the same pipeline immediately. I wonder if they share the RateLimiter.

ryanemerson: Hmm that's a good point, we're explicitly adding a watch to `spec.security.endpointEncryption.certSecretName` in the case of Openshift service based TLS. So we're both requeing and receiving reconciliation events. Maybe we should experiment with removing the explicit requeue when the secret is not available or removing the watch when Openshift TLS is configured.

Crumby:
@ryanemerson I've updated the PR so that the reconciliation event:
* Isn't triggered on Infinispan.status update
* Is triggered when specified secret name is created

The first reduces amount of immediate reconciliation. The second makes the reconciler actually react on defined secret creation. This changes on it's own are enough to make the reconciliation process more predictable and upgrade test is no longer stalling on OpenShift.

If there's a reason why changes to Infinispan.status should still be listened to, let me know. I added full explanation of what and why it's happening to the issue itself.

I wanted to exclude updates to StatefulSet from the events listener as well as it's triggering a lot of extra reconciliation events on it's own but it introduces other flakiness.

ryanemerson:
Thanks for looking into this in such detail, much appreciated!

> If there's a reason why changes to Infinispan.status should still be listened to, let me know. I added full explanation of what and why it's happening to the issue itself.

We do check the value of the the various .status fields to determine whether an upgrade is in progress or not. So this might be a risk change. I noticed that one o the HR rolling upgrade tests failed on the first run, so I have restarted the job, but this could be an indicator that we still require the reconciliation event for Infinispan status updates.

> The second makes the reconciler actually react on defined secret creation. This changes on it's own are enough to make the reconciliation process more predictable and upgrade test is no longer stalling on OpenShift.

Let's see how the testsuite is, but maybe just adding this change is the less risky way to go for now?


> I wanted to exclude updates to StatefulSet from the events listener as well as it's triggering a lot of extra reconciliation events on it's own but it introduces other flakiness.

We also want the reconciliation to be triggered if users start modifying the StatefulSet themselves as we need the Operator to be the source of truth here.

controllers/infinispan.go
```diff
@@ -108,18 +109,16 @@ func (r *InfinispanReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 		r.log.Info("Defaults:", "Annotations", defaultAnnotations, "Labels", defaultLabels)
 	}
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&infinispanv1.Infinispan{}).
+		For(&infinispanv1.Infinispan{}, builder.WithPredicates(predicate.Or(predicate.GenerationChangedPredicate{}, predicate.AnnotationChangedPredicate{}, predicate.LabelChangedPredicate{}))).
 		Owns(&corev1.ConfigMap{}).
-		Owns(&corev1.Secret{}).
+		Owns(&corev1.Secret{}, builder.WithPredicates(predicate.Funcs{CreateFunc: func(ce event.CreateEvent) bool { return false }})).
```
ryanemerson, line 114: This is almost impossible to read :smile: 

Can you update the formatting to be like:

```go
		For(&infinispanv1.Infinispan{},
			builder.WithPredicates(
				predicate.Or(
					predicate.GenerationChangedPredicate{},
					predicate.AnnotationChangedPredicate{},
					predicate.LabelChangedPredicate{},
				),
			),
		).
		Owns(&corev1.ConfigMap{}).
		Owns(&corev1.Secret{},
			builder.WithPredicates(
				predicate.Funcs{
					CreateFunc: func(_ event.CreateEvent) bool { return false }},
			),
		).
```

ryanemerson:
Thanks @Crumby, great work :+1:

Crumby:
> We do check the value of the the various .status fields to determine whether an upgrade is in progress or not. So this might be a risk change. I noticed that one o the HR rolling upgrade tests failed on the first run, so I have restarted the job, but this could be an indicator that we still require the reconciliation event for Infinispan status updates.

Yup, so some of the HotRod Migration pipeline upgrades do not trigger `requeue` but rather rely on updating the status which triggers external event triggering the pipeline again. I presume the same applies for XSite tests as there is a lot of those failing as well.

> Let's see how the testsuite is, but maybe just adding this change is the less risky way to go for now?

Agreed, I kept only the lines affecting the secrets. Removed the create event from the filter and added it to `Own` so the Operator reconciler isn't triggered by secrets it's creating. (I assume this was original intention of the filter). The change on it's own is an improvement as it fixes use case where the user/openshift creates the secret after the Infinispan CR and it triggers the pipeline immediately rather then waiting up to max timeout or other event. From the root cause perspective, this fixes the symptom only. 

> We also want the reconciliation to be triggered if users start modifying the StatefulSet themselves as we need the Operator to be the source of truth here.

I'd say this is a topic for discussion and I'd like to have a separate issue opened for this one. There are two strategies we can adopt regarding general pipeline design: either we react to changes done to all the resources we manage (StatefulSets, Infinispan statuses, services, routes...) and we don't `requeue` on those updates or we react only to spec changes to resources we own CRDs for (and secrets/config maps...). Currently we do both: `requeue` and react to all changes which causes the root cause (unexpected long backoff from user in the controller without the exponential inclination). Just a change to a SS will cause several triggers due to spec changes and follow up status changes as SS is rolled out.

As for the Operator being source of truth, OpenShift will give warning to the user that they are modifying resource they do not own and it can have unpredictable results, other thing is, are we really able to fix any modification to the SS done by user? Generally I think we should adopt one or another approach to the pipeline, and use that but having both will lead to this kind of unpredictable behaviors.

ryanemerson:
> I'd say this is a topic for discussion and I'd like to have a separate issue opened for this one.

Please do :slightly_smiling_face:

I agree with your assessment of the situation and we should chose one strategy and stick with it. My gut feeling is that we probably should go with the latter approach of "react only to spec changes to resources we own CRDs for (and secrets/config maps..."

PR #2236, opened 2025-03-21
[#2229] Apply the versions on templates

### LINKED ISSUE #2229: Utilise template snippets to compose version specific configuration files
#2227 introduced several new *.xml files for the Infinsipan servers. The difference between these templates are minimal, with the primary difference being schema version numbers:

```bash
pkg/templates/templates
├── infinispan-admin-14-0.xml
├── infinispan-admin-15-0.xml
├── infinispan-admin-15-1.xml
├── infinispan-admin-15-2.xml
├── infinispan-base-14-0.xml
├── infinispan-base-15-0.xml
├── infinispan-base-15-1.xml
├── infinispan-base-15-2.xml
├── infinispan-zero-14-0.xml
├── infinispan-zero-15-0.xml
├── infinispan-zero-15-1.xml
├── infinispan-zero-15-2.xml
└── log4j.xml
```

To reduce the overhead of maintaining this duplication, we should create smaller template snippets of common configuration that can be combined to create the overall config.

fax4ever:
Closes #2229


pkg/templates/templates/infinispan-base-15.xml
```diff
@@ -1,9 +1,9 @@
 <infinispan
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="urn:infinispan:config:15.0 https://infinispan.org/schemas/infinispan-config-15.0.xsd
-                        urn:infinispan:server:15.0 https://infinispan.org/schemas/infinispan-server-15.0.xsd"
-    xmlns="urn:infinispan:config:15.0"
-    xmlns:server="urn:infinispan:server:15.0">
+    xsi:schemaLocation="urn:infinispan:config:{{ .Infinispan.Version }} https://infinispan.org/schemas/infinispan-config-{{ .Infinispan.Version }}.xsd
```
ryanemerson, line 3: The amount of code we've reduced is great, but I think we can remove more whilst also providing a solution which is more flexible for when we need to make version specific tweaks to the configuration files.

I think we should utilise the `{{ template "..." }}` functionality to make composable configurations.

For example, the schema header is the same for all configurations, so we could have a `header.xml` like:
```go
{{with $v := printf "%d.%d" .Infinispan.Version.Major .Infinispan.Version.Minor}}
<infinispan
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
        xsi:schemaLocation="urn:infinispan:config:{{$v}} https://infinispan.org/schemas/infinispan-config-{{$v}}.xsd
                        urn:infinispan:server:{{$v}} https://infinispan.org/schemas/infinispan-server-{{$v}}.xsd"
        xmlns="urn:infinispan:config:{{$v}}"
        xmlns:server="urn:infinispan:server:{{$v}}">
{{end}}
```

Then `infinispan-base-14.xml` would be something like:

```go
{{template "header" . }}
<cache-container name="default" statistics="true">
...
```

We can then take this further and extract common xml from the server content into individual templates. For example, the security-realms config is the same for both 14 and 15, so we could have that in a separate template which is then included in `infinispan-base-14.xml` and `infinispan-base-15.xml`.

For this to work, we need to slightly modify `templtes.go` so that we include multiple templates, we can do this like:


```go
tpl, err := template.New(templateName).Funcs(tplFunctions).ParseFS(content, "templates/" + templateName, "templates/common/*.xml")
```

pkg/infinispan/configuration/server/configuration.go
```diff
@@ -95,15 +97,26 @@ type Endpoints struct {
 	ClientCert   string
 }
 
+type Version struct {
```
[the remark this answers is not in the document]
fax4ever: What about `JGroups.Version` ? Shall we use `*semver.Version` also there? I guess so.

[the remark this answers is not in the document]
fax4ever: I added a commit with the change, that we can squash before the merge.

pkg/templates/templates/infinispan-base-15.xml
fax4ever: It is a great idea. Thanks. I'm applying it.

pkg/infinispan/configuration/server/configuration_test.go
```diff
@@ -0,0 +1,29 @@
...
...che-container name=\"default\" statistics=\"true\">\n    <global-state unclean-shutdown-action=\"IGNORE\" />\n    \n    <security>\n        <authorization group-only-mapping=\"false\">\n            \n            <cluster-role-mapper />\n            \n            \n        </authorization>\n    </security>\n    \n    <transport cluster=\"${infinispan.cluster.name:}\" node-name=\"${infinispan.node.name:}\" stack=\"image-tcp\"    \n    />\n</cache-container>\n<server xmlns=\"urn:infinispan:server:15.1\">\n    <socket-bindings default-interface=\"public\" port-offset=\"${infinispan.socket.binding.port-offset:0}\">\n        <socket-binding name=\"default\" port=\"${infinispan.bind.port:11222}\"/>\n    </socket-bindings>\n    <security>\n        \n        \n        <security-realms>\n            <security-realm name=\"default\">\n                <server-identities>\n\t\t\t\t\n                </server-identities>\n                \n            </security-realm>\n            \n        </security-realms>\n    </security>\n    <endpoints>\n        <endpoint socket-binding=\"default\" security-realm=\"default\" require-ssl-client-auth=\"true\" />\n    </endpoints>\n</server>\n</infinispan>\n"
```
ryanemerson, line 12: Can we load this from a file?

It's currently almost impossible to read and we will need to update these values in the future when we inevitably have to modify the config.

pkg/templates/templates.go
```diff
@@ -48,8 +48,7 @@ func LoadAndExecute(templateName string, funcMap template.FuncMap, data interfac
 		tplFunctions[key] = value
 	}
 
-	tplStr := templates[templateName]
-	tpl, err := template.New(templateName).Funcs(tplFunctions).Parse(tplStr)
+	tpl, err := template.New(templateName).Funcs(tplFunctions).ParseFS(content, "templates/"+templateName, "templates/common/*.xml")
```
ryanemerson, line 51: We can now remove the `templates` map and the `init()` method as the former's content is no longer used.

pkg/infinispan/configuration/server/configuration.go
```diff
@@ -147,3 +154,17 @@ func funcMap() template.FuncMap {
 func supportedMajorVersion(v uint64) bool {
 	return v >= 14 && v <= 15
 }
+
+func mapVerstionsToSpec(operand version.Operand, spec *Spec) {
```
ryanemerson, line 158: It's probably beyond the scope of this PR, but I'm wondering if we should add the JGroups version to the Operand metadata provided by `INFINISPAN_OPERAND_VERSIONS`? That way we can handle the schema version automatically without requiring the if/else logic and it means we can move to a new JGroups version without having to make changes to the Operator code. It could also provide benefits for rolling upgrades as the Operator could take into account the JGroups version when allowing/denying a rolling upgrade.

ryanemerson (commented):
The use of `{{ template }} ` is looking good :slightly_smiling_face: 

For the 14.x.y CloudEvents schema, I think we can reuse the common header if we just update `infinispan-base-14.xml` to define the xlmns in place, something like:

```xml
<urn:infinispan:config:cloudevents:14.0 xsi:schemaLocation="urn:infinispan:config:cloudevents:14.0 https://infinispan.org/schemas/infinispan-cloudevents-config-14.0.xsd" ... >
```

I also think we can go further with the code reuse. The `<jgroups>` element for `infinispan-admin.xml` and `infinispan-zero.xml` is the same, as is the `<security`> element for `infinispan-base-15.xml` and `infinispan-base-14.xml`.

pkg/infinispan/configuration/server/configuration_test.go
```diff
@@ -0,0 +1,53 @@
...
+	baseCfg_expected := ReadFile("test/base-14-cfg.xml")
+	adminCfg_expected := ReadFile("test/admin-14-cfg.xml")
+
+	ispn := Infinispan{Authorization: &Authorization{Enabled: true}}
+	spec := Spec{Infinispan: ispn}
+	vers := semver.Version{Major: 14, Minor: 0, Patch: 11}
+	ope := version.Operand{UpstreamVersion: &vers}
+
+	baseCfg, adminCfg, err := Generate(ope, &spec)
+	assert.Equal(t, baseCfg, baseCfg_expected)
```
ryanemerson, line 42: In the various `assert.Equal` calls we have the expected and actual arguments in the wrong order, it should be `Equal(t, expected, actual)`.

pkg/infinispan/configuration/server/configuration.go
fax4ever: I agree with you. We need a declarative way to address it. I opened the followup.
Also, I'm adding the link to the follow up in a TODO.
What do you think of it?

pkg/infinispan/configuration/server/test/base-14-cfg.xml
```diff
@@ -0,0 +1,49 @@
...
+    xmlns:server="urn:infinispan:server:14.0">
+
+<cache-container name="default" statistics="true">
+    
+    <security>
+        <authorization>
+            
+            <cluster-role-mapper />
+            
+
```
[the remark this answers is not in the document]
fax4ever: I agree. Now that we produce the expected configurations we can clearly see those white spaces. Sure, let's use `-` to fix it. Thanks

pkg/infinispan/configuration/server/configuration.go
ryanemerson: Thanks. I've added more information to the issue to try to make it clearer what we want to achieve and why those changes are beneficial.

fax4ever:
Thanks @ryanemerson 
I should have addresses most changes here. I don't know if we can do something else

ryanemerson:
> I also think we can go further with the code reuse. The `<jgroups>` element for `infinispan-admin.xml` and `infinispan-zero.xml` is the same, as is the `<security`> element for `infinispan-base-15.xml` and `infinispan-base-14.xml`.

This is still the case,  we remove this duplication.

pkg/infinispan/configuration/server/configuration_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"os"
+	"testing"
+
+	"github.com/blang/semver"
+	"github.com/infinispan/infinispan-operator/pkg/infinispan/version"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGenerateAndGenerateZeroCapacity(t *testing.T) {
+	baseCfg_expected := ReadFile("test/base-15-cfg.xml")
```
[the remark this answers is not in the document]
fax4ever: TBH I would prefer to keep them, since they can provide a good insights IMO about what we're actually producing as configurations. The changes probably will be happen only when we introduce different versions... But as you prefer.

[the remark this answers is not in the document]
ryanemerson: I'm happy to keep them as well. If they become a pain to maintain in the future we can always re-evaluate.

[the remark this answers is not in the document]
fax4ever: Sounds a great plan

fax4ever:
Thank you @ryanemerson I've done with the last changes.

ryanemerson (commented):
Thanks for the changes @fax4ever. I started to review and then realised that I was making most of the changes locally as I went, so I have just pushed a commit to save you the hassle. I have left the comments that I had already wrote in the review for context. If you disagree with any of the changes, please say :slightly_smiling_face:

pkg/infinispan/configuration/server/configuration.go
```diff
@@ -119,31 +125,43 @@ func GenerateZeroCapacity(operand version.Operand, spec *Spec) (string, error) {
...
-				ret += fmt.Sprintf("%s[%d]", bs.Address, bs.Port)
-			}
-			return ret
-		},
+	mapVerstionsToSpec(operand, spec)
+
+	return templates.LoadAndExecute("infinispan-zero.xml", spec)
+}
+
+func remoteSites(elems []BackupSite) string {
```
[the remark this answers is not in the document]
fax4ever: Sure! I'm not sure that the method is on the `configuration.go` file...

fax4ever:
> Thanks for the changes @fax4ever. I started to review and then realised that I was making most of the changes locally as I went, so I have just pushed a commit to save you the hassle. I have left the comments that I had already wrote in the review for context. If you disagree with any of the changes, please say 🙂

Thanks @ryanemerson. The changes look good to me. I have just a doubt about the `XSite` method.

ryanemerson:
Whoops, I guess I didn't add that locally. I have pushed a new commit

ryanemerson:
Thanks @fax4ever

fax4ever:
Thank you @ryanemerson

PR #2242, opened 2025-04-07
[#2243] Avoid to use inline script with template

fax4ever:
Closes #2243

(Maybe --- or another similar issue)

Issue #2243, opened 2025-04-07
Remove use of internal server templates from code and documentation


fax4ever:
[ISPN-13839](https://issues.redhat.com/browse/ISPN-13839) removed the various `org.infinispan.*` configuration templates from the server in 15.1.0, however we still rely on these templates in parts of the operator test code and our documentation.

We should update the code and all documentation examples to avoid using these templates.

Issue #2234, opened 2025-03-20
Flaky Test: Upgrade tests failing as no entries exist in the persistentCache after upgrade


ryanemerson:
The `upgrade_operands_test.go` is sometimes failing because the number of entries in the "persistentCache" is 0 after an upgrade, when it should be 100 as the cache is configured to use a file-store.

On a GracefulShutdown upgrade, the StatefulSet of the original cluster is deleted but the associated PVCs are kept and a new StatefulSet is created for the new cluster containing the upgraded Infinispan server. The expectation is that the data associated with the cache is written to the PVC and so after cluster upgrade all persisted entries should remain available via the cache.

```go
Next Operand 15.0.13
Upgrading Operand to 15.0.13
GET: http://172.172.0.2:30552/rest/v2/container/health/status
GET: http://172.172.0.2:30552/rest/v2/caches/persistentCache?action=size
Encountered error: previous terminated container "manager" in pod "infinispan-operator-controller-manager-5d85f8984b-9dtqp" not found
Encountered error: previous terminated container "infinispan" in pod "test-operand-upgrades-0" not found
Encountered error: previous terminated container "infinispan" in pod "test-operand-upgrades-1" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-13" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-17" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-19" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-20" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-21" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-24" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-27" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-32" not found
Encountered error: previous terminated container "infinispan" in pod "14-0-9" not found
Encountered error: previous terminated container "infinispan" in pod "15-0-0" not found
Encountered error: previous terminated container "infinispan" in pod "15-0-10" not found
Encountered error: previous terminated container "infinispan" in pod "15-0-11" not found
Encountered error: previous terminated container "infinispan" in pod "15-0-3" not found
Encountered error: previous terminated container "infinispan" in pod "15-0-4" not found
Encountered error: previous terminated container "infinispan" in pod "15-0-5" not found
Encountered error: previous terminated container "infinispan" in pod "15-0-8" not found
[...]

PR #2244, opened 2025-04-07
[#2230] Limit backup scope

fax4ever:
Based on top of https://github.com/infinispan/infinispan-operator/pull/2242
Try to address #2230

### DIFF
--- a/test/e2e/upgrade/common.go
+++ b/test/e2e/upgrade/common.go
@@ -94,6 +94,12 @@ func createBackupAndWaitToSucceed(name string, t *testing.T) *v2.Backup {
 		},
 		Spec: v2.BackupSpec{
 			Cluster: name,
+			Resources: &v2.BackupResources{
+				Caches:       []string{volatileCacheName},
+				CacheConfigs: []string{},
+				Templates:    []string{},
+				ProtoSchemas: []string{},
+			},
 		},
 	}
 	testKube.Create(backup)


test/e2e/upgrade/common.go
```diff
@@ -94,6 +95,12 @@ func createBackupAndWaitToSucceed(name string, t *testing.T) *v2.Backup {
 		},
 		Spec: v2.BackupSpec{
 			Cluster: name,
+			Resources: &v2.BackupResources{
+				Caches:       []string{volatileCacheName},
```
ryanemerson, line 99: This can be simplified to:

```go
		Spec: v2.BackupSpec{
			Cluster: name,
			Resources: &v2.BackupResources{
				Caches: []string{volatileCacheName},
			},
		},
```

ryanemerson (commented):
Thanks @fax4ever.

I looked into the original content of the backup and it seems we were storing all of the internal templates (since deprecated and ignored from backups) in 14.0.8, but also we were backing up the persistent cache which obviously shouldn't have a need to be backed up!

So I think your changes to the Backup spec make sense and I don't think we're missing an important use-case with your changes.

I think we should skip the #2243 changes for now though as the Operator needs to be updated to not use templates in other areas of the code as well as in the documentation. I believe this is a larger effort that's out of scope here.

Crumby:
> ... but also we were backing up the persistent cache which obviously shouldn't have a need to be backed up!

Sounds like solving an issue by removing a use case from the test. What's the expected behavior when user restores (persisted or non-persisted ) cache against a cluster that already has the mentioned cache defined and with data? My gut feeling tells me that expected behavior would be to bring the cache to the state in which the cache was when backup was created.

ryanemerson:
> My gut feeling tells me that expected behavior would be to bring the cache to the state in which the cache was when backup was created.

All cache entries contained with a backup are written to the cache. If the cache contains subsequent K/V not in the Backup, then these remain in place. If the user wants the Backup content to be the only content then they need to clear the cache before executing a restore.

In the test we're not modifying the content between the Backup and the restore, so the content from the file-store and the Backup should be identical, which is why I don't think including the persistent cache in the backup makes sense.

Crumby:
> In the test we're not modifying the content between the Backup and the Restore, so the content from the file-store and the Backup should be identical, which is why I don't think including the persistent cache in the backup makes sense.

It should but the original issue is that the cache content disappears. Given it's fixed by not restoring the persistent cache content from backup it can be assumed there is an incorrect interaction when running backup against cache with a content making it a valid use case. Maybe not exactly in the upgrade tests but it should be investigated and subsequent issue opened so it's not forgotten

ryanemerson:
> It should but the original issue is that the cache content disappears. Given it's fixed by not restoring the persistent cache content from backup it can be assumed there is an incorrect interaction when running backup against cache with a content making it a valid use case. Maybe not exactly in the upgrade tests but it should be investigated and subsequent issue opened so it's not forgotten

Do we know that specific issue is in fact fixed by this change? My understanding is that this change makes the Backup/Restore more stable but unfortunately there are a few observed causes of said failures.


> Maybe not exactly in the upgrade tests but it should be investigated and subsequent issue opened so it's not forgotten

+1 I agree we still need to know the root cause of this failure.

test/e2e/upgrade/common.go
fax4ever: Yeah my idea was to explicitly pass empty lists to be 100% sure about the fact to not import other resources... Anyway, I applied the suggestion. Thanks

Crumby:
> Do we know that specific issue is in fact fixed by this change?

From my understanding we are limiting the backup to only volatile cache, which mean the issua described by #2234 will seemingly disappear but we'll just remove the combination from the test.

ryanemerson:
I don't think that's definitely the case, as the test is failing when we assert the cache size *before* restoring the backup. The test currently verifies the persistenct cache size before and after the Restore is executed for each Operand upgrade. So it's possible the issue isn't related to backup/restore at all.

test/e2e/upgrade/common.go
ryanemerson: I don't think you did :stuck_out_tongue_winking_eye:

Crumby:
> I don't think that's definitely the case, as the test is failing when we assert the cache size before restoring the backup. 

You are right, haven't noticed the issue happening before the actual restore operation.

Still with this PR we are simplifying backup operation to a single cache, reducing the complexity. We may lose detection of any issues caused by incompatibility in internal caches as well.

Nevertheless I understand the need to have the upgrade tests stable and some of the flaky behavior may needs a lot of invested time in investigation. I'd just want to ensure that we are not going just to hide the issue by removing part of the test case.

Crumby:
Let's go ahead with option no 1.

ryanemerson:
Thanks @fax4ever

fax4ever:
Thank you!

PR #2246, opened 2025-04-11
[#2222] Utilize Logging framework in tests

### LINKED ISSUE #2222: Utilize Logging framework in tests
All the tests currently write directly to standard output. Migrating the message to any logging framework would help with debugging and it'd reduce overall clutter generated by tests.

**Logging Requirements:**
* Time
* Configurable Level (Debug, Info)
* File Source

**Frameworks for consideration:**
* go-logr/zap - used in the main code
* slog - Structured Logging introduced as a standard library in Go 1.21

Crumby:
Closes #2222

Utilize Zap Logger in the tests.

### DIFF
--- a/test/e2e/utils/constants.go
+++ b/test/e2e/utils/constants.go
@@ -8,6 +8,7 @@ import (
 
 	ispnv1 "github.com/infinispan/infinispan-operator/api/v1"
 	"github.com/infinispan/infinispan-operator/controllers/constants"
+	"go.uber.org/zap/zapcore"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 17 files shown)


test/e2e/utils/constants.go
```diff
@@ -23,6 +23,8 @@ var (
 
 	LogOutputDir = constants.GetEnvWithDefault("TESTING_LOG_DIR", os.TempDir()+"/infinispan-operator")
 
+	Debug, _ = strconv.ParseBool(constants.GetEnvWithDefault("DEBUG", "false"))
```
ryanemerson, line 26: This would be more flexible if we allow the log level to be specified via the env variable, something like:

```go
var (
...
        LogLevel     = logLevel()
	LogOutputDir = constants.GetEnvWithDefault("TESTING_LOG_DIR", os.TempDir()+"/infinispan-operator")
...
)

func logLevel() zapcore.Level {
	level, err := zapcore.ParseLevel(constants.GetEnvWithDefault("LOG_LEVEL", "info"))
	ExpectNoError(err)
	return level
}
```

Then we can replace `if Debug` with `if Log().Level().Enabled(zapcore.DebugLevel) {...}`

ryanemerson (changes requested):

test/e2e/utils/constants.go
Crumby: That's basically what I had initially, then I came across `DEBUG` option in the `httpClient` and it go me thinking. At the moment ZAP doesn't include possibility to TRACE with debug being the lowest level, so you have by INFO default and you can go only for DEBUG. And I didn't see a use case for enabling only WARN levels and above either. (And running `DEBUG=true go test -v ...` was a bit faster compared to `TESTING_LOG_LEVEL=debug go test -v ...`)

Neverthless I'm pretty happy implement the suggestion as it's a valid solution and should ZAP introduce TRACE logging that it'd make even more sense

Crumby:
Example output with Operator running on the cluster:

ryanemerson:
Thanks @Crumby, this is a big improvement :+1:

PR #2251, opened 2025-04-16
[#2250] Fix TestSpecImage tests using same minor

### LINKED ISSUE #2250: Fix TestSpecImage tests using same minor

fax4ever:
Closes #2250

### DIFF
--- a/test/e2e/infinispan/upgrade_operand_test.go
+++ b/test/e2e/infinispan/upgrade_operand_test.go
@@ -289,15 +289,17 @@ func TestOperandCVEHotRodRolling(t *testing.T) {
 func TestSpecImage(t *testing.T) {
 	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
 
-	// Create Infinispan Cluster using the latest Operand version, but using the image of the preceding Operand
-	// Ensures that creating a cluster with an initial spec.image value does not cause infinite StatefulSet updates
-	versionManager := tutils.VersionManager()
-	operand := versionManager.Operands[len(versionManager.Operands)-2]
+	// Create Infinispan Cluster using a newer Operand version (versionOperand),
+	// but using the image of an older Operand (imageOperand).
+	// Ensures that creating a cluster with an initial spec.image value does not cause infinite StatefulSet updates.
+	// The two version are chosen to be the last possible versions having the same major and minor.
+	imageOperand, versionOperand := specImageOperands()
 	spec := tutils.DefaultSpec(t, testKube, func(i *ispnv1.Infinispan) {
-		i.Spec.Image = pointer.String(operand.Image)
+		i.Spec.Image = pointer.String(imageOperand.Image)
+		i.Spec.Version = versionOperand.Ref()
 	})
 
-	statusVersionRef := versionManager.Latest().Ref()
+	statusVersionRef := versionOperand.Ref()
 	if tutils.OperandVersion != "" {
 		statusVersionRef = tutils.OperandVersion
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/infinispan/upgrade_operand_test.go
```diff
@@ -405,3 +404,15 @@ func TestSpecImageUpdate(t *testing.T) {
...
+	versionManager := tutils.VersionManager()
+	latest := versionManager.Latest()
+	latestMinus1 := *versionManager.Operands[len(versionManager.Operands)-2]
+	if (latest.UpstreamVersion.Major == latestMinus1.UpstreamVersion.Major) && (latest.UpstreamVersion.Minor == latestMinus1.UpstreamVersion.Minor) {
+		return latestMinus1, latest
+	}
+
+	latestMinus2 := *versionManager.Operands[len(versionManager.Operands)-3]
+	return latestMinus2, latestMinus1
+}
```
ryanemerson, line 418: I think we need a loop here. Currently this works because the operand list is:

```json
                  {
                    "upstream-version": "15.1.5",
                    "image": "quay.io/infinispan/server:15.1.5.Final"
                  },
                  {
                    "upstream-version": "15.1.7",
                    "image": "quay.io/infinispan/server:15.1.7.Final"
                  },
                  {
                    "upstream-version": "15.2.1",
                    "image": "quay.io/infinispan/server:15.2.1.Final"
                  }
```
and we're returning `15.1.5` and `15.1.7`.

However, when we add `15.2.2` we will return `15.1.7` and `15.2.2` which have different minor versions and so the test will fail.

We need a loop/retry mechanism so that we ensure that the two operands are in the same major.minor stream.

test/e2e/infinispan/upgrade_operand_test.go
```diff
@@ -292,13 +292,13 @@ func TestSpecImage(t *testing.T) {
 
 	// Create Infinispan Cluster using the latest Operand version, but using the image of the preceding Operand
 	// Ensures that creating a cluster with an initial spec.image value does not cause infinite StatefulSet updates
-	versionManager := tutils.VersionManager()
-	operand := versionManager.Operands[len(versionManager.Operands)-2]
+	imageOperand, versionOperand := specImageOperands()
```
ryanemerson, line 295: The comment needs to be updated to reflect that we're not using the latest operand versions, but rather two operands from the same major.minor stream.

test/e2e/infinispan/upgrade_operand_test.go
```diff
@@ -322,18 +322,17 @@ func TestSpecImageUpdate(t *testing.T) {
 	// release. We can then manually specify spec.image using the FQN of the latest image to simulate a user specifying
 	// custom images
 	replicas := 1
-	versionManager := tutils.VersionManager()
-	operand := versionManager.Operands[len(versionManager.Operands)-2]
+	imageOperand, versionOperand := specImageOperands()
```
ryanemerson, line 325: Maybe we can make these variable names more informative, how about "firstOperand, secondOperand"? I don't think "image" and "version" have much meaning in the context of this test.

The comment also needs updating.

test/e2e/infinispan/upgrade_operand_test.go
fax4ever: When we add `15.2.2` are you sure we won't return ( `15.2.1` `15.2.2` ) ?

ryanemerson: Yes, my mistake. I miss-read the code.

fax4ever:
Rebased and applied the changed. Thanks for the review @ryanemerson

Issue #2254, opened 2025-04-18
Missing secretName on create Ingress


Hijbk304:
When deploying to a regular k8s cluster with Ingress nginx using the following Infinispan configuration:

```yaml
infinispan:
  replicas: 2
  expose:
    type: Route
    host: infinispan-review.example.dev
  security:
    endpointEncryption:
      type: Secret
      certSecretName: infinispan-review-example-dev
....
```
The operator creates an Ingress object but does not add the secretName. I think this can be fixed by adding one line in this section
```diff
index 7899fe3d..46fb2f29 100644
--- a/pkg/reconcile/pipeline/infinispan/handler/provision/services.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/provision/services.go
@@ -259,7 +259,8 @@ func defineExternalIngress(i *ispnv1.Infinispan, ctx pipeline.Context) {
                if i.IsEncryptionEnabled() {
                        ingress.Spec.TLS = []ingressv1.IngressTLS{
                                {
-                                       Hosts: []string{i.Spec.Expose.Host},
+                                       Hosts:      []string{i.Spec.Expose.Host},
+                                       SecretName: i.Spec.Security.EndpointSecretName,
                                },
                        }
                }

```
https://github.com/infinispan/infinispan-operator/blob/6ff6925df4b2edcfda50e96a2108b0bb81dcb489/pkg/reconcile/pipeline/infinispan/handler/provision/services.go#L260

ryanemerson:
Thanks for raising the issue @Hijbk304. The secret referenced should be `i.GetKeystoreSecretName()`. Could you create a PR?

PR #2257, opened 2025-04-29
Add missing secretName for Ingress

Hijbk304:
Add missing secretName on create Ingress. For #2254


Crumby:
@Hijbk304 @ryanemerson I don't think the solution here will be as simple as adding the `secretName` to the `Ingress` specs. According [Ingress TLS documenationt](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls) it requires for the secret to have data in `tls.crt` and `tls.key` format. And while we support that we also support `keystore.p12` format which wouldn't work.

The second and probably more important issue is that Ingress with TLS terminates the encryption and forwards plain text but the server is still configured for the encrypted communication. I suspect the PR as is doesn't work but I didn't get a chance to test it and we run the tests with NodePort configuration in the CI/CD only.

PR #2258, opened 2025-04-30
[test-integration] Update check labels on OCPCertsIT affinity test

Crumby:
`OCPCerts#$antiAffinityTest` will fail on k8s clusters with workers count > 3 as it does not count with listener pod being included in pod filter (getting 4 different nodes instead expected 3).


Crumby:
@ryanemerson Yes, added, thanks

Crumby:
@ryanemerson looks like backport wasn't created

ryanemerson:
@Crumby The backport action needs updating to include the 2.4.x branch: https://github.com/infinispan/infinispan-operator/blob/main/.github/workflows/backport.yaml

PR #2263, opened 2025-05-08
Improve cross-site encryption documentation

### LINKED ISSUE #2262: Improve cross-site encryption documentation
* Document how to create the key store and trust stores using keytool. 
* Validate the Infinspan CR. The router Pod won't start without a Secret when the Infinispan CR with `spec.service.sites.local.expose.type` Route.

pruivo:
Closes #2262


tux-o-matic:
Adding a check and feedback in the webhook is a nice touch, thanks.

PR #2264, opened 2025-05-12
Update version to 2.5.0 ready for Infinispan 16.0

ryanemerson:
@Crumby @pminz This is necessary as we should release the Operator as 2.5.0.CSV when 16.0.0.Final is out. It's also necessary now so that issues are correctly labelled with 2.5.0 by the GH actions when merged.

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 TEST_BUNDLE=${1-main}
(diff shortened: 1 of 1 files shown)


Crumby:
@ryanemerson Looks like we'll need to increase timeout for upgrade tests as the list of the versions got quite long now. Also, is there a reason not to built Server image from 16.0.0.Dev01 and include it in the list so the continuous testing may continue?

Crumby (approved):

PR #2272, opened 2025-05-15
[#2269] GracefulShutdown upgrades should tolerate pods that have already been stopped

### LINKED ISSUE #2269: GracefulShutdown upgrades should tolerate pods that have already been stopped
# Problem
During a GracefulShutdown the Operator does the following:

1. Ping the server to determine the version so that we can create the correct client
2. Disable global rebalancing via the REST endpoint by calling the `-0` pod
3. For each pod in the cluster, call the shutdown endpoint

If the Operator progress is interrupted* during step 3, subsequent attempts to perform the GracefulShutdown could fail as the pod's cache-container has already been shutdown.

Furthermore, the pod list is not guaranteed to be in the same order each time which adds additional non-determinism.

\*Progress maybe interrupted due to the Operator pod being restarted/rescheduled, or an unexpected error from the server.

# Solution
1. When attempting the `GracefulShutdown` we should continue to the next pod if it returns an error response indicating the cache-container has already been stopped. We should output an appropriate log indicating that the pod has already been stopped.
2. We should make sure that all error logs associated with pod specific requests include the name of the pod to ease debugging in the future.
3. We should ensure that the order of pod names returned by `ctx.InfinispanPods()` is deterministic, sorted from the lowest to highest ordinal.

ryanemerson:
Closes #2269

### DIFF
--- a/test/e2e/infinispan/upgrade_operand_test.go
+++ b/test/e2e/infinispan/upgrade_operand_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	ispnv1 "github.com/infinispan/infinispan-operator/api/v1"
+	"github.com/infinispan/infinispan-operator/pkg/infinispan/client"
 	"github.com/infinispan/infinispan-operator/pkg/infinispan/version"
 	"github.com/infinispan/infinispan-operator/pkg/kubernetes"
 	"github.com/infinispan/infinispan-operator/pkg/mime"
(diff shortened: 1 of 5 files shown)


test/e2e/infinispan/upgrade_operand_test.go
```diff
@@ -407,6 +408,50 @@ func TestSpecImageUpdate(t *testing.T) {
...
+		infinispan.Spec.Replicas = 1
+		infinispan.Spec.Security.EndpointAuthentication = pointer.Bool(false)
+	})
+	testKube.CreateInfinispan(i, tutils.Namespace)
+	testKube.WaitForInfinispanPods(1, tutils.SinglePodTimeout, i.Name, tutils.Namespace)
+	i = testKube.WaitForInfinispanCondition(i.Name, i.Namespace, ispnv1.ConditionWellFormed)
+
+	schema := i.GetEndpointScheme()
+	client_ := testKube.WaitForExternalService(i, tutils.RouteTimeout, tutils.NewHTTPClientNoAuth(schema), nil)
+	ispnClient := v15.New(client_)
```
Crumby, line 426: Any reasonable way to use the client based on the version being tested to future proof the test? And ideally disable the test if it's not supported (but I don't think that's the case with Infinispan 14 and we don't support older operands)

ryanemerson: Yes, good catch. We can do `ispnClient := client.New(tutils.CurrentOperand, client_)` and that will use the currently configured Operand to create the client.

ryanemerson:
@Crumby I've addressed all comments re-pushed.

Crumby:
@ryanemerson Please rebase, otherwise LGTM and we can merge once the tests pass

Crumby (approved):

PR #2277, opened 2025-05-16
Prevent graceful shutdown on split brain

### LINKED ISSUE #2273: Prevent a GracefulShutdown if view is != #replicas or any cache is not in a HEALTHY state
If a split-brain has occurred between the cluster pods, we should prevent a scale down to 0 replicas from proceeding until the split has recovered.

To determine if a split-brain is present we should check the following:

- Are any of the Caches in a DEGRADED state?
- Does the CacheTopology/View have less members than `spec.replicas`?

We should also prevent the GracefulShutdown being initiated if a cache is in any state other than HEALTHY as a shutdown during HEALTHY_REBALANCE may also lead to data loss.

ryanemerson:
Closes #2273

### DIFF
--- a/pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
@@ -7,8 +7,8 @@ import (
 
 	ispnv1 "github.com/infinispan/infinispan-operator/api/v1"
 	consts "github.com/infinispan/infinispan-operator/controllers/constants"
+	ispnApi "github.com/infinispan/infinispan-operator/pkg/infinispan/client/api"
 	"github.com/infinispan/infinispan-operator/pkg/infinispan/version"
-	kube "github.com/infinispan/infinispan-operator/pkg/kubernetes"
 	pipeline "github.com/infinispan/infinispan-operator/pkg/reconcile/pipeline/infinispan"
 	"github.com/infinispan/infinispan-operator/pkg/reconcile/pipeline/infinispan/handler/provision"
 	routev1 "github.com/openshift/api/route/v1"
@@ -130,7 +130,8 @@ func GracefulShutdown(i *ispnv1.Infinispan, ctx pipeline.Context) {
 	// Initiate the GracefulShutdown if it's not already in progress
 	if i.Spec.Replicas == 0 {
 		logger.Info(".Spec.Replicas==0")
-		if *statefulSet.Spec.Replicas != 0 {
+		replicas := *statefulSet.Spec.Replicas
+		if replicas != 0 {
 			logger.Info("StatefulSet.Spec.Replicas!=0")
 			// Only send a GracefulShutdown request to the server if it hasn't succeeded already
 			if !i.IsConditionTrue(ispnv1.ConditionStopping) {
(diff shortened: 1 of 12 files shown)


pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -141,18 +142,80 @@ func GracefulShutdown(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+							i.Spec.Replicas,
+							strings.Join(info.ClusterMembers, ","),
+						)
+						ctx.Requeue(err)
+						return
+					}
+
+					health, err := podMeta.client.Container().HealthStatus()
+					if err != nil {
+						ctx.Requeue(fmt.Errorf("unable to retrieve cluster health status for pod '%s': %w", pod.Name, err))
```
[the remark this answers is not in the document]
ryanemerson: It's currently calling `/health/status` and using the `cluster_health` field which is determined by https://github.com/infinispan/infinispan/blob/3caa2cdf0dc191603067ab32353f072a966125a7/core/src/main/java/org/infinispan/health/impl/ClusterHealthImpl.java#L33-L41.

I *think* this should remain using this endpoint as this way we ensure we don't proceed with an upgrade if there is one or more DEGRADED caches.

test/e2e/upgrade/common.go
```diff
@@ -269,7 +272,14 @@ func assertNoDegradedCaches() {
...
 		if strings.Contains(log, "DEGRADED_MODE") {
-			panic("Detected a cache in DEGRADED_MODE, unable to continue with test")
+			return true
 		}
 	}
+	return false
+}
+
+// healDegradedLocksCache forces the org.infinispan.LOCKS cache to AVAILABLE so that subsequent upgrades can proceed
+func healDegradedLocksCache(operand version.Operand, client tutils.HTTPClient) {
```
Crumby, line 282: Do we need this given we opted to start upgrade tests from the  version which should no longer contain LOCKS cache issue? Maybe better keep the code clean

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -141,18 +142,80 @@ func GracefulShutdown(i *ispnv1.Infinispan, ctx pipeline.Context) {
 					return
 				}
 
-				var rebalanceDisabled bool
+				type PodMeta struct {
```
Crumby, line 145: Should we name it `ServerMeta`? I think `PodMeta` is often used in relation to k8s Pod Metadata and I'd expect the information from those in here. May confuse us in the future

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -141,18 +142,80 @@ func GracefulShutdown(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+					health, err := podMeta.client.Container().HealthStatus()
+					if err != nil {
+						ctx.Requeue(fmt.Errorf("unable to retrieve cluster health status for pod '%s': %w", pod.Name, err))
+						return
+					}
+
+					// If any of the caches are not marked as HEALTHY we must prevent a GracefulShutdown to prevent
+					// the cluster from entering an unexpected state
+					if health != ispnApi.HealthStatusHealthy {
+						ctx.Requeue(fmt.Errorf("unable to proceed with GracefulShutdown as the cluster health is '%s'", health))
```
Crumby, line 205: 1. Can we point out which caches are causing issues here to point out to user where to look?
2. I think we should propagate the state to the Infinispan Status and generate warn events for the users to see. The user may not have access to Operator logs. (The same applies for ClusterSize status)

Crumby:
@ryanemerson Tests are completely missing here, I wonder whether we could create two:
* One for split-brain scenario by introducing NetworkPolicy
* One for degraded cache. This can be harder to achieve but it was possible to introduce degraded cache to the cluster by creating a cache with invalid configuration some time ago

test/e2e/upgrade/common.go
ryanemerson: That's your call. Skipping the problematic Operands *should* prevent the issue most of the time, but it's possible that we might still encounter the DEGRADED scenario with later 14.x.y Operands. Do you want a more robust testsuite at the expense of not being aware of the underlying issue (we do log it) or do you want to explicitly fail in these scenarios? The .LOCKS cache can become DEGRADED due to network issues as well as the problems we're skipping in the server.

Crumby: I see. Alright, let's keep it here for now as long as we're supporting and testing any 14.0.x versions to on a safe side. Can you add a comment it's a specific for `14.0.x` operands so it's clear that we can remove the code once we move out of 14.0.x completely?

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
ryanemerson: I have updated the code to fire a warning event when the upgrade is blocked.

> Can we point out which caches are causing issues here to point out to user where to look?

That should be possible if we retrieve the individual cache status from the `/rest/v2/container/health` endpoint.

> 2\. I think we should propagate the state to the Infinispan Status and generate warn events for the users to see. The user may not have access to Operator logs. (The same applies for ClusterSize status)

I believe that's beyond the scope of this PR, I have created https://github.com/infinispan/infinispan-operator/issues/2282 to track this.

ryanemerson:
Writing the last comment made me re-think the problem :slightly_smiling_face:

We already have the `Stopping` condition which is meant to reflect that the status of cluster scale down. So I have made it so that we keep this as `False` with an appropriate error message if the cluster can't scale down due to either a split-brain or a DEGRADED cache.

I have also added a test for the latter case.

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -141,18 +142,94 @@ func GracefulShutdown(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
-				var rebalanceDisabled bool
+				type ServerMeta struct {
+					client ispnApi.Infinispan
+					skip   bool
+					state  string
+				}
+
+				// Loop through all pods to initiate the clients, determine that all pods have the expected
+				// number of cluster members and ensure that the cluster is in a HEALTHY state
+				var podMetaMap = make(map[string]*ServerMeta, len(podList.Items))
```
Crumby, line 153: `serverMetaMap`

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -141,18 +142,94 @@ func GracefulShutdown(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
+				var shutdownAlreadyInitiated bool
 				for _, pod := range podList.Items {
-					ispnClient, err := ctx.InfinispanClientUnknownVersion(pod.Name)
+					podMeta := &ServerMeta{}
```
Crumby, line 156: `serverMeta`

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
```diff
@@ -141,18 +142,94 @@ func GracefulShutdown(i *ispnv1.Infinispan, ctx pipeline.Context) {
...
-							logger.Info("Skipping pod whose cache-container has already been shutdown by the Operator", "pod", pod.Name, "state", state)
+							podMeta.skip, podMeta.state = true, state
+							logger.Info("At least one cache-container has already been shutdown by the Operator, resuming shutdown", "pod", pod.Name, "state", state)
+							shutdownAlreadyInitiated = true
+							// Continue processing other pods here as it's possible that one or more pods still haven't
+							// been shutdown and we need to initiate the client
 							continue
 						}
-						ctx.Requeue(fmt.Errorf("unable to create Infinispan client for cluster being upgraded: %w", err))
+						ctx.Requeue(fmt.Errorf("unable to create Infinispan client to determine if split-brain is present: %w", err))
```
Crumby, line 168: Should the message here be more generic? As far as I understand the code we are trying to get the client to get various information. Not just split brain

test/e2e/infinispan/upgrade_operand_test.go
```diff
@@ -422,6 +424,35 @@ func TestPodAlreadyShutdownOnUpgrade(t *testing.T) {
...
+	cacheHelper := tutils.NewCacheHelper(cacheName, _client)
+	cacheHelper.Create(cacheConfig, mime.ApplicationXml)
+	cacheHelper.Available(false)
+
+	tutils.ExpectNoError(
+		testKube.UpdateInfinispan(ispn, func() {
+			ispn.Spec.Replicas = 0
+		}),
+	)
+	testKube.WaitForInfinispanState(ispn.Name, ispn.Namespace, func(i *ispnv1.Infinispan) bool {
```
Crumby, line 450: Should we extend the test for the second part, where we fix the degraded cache and verify the upgrade will proceed further?

pkg/reconcile/pipeline/infinispan/handler/manage/upgrades.go
ryanemerson: Update to "unable to create Infinispan client to determine if cluster is healthy: %w"

test/e2e/infinispan/upgrade_operand_test.go
ryanemerson: I have updated the test so that we set the Cache to AVAILABLE and then we wait for the cluster to scale down to 0 pods with `GracefulShutdown=True`. We haven't changed the semantics of the rest of the GracefulShutdown upgrade procedure, so it's not necessary to change the remainder of the upgrade workflow.

Crumby (approved):

ryanemerson:
Thanks for reviewing @Crumby

PR #2278, opened 2025-05-19
[#2279] update correct Jgroups version for operand 14.0 and 15.1 and fix jgroups and infinispan operand mapping

### LINKED ISSUE #2279: Fix Jgroups and Infinispan mapping
We were using wrong Jgroups versions for Infinispan 14.x and 15.x and hence mapping between Jgroups and Infinispan was also incorrect.

This will fix the mapping and update the correct Jgroups version.

pminz:
Closes  #2279


ryanemerson:
@pminz Please create a Git Hub issue describing the problem and reference the issue in the PRs commit message using the expected format `[#xxx] <message>`

ryanemerson:
Thanks @pminz

PR #2286, opened 2025-05-21
[#2285]  2.4.x Backport PRs are not triggering CI

### LINKED ISSUE #2285: 2.4.x Backport PRs are not triggering CI
The CI GH actions aren't running for the PRs created when adding the backport labels to `main` PRs.

pminz:
Closes #2285

### DIFF
--- a/.github/workflows/backport.yaml
+++ b/.github/workflows/backport.yaml
@@ -18,7 +18,7 @@ jobs:
         with:
           target-branch: 2.3.x
           pull-request: ${{ github.event.pull_request.url }}
-          auth: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


ryanemerson:
@pminz How does `pull_request_target` fix the issue? I think we should be cautious about allowing this type for all users, as it allows access to secrets which when combined with `checkout` can lead to credentials being exposed. Maybe a solution is to only execute if the user is the `github-action-bot` used to create backport PRs?

https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/

ryanemerson:
Thanks @rigazilla. The secrets are defined organisation wide, so we should be able to use `INFINISPAN_RELEASE_TOKEN` here as well @pminz.

rigazilla (approved):

PR #2293, opened 2025-06-10
Revert configuration

pminz:


pminz:
@ryanemerson  yes. I use VS code don't how it doesn't format it automatically.
Fixed it.

PR #2294, opened 2025-07-03
[#2297] Allow existing Operands in manager.yaml to be overwritten

### LINKED ISSUE #2297: CI Allow existing Operands in manager.yaml to be overwritten
The `main` branch currently has an entry for 16.0.0 with a DevXX image so that we can test Infinispan 16 integration in this repository. However, that causes an issue calling the Operator upgrade tests from the Infinispan repository as we now have two 16.0.0 Operands which the VersionManager does not allow. To work around this, we should update the scripts to allow an Operand in `manager.yaml` to be overwritten if it already exists.

ryanemerson:
Closes #2297 

Can be tested with ` IMAGE=test VERSION=16.0.0 ./scripts/ci/add_operand_to_csv.sh; cat config/manager/manager.yaml`


rigazilla:
thanks @ryanemerson

PR #2317, opened 2025-07-24
[#2290] Allow pod terminationGracePeriodSeconds to be configured via the Infinispan CR

### LINKED ISSUE #2290: Allow pod terminationGracePeriodSeconds to be configured via the Infinispan CR
Currently the Operator relies upon the default `terminationGracePeriodSeconds` of 30s for all clusters.

It should be possible for users to configure a custom `terminationGracePeriodSeconds` value in the event that they have very large caches that need to be persisted on shutdown.

We could expose this by adding the following element to the Infinispan CR:

```yaml
spec:
  service:
    container:
      terminationGracePeriodSeconds: 60s
```

pminz:
Close #2290

### DIFF
--- a/documentation/asciidoc/topics/proc_configuring_pod_termination_grace_period.adoc
+++ b/documentation/asciidoc/topics/proc_configuring_pod_termination_grace_period.adoc
@@ -1,0 +1,21 @@
+[id='configuring-pod-termination-grace-period{context}']
+= Configuring pod termination grace period
+
+[role="_abstract"]
+
+Optionally configure the `TerminationGracePeriodSeconds` of the {brandname} pods.
+
+The {ispn_operator} relies on the Kubernetes default `TerminationGracePeriodSeconds` unless a value is explicitly provided.
+We only recommend providing your own value once you have determined that the default value do not match your requirements.
+
+.Procedure
+. Configure probe values using the `spec.service.container.terminationGracePeriodSeconds` field:
++
+[source,yaml]
+----
+include::yaml/termination_grace_period.yaml[]
+----
++
+. Apply your `Infinispan` CR.
++
+If your cluster is running, {ispn_operator} restarts the {brandname} pods in order for the changes to take effect.
--- a/documentation/asciidoc/stories/assembly_creating_services.adoc
+++ b/documentation/asciidoc/stories/assembly_creating_services.adoc
@@ -19,6 +19,7 @@ include::{topics}/proc_allocating_cpu_memory.adoc[leveloffset=+1]
 include::{topics}/proc_setting_jvm_options.adoc[leveloffset=+1]
 include::{topics}/proc_configuring_pod_probes.adoc[leveloffset=+1]
 include::{topics}/proc_configuring_pod_priority.adoc[leveloffset=+1]
+include::{topics}/proc_configuring_pod_termination_grace_period.adoc[leveloffset=+1]
 
 // Downstream content
 ifdef::downstream[]
--- a/documentation/asciidoc/topics/yaml/termination_grace_period.yaml
+++ b/documentation/asciidoc/topics/yaml/termination_grace_period.yaml
@@ -1,0 +1,4 @@
+spec:
+  service:
+    container:
+      terminationGracePeriodSeconds: 60


ryanemerson (changes requested):
@pminz This is a good start, but we need to pass the configured value to the PodSpec of StatefulSet that we create for the cluster. This should be applied in `pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go`.

We create other pods (zero-controller, gossip-router, batch) but I don't think it makes sense to apply the custom `terminationGracePeriodSeconds` in these contexts as these nodes can't contain large amounts of state that need to be written to disk on shutdown.

It would also be good if we extended `TestBaseFunctionality` so that we specify a custom TerminationGracePeriod and verify that it is indeed set on the created pods. See `verifyScheduling` as an example of how a pod can be retrieved and it's content checked.

api/v1/infinispan_types.go
```diff
@@ -121,6 +121,9 @@ type InfinispanServiceContainerSpec struct {
 	// when it might take a long time to load data or warm a cache, than during steady-state operation.
 	// +optional
 	StartupProbe ContainerProbeSpec `json:"startupProbe,omitempty"`
+	// TerminationGracePeriodSeconds specifies the duration the container is given to shut down gracefully before it is forcefully terminated.
+	// +kubebuilder:default=30
```
ryanemerson, line 125: We should remove this in favour of `+optional` see other comment about not defining a default in the Operator code.

pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
```diff
@@ -85,6 +85,11 @@ func ClusterStatefulSetSpec(statefulSetName string, i *ispnv1.Infinispan, ctx pi
 	statefulSetAnnotations["checksum/podEnvs"] = podEnvsHash
 	annotationsForPod := i.PodAnnotations()
 	annotationsForPod["updateDate"] = time.Now().String()
+	gracePeriod := int64(30) // default fallback
+
+	if i.Spec.Service.Container.TerminationGracePeriodSeconds != nil {
```
ryanemerson, line 90: `i.Spec.Service.Container` is a pointer, so it's possible that this value is `nil`. We need to guard against this to prevent Nil Pointer Deference errors being thrown.

I suggest to add a helper method to `types_util.go` similar to `StorageClassName` that returns `nil` if `spec.Service.Container` is nil otherwise it just returns the value of `TerminationGracePeriodSeconds` (can be nil).

pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
```diff
@@ -112,9 +117,10 @@ func ClusterStatefulSetSpec(statefulSetName string, i *ispnv1.Infinispan, ctx pi
 					Annotations: annotationsForPod,
 				},
 				Spec: corev1.PodSpec{
-					Affinity:                  i.Affinity(),
-					Tolerations:               i.Tolerations(),
-					TopologySpreadConstraints: i.TopologySpreadConstraints(),
+					TerminationGracePeriodSeconds: &gracePeriod,
```
ryanemerson, line 120: `TerminationGracePeriodSeconds` in the PodSpec is a int64 pointer the same as in our types, so there's no need to explicitly check this value is set we can just pass the return value of the `types_util.go` method directly.

It's better for us to leverage the default `TerminationGracePeriodSeconds` determined by Kubernetes, instead of explicitly defining 30s in our code, as different k8s vendors may change this value or the value may be changed in future k8s versions.

test/e2e/infinispan/smoke_test.go
```diff
@@ -65,7 +65,7 @@ func TestBaseFunctionality(t *testing.T) {
 	verifyNoPVCs(assert, ispn)
 	verifyLabelsAndAnnotations(assert, require, ispn)
 	verifyDefaultAuthention(require, ispn)
-	verifyScheduling(assert, require, ispn)
+	verifyScheduling(assert, require, ispn, 30)
```
ryanemerson, line 68: To reflect the changes suggested in this review, we should set an explicit value in the Infinispan CR that is not equal to the default and assert that it has been set.

ryanemerson (changes requested):

api/v1/types_util.go
```diff
@@ -1024,3 +1024,11 @@ func (ispn *Infinispan) InitServiceContainer() {
 	c.ReadinessProbe.AssignDefaults(5, 0, 10, 1, 1)
 	c.StartupProbe.AssignDefaults(600, 3, 1, 1, 1)
 }
+
+func (ispn *Infinispan) TerminationGracePeriodSeconds() *int64 {
+	sc := ispn.Spec.Service.Container
+	if sc != nil && sc.TerminationGracePeriodSeconds != nil {
```
ryanemerson, line 1030: This can  be simplified to:

```go
	if sc != nil {
		return sc.TerminationGracePeriodSeconds
	}
	return nil
```

test/e2e/infinispan/smoke_test.go
```diff
@@ -65,7 +65,7 @@ func TestBaseFunctionality(t *testing.T) {
 	verifyNoPVCs(assert, ispn)
 	verifyLabelsAndAnnotations(assert, require, ispn)
 	verifyDefaultAuthention(require, ispn)
-	verifyScheduling(assert, require, ispn)
+	verifyScheduling(assert, require, ispn, 60)
```
ryanemerson, line 68: The test is failing because you're expecting a non-default value of 60 here, but you have not defined this in the Infinispan CR. You need to explicitly define 60 in the `tutils.DefaultSpec` function that starts on L26.

ryanemerson (changes requested):
Overall LGTM, just one minor comment and the test needs fixing before we can merge.

test/e2e/infinispan/smoke_test.go
```diff
@@ -52,6 +53,7 @@ func TestBaseFunctionality(t *testing.T) {
 				WhenUnsatisfiable: "ScheduleAnyway",
 			}},
 		}
+		i.Spec.Service.Container.TerminationGracePeriodSeconds = pointer.Int64(60)
```
ryanemerson, line 56: @pminz I have pushed a commit to simplify this code. The `pointer` package is very nice for these things as it means you can do assignment with one line instead of two.

ryanemerson (approved):
Thanks @pminz. I'll squash the commits and merge once CI has finished. I have added the `2.4.x` label so that this will automatically be backported.

documentation/asciidoc/topics/proc_configuring_pod_termination_grace_period.adoc
```diff
@@ -0,0 +1,21 @@
+[id='configuring-pod-termination-grace-period{context}']
```
ryanemerson, line 1: @pminz I was about to merge and then I realised we forgot the documentation. I have pushed these real quick. Please review and if you think we have missed something we can address in a future PR

pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
```diff
@@ -112,6 +112,11 @@ func StatefulSetRollingUpgrade(i *ispnv1.Infinispan, ctx pipeline.Context) {
 		}
 	}
 
+	if spec.TerminationGracePeriodSeconds != i.TerminationGracePeriodSeconds() {
```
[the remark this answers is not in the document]
pminz: Thanks @ryanemerson

PR #2336, opened 2025-08-20
[2.4.x] Supported Operand table Operator versions in the wrong order

### LINKED ISSUE #2334: Supported Operand table Operator versions in the wrong order
The Operator versions are displayed in the wrong order, see below:

Image

Instead we should generate the table so that Operator versions appear in ascending order based upon semantic versioning.

ryanemerson:
Closes #2334


pminz:
Merged. Thanks @ryanemerson

PR #2349, opened 2025-09-05
Next Version 2.4.18

pminz:


ryanemerson:
Thanks @pminz, we need the version bump for `main` as well.

PR #2358, opened 2025-09-23
automate bundle release

pminz:


.github/workflows/add_operand.yml
```diff
@@ -52,14 +52,21 @@ jobs:
...
             TAG=${IMAGE#*:}
-            echo VERSION=${TAG%.Final} >> ${GITHUB_ENV}
+            VERSION=${TAG%.Final}
           else
-            echo VERSION=${{ inputs.version }} >> ${GITHUB_ENV}
+            VERSION="$INPUT_VERSION"
           fi
+    
+          echo "VERSION=$VERSION" >> "$GITHUB_ENV"
+          echo "version=$VERSION" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
pminz: yes we don't need it, removed

.github/workflows/add_operand.yml
```diff
@@ -74,3 +81,31 @@ jobs:
...
+               "repository": "${{ inputs.repository }}"
+             }
+
+      - name: Trigger next workflow if branch is main and success
+        if: success() && github.ref_name == 'main' && contains(steps.set_version.outputs.version, '.Final')
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.INFINISPAN_RELEASE_TOKEN }}
+          repository: ${{ inputs.repository }}
+          event-type: trigger-next-workflow
```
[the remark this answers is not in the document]
ryanemerson: Thanks for the explanation. Please change "trigger-next-workflow" to a more meaningful event name, maybe "release-operator"?

.github/workflows/publish_main.yaml
```diff
@@ -0,0 +1,74 @@
+name: Publish to main
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: Branch to wwhich changes needs to be pushed
```
ryanemerson: `s/wwhich/which`

.github/workflows/image_release.yaml
```diff
@@ -41,6 +70,85 @@ jobs:
...
+         path: community-operators-prod
+         ref: main
+
+      - name: Copy contents of bundle/ and issue PRs
+        working-directory: community-operators-prod
+        run: |
+         mkdir -p operators/infinispan/${{ steps.bump_tag.outputs.new_tag }}
+         cp -R ../operator/bundle/* operators/infinispan/${{ steps.bump_tag.outputs.new_tag }}
+         git add operators/infinispan/${{ steps.bump_tag.outputs.new_tag }}
+         git commit -s -m " Operator Infinispan ${{ steps.bump_tag.outputs.new_tag }}"
```
ryanemerson: Unnecessary whitespace before message start.

.github/workflows/publish_main.yaml
```diff
@@ -0,0 +1,74 @@
...
+        required: true
+      rel_version:
+        description: The Operator tag to checkout 
+        required: true
+
+  repository_dispatch:
+    types: [trigger-main-workflow]
```
ryanemerson: "update-version"?

ryanemerson (commented):
@pminz How are you testing these changes?

.github/workflows/image_release.yaml
```diff
@@ -41,6 +70,85 @@ jobs:
...
+         commit-message: 'Operator Infinispan ${{ steps.bump_tag.outputs.new_tag }}'
+         committer: infinispan-qe-bot <q*@infinispan.org>
+         author: infinispan-qe-bot <q*@infinispan.org>
+         branch: pr_request
+         delete-branch: true
+         push-to-fork: infinispan-qe-bot/community-operators-prod
+         title: 'Operator Infinispan ${{ steps.bump_tag.outputs.new_tag }}'
+         body: 'Operator Infinispan ${{ steps.bump_tag.outputs.new_tag }}'
+
+      - name: Trigger next workflow if branch is main and success
```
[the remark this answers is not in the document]
pminz: I have changed the event-type to a meaningful name and fixed all the typo/spaces.
I haven't tested it e2e but  I have verified  repository_dispatch event along with the related scripts and commands which works as expected.
Yes  - name: Trigger next workflow if branch is main and success -The workflow is triggered immediately after a PR is merged, but it first checks whether the operator tag is available in OperatorHub.io before proceeding to the next step.
But I have changed it to the one which we used for gingersnap as the earlier approach caused the runner to stay active for too long.

scripts/ci/bump_version.sh
```diff
@@ -0,0 +1,43 @@
...
+BASE_VERSION=${LATEST_TAG%.Final}
+
+echo "Previous version: $BASE_VERSION"
+
+echo "prev_ver=$BASE_VERSION" >> "$GITHUB_OUTPUT"
+
+IFS='.' read -r MAJOR MINOR PATCH <<< "$BASE_VERSION"
+
+# This logic works for 2.4.x tags, we can tweak it later.
+if [ "$PATCH" -ge 18 ]; then
```
Crumby: Now that 8.5.0 is out, this no longer makes sense. If the release is triggered manually we could add an input option for a new minor. (It'd reset patch and increased minor)

pminz: removed the logic

Crumby:
Great job @pminz, this will save us a bunch of time. I do have a concern regarding all of the process being fully automated though.

The `Add Operand` workflow which triggers release the process is not checking whether Operand is properly working with the Operator and there isn't an regression introduced (eg. if we added 16.0.0 Operand as soon as it was released, we'd release with a broken image). In my opinion this one should rather create a PR which will run the tests against the new Operand before it'll be merged.

The similar situation applies to Operator release, do we have to approve the PR created to the operators repository before they get merged? I think they should be checked at least few first times before we'll trust the automation. Can we trigger smoke tests against the release once it's in?

Opinions @ryanemerson?

Crumby:
Yeah, 1.2b is what I would head for. Main reason so we can say that the newly added version was tested with the Operator (especially now that the tests generally run with main snapshot which is 16.1 atm)

Let me create an issue that describes the steps and ideas for further improvement and then we can merge this PR.

One more thing that we may want to be aware of is that we'll be effectively shorting feature bake time before releasing. We have decided to keep the Operator on 2.5.x for the lifetime of Infinispan 16.x.y releases non restricting ourself to not adding features in the micros. It may happen that feature will be merged and released the day after. Which doesn't give the community time to discover any smaller or bigger subissues with implementation

ryanemerson:
> One more thing, how do we deal with multiple Infinispan releases at once?

I think we should keep it simple to begin as we do not have a way on the Infinispan side to automate multiple releases.

We could make it so that a new Operand PR only triggers an Operator release if a `release` label is present. When the `add_operand` job is triggered we add this label by default but make it configurable. If we modify the release job in Infinispan to have a `operator_release` bool, we can explicitly set the `release` param on the `add_operand` to `false` and only  set to `true` for the last of the three releases in your example.

Crumby (changes requested):
@pminz There's one minor comment and I'd change the label that triggers the release. Once the changes are implemented please rebase and squash the changes into a single commit.

And I'm just thinking whether we should get one more test run of the pipeline just to be on a safe side that the latest changes didn't break anything

PR #2360, opened 2025-09-24
Enable trace logs

ryanemerson:


jabolina:
Thanks, @ryanemerson! We can close this one now that we have collected the needed information.

PR #2363, opened 2025-10-01
Cache and limit the range of Operands tested on CI

### LINKED ISSUE #2365: Limit range of Operands tested on CI
Currently we test all supported Operands from 14.0.23 as part of our update tests, however this is proving to be problematic for several reasons:

1. The GH action runners are starting to run out of space due to the amount of images that need to be pulled
2. We're encountering rate limit related errors from quay.io
3. Older Operands in a stream often contain bugs, which have been fixed in later versions from the same stream, which cause our tests to be flaky
4. CI takes a long time to complete

Instead, we should only test a subset of Operands. Proposal:

- Test all versions in the latest stable stream
- Test the latest Operand version for a supported stream, e.g. 14.0.x -> 14.0.32, 15.0.x -> 15.0.21 etc

ryanemerson:
Closes #2365

### DIFF
--- a/.github/workflows/test_hr_rolling_upgrades.yml
+++ b/.github/workflows/test_hr_rolling_upgrades.yml
@@ -19,6 +19,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Free Disk Space
+        uses: jlumbroso/free-disk-space@v1.3.1
+
       - name: Install Dependencies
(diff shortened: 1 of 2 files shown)


ryanemerson:
Upgrade tests need updating as currently they read the Operand list from the CSV `INFINISPAN_OPERAND_VERSIONS`

PR #2364, opened 2025-10-03
Add encoding configuration to the test caches

Crumby:
After changes introduced by [deprecation removals ](https://github.com/infinispan/infinispan/pull/15370) puts with `Content-Type: application/json` will fail against caches that don't have encoding configured to either `application/json` or `text/plain`

### DIFF
--- a/test/e2e/upgrade/common.go
+++ b/test/e2e/upgrade/common.go
@@ -65,7 +65,7 @@ func subscription(olm tutils.OLMEnv) *coreos.Subscription {
 func createAndPopulateVolatileCache(cacheName string, numEntries int, client tutils.HTTPClient) {
 	c := tutils.NewCacheHelper(cacheName, client)
 	if !c.Exists() {
-		c.Create(`{"distributed-cache":{"mode":"SYNC"}}`, mime.ApplicationJson)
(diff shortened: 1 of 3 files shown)


test/e2e/upgrade/common.go
```diff
@@ -65,7 +65,7 @@ func subscription(olm tutils.OLMEnv) *coreos.Subscription {
 func createAndPopulateVolatileCache(cacheName string, numEntries int, client tutils.HTTPClient) {
 	c := tutils.NewCacheHelper(cacheName, client)
 	if !c.Exists() {
-		c.Create(`{"distributed-cache":{"mode":"SYNC"}}`, mime.ApplicationJson)
+		c.Create(`{"distributed-cache":{"mode":"SYNC", \"encoding\": {\"media-type\": \"application/json\"}}}`, mime.ApplicationJson)
```
ryanemerson, line 68: No need for the backslashes in this file as we're using a ` string in both cases.

We're getting the error "unexpected HTTP status code (400): unexpected error creating cache, response: Invalid JSON at [1,72]" without testing locally, I'm not sure if that's to do with ^ formatting, or if the 14.0.x images don't like the json.

ryanemerson (changes requested):

PR #2372, opened 2025-10-24
[#2371] Operand cache is not updated when new image sha is available

### LINKED ISSUE #2371: Operand cache is not updated when new image sha is available
The Operand cache uses `${{ runner.os }}-operands-${{ steps.images.outputs.sha }}` as a key, which does not take into account an image's actual sha. Consequently, it's possible that the cache contains an older image. We attempt to workaround this by repulling all images even on a cache hit in the "images" job, however the updated `docker save` tar is not saved to the cache as the cache action is not updated when a cache hit occurs. All test jobs utilise the cache "as is" and do not attempt to pull the latest images, to ensure all tests are executed with the same images, so the Operator tests will not be executed with the latest images.

ryanemerson:
Closes #2371

### DIFF
--- a/.github/actions/operand-cache/action.yml
+++ b/.github/actions/operand-cache/action.yml
@@ -27,7 +27,10 @@ runs:
       uses: actions/cache@v4
       with:
         path: ${{ steps.images.outputs.tar }}
-        key: ${{ runner.os }}-operands-${{ steps.images.outputs.sha }}
+        # Utilise a unique key with restore-keys as caches cannot be updated on hit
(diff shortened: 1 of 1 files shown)


Crumby:
@ryanemerson Looks like it won't update the cache on job re-run as `run-id` is still the same. But maybe additional commit or rebasing the PR is enough in such case to get new `run-id`. I'm thinking, now that https://github.com/infinispan/infinispan/pull/15957 was merged, the `TestExternalDependenciesHttp` and `TestCredentialStore` should be passing but they still fail as they use image available when the PR was created.

PR #2378, opened 2025-11-04
[#2377] TestBackupRestore fails due to missing unzip package in Infinispan 16 image

### LINKED ISSUE #2377: TestBackupRestore fails due to missing unzip package in Infinispan 16 image
The test will fail as the new image does not contain the `unzip` package. There's no need to add this package to the image, as it will increase the CVE surface, instead we can utilise an older image in the testsuite that contains this package.

We should also ensure that the "verify-backup-pod" is cleaned up on failure to prevent subsequent test failures.

ryanemerson:
Closes #2377

### DIFF
--- a/test/e2e/backup-restore/backup_restore_test.go
+++ b/test/e2e/backup-restore/backup_restore_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/blang/semver"
 	"github.com/iancoleman/strcase"
 	v1 "github.com/infinispan/infinispan-operator/api/v1"
 	"github.com/infinispan/infinispan-operator/api/v2alpha1"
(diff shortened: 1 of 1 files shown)


test/e2e/backup-restore/backup_restore_test.go
```diff
@@ -80,6 +82,9 @@ func testBackupRestore(t *testing.T, clusterSpec clusterSpec, clusterSize, numEn
 	tutils.ExpectNoError(testKube.Kubernetes.Client.Get(context.TODO(), types.NamespacedName{Namespace: namespace, Name: sourceCluster}, infinispan))
 
 	// Validate the number of entries stored in the someCache backup file
+	// Utilise a 15.0.x image as this still contains the unzip package
+	imageOp, err := tutils.VersionManager().LatestUpstreamPatch(semver.Version{Major: 15, Minor: 0})
```
Crumby: This looks like a temporary solution only. Once we remove 15.0.x from the available image, it'll fail. Even though this situation is a long time ahead of us.

ryanemerson: Yeah, my thoughts that's like years away and we can worry about that at the time. I didn't want to pull another image in, e.g. ubi, as that would then have to be added to the image cache.

ryanemerson:
> Does Infinispan generate zip files from backup? I wonder if it makes sense to have the `zip` utility on the image for possible debugging should it be needed to.

Yes. I thought this as well at first, but we have the `tar` command in the image which allows for us to call `kubectl cp`. So I came to the conclusion it was better to keep the image more lightweight and if users need to inspect the .zips they can just copy it locally first.

PR #2385, opened 2025-11-07
Update IT tests with ISPN16

Crumby:


Crumby:
@ryanemerson So the tests are executed outside of the K8s but they are mostly triggering servlets defined on the Tomcat server with different configurations to test. The reason is that on OCP Route you can't test much with HotRod but passthrough. The servlets from then exeucte client authentication, validation, unencrypted configuration and I believe clustering in the terms that the HotRod can see and connect to all the servers directly

PR #2409, opened 2026-01-02
[#2408] Support additional HTTP digest algorithms

### LINKED ISSUE #2408: Support multiple digest algorithms in the test http client
The test http client is hardcoded to use MD5 authentication. It should also support other algorithms, such as SHA-256 https://datatracker.ietf.org/doc/html/rfc7616

tristantarrant:
Closes #2408


ryanemerson:
@tristantarrant Please rebase on `main` and then we should get a full CI run

PR #2417, opened 2026-01-08
Enable running the tests against latest server release with label

### LINKED ISSUE #2415: Run the tests against newly added Operands via AddOperand action
`Add Operand` action currently directly commits a new server version to the `manager.yaml` file without running and checking the test results. Given we usually release after adding a new Operands and we'll automatically release after adding a new Operand it would be beneficial to execute the tests before having the Operand added to the `manager.yaml` file.

Ideally the action would create a new PR that would be auto-merged once the tests pass while being executed with the newly added version.

**Required changes:**
* The action will create a PR rather then direct commit to a main
* The PR has to run the tests against newly added version rather then latest Infinispan:
  * Adding `TESTING_OPERAND_VERSION` env var to the CI should be sufficient
  * Setting the env var could be either direct by the action or setting `test/operand-version` label which would be then picked up by the triggered tests. (Using the label could be beneficial with other PRs)
* Once the PR passes it'll be auto-merged
* Upon merged the PR, release workflow will be triggered

Crumby:
Closes #2415

### DIFF
--- a/.github/workflows/add_operand.yml
+++ b/.github/workflows/add_operand.yml
@@ -67,10 +67,25 @@ jobs:
           IMAGE: ${{ inputs.image }}
           VERSION: ${{ env.VERSION }}
 
-      - name: Commit changes and push to repository
-        run: |
-          git config user.email "infinispan@infinispan.org"
-          git config user.name "Infinispan"
-          git add config/manager/manager.yaml documentation/asciidoc/topics/attributes/community-attributes.adoc documentation/asciidoc/topics/supported_operands/
-          git commit -m "Add Operand ${{ env.VERSION }}"
-          git push
+      - name: Create Pull Request
+        id: cpr
+        uses: peter-evans/create-pull-request@v8
+        with:
+          labels: test/release
+          token: ${{ secrets.INFINISPAN_RELEASE_TOKEN }}
+          commit-message: 'Add Operand ${{ env.VERSION }}'
+          committer: Infinispan <infinispan@infinispan.org>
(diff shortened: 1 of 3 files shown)


.github/workflows/add_operand.yml
```diff
@@ -67,10 +67,30 @@ jobs:
...
+
+      - name: Create Pull Request
+        id: cpr
+        uses: peter-evans/create-pull-request@v8
+        with:
+          labels: test/release
+          token: ${{ secrets.INFINISPAN_RELEASE_TOKEN }}
+          commit-message: 'Add Operand ${{ env.VERSION }}'
+          committer: infinispan-qe-bot <q*@infinispan.org>
+          author: infinispan-qe-bot <q*@infinispan.org>
```
ryanemerson: We should be consistent here, in our other actions we're using:

```
        git config --global user.email "infinispan@infinispan.org"
        git config --global user.name "Infinispan"
```

.github/workflows/add_operand.yml
```diff
@@ -67,10 +67,30 @@ jobs:
           IMAGE: ${{ inputs.image }}
           VERSION: ${{ env.VERSION }}
 
-      - name: Commit changes and push to repository
+      - name: Configure repository
```
ryanemerson: Do we need this step now that we're specifying the details in the next step?

Crumby: Good point, actually not. Wasn't sure if isn't need for the automerge enablement action

PR #2434, opened 2026-02-05
add concurreny lock and rebase the branch before adding Operand

pminz:


.github/workflows/add_operand.yml
```diff
@@ -51,6 +55,11 @@ jobs:
           repository: ${{ inputs.repository }}
           token: ${{ secrets.INFINISPAN_RELEASE_TOKEN }}
 
+      - name: Rebase onto latest
```
ryanemerson: Is this necessary if the workflow can only be executed sequentially? I would expect that the checkout step automatically pulls the latest.

ryanemerson: I don't understand. In the Infinispan repository when we call `add_operand.yml` we're not passing a sha reference, we're just passing the branch name via `inputs.ref`. The checkout action is not executed until the job starts, so I would expect it should checkout the latest code in that ref and the rebase is redundant.

.github/workflows/image_release.yaml
```diff
@@ -32,7 +29,30 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
```
Crumby: I think we'll need to add `fetch-depth: 2` here otherwise the `check` step won't work properly. Checkout by default fetches only the latest commit and we need access to the history in the `check` step.

https://github.com/actions/checkout?tab=readme-ov-file#usage

.github/workflows/add_operand.yml
```diff
@@ -39,6 +39,10 @@ on:
         required: false
         type: string
 
+concurrency:
```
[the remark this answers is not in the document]
pminz: I think it would still be useful if we were adding two operands simultaneously.

PR #2452, opened 2026-03-06
[#2379] Add Schema CRD for managing Protobuf schemas

### LINKED ISSUE #2379: Schema CR
Use the schema events introduecd by 
https://github.com/infinispan/infinispan/issues/16030

tristantarrant:
Closes #2379


ryanemerson (changes requested):

documentation/asciidoc/topics/yaml/cache_schema_ref.yaml
```diff
@@ -0,0 +1,31 @@
...
+    }
+---
+apiVersion: infinispan.org/v2alpha1
+kind: Cache
+metadata:
+  name: person-cache
+spec:
+  clusterName: infinispan
+  name: persons
+  schemaRef: person-schema
```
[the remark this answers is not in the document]
Crumby: Looks like it is as the cache can index multiple entities

```
<indexed-entities>
    <indexed-entity>org.infinispan.Person</indexed-entity>
</indexed-entities>
```

tristantarrant:
Schema CRs now have a status flag explaining whether bidi sync is enabled (operand >=16.1)

ryanemerson (approved):
LGTM assuming CI turns 🟢

(The previous run failed with an unrelated rate-limiter issue)

test/e2e/schema/schema_test.go
```diff
@@ -0,0 +1,454 @@
...
+	// Create the second Schema CR — cache should become ready
+	authorSchema := schemaCR(authorSchemaName, authorProto, ispn)
+	testKube.Create(authorSchema)
+	testKube.WaitForSchemaConditionReady(authorSchemaName, ispn.Name, tutils.Namespace)
+
+	authorSchemaHelper := tutils.NewSchemaHelper(authorSchemaName, httpClient)
+	authorSchemaHelper.WaitForSchemaToExist()
+
+	// Now the cache should become ready
+	testKube.WaitForCacheConditionReady(cacheName, ispn.Name, tutils.Namespace)
```
Crumby, line 392: The test fails when executed against OpenShift (`16.1` and `snapshot`) with effective error being `ISPN028021: The configured indexed-entity type 'author_sample.Author' must be indexed.` from the server side.

Why the issue doesn't appear in the CI is mystery to me.

config/crd/bases/infinispan.org_caches.yaml
```diff
@@ -96,6 +96,12 @@ spec:
                 description: Name of the cache to be created. If empty ObjectMeta.Name
                   will be used
                 type: string
+              schemaRefs:
+                description: Names of the Schema CRs that must be ready before the
+                  cache is created
+                items:
```
Crumby, line 101: Do we want to go ahead with string array here? It's non-expandable in the future compared to object array which would allow us to add more properties in the future. (I can't think of any though at the moment)

Another note, as it's defined, it'll work reasonably only for the Schema CRs created by the user. Schema CRs created by the config listeners will suffix random alphanumeric numbers

controllers/schema_controller.go
```diff
@@ -0,0 +1,528 @@
...
+		reqLogger.Info(fmt.Sprintf("Infinispan cluster %s not well formed", infinispan.Name))
+		return ctrl.Result{}, nil
+	}
+
+	// Schema reconciliation requires operand >= 16.1
+	operand, err := r.versionManager.WithRef(infinispan.Spec.Version)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("unable to resolve operand version: %w", err)
+	}
+	bidirectionalSync := operand.UpstreamVersion.GTE(constants.MinVersionSchemaReconciliation)
```
Crumby: It also requires Infinispan to have config listener enabled

test/e2e/schema/schema_test.go
```diff
@@ -0,0 +1,454 @@
...
+		cacheHelper.Put(data, data, mime.ApplicationJson)
+	}
+	cacheHelper.AssertSize(numEntries)
+
+	// Verify we can retrieve an entry
+	value, exists := cacheHelper.Get(`{"_type":"book_sample.Book","title":"book0","publicationYear":2020}`)
+	testifyAssert.True(t, exists, "Entry should exist in cache")
+	testifyAssert.Contains(t, value, "book0")
+
+	// Delete the cache first, then the schema
```
Crumby, line 314: What do we expect to happen with a cache if the Schema listed as dependency gets removed? Do we remove the cache from the cluster as well?

test/e2e/schema/schema_test.go
```diff
@@ -0,0 +1,454 @@
...
+	serverSchema := schemaHelper.Get()
+	testifyAssert.Contains(t, serverSchema, "phone")
+
+	assertConfigListenerHasNoErrorsOrRestarts(t, ispn)
+}
+
+// TestSchemaWithServerLifecycle verifies that schemas created/updated/deleted via the
+// REST API are reflected as Schema CRs by the ConfigListener.
+func TestSchemaWithServerLifecycle(t *testing.T) {
+	skipIfBidirectionalSyncNotSupported(t)
```
Crumby: there's a function `skipPriorTo()` in the `assert.go` file. We should reuse it

test/e2e/schema/schema_test.go
tristantarrant: We should not do that

ryanemerson: What happens if the schema is removed on the server while a cache depends on it?

test/e2e/schema/schema_test.go
tristantarrant: Done

test/e2e/schema/schema_test.go
tristantarrant: Nothing

tristantarrant: These integrity aspects should be handled on the server before we approach them from the operator angle

config/crd/bases/infinispan.org_caches.yaml
tristantarrant: I have made it an `[]object` now. As for the second half of the question, I have no idea how we should handle this. We could reference the protobuf package+message names instead ?

controllers/schema_controller.go
tristantarrant: added a check

test/e2e/schema/schema_test.go
ryanemerson: > Nothing

Can you be more specific? If a REST call is made to remove the schema and it is being depended on, is an error thrown 4xx HTTP response etc

> These integrity aspects should be handled on the server before we approach them from the operator angle

That would be ideal!

But the reason for the question is that we might be able to leverage [Kubernetes Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) on the Schema CR to ensure that the resource is not removed until the dependent Cache CR(s) are removed.

tristantarrant: if a REST call is made to remove the schema, it will remove the schema.

ryanemerson: and what state does that leave the cache in?

tristantarrant: The data can still be read/written as `byte[]`. It only means it cannot be transcoded (for example to show it as JSON in the console), queried (even indexless) nor indexed.

ryanemerson: This does seem like a good fit for Finalizers, as it would work for all Operand versions and when the ConfigListener is disabled.

The K8s docs have a good example of how finalizers can work:

> A common example of a finalizer is kubernetes.io/pv-protection, which prevents accidental deletion of PersistentVolume objects. When a PersistentVolume object is in use by a Pod, Kubernetes adds the pv-protection finalizer. If you try to delete the PersistentVolume, it enters a Terminating status, but the controller can't delete it because the finalizer exists. When the Pod stops using the PersistentVolume, Kubernetes clears the pv-protection finalizer, and the controller deletes the volume.

In this case the Schema PR is the PersistentVolume and the Pod is a Cache CR.

api/v2alpha1/schema_types_test.go
```diff
@@ -0,0 +1,106 @@
...
+		{
+			name: "Appends .proto suffix if missing from spec.Name",
+			schema: Schema{
+				ObjectMeta: metav1.ObjectMeta{Name: "my-schema"},
+				Spec:       SchemaSpec{Name: "person"},
+			},
+			expected: "person.proto",
+		},
+		{
+			name: "Does not double-append .proto suffix",
```
Crumby, line 63: The test is identical to `Uses spec.Name when set` test

api/v2alpha1/schema_types_test.go
```diff
@@ -0,0 +1,106 @@
...
+			schema: Schema{
+				ObjectMeta: metav1.ObjectMeta{Name: "my-schema"},
+				Spec:       SchemaSpec{Name: "person.proto"},
+			},
+			expected: "person.proto",
+		},
+		{
+			name: "Appends .proto suffix to ObjectMeta.Name",
```
Crumby: Basically identical to `Falls back to ObjectMeta.Name`

api/v2alpha1/cache_types.go
```diff
@@ -106,6 +115,16 @@ type CacheList struct {
 	Items           []Cache `json:"items"`
 }
 
+// HasSchemaRef returns true if the Cache has a SchemaRef with the given name
+func (c *Cache) HasSchemaRef(name string) bool {
```
Crumby: Any particular reason for the function for Cache to be defined on `cache_types.go`? `*._types.go` files so far contained only data structures and the rest of the helper functions on the types are defined in the `types_util.go` file.

api/v2alpha1/schema_types.go
```diff
@@ -0,0 +1,71 @@
...
+
+type SchemaConditionType string
+
+const (
+	SchemaConditionReady             SchemaConditionType = "Ready"
+	SchemaConditionBidirectionalSync SchemaConditionType = "BidirectionalSync"
+)
+
+// SchemaSpec defines the desired state of Schema
+type SchemaSpec struct {
```
Crumby, line 11: Looks like we are missing update strategy we have for Cache CRs (recreate vs retain), is that intentional?

api/v2alpha1/schema_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+		For(s).
+		Complete()
+}
+
+// +kubebuilder:webhook:path=/validate-infinispan-org-v2alpha1-schema,mutating=false,failurePolicy=fail,sideEffects=None,groups=infinispan.org,resources=schemas,verbs=create;update,versions=v2alpha1,name=vschema.kb.io,admissionReviewVersions={v1,v1beta1}
+
+var _ webhook.Validator = &Schema{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (s *Schema) ValidateCreate() error {
```
Crumby: It's possible to create two Schema CRs that both target the same cluster with the same name. The second will override the first one but both will have condition `Ready` set to `true`. The Cache Webhook validator checks for this clashes and will reject creation of a Cache CR with the already existing combination

config/crd/bases/infinispan.org_caches.yaml
Crumby: We could have `name` and `type` where the type would be either `CustomResource/SchemaCR`, `ServerKey` under which it's stored (eg. `test-schema-cr.proto`), or a package/message combination.

We should ask ourselves how do we want users to use this feature and if it makes sense to implement all or if it's over the top. I don't have an answer to that one though

config/samples/schema/infinispan_v2alpha1_schema.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: infinispan.org/v2alpha1
+kind: Schema
+metadata:
+  name: person-schema
```
Crumby: `example-schema` for consistency with the rest and if I recall correctly, it's in the best practices

controllers/schema_controller.go
```diff
@@ -0,0 +1,573 @@
...
+				}
+
+				for _, item := range schemaList.Items {
+					requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{Namespace: item.GetNamespace(), Name: item.GetName()}})
+				}
+				return requests
+			}),
+	)
+	builder.Watches(
+		&source.Kind{Type: &v2alpha1.Cache{}},
```
Crumby: What's the purpose of triggering Schema reconciliation when Cache is modified?

documentation/asciidoc/topics/proc_creating_caches.adoc
```diff
@@ -17,6 +17,14 @@ If you do not specify a name with the `spec.name` field then the cache uses the
 ====
 +
 . Add a cache configuration with the `spec.template` field.
+. Optionally, if your cache uses `application/x-protostream` encoding, specify `Schema` CR dependencies with the `spec.schemaRefs` field.
+When set, {ispn_operator} waits until all referenced `Schema` CRs are ready before creating the cache.
++
+[source,options="nowrap",subs=attributes+]
+----
+include::yaml/cache_schema_ref.yaml[]
```
Crumby: Including this yaml example results in having a massive yaml file as a first Cache CR example altogether with Schema CR followed up with several smaller ones. Wouldn't it be better for the readability to mention the ability to specify dependencies in a separate follow up section the same way we do for updateStrategies?

test/e2e/schema/schema_test.go
```diff
@@ -0,0 +1,491 @@
...
+	testKube.DeleteSchema(schema)
+
+	// Assert that the schema is removed from the server once the Schema CR has been deleted
+	schemaHelper.WaitForSchemaToNotExist()
+	assertConfigListenerHasNoErrorsOrRestarts(t, ispn)
+}
+
+// TestUpdateSchemaCR verifies that updating the schema content in a Schema CR
+// propagates the change to the server.
+func TestUpdateSchemaCR(t *testing.T) {
```
Crumby: Should we merge this test and `TestSchemaCR` into a single CRUD test? I know this is a parallel to what Cache CR tests have but in my opinion it would make sense to have that one merged as well. K8s integration tests are time consuming

test/e2e/schema/schema_test.go
```diff
@@ -0,0 +1,491 @@
...
+	httpClient = tutils.HTTPClientForCluster(ispn, testKube)
+	schemaHelper = tutils.NewSchemaHelper(schemaName, httpClient)
+	schemaHelper.WaitForSchemaToExist()
+}
+
+// TestCacheSchemaDependency verifies the dependency between Cache and Schema CRs:
+// 1. A Cache CR referencing a non-existent schema should fail (not ready, not on server)
+// 2. Creating the Schema CR should unblock the Cache CR (both become ready on server)
+// 3. Data can be stored and retrieved using the proto types
+func TestCacheSchemaDependency(t *testing.T) {
```
Crumby: TestCacheSchemaDependency and it's multischema variant tests should live in between the cache tests. It's Cache CR spec and tests cache controller behavior.

TestSchemaInUseProtection on the other should probably stay here as it's testing schema controller behavior.

test/e2e/schema/schema_test.go
```diff
@@ -0,0 +1,491 @@
...
+
+message Person {
+  required string name = 1;
+  required int32 id = 2;
+  optional string email = 3;
+  optional string phone = 4;
+}
+`
+
+func TestMain(m *testing.M) {
```
Crumby, line 51: Disabling ConfigListener while having server side created Schemas will keep the respective Schema CRs hanging around. If the ConfigListeners is disable we should remove all the Schema CRs created and managed by the ConfigListener as we do with Caches

api/v2alpha1/schema_types_test.go
```diff
@@ -0,0 +1,106 @@
...
+				ObjectMeta: metav1.ObjectMeta{Name: "person"},
+				Spec:       SchemaSpec{},
+			},
+			expected: "person.proto",
+		},
+		{
+			name: "Appends .proto suffix if missing from spec.Name",
+			schema: Schema{
+				ObjectMeta: metav1.ObjectMeta{Name: "my-schema"},
+				Spec:       SchemaSpec{Name: "person"},
```
Crumby, line 36: If you create a spec with `name: person` it'll be represented as `person.proto` on the server. If you then try to change the spec on the server side the ConfigListener will then fail to update the Schema CR with:  

`ERROR Error encountered for event 'update-schema': unable to Update Schema CR 'broken-schema': admission webhook "vschema.kb.io" denied the request: Schema.infinispan.org "broken-schema" is invalid: spec.name: Forbidden: Schema name is immutable and cannot be updated after initial Schema creation`

We should probably sanitize the `spec.name` in a way so it always contains `.proto` suffix, if the suffix is required by the server.

Crumby (changes requested):

test/e2e/schema/schema_test.go
Crumby: Last changes do not remove the Schema if it's being depended on. On the other if you just remove it directly from the server then it's wiped out (Both CR and from the server). Tbh, i'm not sure if it's good to have this detection on the Operator level but not the server side...

Anyway, I'm also wondering if we could hit an issue where we'd need to recreate a Schema (for what ever reason) but keep the cache around. With the current implementation this will be prevented

api/v2alpha1/schema_types.go
tristantarrant: The recreate/retain update strategy makes sense for Caches because updating a cache definition
  can be destructive (data loss). For schemas, a PUT to the server simply replaces the proto content. I don't think there is a meaningful distinction between "recreate" and "retain". Schemas are idempotent by nature.

controllers/schema_controller.go
tristantarrant: This watch is for the Schema deletion protection flow. 
When a Schema CR is being deleted, the  reconciler calls findReferencingCaches() and blocks deletion if any Cache CRs still reference it. Without it, when  the Cache later removes its schemaRef (or is deleted itself), the Schema controller wouldn't be notified to re-reconcile  and complete the deletion.

documentation/asciidoc/topics/proc_creating_caches.adoc
tristantarrant: Yeah, horrid

test/e2e/schema/schema_test.go
tristantarrant: agreed

api/v2alpha1/schema_types_test.go
tristantarrant: I've removed this one and "expanded" the name of the other to mention this.

api/v2alpha1/schema_types_test.go
tristantarrant: I've removed this one and "expanded" the name of the other to mention this.

api/v2alpha1/cache_types.go
tristantarrant: Moved it

api/v2alpha1/schema_webhook.go
tristantarrant: Should be fixed

config/samples/schema/infinispan_v2alpha1_schema.yaml
tristantarrant: Fixed

test/e2e/schema/schema_test.go
tristantarrant: Fixed

test/e2e/schema/schema_test.go
tristantarrant: This is already handled. The RemoveConfigListener function lists all Schema CRs and marks any that are owned by the Infinispan instance with the ListenerAnnotationDelete annotation, exactly as it does for Cache. So when ConfigListener is disabled, the server-side-created Schema CRs will be cleaned up.

api/v2alpha1/schema_types_test.go
tristantarrant: I believe this scenario is actually handled correctly. The immutability check in the webhook uses GetSchemaName() on both old and new schemas, which normalizes by appending .proto if missing. So `person` → `person.proto` and `person.proto` → `person.proto` compare as equal, and the update goes through.
That said, I agree it would be cleaner to sanitize `spec.name` at creation time (either in the validating webhook's Create method or via a defaulting webhook) to always store the `.proto` suffix. That way the stored value matches what the server  uses and we avoid any confusion. I'll add that.

tristantarrant:
Rebased and green

pkg/reconcile/pipeline/infinispan/handler/provision/config_listener.go
```diff
@@ -138,7 +138,7 @@ func ConfigListener(i *ispnv1.Infinispan, ctx pipeline.Context) {
 			Rules: []rbacv1.PolicyRule{
 				{
 					APIGroups: []string{v2alpha1.GroupVersion.Group},
-					Resources: []string{"caches"},
+					Resources: []string{"caches", "schemas"},
```
Crumby, line 141: We need to update ServiceAccount documentation to include "schemas" in the RBAC

controllers/schema_controller.go
```diff
@@ -0,0 +1,573 @@
...
+	if err != nil {
+		return err
+	}
+
+	k8sClient := sl.Kubernetes.Client
+
+	if existingSchema == nil {
+		// Create new Schema CR with a deterministic name derived from the cluster and schema names
+		sanitizedName := schemaNameRegexp.ReplaceAllString(strcase.ToKebab(schemaName), "-")
+		resourceName := clusterName + "-" + sanitizedName
```
Crumby: We should utilize `meta.generateName` instead of build name with config listener, the same way we do with caches. One reason is obviously consistency with caches. The other to really prevent conflicts. There may be already existing schema which will have the name the same as the one generated.

controllers/cache_controller.go
```diff
@@ -196,6 +221,30 @@ func (r *CacheReconciler) Reconcile(ctx context.Context, request ctrl.Request) (
...
+				if errors.IsNotFound(err) {
+					reqLogger.Info(fmt.Sprintf("Schema %s not found, requeuing", schemaRef.Name))
+					return ctrl.Result{Requeue: true}, cache.update(func() error {
+						instance.SetCondition(v2alpha1.CacheConditionReady, metav1.ConditionFalse, fmt.Sprintf("Schema %s not found", schemaRef.Name))
+						return nil
+					})
+				}
+				return ctrl.Result{}, err
+			}
+			if schemaInstance.GetCondition(v2alpha1.SchemaConditionReady).Status != metav1.ConditionTrue {
```
Crumby: Looks like we are assuming the Schema is configured to target the same cluster. Should we add a check that the clusterNames match?

controllers/schema_controller.go
```diff
@@ -0,0 +1,573 @@
...
+}
+
+func (r *schemaRequest) findReferencingCaches() ([]string, error) {
+	cacheList := &v2alpha1.CacheList{}
+	if err := r.List(r.ctx, cacheList, &client.ListOptions{Namespace: r.schema.Namespace}); err != nil {
+		return nil, fmt.Errorf("unable to list Cache CRs: %w", err)
+	}
+	var names []string
+	for _, c := range cacheList.Items {
+		if c.HasSchemaRef(r.schema.Name) {
```
Crumby: Similar to the one in the cache_controller, should we also filter based on the matching clusterName?

controllers/schema_controller.go
```diff
@@ -0,0 +1,573 @@
...
+				if !errors.IsNotFound(err) {
+					return fmt.Errorf("unable to mark Schema '%s' for deletion: %w", s.Name, err)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func (sl *SchemaListener) CreateOrUpdate(data []byte) error {
```
Crumby, line 415: Found by an accident: If user creates a second proto schema on the server with the same package and message, the server marks the schema as invalid. It'll be marked as Ready on the Schema CR though...

controllers/schema_controller.go
```diff
@@ -0,0 +1,573 @@
...
+			return errors.NewNotFound(schema.ParseGroupResource("schemas.infinispan.org"), s.Name)
+		}
+		if s.Annotations == nil {
+			s.Annotations = make(map[string]string, 1)
+		}
+		s.Annotations[constants.ListenerAnnotationDelete] = "true"
+		return nil
+	})
+	if !errors.IsNotFound(err) {
+		sl.Log.Debugf("Schema CR '%s' not found, nothing todo.", s.Name)
```
Crumby: This line is confusing. If the error is **not** NotFound we debug log not found and return err?

controllers/schema_controller.go
Crumby: Looking at the schema event, it contains errors, is that something to take into account?

Crumby (changes requested):
Looks great, few requested changes, mainly the usage of generatedName and updating the ServiceAccount docs. I'll leave the rest to your judgment

api/v2alpha1/schema_webhook_test.go
```diff
@@ -0,0 +1,185 @@
+package v2alpha1
+
+import (
+	"errors"
+	"time"
+
+	. "github.com/onsi/ginkgo"
```
Crumby:
```suggestion
	. "github.com/onsi/ginkgo/v2"
```

controllers/schema_controller_test.go
```diff
@@ -0,0 +1,59 @@
...
+		{
+			name:      "Empty object",
+			data:      `{}`,
+			expectErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			name, proto, err := unmarshallSchemaEventConfig([]byte(tt.data))
```
Crumby:
```suggestion
			name, proto, _, err := unmarshallSchemaEventConfig([]byte(tt.data))
```

Crumby (changes requested):

tristantarrant:
@Crumby it's green

controllers/schema_controller.go
```diff
@@ -0,0 +1,605 @@
...
+	// Use PUT to create or update the schema
+	response, err := schemaClient.CreateOrUpdate(r.schema.Spec.Schema)
+	if err != nil {
+		err = fmt.Errorf("unable to create or update schema '%s': %w", schemaName, err)
+		r.reqLogger.Error(err, "")
+		return &ctrl.Result{Requeue: true}, err
+	}
+
+	if response.Error != nil {
+		err = fmt.Errorf("schema '%s' has errors: %s", schemaName, response.Error.Message)
```
Crumby: If you create a schema with errors, then the message in the condition will look like:

```
 - message: 'schema ''person2.proto'' has errors: Schema person2.proto has errors'
```

And then the error log message:

```
... "cause": "Duplicate definition of org.infinispan.Person in person.proto and person2.proto", "error": "schema 'person2.proto' has errors: Schema person2.proto has errors"}
```

So maybe we should build err as?
```
err = fmt.Errorf("'%s', %s", response.Error.Message, response.Error.Cause)
```

controllers/schema_controller.go
```diff
@@ -0,0 +1,605 @@
...
+			sl.Log.Errorf("Conflict encountered on Schema CR '%s' update. Retry %d..%d", existingSchema.Name, i, maxRetries)
+		}
+		if err != nil {
+			return fmt.Errorf("unable to Update Schema CR %s after %d attempts", existingSchema.Name, maxRetries)
+		}
+
+		if schemaErrors != "" {
+			sl.Log.Warnf("Schema '%s' has server-side errors: %s", schemaName, schemaErrors)
+			if err := updateSchema(existingSchema, sl.Ctx, k8sClient, func() error {
+				existingSchema.SetCondition(v2alpha1.ConditionReady, metav1.ConditionFalse, schemaErrors)
```
Crumby: This status write by ConfigListener fails due to missing permissions. Not sure if `schemas/status: patch` is enough or more verbs are required for different lifecycle phases of the object

```
ERROR Error encountered for event 'update-schema': unable to update schema example-schema-not: schemas.infinispan.org "example-schema-not" is forbidden: User "system:serviceaccount:openshift-operators:example-infinispan-config-listener" cannot patch resource "schemas/status" in API group "infinispan.org" in the namespace "openshift-operators"
```

controllers/schema_controller.go
```diff
@@ -0,0 +1,605 @@
...
+				return ctrl.Result{}, nil
+			}
+			return ctrl.Result{}, err
+		}
+		reqLogger.Info("Schema CR Removed.")
+		return ctrl.Result{}, nil
+	}
+
+	crDeleted := instance.GetDeletionTimestamp() != nil
+
```
Crumby, line 165: Here was a piece of code to handle reconcilation pausing, was the capability excluded intentionally?

https://github.com/infinispan/infinispan-operator/compare/b036cff1c7b2d8fee083d8ee12d2f09f40a79f10..5b6d370b5cb782f002a1342ad27c46c9953aaa34#diff-89718a8c6c1e2e048a8c7cca5def12d516abd5263cf6ccb715bec91b646ea907L167

controllers/schema_controller.go
Crumby: Don't forget to update both docs reference and the automatically created service account

Crumby (changes requested):
@tristantarrant Few more issues around error handling and missing reconciliation pausing feature in the controller.

controllers/schema_controller.go
```diff
@@ -0,0 +1,611 @@
...
+		sl.Log.Infof("Creating Schema CR for '%s'", schemaName)
+		if err := k8sClient.Create(sl.Ctx, s); err != nil {
+			return fmt.Errorf("unable to create Schema CR for schema '%s': %w", schemaName, err)
+		}
+		sl.Log.Infof("Schema CR '%s' created", s.Name)
+
+		if schemaErrors != "" {
+			sl.Log.Warnf("Schema '%s' has server-side errors: %s", schemaName, schemaErrors)
+			if err := updateSchema(s, sl.Ctx, k8sClient, func() error {
```
Crumby: Now that  the CL has the authorization to patch the status there's an infinite loop once you update/create the schema with error. The failed schema set the condition and increase the generation. Now the resource generation is greater then the one set by the config listener so the Operator thinks it's the user change and updates the server, which then notifies the CL which updates the resource.

I think you need to +1 increase 'constants.ListenerAnnotationGeneration' on the resource with each update call from the configlistner (here and in the else branch). (Other solution would be excluding status changes from the watch)

Crumby (changes requested):

controllers/schema_controller.go
tristantarrant: Done

controllers/schema_controller.go
```diff
@@ -0,0 +1,613 @@
...
+			}
+			sl.Log.Errorf("Conflict encountered on Schema CR '%s' update. Retry %d..%d", existingSchema.Name, i, maxRetries)
+		}
+		if err != nil {
+			return fmt.Errorf("unable to Update Schema CR %s after %d attempts", existingSchema.Name, maxRetries)
+		}
+
+		if schemaErrors != "" {
+			sl.Log.Warnf("Schema '%s' has server-side errors: %s", schemaName, schemaErrors)
+			if err := updateSchema(existingSchema, sl.Ctx, k8sClient, func() error {
```
Crumby: Now the ConfigListener fails crashes the entire pod and enters CrashLoopBackOff if you create Schema CR with invalid schema:

```
panic: assignment to entry in nil map
goroutine 72 [running]:
github.com/infinispan/infinispan-operator/controllers.(*SchemaListener).CreateOrUpdate.func3()
/workspace/controllers/schema_controller.go:499 +0x73
github.com/infinispan/infinispan-operator/controllers.updateSchema.func1()
/workspace/controllers/schema_controller.go:584 +0x67
sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.mutate(0x3091710?, {{0x1b9afaec2928?, 0x1d5c8fa?}, {0x1b9afae16c48?, 0x1eab1c0?}}, {0x1eca898, 0x1b9afb1b6580})
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.7/pkg/controller/controllerutil/controllerutil.go:442 +0x48
sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.CreateOrPatch({0x1eab1c0, 0x1b9afaefddb0}, {0x1ebb920, 0x1b9afaeb0750}, {0x1eca898, 0x1b9afb1b6580}, 0x1b9afae29948)
/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.7/pkg/controller/controllerutil/controllerutil.go:377 +0x4d2
github.com/infinispan/infinispan-operator/controllers.updateSchema(0x1b9afb1b6580, {0x1eab1c0?, 0x1b9afaefddb0?}, {0x1ebb920?, 0x1b9afaeb0750?}, 0x1d97b69?)
/workspace/controllers/schema_controller.go:580 +0x75
github.com/infinispan/infinispan-operator/controllers.(*SchemaListener).CreateOrUpdate(0x1b9afb40ae40, {0x1b9afb1fa000?, 0x0?, 0x1b9afb18b300?})
/workspace/controllers/schema_controller.go:498 +0xd3f
github.com/infinispan/infinispan-operator/launcher/listener.New.func4.1(0x1b9afb3f0000)
/workspace/launcher/listener/listener.go:168 +0x285
github.com/r3labs/sse/v2.(*Client).SubscribeWithContext.func1()
/opt/app-root/src/go/pkg/mod/github.com/r3labs/sse/v2@v2.10.0/client.go:108 +0x223
gopkg.in/cenkalti/backoff%2ev1.RetryNotify(0x1b9afae29ef0, {0x1e94a70, 0x1b9afb1657f0}, 0x1b9afb31c870)
/opt/app-root/src/go/pkg/mod/gopkg.in/cenkalti/backoff.v1@v1.1.0/retry.go:37 +0x156
github.com/r3labs/sse/v2.(*Client).SubscribeWithContext(0x0?, {0x1eab1c0?, 0x1b9afaefddb0?}, {0x0?, 0x1b9afae7c380?}, 0x1b9afaecc210?)
/opt/app-root/src/go/pkg/mod/github.com/r3labs/sse/v2@v2.10.0/client.go:116 +0x69
github.com/r3labs/sse/v2.(*Client).SubscribeRawWithContext(...)
/opt/app-root/src/go/pkg/mod/github.com/r3labs/sse/v2@v2.10.0/client.go:258
github.com/infinispan/infinispan-operator/launcher/listener.New.func4()
/workspace/launcher/listener/listener.go:158 +0x75
created by github.com/infinispan/infinispan-operator/launcher/listener.New in goroutine 1
/workspace/launcher/listener/listener.go:157 +0x1637

File: controllers/schema_controller.go (hunk 30 above)
Comment (Crumby) [reply controllers/schema_controller.go]: So, increasing the the annotation in case of status and annotation updates is not the solution. Annotation and status updates do not increase generation. So with this in place we now have an issue that follow up spec update on the Schema CR isn't reflected on the server

(Issue scenario: user creates failed schema, configlistener generation ends up on two but the object generation is still one. User updates the schema via CR, generation gets bumped to two but the Operator checks if the generation is greater, which it isn't so the updated schema isn't applied)

File: controllers/schema_controller.go [hunk 32]
```diff
@@ -0,0 +1,613 @@
...
+			return err
+		}
+
+		sl.Log.Infof("Creating Schema CR for '%s'", schemaName)
+		if err := k8sClient.Create(sl.Ctx, s); err != nil {
+			return fmt.Errorf("unable to create Schema CR for schema '%s': %w", schemaName, err)
+		}
+		sl.Log.Infof("Schema CR '%s' created", s.Name)
+
+		if schemaErrors != "" {
```
Crumby, line 463: The original issue with invalid schema created on the server side being marked as Ready is still in place. What I believe is happening is Operator and ConfigListener fighting for the Schema status. ConfigListener creates schema -> Operator sees Schema create, it checks the CL annotation skip server reconcilation, marks as ready -> CL then marks the resource not ready, with invalid schema configuration -> Operator gets an event Schema CR being updated and again skips the reconcilation on server and just marks it as ready.

It's traceable in the order in which the changes are applied to the resource after being created on the server side:
```
apiVersion: infinispan.org/v2alpha1
kind: Schema
metadata:
  annotations:
    infinispan.org/listener-generation: "1"
  creationTimestamp: "2026-08-31T10:07:04Z"
  finalizers:
  - finalizer.infinispan.org
  generateName: fail-proto-
  generation: 1
  name: fail-proto-6cwn8
  namespace: datagrid
  ownerReferences:
  - apiVersion: infinispan.org/v1
    kind: Infinispan
    name: infinispan
    uid: 0c267af9-3619-49bc-8845-979c3e647076
  resourceVersion: "71707"
  uid: f126658c-5364-408b-a41f-e744054f1706
spec:
  clusterName: infinispan
  name: fail.proto
  schema: |2-
        package org.infinispan;

        message Person {
            string name = 1;
            int32 age = 2;
            string email = 3;
            eaf
        }
---
apiVersion: infinispan.org/v2alpha1
kind: Schema
metadata:
  annotations:
    infinispan.org/listener-generation: "1"
  creationTimestamp: "2026-08-31T10:07:04Z"
  finalizers:
  - finalizer.infinispan.org
  generateName: fail-proto-
  generation: 1
  name: fail-proto-6cwn8
  namespace: datagrid
  ownerReferences:
  - apiVersion: infinispan.org/v1
    kind: Infinispan
    name: infinispan
    uid: 0c267af9-3619-49bc-8845-979c3e647076
  resourceVersion: "71708"
  uid: f126658c-5364-408b-a41f-e744054f1706
spec:
  clusterName: infinispan
  name: fail.proto
  schema: |2-
        package org.infinispan;

        message Person {
            string name = 1;
            int32 age = 2;
            string email = 3;
            eaf
        }
status:
  conditions:
  - status: "True"
    type: Ready
  - status: "True"
    type: BidirectionalSync
---
apiVersion: infinispan.org/v2alpha1
kind: Schema
metadata:
  annotations:
    infinispan.org/listener-generation: "2"
  creationTimestamp: "2026-08-31T10:07:04Z"
  finalizers:
  - finalizer.infinispan.org
  generateName: fail-proto-
  generation: 1
  name: fail-proto-6cwn8
  namespace: datagrid
  ownerReferences:
  - apiVersion: infinispan.org/v1
    kind: Infinispan
    name: infinispan
    uid: 0c267af9-3619-49bc-8845-979c3e647076
  resourceVersion: "71709"
  uid: f126658c-5364-408b-a41f-e744054f1706
spec:
  clusterName: infinispan
  name: fail.proto
  schema: |2-
        package org.infinispan;

        message Person {
            string name = 1;
            int32 age = 2;
            string email = 3;
            eaf
        }
status:
  conditions:
  - status: "True"
    type: Ready
  - status: "True"
    type: BidirectionalSync
---
apiVersion: infinispan.org/v2alpha1
kind: Schema
metadata:
  annotations:
    infinispan.org/listener-generation: "2"
  creationTimestamp: "2026-08-31T10:07:04Z"
  finalizers:
  - finalizer.infinispan.org
  generateName: fail-proto-
  generation: 1
  name: fail-proto-6cwn8
  namespace: datagrid
  ownerReferences:
  - apiVersion: infinispan.org/v1
    kind: Infinispan
    name: infinispan
    uid: 0c267af9-3619-49bc-8845-979c3e647076
  resourceVersion: "71710"
  uid: f126658c-5364-408b-a41f-e744054f1706
spec:
  clusterName: infinispan
  name: fail.proto
  schema: |2-
        package org.infinispan;

        message Person {
            string name = 1;
            int32 age = 2;
            string email = 3;
            eaf
        }
status:
  conditions:
  - message: 'org.infinispan.protostream.DescriptorParserException: Syntax error in
      fail.proto at 8:5: unexpected label: }'
    status: "False"
    type: Ready
---
apiVersion: infinispan.org/v2alpha1
kind: Schema
metadata:
  annotations:
    infinispan.org/listener-generation: "2"
  creationTimestamp: "2026-08-31T10:07:04Z"
  finalizers:
  - finalizer.infinispan.org
  generateName: fail-proto-
  generation: 1
  name: fail-proto-6cwn8
  namespace: datagrid
  ownerReferences:
  - apiVersion: infinispan.org/v1
    kind: Infinispan
    name: infinispan
    uid: 0c267af9-3619-49bc-8845-979c3e647076
  resourceVersion: "71711"
  uid: f126658c-5364-408b-a41f-e744054f1706
spec:
  clusterName: infinispan
  name: fail.proto
  schema: |2-
        package org.infinispan;

        message Person {
            string name = 1;
            int32 age = 2;
            string email = 3;
            eaf
        }
status:
  conditions:
  - status: "True"
    type: Ready
  - status: "True"
    type: BidirectionalSync

**Review by Crumby — CHANGES_REQUESTED** [review]
@tristantarrant So increasing the CL annotation wasn't the way to go and probably not even the source of the issue. Another issues popped up

File: controllers/schema_controller.go (hunk 30 above)
Comment (Crumby) [reply controllers/schema_controller.go]: I wonder if should let the Operator to always fetch the status of the schema and just keep the CL for creating/updating the meta/spec part of the schema

Comment (tristantarrant) [reply controllers/schema_controller.go]: Agreed — pushed a commit that does exactly this. The Operator now always calls `checkSchemaStatus()` to fetch the schema status from the server before setting the Ready condition, and the CL no longer touches status at all (just logs a warning if `schemaErrors` is non-empty).

This fixes all three issues:
- **Nil map panic**: the code paths that set annotations in the error-handling closures are removed
- **Generation tracking**: the CL no longer bumps `listener-generation` on status-only updates, so the annotation stays in sync with actual spec generation
- **Status race**: only the Operator writes status conditions, so the CL can't overwrite them

Also fixed a potential nil map in `RemoveStaleResources` when marking stale schemas for deletion.

File: controllers/schema_controller.go (hunk 31 above)
Comment (tristantarrant) [reply controllers/schema_controller.go]: Fixed — the code paths that were doing `s.Annotations[...] = ...` inside `updateSchema` closures are removed entirely. The CL no longer sets status or bumps the listener-generation annotation after the initial create/update, so these closures are gone.

Also added a nil-map guard in `RemoveStaleResources` which had the same pattern.

File: controllers/schema_controller.go (hunk 30 above)
Comment (tristantarrant) [reply controllers/schema_controller.go]: Fixed by the same commit — the CL no longer bumps `listener-generation` on status-only updates (those `updateSchema` calls are removed). The annotation is now only set in two places:
- On create: hardcoded to `"1"` (matching initial generation)
- On spec update: set to `generation+1` inside the `CreateOrPatch` mutate function, which is a spec change that *does* bump generation

So the annotation and generation stay in sync.

File: controllers/schema_controller.go (hunk 32 above)
Comment (tristantarrant) [reply controllers/schema_controller.go:463]: Fixed — the Operator is now the sole owner of status. Every reconciliation calls `checkSchemaStatus()` which fetches the schema's status from the server before setting the Ready condition. The CL only logs a warning when `schemaErrors` is non-empty but does not touch the CR's status. This eliminates the race entirely.

File: pkg/reconcile/pipeline/infinispan/handler/provision/config_listener.go [hunk 33]
```diff
@@ -138,7 +138,7 @@ func ConfigListener(i *ispnv1.Infinispan, ctx pipeline.Context) {
 			Rules: []rbacv1.PolicyRule{
 				{
 					APIGroups: []string{v2alpha1.GroupVersion.Group},
-					Resources: []string{"caches"},
+					Resources: []string{"caches", "caches/status", "schemas", "schemas/status"},
```
Crumby:
```suggestion
					Resources: []string{"caches", "schemas"},
```

documentation/asciidoc/topics/yaml/config_listener_rbac.yaml
```diff
@@ -11,7 +11,7 @@ metadata:
   namespace: <namespace>
 rules:
   - apiGroups: ["infinispan.org"]
-    resources: ["caches"]
+    resources: ["caches", "caches/status", "schemas", "schemas/status"]
```
Crumby:
```suggestion
    resources: ["caches", "schemas"]
```

Crumby (commented):
@tristantarrant Let's remove the previously added `resource/status` verbs to Config Listener (and documentation) added previously and we are good to go.

Crumby (approved):

Crumby:
Great job, thanks @tristantarrant!

The test failure is unrelated to the PR, fails only on the latest dev version.

PR #2458, opened 2026-03-09
Fix flaky TestCacheResourcesCleanedUpOnDisable

### LINKED ISSUE #2457: Fix flaky TestCacheResourcesCleanedUpOnDisable
## Description

`TestCacheResourcesCleanedUpOnDisable` is flaky due to a race condition between the test and the Infinispan controller's `RemoveConfigListener` cleanup.

## Root Cause

After disabling the ConfigListener, `RemoveConfigListener` marks all Cache CRs owned by the Infinispan CR with `ListenerAnnotationDelete`. The test waits for the first CR to be removed, then immediately creates a new CR with the same owner reference. However, the Infinispan controller can re-reconcile (triggered by the new CR creation) and re-run `RemoveConfigListener`, which marks the new CR for deletion before the cache controller can set it to Ready — causing a timeout at `WaitForCacheConditionReady`.

## Fix

Wait for the Infinispan controller to complete its reconcile cycle (`WaitForInfinispanCondition(WellFormed)`) after the first CR cleanup, before creating the new Cache CR.

## Example failure

https://github.com/infinispan/infinispan-operator/actions/runs/22845471413/job/66261358906?pr=2452

tristantarrant:
## Summary
- Fix race condition in `TestCacheResourcesCleanedUpOnDisable` where the Infinispan controller re-runs `RemoveConfigListener` and marks a newly created Cache CR for deletion before the test can observe it as Ready
- Add `WaitForInfinispanCondition(WellFormed)` after listener cleanup to let the controller settle before creating the new CR

Fixes #2457

### DIFF
--- a/test/e2e/cache/cache_test.go
+++ b/test/e2e/cache/cache_test.go
@@ -532,6 +532,12 @@ func TestCacheResourcesCleanedUpOnDisable(t *testing.T) {
 	// Assert that the listener created Cache CR is removed by the Infinispan controller when the ConfigListener is disabled
 	testKube.WaitForResourceRemoval(cache.Name, tutils.Namespace, &v2alpha1.Cache{})
 
+	// Wait for the Infinispan controller to finish its reconcile cycle after cleaning up
+	// listener resources. Without this, the controller may re-run RemoveConfigListener on a
+	// subsequent reconcile and mark the newly created CR for deletion before the test can
+	// observe it as Ready.
+	testKube.WaitForInfinispanCondition(ispn.Name, ispn.Namespace, v1.ConditionWellFormed)
+
 	// Manually recreate the Cache CR and set its owner reference
 	cr := cacheCR(cacheName, ispn)
 	cr.ObjectMeta.Annotations = map[string]string{constants.ListenerAnnotationGeneration: "1"}


test/e2e/cache/cache_test.go
```diff
@@ -532,6 +532,12 @@ func TestCacheResourcesCleanedUpOnDisable(t *testing.T) {
 	// Assert that the listener created Cache CR is removed by the Infinispan controller when the ConfigListener is disabled
 	testKube.WaitForResourceRemoval(cache.Name, tutils.Namespace, &v2alpha1.Cache{})
 
+	// Wait for the Infinispan controller to finish its reconcile cycle after cleaning up
+	// listener resources. Without this, the controller may re-run RemoveConfigListener on a
+	// subsequent reconcile and mark the newly created CR for deletion before the test can
+	// observe it as Ready.
+	testKube.WaitForInfinispanCondition(ispn.Name, ispn.Namespace, v1.ConditionWellFormed)
```
Crumby: This wait adds nothing to the test as it's already checked at the beginning of the test (in the `initCluster` func). WellFormed condition is then true for the remainder of the test (as it rather represent correct cluster view status rather then reconciliation), so when it reaches this point it just confirms that it's really set (there's no real waiting).

Crumby:
The wait does not change the test behavior as the condition is already checked before

PR #2461, opened 2026-03-11
feat: support propagation of external labels (e.g. Kyverno) to StatefulSet pod selector 

mcarbonneaux:
## Summary

This change allows the Infinispan operator to handle labels injected by Kyverno (or any external controller) by enabling them to be included in the StatefulSet's immutable pod selector.

## Changes

- Added a new internal annotation `infinispan.org/podSelectorLabels` in `api/v1/types_util.go`.
- Updated `PodSelectorLabels()` to include labels listed in this annotation when they are present on the Infinispan CR.
- Updated `ServiceSelectorLabels()` to include those same labels, ensuring consistent pod discovery across services (ping, etc.).
- Together, these changes allow users to explicitly declare which CR labels should be propagated to both Pod labels and the StatefulSet's immutable `Spec.Selector.MatchLabels`.

## Notes

To use this feature, add the desired label to your Infinispan CR and list it in the `infinispan.org/podSelectorLabels` annotation:
```yaml
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: example-infinispan
  labels:
    my-kyverno-label: my-value
  annotations:
    infinispan.org/podSelectorLabels: my-kyverno-label
spec:
  replicas: 3
```

> ⚠️ Because `Spec.Selector` is immutable on StatefulSets, the labels listed in this annotation must be set on the CR **before** the StatefulSet is first created. Changing them on an existing StatefulSet will require its deletion and recreation.

### DIFF
--- a/.github/workflows/image_release.yaml
+++ b/.github/workflows/image_release.yaml
@@ -2,7 +2,7 @@ name: Image Release
 
 on:
   schedule:
-    - cron: '30 4 * * *'
+    - cron: '0 0 * * *'
 
   workflow_dispatch:
     inputs:
@@ -31,26 +31,54 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v6
         with:
-          fetch-depth: 2
+          fetch-depth: 0
           ref: ${{ github.event.inputs.ref || 'main' }}
           repository: ${{ github.repository }}
           token: ${{ secrets.INFINISPAN_RELEASE_TOKEN }}
           path: operator
 
       - name: Check if new operand added
+        id: check_operand
         working-directory: operator
-        id: check
         run: |
-          PREV_COMMIT=$(git rev-parse HEAD~1 2>/dev/null || echo HEAD)
-          if git diff --quiet $PREV_COMMIT HEAD -- config/manager/manager.yaml; then
-            echo "changed=false" >> $GITHUB_OUTPUT
-          else
+          LAST_TAG=$(git describe --tags --abbrev=0)
+          echo "Last release tag: $LAST_TAG"
+          ADDED=$(git diff $LAST_TAG HEAD -- config/manager/manager.yaml | grep '^+' | grep '"upstream-version"')
+          echo "Added operand:"
+          echo "$ADDED"
+          if [ -n "$ADDED" ]; then
             echo "changed=true" >> $GITHUB_OUTPUT
+          else
+            echo "changed=false" >> $GITHUB_OUTPUT
           fi
+
+      - name: Check open PRs from forks repository
+        id: check_pr_status
+        env:
+          GH_TOKEN: ${{ secrets.FORK_PAT }}
+        run: |
+          PR1=$(gh pr list \
+            --repo infinispan-qe-bot/community-operators \
+            --state open \
+            --json number \
+            --jq 'length')
+          PR2=$(gh pr list \
+            --repo infinispan-qe-bot/community-operators-prod \
+            --state open \
+            --json number \
+            --jq 'length')
+          echo "Repo infinispan-qe-bot/community-operators PR count: $PR1"
+          echo "Repo infinispan-qe-bot/community-operators-prod  PR count: $PR2"
+          if [ "$PR1" -gt 0 ] || [ "$PR2" -gt 0 ]; then
+            echo "skip=true" >> $GITHUB_OUTPUT
+          else
+            echo "skip=false" >> $GITHUB_OUTPUT
+          fi  
+
 
   release:
     needs: check_operand
-    if: needs.check_operand.outputs.changed == 'true' || github.event_name == 'workflow_dispatch'
+    if: (needs.check_operand.outputs.changed == 'true' && needs.check_pr_status.outputs.skip != 'true') || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     steps:
       - name: Configure Git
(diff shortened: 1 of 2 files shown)


mcarbonneaux:
The existing CRDs for labels (like `spec.service.metadata`) only affect the **metadata** of auxiliary resources. They do **not** solve the 'Infinispan-Operator vs. Admission Controller' race condition.

Here is the technical deadlock we face:

1.  **The Trigger:** An enterprise policy (Kyverno) is configured to mutate all workloads to add mandatory labels (e.g., `security-zone`, `cost-center`).
2.  **The Conflict:** When the Operator creates the `StatefulSet`, Kyverno's Webhook injects these labels into the `spec.template.metadata.labels`.
3.  **The Failure:** Since the Operator didn't know about these labels, it didn't include them in the `spec.selector.matchLabels`.
4.  **The Result:** Kubernetes rejects the `StatefulSet` creation because the **Selector must match the Pod Template labels**.
5.  **The Immutability Trap:** Even if we try to fix it later, the `spec.selector` is **immutable**. We cannot 'patch' it.



**Why the current CRD implementation isn't enough:**

The Operator must be aware of these 'mandatory' labels *at the moment of the initial API call* to build a valid `StatefulSet` object where `selector` == `template.labels`. If the Operator doesn't allow us to define these in the Infinispan CR, we are forced to disable security policies cluster-wide, which is not an option in production environments."

Crumby:
> 4. The Result: Kubernetes rejects the StatefulSet creation because the Selector must match the Pod Template labels.

@mcarbonneaux The `spec.selector.matchLabels` needs to contain a subset of labels defined by `spec.template.metadata.labels`. It does not need to match it 1:1.

If you create Infinsipan CR with:
```
annotations:
  infinispan.org/podTargetLabels: 'pod-label1, pod-label2'
labels:
    pod-label1: value1
    pod-label2: value2
```

then you can see the first SS revision created by the Operator does not have 1:1 matching labels:
```
spec:
  selector:
    matchLabels:
      app: infinispan-pod
      app.kubernetes.io/created-by: infinispan
      clusterName: infinispan
      infinispan_cr: infinispan
  template:
    metadata:
      labels:
        app: infinispan-pod
        app.kubernetes.io/created-by: infinispan
        clusterName: infinispan
        infinispan_cr: infinispan
        pod-label1: value1
        pod-label2: value2
```

Anyway, looking at the issue as described on https://github.com/infinispan/infinispan-helm-charts/pull/146, it looks like the issue is not addition of a new labels but rather modification of existing ones that the Infinispan Operator uses as a selector (`app`, `clusterName`).

I've briefly tested the PR and there's several issues I have with it, assuming you are changing the values of the selector labels (if not, then I'd reconfigure your Kyverno policies to work with `infinispan.org/podTargetLabels`):
* It does not account for the changed selectors in affinity configuration, which break the defaults as it relies on original labels
* `app` label is used in config listener and gossip router deployments (and related services) as well with specific value, if your Kyverno policy changes the `app` label for all the resources you might break those two deployments
* The whole design feels quite cumbersome. Aside from having to define the policy and then configure the overwrite labels on the Infinsipan CR so they get properly propagated when injected by Kyverno, I can't really tell what the impact will be during cluster lifecycle (upgrades) and how it'd interact with other functionality (backup-restore, caches). It'd need thorough testing.

 I do understand though that having general labels (`app`, `clusterName`) as a selector isn't optimal and I like the solution type you applied for the Helm Charts. So if we'd implement second version of the labels that would be namespaced help you?

Instead of:
``` 
app: infinispan-pod
app.kubernetes.io/created-by: infinispan
clusterName: infinispan
infinispan_cr: infinispan
```

Something like:
```
* app.kubernetes.io/name: infinispan
* app.kubernetes.io/instance: cr.Name
* app.kubernetes.io/managed-by: infinispan-operator
```
or
```
* operator.infinispan.org/name: infinispan
* operator.infinsipan.org/instance: cr.Name
```

Obviously implementation here won't be simple either as we would have ensure support for previous set of labels alongside a new ones as well as allow for a migration

Crumby:
@mcarbonneaux Any opinions on migrating default labels to `app.kubernetes.io` or `operator.infinispan.org`?

mcarbonneaux:
> @mcarbonneaux Any opinions on migrating default labels to `app.kubernetes.io` or `operator.infinispan.org`?


Hi,

Thanks for the detailed feedback and for clarifying the behavior of the StatefulSet selector. You are right: the conflict arises because our Kyverno policies enforce corporate standards on labels like `app`, which unfortunately overlap with the labels the Operator uses for its internal logic.

Regarding your points and the proposed solution:

* **On the current design:** I agree that the current approach in the PR is a bit cumbersome. The main goal was to find a workaround for the rigidness of the current label selection, but it clearly introduces risks regarding affinity and component communication.
* **On Namespaced Labels:** Your proposal to move towards `operator.infinispan.org/name` or `app.kubernetes.io/managed-by` is **the best path forward**. It aligns with CNCF best practices and would solve our collision issues with Kyverno by separating the Operator's internal wiring from generic labels.
* **Configurable Label Keys:** To make this even more robust, it would be ideal if these label keys were **configurable** in the Infinispan CR or the Operator settings. Providing a way to customize the label names (while keeping sensible defaults) would offer the ultimate flexibility for environments with very specific governance rules.
* **On Migration:** I'm aware that this is a significant change. If you decide to implement this, a phased migration or "dual-label" support during a transition period would be ideal to avoid breaking existing clusters.

**Conclusion:**
Yes, implementing namespaced labels—especially if they are configurable—would definitely help us. It provides a much cleaner separation of concerns between the Operator's management and the user's metadata.

I'm looking forward to seeing how this could be integrated into the Operator's roadmap!

Best regards,

Crumby:
I'm closing the PR for the moment as we've agreed on different implementation. The target release has to be yet decided.

PR #2462, opened 2026-03-11
[#2443] Introduce skipRange usage

### LINKED ISSUE #2443: Utilize skipRange to allow for swifter Operator upgrades
We should start utilizing `skipRange` alongside replaces to allow users skip releases rather then being force to go through them all.

**Note:** Utilizing `skipRange` should be done alongside current `replaces` where we replaces the versions one by one otherwise the skipped versions would be pruned from the catalog, impossible to be installed.

**Note:** Due to a nature of Infinsipan Operator versioning we should probably utilize skipRange on micro stream level and force the users to go through the latest releases per minor.

Examples:
```
metadata:
  name: infinispan-operator.v2.5.3
  annotations:
    olm.skipRange: '>=2.5.0 <2.5.3'
spec:
  version: 2.5.3
  replaces: my-operator.v2.5.2
```

```
metadata:
  name: infinispan-operator.v2.5.4
  annotations:
    olm.skipRange: '>=2.5.0 <2.5.4'
spec:
  version: 2.5.4
  replaces: my-operator.v2.5.3
```

Crumby:
Closes #2443.

Adds `skipRange` to CSV, test catalog and related automation for version bumping.


ryanemerson:
Thanks @Crumby

PR #2488, opened 2026-04-01
Add configuration for auto-release of File-Based Catalog

pminz:


release-config.yaml
```diff
@@ -0,0 +1,5 @@
+---
+catalog_templates:
+  - template_name: basic.yaml
+    channels: stable
+    replaces: infinispan-operator.v2.5.5
```
ryanemerson: Do we need to update our GH workflows so that this is automatically changed after a release?

pminz: > Do we need to update our GH workflows so that this is automatically changed after a release?

yes . I will add this to the post-release job.

pminz: Updated the replaces version to infinispan-operator.v2.5.6, as the post-release job will run after the 2.5.7 release.
For the current PR against community-operators-prod will update it to  infinispan-operator.v2.5.5

.github/workflows/image_release.yaml
```diff
@@ -184,6 +184,7 @@ jobs:
         run: |
           mkdir -p operators/infinispan/${{ steps.bump_tag.outputs.new_tag }}
           cp -R ../operator/bundle/* operators/infinispan/${{ steps.bump_tag.outputs.new_tag }}
+          cp ../operator/release-config.yaml operators/infinispan/${{ steps.bump_tag.outputs.new_tag }}
```
Crumby, line 187: So this one here is valid only when creating the PR to the red hat repository, the k8s one is still SQLite based and it shouldn't be there

Crumby: And I can see it's only executed for the prod repo, ignore me...

Crumby (approved):

PR #2491, opened 2026-04-07
[#2416] Log Info by default, allow JSON formmated output

### LINKED ISSUE #2416: Disable debug logs in the production
Operator currently enables debug logging by default in the production deployment. We should be logging INFO above only by default with the possibility to enabled debug logging.

### LINKED ISSUE #2489: Incorrectly formmated log entries
Some of the entries in the Operator log do not follow the format:

```
I0407 13:43:57.051738 1 request.go:690] Waited for 1.039767795s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/whereabouts.cni.cncf.io/v1alpha1?timeout=32s
...
I0407 13:43:58.872276 1 leaderelection.go:248] attempting to acquire leader lease openshift-operators/632512e4.infinispan.org...
I0407 13:44:16.406810 1 leaderelection.go:258] successfully acquired lease openshift-operators/632512e4.infinispan.org
```

vs 

```
2026-04-07T13:44:16.407Z INFO Starting EventSource {"controller": "infinispan", "controllerGroup": "infinispan.org", "controllerKind": "Infinispan", "source": "kind source: *v1.Infinispan"}
2026-04-07T13:44:16.407Z INFO Starting EventSource {"controller": "restore", "controllerGroup": "infinispan.org", "controllerKind": "Restore", "source": "kind source: *v2alpha1.Restore"}
2026-04-07T13:44:16.407Z INFO Starting EventSource {"controller": "restore", "controllerGroup": "infinispan.org", "controllerKind": "Restore", "source": "kind source: *v1.Pod"}
```

the issue is in the client-go library that uses `klog` instead of `zap`. Setting the klog logger to zap should fix the issue

### LINKED ISSUE #2490: JSON formated logging
JSON formatted logging is a standard in k8s payloads. We should allow the format to be configured.

Note: It'd be ideal to have the JSON formatting as default, unfortunately, that's a breaking change and shouldn't be introduced mid release cycle.

Crumby:
Closes #2416, closes #2490, closes #2489

### DIFF
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -13,6 +13,10 @@ import (
 )
 
 var (
+	// Logging configuration, default to json once we get to a next minor release
+	OperatorLogFormat = GetEnvWithDefault("INFINISPAN_OPERATOR_LOG_FORMAT", "console")
+	OperatorLogLevel  = os.Getenv("INFINISPAN_OPERATOR_LOG_LEVEL")
+
 	// InitContainerImageName allows a custom initContainer image to be used
 	InitContainerImageName = GetEnvWithDefault("INITCONTAINER_IMAGE", "registry.access.redhat.com/ubi8-micro")
 
--- a/documentation/asciidoc/topics/proc_configuring_operator_logging.adoc
+++ b/documentation/asciidoc/topics/proc_configuring_operator_logging.adoc
@@ -1,0 +1,49 @@
+[id='configuring-operator-logging_{context}']
+= Configure Operator logging
+
+[role="_abstract"]
+Configure the {brandname} Operator to output logs in JSON format for improved machine processing and log aggregation. You can also adjust the logging level to increase verbosity for debugging reconciliation issues.
+
+`INFINISPAN_OPERATOR_LOG_FORMAT` variable:: Specifies the output format.
+* `console` (default): Human-readable text.
+* `json`: Structured JSON format for log aggregators.
+`INFINISPAN_OPERATOR_LOG_LEVEL` variable:: Sets the verbosity of logs.
+* Valid values are `debug`, `info` (default), `warn`, or `error`.
+
+[TIP]
+====
+In production environments, it is highly recommended to set `INFINISPAN_OPERATOR_LOG_FORMAT` to `json`. This ensures logs are easily indexed by monitoring tools like Splunk, Elasticsearch, or Datadog.
+====
+
+.Procedure
+
+. Open your {brandname} Operator Subscription for editing
+. Add the logging environment variables under the spec.config.env field:
++
+[source,options="nowrap",subs=attributes+]
+----
+include::yaml/logging_subscription.yaml[]
+----
++
+. Apply the changes.OLM will automatically restart the Operator pod to apply the new configuration.
+. Retrieve logs from Operator pod as required.
++
+[source,options="nowrap",subs=attributes+]
+----
+{oc_logs} -f deployment/infinispan-operator-controller-manager
+----
+
+[discrete]
+== Logging format comparison
+
+.Console
+[source,text,options="nowrap",role="primary"]
+----
+2026-04-14T09:28:48.546Z INFO controllers.Infinispan Triggering StatefulSet Rolling upgrade {"infinispan": "infinispan/infinispan"}
+----
+
+.JSON
+[source,json,options="nowrap",role="secondary"]
+----
+{"level":"info","ts":"2026-04-14T09:33:53.960Z","logger":"controllers.Infinispan","msg":"Triggering StatefulSet Rolling upgrade","infinispan":"infinispan/infinispan"}
+----
--- a/documentation/asciidoc/topics/yaml/logging_subscription.yaml
+++ b/documentation/asciidoc/topics/yaml/logging_subscription.yaml
@@ -1,0 +1,17 @@
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: {subscription_name}
+  namespace: {subscription_namespace}
+spec:
+  channel: {operator_channel}
+  installPlanApproval: Automatic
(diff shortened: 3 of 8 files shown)


documentation/asciidoc/topics/yaml/logging_subscription.yaml
```diff
@@ -0,0 +1,17 @@
...
+  channel: {operator_channel}
+  installPlanApproval: Automatic
+  name: {subscription_name}
+  source: {operator_source}
+  sourceNamespace: {source_namespace}
+  config:
+    env:
+    - name: INFINISPAN_OPERATOR_LOG_FORMAT
+      value: json
+    - name: INFINSIPAN_OPERATOR_LOG_LEVEL
```
rigazilla:
```suggestion
    - name: INFINISPAN_OPERATOR_LOG_LEVEL
```

controllers/constants/constants.go
```diff
@@ -177,3 +182,16 @@ func GetEnvBool(name string) bool {
...
+func GetLogLevel() zapcore.Level {
+	level, err := zapcore.ParseLevel(OperatorLogLevel)
+	if err != nil {
+		// Default to Info so the Operator doesn't crash
+		return zapcore.InfoLevel
+	}
+	return level
+}
+
+func IsDevel() bool {
```
rigazilla: - IsDevel name is misleading for a global function, the func name should suggest it's related to logging mode/format.

- the tight coupling "!json -> zap devel mode" should be mentioned in the doc so users understand what to expect when choosing a format.

rigazilla (changes requested):

documentation/asciidoc/topics/yaml/logging_subscription.yaml
Crumby: Thanks, fixed

controllers/constants/constants.go
Crumby: I've played with the configuration a bit more and made bunch of changes around:
* Replaced global `IsDevel()` with local `getLogEncoder()`
* `getLogLevel()` is now local as well and only applied if defined.
* The production config is used for both console and json outputs
  * The above changes allow for eventual `--zap-devel=true` usage as an argument on startup and it'll change the log level and stacktrace level. The console/json will be still used based on env. (Console is default so it doesn't really matter and once we move to json by default, we could apply the same approach as with `log-level`)
* I added `json` vs `console` output comparison to the documentation

documentation/asciidoc/topics/proc_configuring_operator_logging.adoc
```diff
@@ -0,0 +1,49 @@
+[id='configuring-operator-logging_{context}']
+= Configure Operator logging
+
+[role="_abstract"]
+Configure the {brandname} Operator to output logs in JSON format for improved machine processing and log aggregation. You can also adjust the logging level to increase verbosity for debugging reconciliation issues.
+
+`INFINISPAN_OPERATOR_LOG_FORMAT` variable:: Specifies the output format.
+* `console` (default): Human-readable text.
+* `json`: Structure JSON format for log aggregators.
+`INFINISPAN_OPERATOR_LOG_LEVEL` variable:: Sets the verbosity of logs.
```
rigazilla, line 10: Does it worth to mention that if level is not set, zap options could be applied? I don't know if olm allows it anyway

documentation/asciidoc/topics/proc_configuring_operator_logging.adoc
```diff
@@ -0,0 +1,49 @@
+[id='configuring-operator-logging_{context}']
+= Configure Operator logging
+
+[role="_abstract"]
+Configure the {brandname} Operator to output logs in JSON format for improved machine processing and log aggregation. You can also adjust the logging level to increase verbosity for debugging reconciliation issues.
+
+`INFINISPAN_OPERATOR_LOG_FORMAT` variable:: Specifies the output format.
+* `console` (default): Human-readable text.
+* `json`: Structure JSON format for log aggregators.
```
rigazilla:
```suggestion
* `json`: Structured JSON format for log aggregators.
```

rigazilla (commented):
looks good,
some small comments

documentation/asciidoc/topics/proc_configuring_operator_logging.adoc
Crumby: I don't think that's reasonably possible with OLM. You can overwrite the CSV or the deployment directly but that change won't hold after you upgrade, so for OLM I think the Subscriptions `spec.config.env` is the way to go.

On the other hand for local development and one day if we provide Helm Charts for the Operator, it may make sense. But I'd say that's irrelevant for the documentation at the moment

Crumby: > zap options could be applied

I assume you mean zap flags here that could be applied as an argument. I didn't find any zap envs

documentation/asciidoc/topics/proc_configuring_operator_logging.adoc
Crumby: Fixed

documentation/asciidoc/topics/proc_configuring_operator_logging.adoc
Crumby: Maybe one thing we could play around with is to add `$(OPERATOR_EXTRA_ARGS)` to Operators deployment arg list which would allow to define `--zap-devel` on Subscription (The implmentation might be a bit more complex then just adding the env var to the list).

But I don't think that this on it's own would invalidate this PR's solution though

rigazilla: yep, PR is ok

rigazilla (approved):

rigazilla:
thanks @Crumby !

PR #2493, opened 2026-04-08
[#2492] Proper version comparison when using compare api

### LINKED ISSUE #2492: Operator triggers `?compare` rest api only if the micro version is greater then 6
The part of the code that triggers `?action=compare` to compare cache configurations doesn't compare the versions properly. It expects the patch version to be always greater then 6, no matter the major version.

Aside from version check being obviously wrong, it resulted into undetected regression being implemented on the Infinsipan side as the current snapshot version is `16.2.0`, which is suitable for compare endpoint execution but the version but the patch version being 0 makes the controller use older logic

Crumby:
Closes #2492

### DIFF
--- a/controllers/cache_controller.go
+++ b/controllers/cache_controller.go
@@ -669,7 +669,8 @@ func configChanged(a, b string, i *v1.Infinispan, caches api.Caches, vm *version
 	// If Operand is >= 14.0.7.Final we can use the cache compare endpoint to avoid unnecessary updates to the
 	// Infinispan spec.template field
 	operand, _ := vm.WithRef(i.Spec.Version)
-	if operand.UpstreamVersion.Major >= 14 && operand.UpstreamVersion.Patch >= 7 {
+
+	if operand.UpstreamVersion.GE(constants.MinVersionCacheCompareEndpoint) {
 		// Check if Cache configuration is semantically the same as that already present in the Cache CR
 		equalConfig, err := caches.EqualConfiguration(a, b)
 		if err != nil {
--- a/controllers/constants/constants.go
+++ b/controllers/constants/constants.go
@@ -57,6 +57,13 @@ var (
 		Minor: 0,
 		Patch: 7,
 	}
+
+	// 14.0.7.Final is required for ?action=compare endpoint
+	MinVersionCacheCompareEndpoint = semver.Version{
+		Major: 14,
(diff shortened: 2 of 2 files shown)


controllers/cache_controller.go
```diff
@@ -669,7 +670,11 @@ func configChanged(a, b string, i *v1.Infinispan, caches api.Caches, vm *version
 	// If Operand is >= 14.0.7.Final we can use the cache compare endpoint to avoid unnecessary updates to the
 	// Infinispan spec.template field
 	operand, _ := vm.WithRef(i.Spec.Version)
-	if operand.UpstreamVersion.Major >= 14 && operand.UpstreamVersion.Patch >= 7 {
+
+	minVersion := semver.MustParse("14.0.7")
+	currentVersion, err := semver.Parse(operand.UpstreamVersion.String())
+
+	if err == nil && currentVersion.GE(minVersion) {
```
ryanemerson: Can't we just do `if operand.UpstreamVersion.GE(minVersion) {`?

controllers/cache_controller.go
```diff
@@ -669,7 +670,11 @@ func configChanged(a, b string, i *v1.Infinispan, caches api.Caches, vm *version
 	// If Operand is >= 14.0.7.Final we can use the cache compare endpoint to avoid unnecessary updates to the
 	// Infinispan spec.template field
 	operand, _ := vm.WithRef(i.Spec.Version)
-	if operand.UpstreamVersion.Major >= 14 && operand.UpstreamVersion.Patch >= 7 {
+
+	minVersion := semver.MustParse("14.0.7")
```
ryanemerson: Should we create a var in `constants.go` so that we're not creating this semver on every reconcilliation?

ryanemerson (changes requested):

controllers/cache_controller.go
Crumby: Oh, yeah, that's way better

ryanemerson:
`TestUpdateCacheCR` is still failing, do we also need server-side fixes? There's no information in the linked issue, so I'm missing the context here.

The commit message also needs to be updated to reference the correct issue.

Crumby:
@ryanemerson Fixed the commit and updated issue description

ryanemerson:
Thanks @Crumby

PR #2501, opened 2026-04-16
[#2500] Filter the test activeTargets by infinispan_cr label

### LINKED ISSUE #2500: DataGridServiceIT#serviceMonitorTest does not account for other payload on the cluster
If there are other monitored entities by user workload prometheus, the test will fail as it expects Infinispan cluster to be the only running monitored entity.

Solution: Filter the active targets by Infinispan specific label

Crumby:
Closes #2500.


ramdr:
Changes have been tested. it is working fine now.

PR #2506, opened 2026-04-23
[#2324] Allow to specify environment variables per CR

### LINKED ISSUE #2324: Allow adding environment variables to Infinispan pods per namespace or per Infinispan object
Hello,

Currently we can add environment variables to Infinispan pods deployed by the Operator using Subscription
```
apiVersion: operators.coreos.com/v1alpha1
kind: Subscription
metadata:
  name: infinispan
  namespace: operators
spec:
  channel: stable
  installPlanApproval: Automatic
  name: infinispan
  source: operatorhubio-catalog
  sourceNamespace: operator-lifecycle-manager
  config:
    env:
    - name: ADDITIONAL_VARS
      value: "[\"TEST_VAR\"]"
    - name: TEST_VAR
      valueFrom:
        secretKeyRef:
          name: test-operator-secrets
          key: TEST_KEY
```
However, in this example, the Kubernetes secret used to set the environment variable value must be located in the operator's namespace, which means the same value will be added to all Infinispan pods running in the cluster.

It would be useful to have a way to add environment variables to Infinispan pods on a per-namespace basis, or perhaps simply allow environment variables to be added directly to the Infinispan object, similar to this:
```
apiVersion: infinispan.org/v1
kind: Infinispan
metadata:
  name: infinispan
spec:
  replicas: 2
  version: <Infinispan_version>
  service:
    type: DataGrid
  container:
    env:
    - name: TEST_VAR
      valueFrom:
        secretKeyRef:
          name: test-operator-secrets
          key: TEST_KEY
    - name: POD_IP
      valueFrom:
        fieldRef:
          fieldPath: status.podIP
```

Crumby:
Closes #2324


documentation/asciidoc/topics/yaml/custom_envvars.yaml
```diff
@@ -0,0 +1,11 @@
+
+spec:
+  container:
+    env:
+      - name: JAVA_OPTS
```
tristantarrant: In the context of https://github.com/infinispan/infinispan/pull/17311 I would not show an example with `JAVA_OPTS` because it would break the server. Maybe use `JAVA_OPTIONS` which is just appended, will be compatible with older images and is safer.

pkg/reconcile/pipeline/infinispan/handler/provision/pods.go
```diff
@@ -182,6 +182,10 @@ func PodEnv(i *ispnv1.Infinispan, systemEnv *[]corev1.EnvVar) []corev1.EnvVar {
 		}
 	}
 
+	if i.Spec.Container.Env != nil {
```
tristantarrant, line 191: Maybe we should validate that we don't stomp on env vars?

tristantarrant (approved):
LGTM, two suggestions

documentation/asciidoc/topics/yaml/custom_envvars.yaml
Crumby: Yeah, it occurred to me, that this one is not very useful example. I was thinking to wait for the JAVA_OPTS PR rework but and using `JAVA_OPTS_MEMORY` but that'll work only with 16.0.x+.

I did think about JAVA_OPTIONS as well but we might probably want to add a note about it's interference with `spec.container.extraJvmOpts` which effectively sets the same env.

pkg/reconcile/pipeline/infinispan/handler/provision/pods.go
Crumby: Which one? Just the ones used internally by the Operator, all of which are defined on system level?

I'm rather inclined not to check them, chances of user overriding the internal env vars are quite low as they pretty specific with an exception of `JAVA_OPTIONS` (and `CLI_JAVA_OPTIONS`), those are set according to `spec.container.extraJvmOpts` and respective cli counter part. If not set they are configured as empty and tbh. I'm not really sure what happens if you pass a slice to k8s with duplicate names. Whether the change is going to be rejected or one of the duplicates dumped.

One thing we probably could/should do is to ensure that no env exist twice in the slice and apply them with a priority. At the moment there are the env vars may come from three different sources. Internal (to configure the image), user defined on the Operator level applied to all the deployments and the user ones set by `spec.container.env` (this PR). If we applly them in particular order some will have higher priority then others, making them effectively un-overwriteable

tristantarrant: Let's not go overboard :)

documentation/asciidoc/topics/yaml/custom_envvars.yaml
Crumby: The docs now mention `JAVA_OPTIONS` and add an important note on possible issues using them in parallel with `spec.container.extraJvmOpts`

pkg/reconcile/pipeline/infinispan/handler/provision/pods.go
Crumby: I've modified the behavior slightly so the `JAVA_OPTIONS` and `CLI_JAVA_OPTIONS` are not set empty and add a comment in the docs. The rest can be tuned later if issues arise with env conflicts

Crumby:
@tristantarrant PR updated

documentation/asciidoc/topics/proc_adding_environments.adoc
```diff
@@ -0,0 +1,34 @@
+[id='adding-custom-env-vars{context}']
```
tristantarrant, line 1: We should link to the container image docs: https://infinispan.org/docs/stable/titles/container_image/container_image.html

Crumby: I added the link to JVM Settings section rather then the container one. As that section contains all the settings we want to share while not containing the list of envs that Operator is already leveraging.

tristantarrant (approved):

tristantarrant:
Thanks @Crumby

PR #2512, opened 2026-04-29
update release-config.yaml with correct template name

pminz:


release-config.yaml
```diff
@@ -1,6 +1,56 @@
 ---
 catalog_templates:
-  - template_name: basic.yaml
-    channels: stable
+  - template_name: v4.12.yaml
```
Crumby: @pminz Wouldn't it be better to convert [ci.yaml](https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/main/operators/infinispan/ci.yaml) into a setup similar to what they have with kube-green ([ci.yaml](https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/main/operators/kube-green/ci.yaml))? This looks like a lot of unnecessary duplication on both sides

pminz: yes this looks simpler but I didn't want to touch `catalog-templates` folder.
I think we will have to regenerate the catalog template.
I will check the on-boarding document and confirm.

Crumby (approved):

PR #2524, opened 2026-05-29
[#2404] Custom ServiceAccount in Infinispan CR

### LINKED ISSUE #2404: Allow the Infinispan pod's ServiceAccount to be configured in the Infinispan CR
Hi Team ,

Need a placeholder in ISPN pod config to specify the service account name to be used while deploying.

https://infinispan.zulipchat.com/#narrow/channel/118645-infinispan/topic/where.20to.20specify.20service.20account.20name.20in.20the.20infinispan.20spec/with/563973410

tristantarrant:
Closes #2404

### DIFF
--- a/controllers/zero_controller.go
+++ b/controllers/zero_controller.go
@@ -323,7 +323,8 @@ func (z *zeroCapacityController) zeroPodSpec(name, namespace string, podSecurity
 			Annotations: ispn.PodAnnotations(),
 		},
 		Spec: corev1.PodSpec{
-			SecurityContext: podSecurityCtx,
+			ServiceAccountName: ispn.Spec.ServiceAccountName,
+			SecurityContext:    podSecurityCtx,
 			Containers: []corev1.Container{{
 				Image:         ispn.ImageName(),
 				Name:          provision.InfinispanContainer,
--- a/documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
+++ b/documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
@@ -1,0 +1,41 @@
+[id='configuring-service-accounts_{context}']
+= Configuring service accounts
+
+[role="_abstract"]
+Specify the service account that {brandname} pods use instead of the default service account in the namespace.
+This is useful when you need pods to use a specific service account for pulling images from private container registries or for fine-grained RBAC control.
+
+.Prerequisites
+* Have a service account created in the same namespace as the `Infinispan` CR (for example, `custom-service-account`).
+* You have configured any necessary image pull secrets on that service account if using a private registry.
+
+.Procedure
+. Configure the service account name in the `Infinispan` CR.
++
+.Infinispan CR
+[source,yaml]
+----
+include::yaml/service_account_name.yaml[]
+----
++
+The `spec.serviceAccountName` field sets the service account for {brandname} server pods, GossipRouter pods, and Batch, Backup, and Restore pods.
++
+The `spec.configListener.serviceAccountName` field sets the service account for the ConfigListener pod.
+The ConfigListener requires specific RBAC permissions to manage Cache CRs, so a separate service account is supported.
++
+. If you specify `spec.configListener.serviceAccountName`, you must create the required RBAC resources for the ConfigListener yourself.
+The following ServiceAccount, Role, and RoleBinding must be created and bound to the specified service account:
++
+.ConfigListener RBAC
+[source,yaml]
+----
+include::yaml/config_listener_rbac.yaml[]
+----
++
+. Apply your `Infinispan` CR.
++
+If your cluster is running, {ispn_operator} restarts the {brandname} pods in order for the changes to take effect.
+
+NOTE: When `spec.configListener.serviceAccountName` is not specified, {ispn_operator} automatically creates and manages the ServiceAccount, Role, and RoleBinding for the ConfigListener.
+
+NOTE: Changing `spec.configListener.serviceAccountName` on an already running cluster requires disabling and re-enabling the ConfigListener (`spec.configListener.enabled`) for the change to take effect. This limitation will be addressed in a future release.
--- a/documentation/asciidoc/topics/yaml/config_listener_rbac.yaml
+++ b/documentation/asciidoc/topics/yaml/config_listener_rbac.yaml
@@ -1,0 +1,41 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: <config-listener-sa-name> # MUST match spec.configListener.serviceAccountName
+  namespace: <namespace>
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: config-listener-role
+  namespace: <namespace>
+rules:
+  - apiGroups: ["infinispan.org"]
+    resources: ["caches"]
+    verbs: ["create", "delete", "get", "list", "patch", "update", "watch"]
+  - apiGroups: ["infinispan.org"]
+    resources: ["infinispans"]
+    verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["list"]
+  - apiGroups: [""]
+    resources: ["pods/exec"]
+    verbs: ["create"]
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: config-listener-rolebinding
+  namespace: <namespace>
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: config-listener-role
+subjects:
+  - kind: ServiceAccount
+    name: <config-listener-sa-name> # MUST match the ServiceAccount name above
+    namespace: <namespace>
--- a/documentation/asciidoc/topics/yaml/service_account_name.yaml
+++ b/documentation/asciidoc/topics/yaml/service_account_name.yaml
@@ -1,0 +1,6 @@
+kind: Infinispan
+...
+spec:
+  serviceAccountName: custom-service-account
+  configListener:
+    serviceAccountName: custom-config-listener-sa
--- a/pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
@@ -132,6 +132,14 @@ func StatefulSetRollingUpgrade(i *ispnv1.Infinispan, ctx pipeline.Context) {
 		updateNeeded = true
 	}
 
+	// Note: removing serviceAccountName entirely (setting to "") may not take effect because
+	// K8s auto-populates the deprecated spec.ServiceAccount field, which then re-sets ServiceAccountName.
+	// Users can work around this by explicitly setting serviceAccountName to "default".
+	if spec.ServiceAccountName != i.Spec.ServiceAccountName {
+		spec.ServiceAccountName = i.Spec.ServiceAccountName
+		updateNeeded = true
+	}
+
 	if updateCmdArgs, err := updateStartupArgs(container, configFiles); err != nil {
 		ctx.Requeue(err)
 		return
--- a/pkg/reconcile/pipeline/infinispan/handler/provision/provision_test.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/provision/provision_test.go
@@ -74,4 +74,50 @@ var _ = Describe("Provision", func() {
 		Expect(err).Should(BeNil())
 		Expect(ss.Spec.Template.Spec.PriorityClassName).Should(BeEmpty())
 	})
+
+	It("should correctly set StatefulSet ServiceAccountName when defined", func() {
+		mockCtrl := gomock.NewController(GinkgoT())
+		resources := infinispan.NewMockResources(mockCtrl)
+
+		ctx := infinispan.NewMockContext(mockCtrl)
+		ctx.EXPECT().ConfigFiles().AnyTimes().Return(
+			&infinispan.ConfigFiles{
+				AdminIdentities: &infinispan.AdminIdentities{},
+			},
+		)
+		ctx.EXPECT().Resources().AnyTimes().Return(resources)
+		ctx.EXPECT().Operand().AnyTimes().Return(version.Operand{UpstreamVersion: &semver.Version{Major: 15, Minor: 0, Patch: 0}})
+
+		ispn := &ispnv1.Infinispan{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      key.Name,
+				Namespace: key.Namespace,
+			},
+			Spec: ispnv1.InfinispanSpec{
+				Replicas:           1,
+				ServiceAccountName: "custom-sa",
+				Container: ispnv1.InfinispanContainerSpec{
+					Memory: "1Gi",
+				},
+				Service: ispnv1.InfinispanServiceSpec{
+					Type: ispnv1.ServiceTypeDataGrid,
+					Container: &ispnv1.InfinispanServiceContainerSpec{
+						EphemeralStorage: true,
+					},
+				},
+				Version: "IGNORED. Required so we can call Default()",
+			},
+		}
+		ispn.Default()
+
+		ss, err := ClusterStatefulSetSpec("statefulset", ispn, ctx)
+		Expect(err).Should(BeNil())
+		Expect(ss.Spec.Template.Spec.ServiceAccountName).Should(Equal("custom-sa"))
+
+		// Assert ServiceAccountName is empty when not specified
+		ispn.Spec.ServiceAccountName = ""
+		ss, err = ClusterStatefulSetSpec("statefulset", ispn, ctx)
+		Expect(err).Should(BeNil())
+		Expect(ss.Spec.Template.Spec.ServiceAccountName).Should(BeEmpty())
+	})
 })
--- a/pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
@@ -51,6 +51,13 @@ func ClusterStatefulSet(i *ispnv1.Infinispan, ctx pipeline.Context) {
 		return
 	}
 
+	// Validate that the user-specified ServiceAccount exists before creating the StatefulSet
+	if i.Spec.ServiceAccountName != "" {
+		if err := ctx.Resources().Load(i.Spec.ServiceAccountName, &corev1.ServiceAccount{}, pipeline.RetryOnErr); err != nil {
+			return
+		}
+	}
+
 	statefulSet, err := ClusterStatefulSetSpec(statefulSetName, i, ctx)
 	if err != nil {
 		ctx.Requeue(fmt.Errorf("unable to create StatefulSet spec: %w", err))
@@ -113,6 +120,7 @@ func ClusterStatefulSetSpec(statefulSetName string, i *ispnv1.Infinispan, ctx pi
 					Annotations: annotationsForPod,
 				},
 				Spec: corev1.PodSpec{
+					ServiceAccountName:            i.Spec.ServiceAccountName,
 					TerminationGracePeriodSeconds: i.TerminationGracePeriodSeconds(),
 					Affinity:                      i.Affinity(),
 					Tolerations:                   i.Tolerations(),
--- a/test/e2e/infinispan/service_account_test.go
+++ b/test/e2e/infinispan/service_account_test.go
@@ -1,0 +1,238 @@
+package infinispan
+
+import (
+	"context"
+	"testing"
+
+	v1 "github.com/infinispan/infinispan-operator/api/v1"
+	v2 "github.com/infinispan/infinispan-operator/api/v2alpha1"
+	kube "github.com/infinispan/infinispan-operator/pkg/kubernetes"
+	"github.com/infinispan/infinispan-operator/pkg/reconcile/pipeline/infinispan/handler/provision"
+	tutils "github.com/infinispan/infinispan-operator/test/e2e/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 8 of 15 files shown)


documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
```diff
@@ -0,0 +1,29 @@
...
++
+.Infinispan CR
+[source,yaml]
+----
+include::yaml/service_account_name.yaml[]
+----
++
+The `spec.serviceAccountName` field sets the service account for {brandname} server pods and GossipRouter pods.
+The `spec.configListener.serviceAccountName` field sets the service account for the ConfigListener pod.
+Because the ConfigListener uses the Operator image rather than the {brandname} server image, it may require a different service account.
```
Crumby: Different image may be one reason if for whatever reason the images are in different repositories but mainly the config listener service account needs access (roles) to manage Cache CRs. Cluster and Gossip service accounts do not require any roles (at least at the moment)

documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
```diff
@@ -0,0 +1,29 @@
...
++
+When you specify `spec.configListener.serviceAccountName`, {ispn_operator} does not create its own service account for the ConfigListener.
+However, it still creates the necessary Role and RoleBinding resources bound to the specified service account to ensure the ConfigListener has the required API permissions.
```
Crumby: This is anti pattern, you are giving away authorization secretly. If user want's to use his own service account, he needs to give it correct rbac roles and we need to document which ones are required and how to do that.

pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
```diff
@@ -113,6 +113,7 @@ func ClusterStatefulSetSpec(statefulSetName string, i *ispnv1.Infinispan, ctx pi
 					Annotations: annotationsForPod,
 				},
 				Spec: corev1.PodSpec{
+					ServiceAccountName:            i.Spec.ServiceAccountName,
```
Crumby, line 123: We might want to do a validation of SA existence and either emit an event if the SA does not exist or set a condition the SA is missing and loop reconciliation till it exists.

At the moment it just creates a StatefulSet with a error pod with missing ServiceAccount (Not 100% sure). Which is okay as well and we may keep it as is for now but generally I'd say it's better to accumulate issues directly at the Infinispan CR if we are able to detect them.

test/e2e/infinispan/service_account_test.go
```diff
@@ -0,0 +1,120 @@
...
+	ss := testKube.GetStatefulSet(ispn.GetStatefulSetName(), ispn.Namespace)
+	assert.Equal(saName, ss.Spec.Template.Spec.ServiceAccountName)
+
+	// Verify the actual pod uses the custom ServiceAccountName
+	pod := corev1.Pod{}
+	require.NoError(testKube.Kubernetes.Client.Get(context.TODO(), types.NamespacedName{Name: ispn.Name + "-0", Namespace: tutils.Namespace}, &pod))
+	assert.Equal(saName, pod.Spec.ServiceAccountName)
+}
+
+func TestServiceAccountNameUpdate(t *testing.T) {
```
Crumby, line 64: This test is failing because required changes to `pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go` reponsible for taking care of updates are missing

documentation/asciidoc/topics/yaml/service_account_name.yaml
```diff
@@ -0,0 +1,6 @@
+kind: Infinispan
```
Crumby, line 1: Let's not forget on ServiceAccount for Batches, Backups and Restores which spin up pods as well.

For those I'd say the general strategy should be to use any `serviceAccount` configured for target cluster (Infinispan CR) if present and maybe add an option to override with `spec.serviceAccountName`. Question is should we add an option to override the serviceAccount for the GossipRouter as well? I'd say not mandatory, at least not now.

Crumby (changes requested):

controllers/zero_controller.go
```diff
@@ -325,7 +325,8 @@ func (z *zeroCapacityController) zeroPodSpec(name, namespace string, podSecurity
 			Annotations: ispn.PodAnnotations(),
 		},
 		Spec: corev1.PodSpec{
-			SecurityContext: podSecurityCtx,
+			ServiceAccountName: ispn.Spec.ServiceAccountName,
+			SecurityContext:     podSecurityCtx,
```
Crumby: Improperly formatted code. As part of the `unit` CIs, `make test` runs `go fmt ` which formats the go code. `git diff` then later detects a change and fails.

Crumby:
```suggestion
			SecurityContext:    podSecurityCtx,
```

pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
```diff
@@ -132,6 +132,11 @@ func StatefulSetRollingUpgrade(i *ispnv1.Infinispan, ctx pipeline.Context) {
 		updateNeeded = true
 	}
 
+	if spec.ServiceAccountName != i.Spec.ServiceAccountName {
```
Crumby, line 138: Removing `spec.serviceAccountName` from the yaml doesn't do anything. It should remove the `serviceAccountName` from the StatefulSet template definition

pkg/reconcile/pipeline/infinispan/handler/provision/provision_test.go
```diff
@@ -74,4 +74,50 @@ var _ = Describe("Provision", func() {
...
+			},
+		}
+		ispn.Default()
+
+		ss, err := ClusterStatefulSetSpec("statefulset", ispn, ctx)
+		Expect(err).Should(BeNil())
+		Expect(ss.Spec.Template.Spec.ServiceAccountName).Should(Equal("custom-sa"))
+
+		// Assert ServiceAccountName is empty when not specified
+		ispn.Spec.ServiceAccountName = ""
```
Crumby, line 118: Test doesn't validate properly against OCP real behavior. If you set `spec.serviceAccountName`  to `""` then it removes the field completely and ignores it's removal as mentioned in other of my comments

pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
Crumby: I've run manual test to see what happen if you 'mistype' serviceAccount name and the only thing you get is an event directly on the StatefulSet: `create Pod infinispan-1 in StatefulSet infinispan failed error: pods "infinispan-1" is forbidden: error looking up service account datagrid/non-existent-ispn-sa: serviceaccount "non-existent-ispn-sa"`.

You won't see pods starting up and looking at the Infinispan CR you won't see any related event. What we should at least  do, is to emit an event that the ServiceAccount can't be found and thus we can't continue with deployment. Ideally we should have a condition as well telling user we can't continue with reconciliation. (Btw. I've checked the common conditions used and it seems the standard is to use `Degraded`, `Available` and `Progressing` conditions. Even though implementing them properly is out of scope of this PR.)

Emitting an event if the SA can't be found an postponing reconciliation is inline with what we do when secret with authentication config does not exist

test/e2e/infinispan/service_account_test.go
```diff
@@ -0,0 +1,120 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestServiceAccountName(t *testing.T) {
+	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
```
Crumby, line 26: Add `t.Parallel()` at the beginning of the tests to allow for parallel test execution. (See other tests, most of them have it)

test/e2e/infinispan/service_account_test.go
```diff
@@ -0,0 +1,120 @@
...
+		assert.Equal(t, saName, ss.Spec.Template.Spec.ServiceAccountName)
+	}
+	genericTestForContainerUpdated(*spec, modifier, verifier)
+}
+
+func TestConfigListenerServiceAccountName(t *testing.T) {
+	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
+
+	clSaName := "custom-cl-sa"
+	sa := &corev1.ServiceAccount{
```
Crumby, line 95: I wonder if it's worth it to go all the way and create Role and RoleBinding as specified in the docs and validate that the ConfigListener is still able manage Caches. It would only make sense if the test would be able to catch any requirement changes on the ServiceAccount Roles

test/e2e/infinispan/service_account_test.go
```diff
@@ -0,0 +1,120 @@
...
+	var verifier = func(ispn *v1.Infinispan, ss *appsv1.StatefulSet) {
+		assert.Equal(t, saName, ss.Spec.Template.Spec.ServiceAccountName)
+	}
+	genericTestForContainerUpdated(*spec, modifier, verifier)
+}
+
+func TestConfigListenerServiceAccountName(t *testing.T) {
+	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
+
+	clSaName := "custom-cl-sa"
```
Crumby, line 93: We should verify custom serviceAccounts won't get deleted when Infinspan CR is deleted.

test/e2e/infinispan/service_account_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestServiceAccountName(t *testing.T) {
```
Crumby, line 24: Let's add some trivial verification for Batch, Backup and Restore as well to the test to check it properly pulls SA from the target cluster

documentation/asciidoc/topics/yaml/service_account_name.yaml
```diff
@@ -0,0 +1,6 @@
...
+...
+spec:
+  serviceAccountName: "custom-service-account"
```
Crumby: We should use idiomatic yaml for better readability, remove the quotes around the SA names

documentation/asciidoc/topics/yaml/config_listener_rbac.yaml
```diff
@@ -0,0 +1,35 @@
...
+metadata:
+  name: config-listener-rolebinding
+  namespace: <namespace>
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: config-listener-role
+subjects:
+  - kind: ServiceAccount
+    name: custom-config-listener-sa
```
Crumby: Should we change it to: `name: <config-listener-sa-name> # MUST match the ServiceAccount name above`? So it doesn't get lost in copy pasting

documentation/asciidoc/topics/yaml/config_listener_rbac.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
Crumby, line 7: Should we add a SA example for copy pasting and for the user to see how the names map across all the resources around?

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: <config-listener-sa-name> # MUST match spec.configListener.serviceAccountName
  namespace: <namespace>

File: documentation/asciidoc/topics/proc_configuring_service_accounts.adoc [hunk 16]
```diff
@@ -0,0 +1,38 @@
+[id='configuring-service-accounts_{context}']
+= Configuring service accounts
+
+[role="_abstract"]
+Specify the service account that {brandname} pods use instead of the default service account in the namespace.
+This is useful when you need pods to use a specific service account for pulling images from private container registries or for fine-grained RBAC control.
+
+.Prerequisites
+* Have a service account created in the same namespace as the `Infinispan` CR.
```
Crumby: Should we add (for example, `custom-service-account`) at the end to create a binding with the example?

One more prerequisite bullet that we could have here is: `* You have configured any necessary image pull secrets on that server service account if using a private registry.`

documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
```diff
@@ -0,0 +1,38 @@
...
++
+.ConfigListener RBAC
+[source,yaml]
+----
+include::yaml/config_listener_rbac.yaml[]
+----
++
+. Apply your `Infinispan` CR.
++
+If your cluster is running, {ispn_operator} restarts the {brandname} pods in order for the changes to take effect.
```
Crumby, line 37: Actually applies only for the cluster itself. We are completely missing ConfigListener manage reconciler and it doesn't react on any changes done to `spec.configListener` beside changing the `enabled` to  true/false. So if user has already running config listener and they want to change the SA, they need to disable/enable the listener. Given it's not specific to serviceAccountName we may want to add a note for now and fix this is in separate issue/pr

Crumby (changes requested):

aman-kumar-2:
Hi @Crumby, any further progress on this PR?

Crumby:
Hey @aman-kumar-2, @tristantarrant needs to address the required changes

documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
tristantarrant: Agreed. Added a separate spec.configListener.serviceAccountName field specifically  because the configListener requires distinct RBAC permissions (manage Cache CRs, get Infinispans, list/exec pods, get  secrets) that the cluster/gossip SAs don't need. The documentation explains this separation.

documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
tristantarrant: Fixed. When a user specifies spec.configListener.serviceAccountName, the operator now skips automatic creation of  the ServiceAccount, Role, and RoleBinding (see config_listener.go). The user is responsible for creating the required RBAC  resources themselves. The documentation includes a complete config_listener_rbac.yaml example with the ServiceAccount,  Role, and RoleBinding showing all required permissions.

pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
tristantarrant: Addressed. Added a validation check in ClusterStatefulSet that calls ctx.Resources().Load() with
  pipeline.RetryOnErr when a custom serviceAccountName is specified. If the SA doesn't exist, reconciliation is postponed until it does, consistent with how we handle missing auth secrets. This prevents the StatefulSet from being  created with a non-existent SA.

test/e2e/infinispan/service_account_test.go
tristantarrant: Fixed

documentation/asciidoc/topics/yaml/service_account_name.yaml
tristantarrant: Done

controllers/zero_controller.go
tristantarrant: Fixed

tristantarrant: Fixed

pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
tristantarrant: Fixed

pkg/reconcile/pipeline/infinispan/handler/provision/provision_test.go
tristantarrant: Fixed. The unit test now validates both directions

test/e2e/infinispan/service_account_test.go
tristantarrant: Done

test/e2e/infinispan/service_account_test.go
tristantarrant: We could address it in a separate issue

test/e2e/infinispan/service_account_test.go
tristantarrant: Done. TestServiceAccountName now explicitly deletes the Infinispan CR and then asserts the custom ServiceAccount  still exists

test/e2e/infinispan/service_account_test.go
tristantarrant: Done for batch. Let's do Backup/restore in a follow-up

documentation/asciidoc/topics/yaml/service_account_name.yaml
tristantarrant: Done

documentation/asciidoc/topics/yaml/config_listener_rbac.yaml
tristantarrant: Done

documentation/asciidoc/topics/yaml/config_listener_rbac.yaml
tristantarrant: Done in config_listener_rbac.yaml

documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
tristantarrant: Done

documentation/asciidoc/topics/proc_configuring_service_accounts.adoc
tristantarrant: Added a note to the docs

pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
Crumby: So, this still does not work as expected and the reason is that the Pods have deprecated `ServiceAccount` field which is autopopulated for backward compatibility. That allows for changes through `ServiceAccountName` but when you decide to remove it completely, the OCP/K8s will use the autopopulated value from `ServiceAccount` and sets the `ServiceAccountName` back to what it was.

The fix could be as simple as adding `spec.DeprecatedServiceAccount = i.Spec.ServiceAccountName` (I guess ideally only in the specific condition of `i.Spec.ServiceAccountName` being an empty string ). But given the edge case of this scenario and the fact that we'd need to interact with deprecated fields I'm probably more inclined to keep it the way it is an just add a comment  above the if we happen to stumble across the issue. You can still set `serviceAccountName: default` explicitly

pkg/reconcile/pipeline/infinispan/handler/provision/statefulsets.go
Crumby: The issue is still present in the manage section of the StatefulSets, changing the existing `spec.serviceAccountName` to nonexisting one will just replace that on StatefulSet which scales down emits an event and the issue is invisible from the Infinispan CR.

The current solution with  `... resource 'non-existent' not ready` used in provision and other resources is really insufficient and will need a rework. (It's hard to spot the real issue in between all the not ready events and no condition) so I'm willing to let this slide for now

test/e2e/infinispan/service_account_test.go
```diff
@@ -0,0 +1,233 @@
...
+	defer testKube.CleanNamespaceAndLogOnPanic(t, tutils.Namespace)
+
+	saName := "custom-ispn-sa"
+	sa := &corev1.ServiceAccount{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      saName,
+			Namespace: tutils.Namespace,
+		},
+	}
+	tutils.ExpectNoError(testKube.Kubernetes.Client.Create(context.TODO(), sa))
```
Crumby, line 35: All the test ServiceAccounts hang around after test is done. Needs `defer testKube.DeleteServiceAccount()` on the next line... (needs to be implemented. Look eg. on ConfigMap functions in kubernetes.go )

Crumby: Same applies for the Roles and RoleBindings in the ConfigListenerTest

test/e2e/infinispan/service_account_test.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// Wait for the Batch Job to be created and verify it uses the custom SA
+	var job batchv1.Job
+	err := wait.PollUntilContextTimeout(context.Background(), tutils.DefaultPollPeriod, tutils.SinglePodTimeout, false, func(ctx context.Context) (bool, error) {
+		return testKube.AssertK8ResourceExists(spec.Name, tutils.Namespace, &job), nil
+	})
+	require.NoError(t, err, "Batch Job should be created")
+	assert.Equal(t, saName, job.Spec.Template.Spec.ServiceAccountName)
+
+	testKube.DeleteBatch(batch)
```
Crumby: If the test fails after creation of the batch then the batch will become dangling resource. So either `defer testKube.DeleteBatch(batch)` right after `testKube.Create(batch)` or add a 
`"test-name" = TestName(t)` label (see DefaultSpec() lables) and it'll be auto cleared up by CleanNamespaceAndLogOnPanic().

Crumby (changes requested):
@tristantarrant Few more changes that need to be addressed, mainly the tests handle user resource removal after the test is finished

test/e2e/infinispan/service_account_test.go
tristantarrant: done

test/e2e/infinispan/service_account_test.go
tristantarrant: Done

Crumby:
Great job, thanks @tristantarrant

aman-kumar-2:
@Crumby, thanks team, just a quick question: which image version will this be implemented in, so that i can pull and use that image?

aman-kumar-2:
@tristantarrant, Which image version will this change be inherited by? Also, where in the CR can we specify the custom service account parameter? Could you please provide a sample template for the updated CR? Thanks.

Crumby:
@aman-kumar-2 Ability to configure serviceAccount is already available in the 2.5.14 release. For information and example on how to configure custom serviceAccount please take a look at https://infinispan.org/docs/infinispan-operator/main/operator.html#configuring-service-accounts_services.

PR #2540, opened 2026-06-24
[#2535] Update StartingCSV in upgrade tests to latest 2.4 release

### LINKED ISSUE #2535: Refresh upgrade test matrix
We are currently starting upgrade tests starting with CSV 2.3.7 and cycle through all the versions till we hit the current one. There's several issues with this approach:
* Redundancy testing: Testing upgrades from released to released version is redundant as it was already tested during the release cycle
* There's around ~20 intermediate Operator versions that the tests need to go through and upgrading through all of them takes a lot of time for OLM
* Some of the already released versions are flaky and cause CI failures. CI then need to be rerun which delays PR merges

Goal: Update the matrix to start with latest 2.4.x release (2.4.18) and it's respective releases. We allow for skipRange starting with 2.4.18 which greatly speeds up the test

Crumby:
Closes #2535 

Summary:
* Update StartingCSV to 2.4.18 (Last 2.4 release)
* `TestDroppedOperand` is the only test that explicitly uses installs 2.3.7.CSV
* Remove Infinispan 13 and 14 specific workaround from tests as we no longer test upgrades within neither
* Split upgrade test in two since we now allow for skipRange
  * Upgrade from 2.4.18 we allow the first version to skip from (and effectively from Infinispan 15.2.6)
  * Upgrade from last released (from latest available Infinispan)
* Add a shutdown/restart at the end of the upgrade test to verify the upgrade hasn't corrupted any files in addition to pod kill

### DIFF
--- a/.github/workflows/test_hr_rolling_upgrades.yml
+++ b/.github/workflows/test_hr_rolling_upgrades.yml
@@ -68,7 +68,7 @@ jobs:
           export TESTING_OPERAND_ALLOW_LIST=$(./scripts/ci/operand_versions.sh ,)
           make hotrod-upgrade-test
         env:
-          SUBSCRIPTION_STARTING_CSV: infinispan-operator.v2.3.7
+          SUBSCRIPTION_STARTING_CSV: infinispan-operator.v2.4.18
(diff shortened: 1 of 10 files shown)


tristantarrant:
Thanks @Crumby

PR #2544, opened 2026-07-02
Controller Runtime 0.19

### LINKED ISSUE #2210: Update the Operator SDK to version 1.26.1
We want to update to [Operator SDK v1.26.1](https://github.com/operator-framework/operator-sdk/releases/tag/v1.26.1),
that still supports Kubernetes 1.25.

### LINKED ISSUE #2543: Created ServiceMonitor is rejected by Prometheus Operator on OpenShift 4.22
The issue lies within the Prometheus operator library we use to create the ServiceMonitor resource. Version in place uses direct struct reference for `bearerTokenSecret` which is alway initialized with empty strings. On OpenShift 4.22 this gets rejected as we create the object with `basicAuth` and both can't be configured at the same time.

### LINKED ISSUE #2545: Update k8s.io dependency stack to 1.31
The update is required so we can fix #2543. We'll still be testing against k8s 1.25

Crumby:
Closes #2545
Closes #2543
Closes #2210


tristantarrant:
Merged, thanks @Crumby

PR #2575, opened 2026-07-31
Add Operand 16.2.2

infinispanrelease:
Add Operand 16.2.2


Crumby:
@pminz #2577 is required to fix the issue in the CI, so this PR can pass

PR #2585, opened 2026-08-18
Reconcilation pause annotation

### LINKED ISSUE #2578: Reconcilation pausing
**Summary:** Introduce the ability to pause Operator reconciliation on managed resources to easy debugging in case of unexpected failures.

Crumby:
Closes #2578

  **User-facing changes**
  
  - Added support for pausing reconciliation on any Infinispan CR (Infinispan, Cache, Batch, Backup, Restore) by setting the annotation infinispan.org/paused: "true".
  - While paused, the operator skips all reconciliation logic for the annotated resource — spec changes are not applied to managed resources.
  - A ReconciliationPaused status condition is set on the resource while paused, visible via kubectl get.
  - Removing the annotation (or setting it to any value other than "true") resumes normal reconciliation immediately.
  - Note: pausing also prevents deletion processing. A paused CR with finalizers will stay in Terminating state until unpaused.

  **Developer-facing changes**
  
  - Introduced a unified Condition type in v2alpha1 (replacing the cache-specific CacheCondition/CacheConditionType), shared across Cache, Batch, Backup, and Restore status types.
  - Added conditions field to BackupStatus, BatchStatus, and RestoreStatus (Cache already had it).
  - Extracted shared condition logic (setCondition, removeCondition, getCondition) into api/v2alpha1/condition_types.go.
  - Added Pausable interface and HandleReconciliationPause function in controllers/reconcile_pause.go, called at the top of each controller's Reconcile method.
  - Added ConditionReconciliationPaused to both v1 and v2alpha1 condition types.
  - Added e2e test (TestReconciliationPause) verifying pause blocks scaling and resume re-enables it.
  - Updated CRD manifests for all affected resources.

Created with the help of Claude Code.


controllers/reconcile_pause.go
```diff
@@ -0,0 +1,43 @@
...
+	logger logr.Logger,
+) (paused bool, err error) {
+	if obj.GetAnnotations()[consts.AnnotationPaused] == "true" {
+		logger.Info("Reconciliation paused via annotation 'infinispan.org/paused'")
+		if obj.SetPausedCondition(consts.ConditionMessageReconciliationPaused) {
+			if err := statusClient.Status().Update(ctx, obj); err != nil {
+				return false, err
+			}
+		}
+		eventRec.Event(obj, corev1.EventTypeNormal, consts.EventReasonReconciliationPaused, consts.EventMessageReconciliationPaused)
```
tristantarrant: This emits the event for every run. Maybe it should be moved inside the `if obj.SetPausedCondition(...)` block, so it only fires when the condition actually changes

controllers/zero_controller.go
```diff
@@ -143,6 +143,10 @@ func (z *zeroCapacityController) Reconcile(ctx context.Context, request reconcil
 		return reconcile.Result{}, fmt.Errorf("unable to fetch %s CR '%s': %w", resource, request.Name, err)
 	}
 
+	if paused, err := HandleReconciliationPause(ctx, instance.AsMeta().(Pausable), z.Client, z.EventRec, reqLogger); err != nil || paused {
```
tristantarrant: Make this a bit more robust in case we forget to add Pausable to resource types.

api/v2alpha1/condition_types.go
```diff
@@ -0,0 +1,66 @@
...
+	Type ConditionType `json:"type"`
+	// Status is the status of the condition.
+	Status metav1.ConditionStatus `json:"status"`
+	// Human-readable message indicating details about last transition.
+	// +optional
+	Message string `json:"message,omitempty"`
+}
+
+func (a ConditionType) equals(b ConditionType) bool {
+	return strings.EqualFold(strings.ToLower(string(a)), strings.ToLower(string(b)))
```
tristantarrant: `EqualFold` is already case insensitive. No need for the `ToLower`

controllers/infinispan.go
```diff
@@ -223,6 +226,10 @@ func (r *InfinispanReconciler) Reconcile(ctx context.Context, ctrlRequest ctrl.R
 		return reconcile.Result{}, fmt.Errorf("unable to fetch Infinispan CR %w", err)
 	}
 
+	if paused, err := HandleReconciliationPause(ctx, instance, r.Client, r.eventRec, reqLogger); err != nil || paused {
```
tristantarrant: I know you have documented the fact that paused CRs with finalizers will be stuck in Terminating.
What if the check is moved after the deletion check or log a warning when a paused resource has a deletiontimestamp

tristantarrant (changes requested):
Very nice.
I would like some changes.

controllers/infinispan.go
Crumby: I moved the checks and removed the section from documentation, the deletion now always goes through. The thing with finalizers mainly affects Cache CR as they required server side action after deletion.

tristantarrant:
Merged, thanks

PR #2590, opened 2026-08-20
[#2587] TerminationGracePeriod deepEquals and default check

### LINKED ISSUE #2587: Unset 'terminationGracePeriodSeconds' makes Operator fight with K8s for the default
There are two issues in TerminationGracePeriod checks:
1. It compares references rather then the values. Even if the values are the same, the check will fail as they live in different objects.
2. k8s sets TerminationGracePeriod to 30s if unset, which will make the check fail everytime it's unset, attempting to update SS spec on every reconciliation

```
...
if spec.TerminationGracePeriodSeconds != i.TerminationGracePeriodSeconds() {
	spec.TerminationGracePeriodSeconds = i.TerminationGracePeriodSeconds()
	updateNeeded = true
}
...
```

Crumby:
Closes #2587 

Three changes come within this PR:
* DeepEquals for TerminatinationGracePeriod and check against the default rather then direct set to nil
* Fix inverted DeepEqual check on encryption volumeMount check. Currently it sets updated to true if the volume mounts are equal, resulting in StatefulSet update with nothing to be updated.
* Print reason for the StatefulSet update in the Info log

### DIFF
--- a/pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
+++ b/pkg/reconcile/pipeline/infinispan/handler/manage/statefulset_updates.go
@@ -3,6 +3,7 @@ package manage
 import (
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	ispnv1 "github.com/infinispan/infinispan-operator/api/v1"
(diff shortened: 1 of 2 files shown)


tristantarrant:
merged, thanks

PR #2599, opened 2026-09-01
[#2598] Return the error instead of swallowing it

### LINKED ISSUE #2598: 'passwordFromSecret()' swallows Secret-read errors
We should return the error rather then swallow the error.

Crumby:
Closes #2598


tristantarrant:
Merged, thanks

PR #2606, opened 2026-09-07
fix(gossiprouter): Propagate spec.scheduling to Gossip Router Deployment

Hijbk304:
**What's the problem?**

Right now, if you define spec.scheduling in your Infinispan CR, the main StatefulSet pods happily respect all those nice affinity rules and tolerations. But the Gossip Router Deployment? [It totally ignores them](https://github.com/infinispan/infinispan-operator/blob/84f60dfd8f3246d087966405488210d9ae91ecb4/pkg/reconcile/pipeline/infinispan/handler/provision/gossiprouter.go#L171).

This creates a pretty annoying situation: the Gossip Router pod can land on a node that doesn't actually meet the requirements (e.g. missing a CPU feature, or tainted for maintenance), hang in Pending forever, and leave the whole cluster showing GossipRouterReady=False. Meanwhile, the StatefulSet pods are scheduled just fine — so you spend way too long debugging why everything looks correct but the cluster won't become ready.

**How does this PR fix it?**

Instead of building the Gossip Router PodSpec from scratch, I made it use the same helpers we already rely on for the StatefulSet — Affinity(), Tolerations(), TopologySpreadConstraints(), and PriorityClassName(). So now both workloads consistently respect the scheduling config you set in the CR. No more surprises.

**Testing**

I also updated provision_test.go to cover the new fields in the Gossip Router deployment


Crumby:
Hi @Hijbk304, thanks for opening the PR, I'll take a look at the changes later. For now I've enabled CI to get early results

Hijbk304:
Hi @Crumby, I see that Ci and the tests were successful. When you have time, take a look at this small PR =)

> Hi @Hijbk304, thanks for opening the PR, I'll take a look at the changes later. For now I've enabled CI to get early results

PR #2614, opened 2026-09-17
[#2368] StorageClassName immutabilty webhook checks and docs

### LINKED ISSUE #2368: Document that StorageClass is immutable field
Once the cluster was initialized with a particular StorageClass (`spec.service.container.storageClassName`) or ephemeral storage, it's not possible to change it and Operator won't react to it. It'll store the new value in the StorageClassName causing possible mismatch between desired configuration and actual state.

Ideally configure webhooks to not allow changes to the fields after initial deployment.

Crumby:
Closes #2368


api/v1/infinispan_webhook.go
```diff
@@ -234,8 +234,18 @@ func (i *Infinispan) ValidateUpdate(oldRuntimeObj runtime.Object) (admission.War
 		allErrs = append(allErrs, err)
 	}
 
-	if i.Spec.Service.Container != nil && i.Spec.Service.Container.Storage != nil && *old.Spec.Service.Container.Storage != *i.Spec.Service.Container.Storage {
-		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec").Child("service").Child("container").Child("storage"), "Storage configuration is immutable and cannot be updated after initial Infinispan creation"))
+	if i.Spec.Service.Container != nil && old.Spec.Service.Container != nil {
+		if *old.Spec.Service.Container.Storage != *i.Spec.Service.Container.Storage {
```
rigazilla, line 238: Storage could be nil if isDataGrid is false, not sure how is it likely but it could create panic

Crumby: Somebody would still have to be running `Cache` service for that and we do not allow new cluster with that service type for over two years. And for DataGrid we always default to 1Gi if not present.

Anyway I've replaced the check with proper utility check

rigazilla (approved):

rigazilla:
thanks @Crumby !

